PR #7, opened 2024-10-22
[PM-5693] KeyStore implementation

dani-garcia:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-5693

## 📔 Objective

Migrated from https://github.com/bitwarden/sdk/pull/979

I've been looking into using `memfd_secret` to protect keys in memory on Linux.

The `memfd_secret` API is similar to malloc in that we get some memory range allocated for us, but this memory is only visible to the process that has access to the file descriptor, which should protect us from any memory dumping from anything other than a kernel driver.

https://man.archlinux.org/man/memfd_secret.2.en

Using this requires changing how we store keys, so this is the perfect moment to go through removing the raw keys from the API surface.

## Changes
- Added a `KeyRef` trait and `SymmetricKeyRef`/`AsymmetricKeyRef` subtraits to represent the key types. Also a small macro to quickly implement them. `KeyRef` implementations are user defined types that implements `Eq+Hash+Copy`, and don't contain any key material
- `KeyEncryptable/KeyDecryptable` are now `Decryptable/Encryptable`, and instead of taking the key as parameter, they take a `CryptoServiceContext` and a `KeyRef`. This way keys are never exposed to the clients.
- `KeyLocator` trait is replaced by `UsesKey`. This trait only returns the `KeyRef` of the key required to decrypt it, instead of fetching the key itself.
- Created a `KeyStore` trait, with some simple CRUD methods. We have two implementations, one based on `memfd_secret` for Linux, and another one based on a Rust boxed slice, that also applies `mlock` if possible.
-  Because both `mlock` and `memfd_secret` apply their protection over a specified memory area, we need a Rust data structure that is laid out linearly and also doesn't reallocate by itself. Also because `memfd_secret` allocates the memory for us, we can't use a std type like Vec (reason is the Vec must be allocated by the Rust allocator [[ref](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.from_raw_parts)]). This basically only leaves us with a Rust slice, on top of which we'd need to implement insert/get/delete. To allow for reuse and easier testing I've created `SliceKeyStore`, which implements the CRUD methods from the `KeyStore` trait on top of a plain slice. Then the actual `mlock` and `memfd_secret` implementations just need to implement a trait casting their data to a slice. The data is stored as a slice of `Option<(KeyRef, KeyMaterial)>`, and the keys are unique and sorted in the slice for easier lookups.
- Created `CryptoService`, which contains the `KeyStore`s and some encrypt/decrypt utility functions. From a `CryptoService` you can also initialize a `CryptoServiceContext`
- A `CryptoServiceContext` contains a read only view of the keys inside the `CryptoService`, plus a read-write ephemeral key store, for use by `Decryptable/Encryptable` implementations when they need to temporarily store some keys (Cipher keys, attachment keys, send keys...).

Migrated the codebase to use these changes in a separate PR: https://github.com/bitwarden/sdk/pull/1117

## ⏰ Reminders before review

- CI builds passed
  team

## 🦮 Reviewer guidelines


crates/bitwarden-crypto/Cargo.toml
```diff
@@ -49,6 +49,11 @@ uuid = { workspace = true }
 wasm-bindgen = { workspace = true, optional = true }
 zeroize = { version = ">=1.7.0, <2.0", features = ["derive", "aarch64"] }
 
+[target.'cfg(not(target_arch = "wasm32"))'.dependencies]
+memsec = { version = "0.7.0", features = [
+    "alloc_ext",
+], git = "https://github.com/dani-garcia/memsec", rev = "3d2e272d284442637bac0a7d94f76883960db7e2" }
```
Hinton: We should try and get this merged upstream.

crates/bitwarden-crypto/src/service/context.rs
```diff
@@ -0,0 +1,357 @@
...
+impl<'a, SymmKeyRef: SymmetricKeyRef, AsymmKeyRef: AsymmetricKeyRef>
+    GlobalAccessMode<'a, SymmKeyRef, AsymmKeyRef> for ReadOnlyGlobal<'a, SymmKeyRef, AsymmKeyRef>
+{
+    fn get(&self) -> &Keys<SymmKeyRef, AsymmKeyRef> {
+        &self.0
+    }
+
+    fn get_mut(&mut self) -> Result<&mut Keys<SymmKeyRef, AsymmKeyRef>> {
+        Err(crate::CryptoError::ReadOnlyCryptoStore)
+    }
```
Hinton: Do we still need `get_mut` if we can't always access it? Or should we expose a different trait which can't use this to get compile time safety.

crates/bitwarden-crypto/src/error.rs
```diff
@@ -23,6 +23,10 @@ pub enum CryptoError {
     MissingKey(Uuid),
     #[error("The item was missing a required field: {0}")]
     MissingField(&'static str),
+    #[error("Missing Key for Ref. {0}")]
+    MissingKey2(String),
```
Hinton: `MissingKey2` 😭 Will this be removed once we remove all usages of the old interfaces?

crates/bitwarden-crypto/src/service/mod.rs
```diff
@@ -0,0 +1,197 @@
...
+        &self,
+        data: &[Data],
+    ) -> Result<Vec<Output>, crate::CryptoError> {
+        use rayon::prelude::*;
+
+        // We want to split all the data between available threads, but at the
+        // same time we don't want to split it too much if the amount of data is small.
+
+        // In this case, the minimum chunk size is 50.
+        let chunk_size = usize::max(1 + data.len() / rayon::current_num_threads(), 50);
```
Hinton: **suggestion**: Move this to a function to ensure consistency between both encrypt and decrypt.

crates/bitwarden-crypto/src/service/key_store/implementation/rust_slice.rs
Hinton, line 1: Do we have any documentation for why we need to implement our own slice vs using built in?

crates/bitwarden-crypto/src/service/mod.rs
```diff
@@ -0,0 +1,197 @@
...
+    /// Initiate an encryption/decryption context. This context will have MUTABLE access to the
+    /// global keys, and will have its own local key stores with read/write access. This
+    /// context-local store will be cleared up when the context is dropped.
+    ///
+    /// This is an advanced API, use with care and ONLY when needing to modify the global keys.
+    ///
+    /// The same pitfalls as `context` apply here, but with the added risk of accidentally
+    /// modifying the global keys and leaving the service in an inconsistent state.
+    ///
+    /// TODO: We should work towards making this pub(crate)
```
Hinton: What is preventing us from doing this now? Do we have a story to capture this work?

crates/bitwarden-crypto/src/service/mod.rs
```diff
@@ -0,0 +1,197 @@
...
+    /// global keys, and will have its own local key stores with read/write access. This
+    /// context-local store will be cleared up when the context is dropped.
+    ///
+    /// This is an advanced API, use with care. Prefer to instead use
+    /// `encrypt`/`decrypt`/`encrypt_list`/`decrypt_list` methods.
+    ///
+    /// One of the pitfalls of the current implementations is that keys stored in the context-local
+    /// store only get cleared automatically when dropped, and not between operations. This
+    /// means that if you are using the same context for multiple operations, you may want to
+    /// clear it manually between them.
```
Hinton: Do you have any examples of when this would be desired?

crates/bitwarden-crypto/src/keys/encryptable.rs
```diff
@@ -0,0 +1,238 @@
...
+>
+{
+    fn decrypt(
+        &self,
+        ctx: &mut CryptoServiceContext<SymmKeyRef, AsymmKeyRef>,
+        key: Key,
+    ) -> Result<Output, crate::CryptoError>;
+}
+
+// Basic Encryptable/Decryptable implementations to and from bytes
```
Hinton: Should these be doc comments?

crates/bitwarden-crypto/src/service/context.rs
```diff
@@ -0,0 +1,357 @@
...
+    derive_shareable_key,
+    service::{key_store::KeyStore, AsymmetricKeyRef, SymmetricKeyRef},
+    AsymmetricCryptoKey, AsymmetricEncString, CryptoError, EncString, Result, SymmetricCryptoKey,
+};
+
+// This is to abstract over the read-only and read-write access to the global keys
+// inside the CryptoServiceContext. The read-write access should only be used internally
+// in this crate to avoid users leaving the crypto store in an inconsistent state,
+// but for the moment we have some operations that require access to it.
+pub trait GlobalAccessMode<'a, SymmKeyRef: SymmetricKeyRef, AsymmKeyRef: AsymmetricKeyRef> {
```
Hinton: **question**: Is there a purpose in making the types generic?

crates/bitwarden-crypto/src/service/mod.rs
```diff
@@ -0,0 +1,197 @@
...
+
+mod key_store;
+
+use context::ReadWriteGlobal;
+pub use context::{CryptoServiceContext, ReadOnlyGlobal};
+pub use key_store::create_key_store;
+use key_store::KeyStore;
+
+#[derive(Clone)]
+pub struct CryptoService<SymmKeyRef: SymmetricKeyRef, AsymmKeyRef: AsymmetricKeyRef> {
```
Hinton: **question**: Should this be called a service, or something more generic like key store? I think we can drop the references to service.

crates/bitwarden-crypto/src/service/mod.rs
```diff
@@ -0,0 +1,197 @@
...
+    // We use an Arc<> to make it easier to pass this service around, as we can
+    // clone it instead of passing references
+    key_stores: Arc<RwLock<Keys<SymmKeyRef, AsymmKeyRef>>>,
```
Hinton: **question**: This seems to be one instance of Keys, is there a reason to use a plural name here?

crates/bitwarden-crypto/src/service/mod.rs
Hinton: **question**: We define a Keys struct which essentially acts as the KeyStore. Would it be a cleaner interface to define a KeyStore trait which both CryptoService and Keys implements, and CryptoService simply forwards calls to the inner store.

Hinton (commented):
I've taken an initial look at this. I think the general concept is solid, there are some potential areas for improvements.

However the biggest pieces would be:
1. Documentation, we need to ensure we write solid documentation on the difference areas and ensure we are clear about how it's intended to be used.
2. Improve test coverage, since we're dealing with cryptography we want to be really sure nothing breaks in the future.

crates/bitwarden-crypto/src/error.rs
dani-garcia: Yeah, that's removed in the other PR that starts using this code in the rest of the SDK:
https://github.com/bitwarden/sdk-internal/pull/8/files#diff-3a39fb8ed55a85cd232b4ac7681c85bb81462727f6d311b1a4c33be7c5b7f4ed

I didn't want to update it here to avoid needing to touch up code which was going to need to be updated in the other PR anyway

crates/bitwarden-crypto/src/service/mod.rs
dani-garcia: The only current use I have for reusing the context is during calls to `decrypt_list/encrypt_list`, which would helps avoid allocations/deallocations. In those calls we're manually clearing the context between iterations though.

In most cases you're only doing one operation at a time, so there's no benefit to reusing.

crates/bitwarden-crypto/src/service/mod.rs
dani-garcia: We're already using `KeyStore` for the trait that only deals with the key storage, so we'd need to rename that or choose a different name. 

But yeah I'm happy to rename it if we have a better name.

crates/bitwarden-crypto/src/service/context.rs
dani-garcia: So initially, the way I thought about the `CryptoContext` was that it would have a read-only view of the global keys, and a read-write exclusive store to deal with intermediate and ephemeral keys, like cipher keys.

The reason to enforce the read-only view was that it would allow you to parallelize decryption by just creating a bunch of independent contexts. This works great for decryptable/encryptable implementations, which don't have to modify the global keys ever.

What I didn't initially consider was that this context concept would also be useful when initializing the global keys themselves. For example, the context provides utilities for decrypting the org keys directly into the key store, but of course, as-is, the context can't store things into the global store. 

So to avoid duplicating the context code, I made the `Context` generic over the read/read-write status over the global store. This is a bit awkward and leads to strange things like the always-failing `get_mut` you mention here: https://github.com/bitwarden/sdk-internal/pull/7#discussion_r1824615486. I'm now thinking it would have been simpler with an `enum GlobalAccessMode { ReadOnly(RWLockReadGuard), ReadWrite(RWLockWriteGuard) }`

I think this change, combined with the local key refs not being a separate type and instead being a runtime check, makes things a bit weird. I want to look into improving the story around local keys but that's another thing.

crates/bitwarden-crypto/src/service/mod.rs
dani-garcia: I think I've refactored this a couple of times and this was from a time where things were structured differently, I can change it.

crates/bitwarden-crypto/src/keys/encryptable.rs
dani-garcia: I've mostly added these to split up and organize all the Encryptable/Decryptable implementations.

They all look very similar and when I didn't have these comments is was quite annoying to find one in particular. Maybe what that is telling us is that we should move those impls into separate modules or separate files.

dani-garcia:
I've made a few changes while documenting this a bit, plus some of the review suggestions are included here as well:

- `memsec` is disabled on windows for now, and we're not using my fork
- There's been a few renames:
    - `CryptoService` -> `KeyStore`. to move away from the service name
    - `Keys` -> `KeyStoreInner`. This only exists to be wrapped with the RwLock, and everything in `KeyStore` is pretty much delegated to it, so it seems like a better name.
    - `CryptoServiceContext` -> `KeyStoreContext`. To match the service rename
    - `KeyStore` -> `StoreBackend`. To signify there are different backends per platform
    - `LinuxMemfdSecretKeyStore` -> `LinuxMemfdSecretBackend` and `RustStore` -> `RustBackend`, to match the trait above.
- Changed the generic params of the `KeyStore` to be a single `KeyRefs`, which makes the generic types everywhere much less verbose (Instead of having to do `<SymmetricKeyRef, AsymmetricKeyRef>` everywhere you can do `<KeyRefs>`.
- Removed the generic type over `KeyStoreContext` global keys, and moved it to an enum.

-------------------

The next steps for the crypto refactor after this are:
- Move all the `Encryptable` implementations to the new trait: https://github.com/bitwarden/sdk-internal/pull/8
- Migrate existing APIs that deal with `SymmetricCryptoKey/AsymmetricCryptoKey` directly to deal with references instead (`init_crypto` functions, `MasterKey`, `UserKey`, `PinKey`, Key fingerprint etc). 
- Remove uses of `KeyStoreContext::get_key|set_key` from the non crypto crates, which shouldn't have any uses left after the previous step (maybe?).
- Once no other uses of `SymmetricCryptoKey/AsymmetricCryptoKey` are left, make them private to `bitwarden-crypto`
- Remove the now unnecessary boxing around `SymmetricCryptoKey/AsymmetricCryptoKey`

-------------------

Some future API improvements:
- Improve the design around local keys, instead of accepting any user type, create an opaque token and return it in the functions, so the only way for a user to get a valid local key ref is to obtain it from a function.
- Currently the context implementation can be read-only or read-write over the global keys, depending on how it's initialized, and it's behavior changes slightly in both cases. I think it might be a better design to instead do something like SQL transactions. Contexts are always read-only, but they have a `commit` function that consumes them and applies the changes to the global `KeyStore`

crates/bitwarden-crypto/src/service/mod.rs
dani-garcia: Initially I think I've tried to do something like that, but it was hard to get a good API when using a trait here and then also trying to be generic over key type / Encryptable implementation.

crates/bitwarden-crypto/src/service/mod.rs
dani-garcia: We need to update the types that generate these keys to insert them directly into the `KeyStore`. By that I mean types like `MasterKey`, `PinKey` etc.

Once those are updated to insert the keys directly instead of returning them, we won't need to use this from outside crypto again.

crates/bitwarden-crypto/src/service/key_store/implementation/rust_slice.rs
dani-garcia: Added some doc comments here: https://github.com/bitwarden/sdk-internal/blob/e7bfdc32aef6f8e6eaa91ebdb22e85d0992ecf70/crates/bitwarden-crypto/src/store/backend/slice_backend.rs#L40-L44

crates/bitwarden-crypto/Cargo.toml
```diff
@@ -50,6 +50,9 @@ wasm-bindgen = { workspace = true, optional = true }
 zeroize = { version = ">=1.7.0, <2.0", features = ["derive", "aarch64"] }
 zeroizing-alloc = ">=0.1.0, <0.2"
 
+[target.'cfg(all(not(target_arch = "wasm32"), not(windows)))'.dependencies]
+memsec = { version = "0.7.0", features = ["alloc_ext"] }
```
Hinton: Did we explore the alternative crate (`memfd`) which has more downloads?

crates/bitwarden-crypto/Cargo.toml
```diff
@@ -50,6 +50,9 @@ wasm-bindgen = { workspace = true, optional = true }
 zeroize = { version = ">=1.7.0, <2.0", features = ["derive", "aarch64"] }
 zeroizing-alloc = ">=0.1.0, <0.2"
 
+[target.'cfg(all(not(target_arch = "wasm32"), not(windows)))'.dependencies]
```
Hinton: Do we use memsec on mac?

crates/bitwarden-crypto/src/error.rs
Hinton: Can we rename this to `MissingKeyRef`? Put it under `MissingKey` and add a comment to rename this once we refactor the usage.

crates/bitwarden-crypto/src/lib.rs
```diff
@@ -80,6 +80,8 @@ mod util;
 pub use util::{generate_random_alphanumeric, generate_random_bytes, pbkdf2};
 mod wordlist;
 pub use wordlist::EFF_LONG_WORD_LIST;
+pub mod store;
```
Hinton: **question**: Do we want to use `pub mod` here?

crates/bitwarden-crypto/src/keys/mod.rs
```diff
@@ -1,5 +1,9 @@
 mod key_encryptable;
 pub use key_encryptable::{CryptoKey, KeyContainer, KeyDecryptable, KeyEncryptable, LocateKey};
+mod encryptable;
+pub use encryptable::{Decryptable, Encryptable, UsesKey};
+pub mod key_ref;
```
Hinton: **question**: `keys` isn't public so `key_ref` should probably not be public either?

```suggestion
pub key_ref;
```

crates/bitwarden-crypto/src/keys/key_ref.rs
```diff
@@ -0,0 +1,142 @@
...
+/// - One implementing [KeyRefs]
+pub trait KeyRef:
+    Debug + Clone + Copy + Hash + Eq + PartialEq + Ord + PartialOrd + Send + Sync + 'static
+{
+    type KeyValue: CryptoKey + Send + Sync + ZeroizeOnDrop;
+
+    /// Returns whether the key is local to the current context or shared globally by the
+    /// key store. See [crate::store::KeyStoreContext] for more information.
+    fn is_local(&self) -> bool;
+}
```
Hinton: **question:** Is this a `KeyRef` or `KeyIdentifier`?

I consider references to be more aligned with pointers while identifiers to be more of a GUID.

crates/bitwarden-crypto/src/keys/encryptable.rs
```diff
@@ -0,0 +1,191 @@
+use crate::{store::KeyStoreContext, AsymmetricEncString, CryptoError, EncString, KeyRef, KeyRefs};
+
+/// Types implementing [UsesKey] are capable of knowing which cryptographic key is
+/// needed to encrypt/decrypt them.
+pub trait UsesKey<Key: KeyRef> {
+    fn uses_key(&self) -> Key;
+}
```
Hinton: **suggestion:** I think the name `UsesKey` can be a bit misleading. It's not actually using a key rather presenting a identifier that should be used.

I can't really think of a better name though. `EncryptedBy` would work when decrypting but it's misleading when encrypting.

crates/bitwarden-crypto/src/keys/encryptable.rs
```diff
@@ -0,0 +1,191 @@
...
+/// the type.
+pub trait Decryptable<Refs: KeyRefs, Key: KeyRef, Output> {
+    fn decrypt(
+        &self,
+        ctx: &mut KeyStoreContext<Refs>,
+        key: Key,
+    ) -> Result<Output, crate::CryptoError>;
+}
+
+// Basic Encryptable/Decryptable implementations to and from bytes
```
Hinton: It's not obvious this comment is meant for multiple impl below. Any idea how we could improve it? Moving to a separate file would be the obvious one, but annoying.

crates/bitwarden-crypto/src/keys/encryptable.rs
Hinton, line 1: Can we improve the test coverage of the implementations?

Hinton, line 1: Can we rename this file? It also includes decryptable traits.

crates/bitwarden-crypto/src/store/backend/implementation/linux_memfd_secret.rs
```diff
@@ -0,0 +1,111 @@
...
+
+use super::{
+    slice_backend::{SliceBackend, SliceLike},
+    KeyRef,
+};
+
+// This is an in-memory key store that is protected by memfd_secret on Linux 5.14+.
+// This should be secure against memory dumps from anything except a malicious kernel driver.
+// Note that not all 5.14+ systems have support for memfd_secret enabled, so
+// LinuxMemfdSecretKeyStore::new returns an Option.
```
Hinton: Should these be docblocks?

Hinton: Link?

crates/bitwarden-crypto/src/store/backend/implementation/linux_memfd_secret.rs
```diff
@@ -0,0 +1,111 @@
...
+pub(crate) type LinuxMemfdSecretBackend<Key> = SliceBackend<Key, MemfdSecretImplKeyData>;
+
+pub(crate) struct MemfdSecretImplKeyData {
+    ptr: std::ptr::NonNull<[u8]>,
+    capacity: usize,
+}
```
Hinton: Can we document this struct?

crates/bitwarden-crypto/src/store/backend/implementation/linux_memfd_secret.rs
Hinton: Maybe move some of them to be module documentation?

crates/bitwarden-crypto/src/store/backend/implementation/linux_memfd_secret.rs
```diff
@@ -0,0 +1,111 @@
...
+
+// For Send+Sync to be safe, we need to ensure that the memory is only accessed mutably from one
+// thread. To do this, we have to make sure that any funcion in `MemfdSecretImplKeyData` that
+// accesses the pointer mutably is defined as &mut self, and that the pointer is never copied or
```
Hinton:
```suggestion
// accesses the pointer mutably is defined as `&mut self`, and that the pointer is never copied or
```

crates/bitwarden-crypto/src/store/backend/implementation/linux_memfd_secret.rs
```diff
@@ -0,0 +1,111 @@
...
+            true
+        })
+    }
+
+    fn with_capacity(capacity: usize) -> Self {
+        let entry_size = std::mem::size_of::<Option<(Key, Key::KeyValue)>>();
+
+        unsafe {
+            let ptr: NonNull<[u8]> = memsec::memfd_secret_sized(capacity * entry_size)
+                .expect("memfd_secret_sized failed");
```
Hinton: When would memfd fail? If we run out of memory?

crates/bitwarden-crypto/src/store/backend/implementation/linux_memfd_secret.rs
```diff
@@ -0,0 +1,111 @@
...
+
+            // Initialize the array with Nones using MaybeUninit
+            let uninit_slice: &mut [MaybeUninit<_>] = std::slice::from_raw_parts_mut(
+                ptr.as_ptr() as *mut MaybeUninit<Option<(Key, Key::KeyValue)>>,
+                capacity,
+            );
+            for elem in uninit_slice {
+                elem.write(None);
+            }
```
Hinton: This is probably unsafe? https://doc.rust-lang.org/std/slice/fn.from_raw_parts_mut.html

crates/bitwarden-crypto/src/store/backend/implementation/linux_memfd_secret.rs
```diff
@@ -0,0 +1,111 @@
...
+
+            MemfdSecretImplKeyData { ptr, capacity }
+        }
+    }
+
+    fn get_key_data(&self) -> &[Option<(Key, Key::KeyValue)>] {
+        let ptr = self.ptr.as_ptr() as *const Option<(Key, Key::KeyValue)>;
+        // SAFETY: The pointer is valid and points to a valid slice of the correct size.
+        // This function is &self so it only takes a immutable *const pointer.
```
Hinton: Can you elaborate on how we know it's valid?

crates/bitwarden-crypto/src/store/backend/implementation/mod.rs
```diff
@@ -0,0 +1,35 @@
...
+pub(crate) mod rust;
+
+/// Initializes a key store backend with the best available implementation for the current platform
+pub fn create_store<Key: KeyRef>() -> Box<dyn StoreBackend<Key>> {
+    #[cfg(all(target_os = "linux", not(feature = "no-memory-hardening")))]
+    if let Some(key_store) = linux_memfd_secret::LinuxMemfdSecretBackend::<Key>::new() {
+        return Box::new(key_store);
+    }
+
+    Box::new(rust::RustBackend::new().expect("RustKeyStore should always be available"))
```
Hinton:
```suggestion
    Box::new(rust::RustBackend::new().expect("RustKeyStore to always be available"))
```

Hinton (commented):
Still need to review the rust backend.

crates/bitwarden-crypto/src/keys/encryptable.rs
dani-garcia: We could do something like:
```rust
// #region: Implementation for x 

impl X for Y {
}

// #endregion
```

Which some IDEs support for code folding, but it's nonstandard, and I couldn't get it to work in VSCode right now.

crates/bitwarden-crypto/Cargo.toml
dani-garcia: Yeah, the `mlock`/`munlock` APIs work on any unix system, including Mac and BSDs. `memfd_secret` is Linux only

crates/bitwarden-crypto/src/keys/encryptable.rs
dani-garcia: I was mostly following what we were doing with `key_encryptable.rs` at the moment. Maybe this can be tied together with https://github.com/bitwarden/sdk-internal/pull/7#discussion_r1913072030, and we split the implementation in two files: `encryptable.rs` and `decryptable.rs`?

They'll be very similar, but at least it's a more clear split than moving only some implementations to other places.

crates/bitwarden-crypto/src/keys/key_ref.rs
dani-garcia: Yeah, when I started this I had three names going around my head: `KeyReference/KeyRef`, `KeyIdentifier/KeyId` and `KeyDescriptor`. I discarded that last one as I didn't think the `file descriptor <-> key descryptor` analogy i was going for was that useful.

Initially my design was more reference oriented, where holding a `KeyRef` would ensure that the key in the store could not be deleted until the reference was dropped, but it turned out into a lifetime/ownership hell for not that much benefit. 

Yeah we can rename to `KeyIdentifier/KeyId`. I'd agree it's more aligned with the current code. I'd use `KeyId` instead of `KeyIdentifier` though, as the type is used as a generic in lots of other types, so a short name is more convenient. Thoughts?

crates/bitwarden-crypto/src/keys/encryptable.rs
dani-garcia: I changed the name from the previous `LocatesKey` because of similar reasons yeah.

Some other names I can think of:
- `IdentifyKey` (would probably be great if we rename ref -> identifier)
- `SelectKey`
- `ProvideKey/ProvideKeyRef`

Any thoughts?

crates/bitwarden-crypto/src/store/backend/implementation/linux_memfd_secret.rs
dani-garcia: According to the docs (https://man7.org/linux/man-pages/man2/memfd_secret.2.html):
```
       EINVAL flags included unknown bits.

       EMFILE The per-process limit on the number of open file
              descriptors has been reached.

       EMFILE The system-wide limit on the total number of open files
              has been reached.

       ENOMEM There was insufficient memory to create a new anonymous
              file.

       ENOSYS memfd_secret() is not implemented on this architecture, or
              has not been enabled on the kernel command-line with
              secretmem_enable=1.
```

So when the system is out of memory or the process has reached the file descriptor limit, or if the feature is disabled and not supported (but we check for that the first time)

Hinton: Should we provide a better error message? This seems like an unrecoverable situation but it also sounds like we need to test what happens if we panic on desktop with napi?

crates/bitwarden-crypto/src/store/mod.rs
```diff
@@ -0,0 +1,316 @@
...
+}
+
+struct KeyStoreInner<Refs: KeyRefs> {
+    symmetric_keys: Box<dyn StoreBackend<Refs::Symmetric>>,
+    asymmetric_keys: Box<dyn StoreBackend<Refs::Asymmetric>>,
+}
+
+impl<Refs: KeyRefs> KeyStore<Refs> {
+    /// Create a new key store with the best available implementation for the current platform.
+    #[allow(clippy::new_without_default)]
```
Hinton: Any reason for not just implementing default?

crates/bitwarden-crypto/src/store/mod.rs
```diff
@@ -0,0 +1,316 @@
...
+    /// TODO: We should work towards making this pub(crate), and instead providing a safe API for
+    /// modifying the global keys. (i.e. `derive_master_key`, `derive_user_key`, etc.)
+    pub fn context_mut(&'_ self) -> KeyStoreContext<'_, Refs> {
+        KeyStoreContext {
+            global_keys: GlobalKeys::ReadWrite(self.inner.write().expect("RwLock is poisoned")),
+            local_symmetric_keys: create_store(),
+            local_asymmetric_keys: create_store(),
+            _phantom: std::marker::PhantomData,
+        }
+    }
```
Hinton: **question**: Can we provide some examples of when it's alright to use these APIs?

crates/bitwarden-crypto/src/store/mod.rs
```diff
@@ -0,0 +1,316 @@
...
+    ///
+    /// One of the pitfalls of the current implementations is that keys stored in the context-local
+    /// store only get cleared automatically when the context is dropped, and not between
+    /// operations. This means that if you are using the same context for multiple operations,
+    /// you may want to clear it manually between them.
+    pub fn context(&'_ self) -> KeyStoreContext<'_, Refs> {
+        KeyStoreContext {
+            global_keys: GlobalKeys::ReadOnly(self.inner.read().expect("RwLock is poisoned")),
+            local_symmetric_keys: create_store(),
+            local_asymmetric_keys: create_store(),
```
Hinton, line 166: **question:** Are there performance concerns with always initiating local stores? And if so would we be better of being lazy about these?

Currently we should rarely have local keys, once we get cipher key we'll regularly have a single item in local symmetric keys. I have a harder time imagining us creating many local keys?

crates/bitwarden-crypto/src/store/mod.rs
```diff
@@ -0,0 +1,316 @@
...
+    }
+}
+
+fn batch_chunk_size(len: usize) -> usize {
+    // We want to split all the data between available threads, but at the
+    // same time we don't want to split it too much if the amount of data is small.
+
+    // In this case, the minimum chunk size is 50. This was chosen pretty arbitrarily,
+    // but it seems to work well in practice.
+    usize::max(1 + len / rayon::current_num_threads(), 50)
```
Hinton: **issue:** Magic number, extract it as a constant with a good name?

crates/bitwarden-crypto/src/store/mod.rs
```diff
@@ -0,0 +1,316 @@
...
+        Data: Encryptable<Refs, Key, Output> + UsesKey<Key> + Send + Sync,
+        Output: Send + Sync,
+    >(
+        &self,
+        data: &[Data],
+    ) -> Result<Vec<Output>, crate::CryptoError> {
+        let res: Result<Vec<_>, _> = data
+            .par_chunks(batch_chunk_size(data.len()))
+            .map(|chunk| {
+                let mut ctx = self.context();
```
Hinton, line 268: **thought:** This will share the same context for the operations? If we use cipher key does this mean every cipher key is added to the context?

That could potentially result in inconsistent behaviour where we can have access to more keys than we should. It shouldn't cause issues but if we for some reason starts referencing the wrong key in the wrong ciphers this could result in weird behaviour?

crates/bitwarden-crypto/src/store/backend/implementation/rust.rs
```diff
@@ -0,0 +1,111 @@
...
+        not(feature = "no-memory-hardening")
+    ))]
+    {
+        use std::mem::MaybeUninit;
+        unsafe {
+            memsec::munlock(data.as_mut_ptr() as *mut u8, std::mem::size_of_val(data));
+
+            // Note: munlock is zeroing the memory, which leaves the data in an undefined
+            // state, so we set it to None/Default again to avoid UB in the Drop implementation.
+            let uninit_slice: &mut [MaybeUninit<_>] = std::slice::from_raw_parts_mut(
```
Hinton: **question:** Just checking but we fulfill all safety criteria defined in https://doc.rust-lang.org/std/slice/fn.from_raw_parts_mut.html?

crates/bitwarden-crypto/src/store/backend/implementation/rust.rs
```diff
@@ -0,0 +1,111 @@
...
+                data.as_mut_ptr() as *mut MaybeUninit<T>,
+                data.len(),
+            );
+            for elem in uninit_slice {
+                elem.write(T::default());
+            }
```
Hinton: **question:** Why is munlock zeroizing memory undefined state?

And do we have documentation that we're re-initializing it correctly.

crates/bitwarden-crypto/src/store/backend/slice_backend.rs
```diff
@@ -0,0 +1,743 @@
...
+impl<Key: KeyRef, Data: SliceLike<Key>> ZeroizeOnDrop for SliceBackend<Key, Data> {}
+
+impl<Key: KeyRef, Data: SliceLike<Key>> Drop for SliceBackend<Key, Data> {
+    fn drop(&mut self) {
+        self.clear();
+    }
+}
+
+impl<Key: KeyRef, Data: SliceLike<Key>> StoreBackend<Key> for SliceBackend<Key, Data> {
+    fn insert(&mut self, key_ref: Key, key: Key::KeyValue) {
```
Hinton: **issue:** This sounds like an upsert. I think we should name it accurately if that is the expected behaviour.

crates/bitwarden-crypto/src/store/backend/slice_backend.rs
```diff
@@ -0,0 +1,743 @@
...
+///
+/// Internally this is represented as a slice of `Option<(Key, Key::KeyValue)>`
+/// and elements are sorted based on Key for performance. In essence, this is almost a homegrown
+/// `HashMap`.
+///
+/// The reason why we're not using a Rust collection like `Vec` or `HashMap` is that those types
+/// expect their memory to be allocated by Rust, and they will deallocate/reallocate it as needed.
+/// That will not work for our usecases where we want to have control over allocations/deallocations
+/// and where some of our strategies rely on using system-allocated secure memory for the storage,
+/// like the Linux-only `memfd_secret` API.
```
Hinton: **question:** We seem to rely on this being an ordered list but it's not documented? Can we add this.

Hinton (commented):
Some more comments, still need to review slice_backend properly.

crates/bitwarden-crypto/src/store/mod.rs
dani-garcia: The operation itself follows the strategy of std types like Vec/HashMap/String, where the initial creation has capacity 0 and doesn't allocate, so it should be fairly cheap to do.

Once the first local key is inserted the store allocates for the first time.

crates/bitwarden-crypto/src/store/mod.rs
dani-garcia: It should be doing `ctx.clear();` inside the loop after every item is decrypted, which will clear all the keys so every item being decrypted starts with a clean slate.

This is one of the reasons why I have the recommendations to not use the context API directly, to avoid users dealing with these inconsistencies.

crates/bitwarden-crypto/src/store/backend/implementation/rust.rs
dani-garcia: Because the items inside the store can implement Drop themselves, and if they try to access their values after they have been zeroed it would cause UB.

crates/bitwarden-crypto/src/store/backend/slice_backend.rs
dani-garcia: The ordering is an implementation detail of the backend, I chose it because it allows us to do binary searches for lookup performance. It shouldn't be exposed outside its own API.

That said it would be good to document it yes.

crates/bitwarden-crypto/src/store/backend/slice_backend.rs
dani-garcia: Yeah, I based it on the std HashMap API, where the `insert` does insert and replace. I'll rename it

dani-garcia:
Split the secure storage implementations to https://github.com/bitwarden/sdk-internal/pull/125 so this is easier to review and merge.

crates/bitwarden-crypto/src/store/backend/implementation/basic.rs
```diff
@@ -0,0 +1,49 @@
...
+    fn clear(&mut self) {
+        self.keys.clear();
+    }
+
+    fn retain(&mut self, f: fn(Key) -> bool) {
+        self.keys.retain(|k, _| f(*k));
+    }
+}
+
+// Key::KeyValue already implements ZeroizeOnDrop,
```
Hinton:
```suggestion
// [Key::KeyValue] already implements ZeroizeOnDrop,
```

crates/bitwarden-crypto/src/store/backend/mod.rs
```diff
@@ -0,0 +1,34 @@
...
+
+use crate::store::KeyId;
+
+mod implementation;
+
+pub use implementation::create_store;
+
+/// This trait represents a platform that can store and return keys. If possible,
+/// it will try to enable as many security protections on the keys as it can.
+/// The keys themselves implement ZeroizeOnDrop, so the store will only need to make sure
```
Hinton:
```suggestion
/// The keys themselves implement [ZeroizeOnDrop], so the store will only need to make sure
```

crates/bitwarden-crypto/src/store/backend/mod.rs
```diff
@@ -0,0 +1,34 @@
...
+/// guarantees.
+///
+/// We have other implementations in testing using `mlock` and `memfd_secret` for protecting keys in
+/// memory.
+///
+/// Other implementations could use secure enclaves, HSMs or OS provided keychains.
+pub trait StoreBackend<Key: KeyId>: ZeroizeOnDrop + Send + Sync {
+    /// Inserts a key into the store. If the key already exists, it will be replaced.
+    fn upsert(&mut self, key_id: Key, key: Key::KeyValue);
+    /// Retrieves a key from the store.
```
Hinton, line 25: **suggestion:** New line between functions?

crates/bitwarden-crypto/src/store/context.rs
```diff
@@ -0,0 +1,531 @@
...
+    derive_shareable_key, store::backend::StoreBackend, AsymmetricCryptoKey, AsymmetricEncString,
+    CryptoError, EncString, KeyId, KeyIds, Result, SymmetricCryptoKey,
+};
+
+/// The context of a crypto operation using [super::KeyStore]
+///
+/// This will usually be accessed from an implementation of [crate::Decryptable] or
+/// [crate::Encryptable], but can also be obtained through [super::KeyStore::context]
+///
+/// This context contains access to the user keys stored in the [super::KeyStore] (sometimes refered
```
Hinton:
```suggestion
/// This context contains access to the user keys stored in the [super::KeyStore] (sometimes referred
```

crates/bitwarden-crypto/src/store/context.rs
```diff
@@ -0,0 +1,531 @@
...
+    pub(super) local_asymmetric_keys: Box<dyn StoreBackend<Ids::Asymmetric>>,
+
+    // Make sure the context is !Send & !Sync
+    pub(super) _phantom: std::marker::PhantomData<(Cell<()>, RwLockReadGuard<'static, ()>)>,
+}
+
+// A KeyStoreContext is usually limited to a read only access to the global keys,
+// which allows us to have multiple read only contexts at the same time and do multitheaded
+// encryption/decryption. We also have the option to create a read/write context, which allows us to
+// modify the global keys, but only allows one context at a time. This is controlled by a RWLock on
```
Hinton:
```suggestion
// modify the global keys, but only allows one context at a time. This is controlled by a [RwLock] on
```

crates/bitwarden-crypto/src/store/context.rs
```diff
@@ -0,0 +1,531 @@
...
+impl<Ids: KeyIds> KeyStoreContext<'_, Ids> {
+    /// Clears all the local keys stored in this context
+    /// If this context has write access to the global keys,
+    /// it will also clear them
+    pub fn clear(&mut self) {
+        // Clear global keys if we have write access
+        if let Ok(keys) = self.global_keys.get_mut() {
+            keys.symmetric_keys.clear();
+            keys.asymmetric_keys.clear();
+        }
```
Hinton: **question:** Can this be confusing? Chances are you accidentally clear the global keys and we might be better of splitting this into two?

And similar thoughts on the other methods.

crates/bitwarden-crypto/src/traits/key_id.rs
```diff
@@ -0,0 +1,121 @@
+use std::{fmt::Debug, hash::Hash};
+
+use zeroize::ZeroizeOnDrop;
+
+use crate::{AsymmetricCryptoKey, CryptoKey, SymmetricCryptoKey};
+
+/// Represents a key identifier that can be used to identify cryptographic keys in the
+/// key store. It is used to avoid exposing the key material directly in the public API.
+///
+/// This trait is user-implemented, and our recommended implementation is using enums with variants
```
Hinton:
```suggestion
/// This trait is user-implemented, and the recommended implementation is using enums with variants
```

crates/bitwarden-crypto/src/traits/encryptable.rs
```diff
@@ -0,0 +1,250 @@
...
+
+impl<Ids: KeyIds> Encryptable<Ids, Ids::Asymmetric, AsymmetricEncString> for &str {
+    fn encrypt(
+        &self,
+        ctx: &mut KeyStoreContext<Ids>,
+        key: Ids::Asymmetric,
+    ) -> Result<AsymmetricEncString, CryptoError> {
+        self.as_bytes().encrypt(ctx, key)
+    }
+}
```
Hinton, line 68: I believe we can generalize this code slightly.

```suggestion
impl<Ids, K, E> Encryptable<Ids, K, E> for &str
where
    Ids: KeyIds,
    K: KeyId,
    E: EncStringType,
    [u8]: Encryptable<Ids, K, E>,
{
    fn encrypt(&self, ctx: &mut KeyStoreContext<Ids>, key: K) -> Result<E, CryptoError> {
        self.as_bytes().encrypt(ctx, key)
    }
}
```

And define a common trait for `EncString`.
```rust
pub trait EncStringType {}
impl EncStringType for EncString {}
impl EncStringType for AsymmetricEncString {}
```

crates/bitwarden-crypto/src/store/mod.rs
Hinton, line 1: Can we add some quick module docs? I.e. `//!`

crates/bitwarden-crypto/src/store/mod.rs
```diff
@@ -0,0 +1,352 @@
...
+/// let encrypted = store.encrypt(decrypted).unwrap();
+/// ```
+#[derive(Clone)]
+pub struct KeyStore<Ids: KeyIds> {
+    // We use an Arc<> to make it easier to pass this store around, as we can
+    // clone it instead of passing references
+    inner: Arc<RwLock<KeyStoreInner<Ids>>>,
+}
+
+impl<Ids: KeyIds> std::fmt::Debug for KeyStore<Ids> {
```
Hinton, line 95:
```suggestion
// [KeyStore] contains sensitive data, provide a dummy [Debug] implementation.
impl<Ids: KeyIds> std::fmt::Debug for KeyStore<Ids> {
```

crates/bitwarden-crypto/src/store/context.rs
```diff
@@ -0,0 +1,531 @@
...
+    /// Decrypt a symmetric key into the context by using an already existing symmetric key
+    ///
+    /// # Arguments
+    ///
+    /// * `encryption_key` - The key id used to decrypt the `encrypted_key`. It must already exist
+    ///   in the context
+    /// * `new_key_id` - The key id where the decrypted key will be stored. If it already exists, it
+    ///   will be overwritten
+    /// * `encrypted_key` - The key to decrypt
+    pub fn decrypt_symmetric_key_with_symmetric_key(
```
Hinton, line 139: I wonder if the naming here is slightly misleading. We're not decrypting it we're adding it to the context?

dani-garcia: Well, we're doing both: We decrypt the key and add it to the context. Originally i was planning to make it a generic function like `decrypt_key_into_store`, but I couldn't get the typing to do what I wanted, which explains the long names. (And calling it `decrypt_symmetric_key_into_store_using_symmetric_key` is just way too long)

crates/bitwarden-crypto/src/store/context.rs
dani-garcia: `HTTP Referer` strikes back 😄

crates/bitwarden-crypto/src/traits/encryptable.rs
dani-garcia: This is something Matt also brought up when trying to do authenticated data. I think the best path forward would be to separate encryptable/decrytable into encryptable/decryptable and encodable/decodable, something like this.

```rust
impl <T: Encodable<Vec<u8>> Encryptable<EncString> for T
impl <T: Encodable<Vec<u8>> Encryptable< AsymmetricEncString> for T

impl Encodable<Vec<u8>> for &str
impl Encodable<Vec<u8>> for String
```

crates/bitwarden-crypto/src/store/context.rs
dani-garcia: Yeah, ultimately I'd want to remove the mutable access from the context once we have a better way to initialize the user keys into the KeyStore, so I didn't want to split the methods into two to then remove them anyway, I can look into it if you think it's important.

Hinton (approved):

PR #8, opened 2024-10-22
[PM-5693] Migrate SDK to KeyStore

dani-garcia:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-5693

## 📔 Objective

Migrated from https://github.com/bitwarden/sdk/pull/1117

Migrate the codebase to the new KeyStore introduced in https://github.com/bitwarden/sdk-internal/pull/7.

EncryptionSettings was removed from Client, though it is still used to initialize the KeyStore. Ideally we'd move that initialization code over directly into the crypto crate but this PR is big enough as it is.

There are still some things using keys directly and KeyEncryptable/KeyDecryptable, like MasterKey, PinKey, DeviceKey, private key fingerprint. Those would need to be migrated over on a separate PR. 

We need to remove the rest of the uses of SymmetricCryptoKey from the client crates, then we can remove the internal boxing they are doing, as the keys would be protected by the KeyStore instead.

Secrets Manager code should be moved to using the Encryptable/Decryptable interface, as it's encrypting fields one by one, I'll look into doing that on a separate PR.

## ⏰ Reminders before review

- CI builds passed
  team

## 🦮 Reviewer guidelines

### DIFF
--- a/bitwarden_license/bitwarden-sm/src/projects/list.rs
+++ b/bitwarden_license/bitwarden-sm/src/projects/list.rs
@@ -1,5 +1,6 @@
 use bitwarden_api_api::models::ProjectResponseModelListResponseModel;
-use bitwarden_core::client::{encryption_settings::EncryptionSettings, Client};
+use bitwarden_core::{client::Client, key_management::KeyIds};
+use bitwarden_crypto::KeyStoreContext;
 use schemars::JsonSchema;
 use serde::{Deserialize, Serialize};
 use uuid::Uuid;
@@ -24,9 +25,9 @@ pub(crate) async fn list_projects(
     )
     .await?;
 
-    let enc = client.internal.get_encryption_settings()?;
+    let key_store = client.internal.get_key_store();
 
-    ProjectsResponse::process_response(res, &enc)
+    ProjectsResponse::process_response(res, &mut key_store.context())
 }
 
 #[derive(Serialize, Deserialize, Debug, JsonSchema)]
@@ -38,14 +39,14 @@ pub struct ProjectsResponse {
 impl ProjectsResponse {
     pub(crate) fn process_response(
         response: ProjectResponseModelListResponseModel,
-        enc: &EncryptionSettings,
+        ctx: &mut KeyStoreContext<KeyIds>,
     ) -> Result<Self, SecretsManagerError> {
         let data = response.data.unwrap_or_default();
 
         Ok(ProjectsResponse {
             data: data
                 .into_iter()
-                .map(|r| ProjectResponse::process_response(r, enc))
+                .map(|r| ProjectResponse::process_response(r, ctx))
                 .collect::<Result<_, _>>()?,
         })
     }
--- a/bitwarden_license/bitwarden-sm/src/secrets/create.rs
+++ b/bitwarden_license/bitwarden-sm/src/secrets/create.rs
@@ -1,6 +1,6 @@
 use bitwarden_api_api::models::SecretCreateRequestModel;
-use bitwarden_core::Client;
-use bitwarden_crypto::KeyEncryptable;
+use bitwarden_core::{key_management::SymmetricKeyId, Client};
+use bitwarden_crypto::Encryptable;
 use schemars::JsonSchema;
 use serde::{Deserialize, Serialize};
 use uuid::Uuid;
@@ -34,16 +34,23 @@ pub(crate) async fn create_secret(
 ) -> Result<SecretResponse, SecretsManagerError> {
     input.validate()?;
 
-    let enc = client.internal.get_encryption_settings()?;
-    let key = enc.get_key(&Some(input.organization_id))?;
+    let key_store = client.internal.get_key_store();
+    let key = SymmetricKeyId::Organization(input.organization_id);
+    let mut ctx = key_store.context();
 
     let secret = Some(SecretCreateRequestModel {
-        key: input.key.clone().trim().encrypt_with_key(key)?.to_string(),
-        value: input.value.clone().encrypt_with_key(key)?.to_string(),
-        note: input.note.clone().trim().encrypt_with_key(key)?.to_string(),
+        key: input.key.clone().trim().encrypt(&mut ctx, key)?.to_string(),
+        value: input.value.clone().encrypt(&mut ctx, key)?.to_string(),
+        note: input
+            .note
+            .clone()
+            .trim()
+            .encrypt(&mut ctx, key)?
+            .to_string(),
         project_ids: input.project_ids.clone(),
         access_policies_requests: None,
     });
+    drop(ctx);
 
     let config = client.internal.get_api_configurations().await;
     let res = bitwarden_api_api::apis::secrets_api::organizations_organization_id_secrets_post(
@@ -53,7 +60,7 @@ pub(crate) async fn create_secret(
     )
     .await?;
 
-    SecretResponse::process_response(res, &enc)
+    SecretResponse::process_response(res, &mut key_store.context())
 }
 
 #[cfg(test)]
--- a/bitwarden_license/bitwarden-sm/src/secrets/create.rs
+++ b/bitwarden_license/bitwarden-sm/src/secrets/create.rs
@@ -36,21 +36,22 @@ pub(crate) async fn create_secret(
 
     let key_store = client.internal.get_key_store();
     let key = SymmetricKeyId::Organization(input.organization_id);
-    let mut ctx = key_store.context();
 
-    let secret = Some(SecretCreateRequestModel {
-        key: input.key.clone().trim().encrypt(&mut ctx, key)?.to_string(),
-        value: input.value.clone().encrypt(&mut ctx, key)?.to_string(),
-        note: input
-            .note
-            .clone()
-            .trim()
-            .encrypt(&mut ctx, key)?
-            .to_string(),
-        project_ids: input.project_ids.clone(),
-        access_policies_requests: None,
-    });
-    drop(ctx);
+    let secret = {
+        let mut ctx = key_store.context();
+        Some(SecretCreateRequestModel {
+            key: input.key.clone().trim().encrypt(&mut ctx, key)?.to_string(),
+            value: input.value.clone().encrypt(&mut ctx, key)?.to_string(),
+            note: input
+                .note
+                .clone()
+                .trim()
+                .encrypt(&mut ctx, key)?
+                .to_string(),
+            project_ids: input.project_ids.clone(),
+            access_policies_requests: None,
+        })
+    };
 
     let config = client.internal.get_api_configurations().await;
     let res = bitwarden_api_api::apis::secrets_api::organizations_organization_id_secrets_post(
--- a/crates/bitwarden-core/src/auth/auth_client.rs
+++ b/crates/bitwarden-core/src/auth/auth_client.rs
@@ -189,6 +189,7 @@ fn trust_device(client: &Client) -> Result<TrustDeviceResponse, TrustDeviceError
 
     let key_store = client.internal.get_key_store();
     let ctx = key_store.context();
+    // FIXME: [PM-18099] Once DeviceKey deals with KeyIds, this should be updated
     #[allow(deprecated)]
     let user_key = ctx.dangerous_get_symmetric_key(SymmetricKeyId::User)?;
 
--- a/crates/bitwarden-core/src/auth/auth_client.rs
+++ b/crates/bitwarden-core/src/auth/auth_client.rs
@@ -185,9 +185,12 @@ pub enum TrustDeviceError {
 
 #[cfg(feature = "internal")]
 fn trust_device(client: &Client) -> Result<TrustDeviceResponse, TrustDeviceError> {
-    let enc = client.internal.get_encryption_settings()?;
+    use crate::key_management::SymmetricKeyId;
 
-    let user_key = enc.get_key(&None)?;
+    let key_store = client.internal.get_key_store();
+    let ctx = key_store.context();
+    #[allow(deprecated)]
+    let user_key = ctx.dangerous_get_symmetric_key(SymmetricKeyId::User)?;
 
     Ok(DeviceKey::trust_device(user_key)?)
 }
--- a/crates/bitwarden-core/src/client/encryption_settings.rs
+++ b/crates/bitwarden-core/src/client/encryption_settings.rs
@@ -66,6 +66,7 @@ impl EncryptionSettings {
             // )
         };
 
+        // FIXME: [PM-18098] When this is part of crypto we won't need to use deprecated methods
         #[allow(deprecated)]
         {
             let mut ctx = store.context_mut();
@@ -82,6 +83,7 @@ impl EncryptionSettings {
     /// This is used only for logging in Secrets Manager with an access token
     #[cfg(feature = "secrets")]
     pub(crate) fn new_single_key(key: SymmetricCryptoKey, store: &KeyStore<KeyIds>) {
+        // FIXME: [PM-18098] When this is part of crypto we won't need to use deprecated methods
         #[allow(deprecated)]
         store
             .context_mut()
--- a/crates/bitwarden-core/src/client/encryption_settings.rs
+++ b/crates/bitwarden-core/src/client/encryption_settings.rs
@@ -1,15 +1,16 @@
-use std::collections::HashMap;
-
-use bitwarden_crypto::{AsymmetricCryptoKey, CryptoError, KeyContainer, SymmetricCryptoKey};
+use bitwarden_crypto::{AsymmetricCryptoKey, KeyStore, SymmetricCryptoKey};
 #[cfg(feature = "internal")]
-use bitwarden_crypto::{AsymmetricEncString, EncString, MasterKey};
+use bitwarden_crypto::{AsymmetricEncString, EncString};
 use bitwarden_error::bitwarden_error;
 use thiserror::Error;
 use uuid::Uuid;
 
 #[cfg(feature = "internal")]
 use crate::error::Result;
-use crate::VaultLockedError;
+use crate::{
+    key_management::{AsymmetricKeyId, KeyIds, SymmetricKeyId},
+    VaultLockedError,
+};
 
 #[bitwarden_error(flat)]
 #[derive(Debug, Error)]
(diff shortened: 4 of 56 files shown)


bitwarden_license/bitwarden-sm/src/secrets/create.rs
```diff
@@ -34,16 +34,23 @@ pub(crate) async fn create_secret(
...
+        note: input
+            .note
+            .clone()
+            .trim()
+            .encrypt(&mut ctx, key)?
+            .to_string(),
         project_ids: input.project_ids.clone(),
         access_policies_requests: None,
     });
+    drop(ctx);
```
Hinton: Is this to avoid keeping them in memory longer than needed? Can we document it if that's the case.

Would it be nicer to use a block scope?

```
let secret = {
    let mut ctx = key_store.context();

    Some(SecretCreateRequestModel {
        key: input.key.clone().trim().encrypt(&mut ctx, key)?.to_string(),
        value: input.value.clone().encrypt(&mut ctx, key)?.to_string(),
        note: input
            .note
            .clone()
            .trim()
            .encrypt(&mut ctx, key)?
            .to_string(),
        project_ids: input.project_ids.clone(),
        access_policies_requests: None,
    })
}; 
```

dani-garcia: Because we're using RwLocks inside the context, if you hold on to them across an await point, that makes the async future !Send, which causes compilation issues in the wasm and uniffi bindings.

Dropping the context before we call the API endpoint is one way to solve it, I agree that looks nicer. I'll change it.

I'll add some docs mentioning that it's not recommended to hold the context across await points too.

bitwarden_license/bitwarden-sm/src/projects/list.rs
```diff
@@ -1,5 +1,6 @@
 use bitwarden_api_api::models::ProjectResponseModelListResponseModel;
-use bitwarden_core::client::{encryption_settings::EncryptionSettings, Client};
+use bitwarden_core::{client::Client, key_management::KeyIds};
+use bitwarden_crypto::KeyStoreContext;
```
Hinton, line 3: We depend on `KeyStoreContext` and `KeyIds` more or less everywhere. Would it make sense to expose a syntax sugar from `core` that is `KeyStoreContext<KeyIds>`?

bitwarden_license/bitwarden-sm/src/secrets/create.rs
dani-garcia: Resolved in https://github.com/bitwarden/sdk-internal/commit/d454cf4c0228d0c0921662c603347aa082de324f?w=1

crates/bitwarden-exporters/src/models.rs
```diff
@@ -83,7 +82,7 @@ fn from_login(
             .collect(),
         totp: l.totp,
         fido2_credentials: l.fido2_credentials.as_ref().and_then(|_| {
-            let credentials = view.get_fido2_credentials(enc).ok()?;
+            let credentials = view.get_fido2_credentials(&mut key_store.context()).ok()?;
```
Hinton, line 85: **question:** We call `context` extensively and there are warnings that it is an advanced API. Should we rethink the phrasing there?

crates/bitwarden-core/src/auth/auth_client.rs
```diff
@@ -185,9 +185,12 @@ pub enum TrustDeviceError {
...
 #[cfg(feature = "internal")]
 fn trust_device(client: &Client) -> Result<TrustDeviceResponse, TrustDeviceError> {
-    let enc = client.internal.get_encryption_settings()?;
+    use crate::key_management::SymmetricKeyId;
 
-    let user_key = enc.get_key(&None)?;
+    let key_store = client.internal.get_key_store();
+    let ctx = key_store.context();
+    #[allow(deprecated)]
+    let user_key = ctx.dangerous_get_symmetric_key(SymmetricKeyId::User)?;
```
Hinton, line 194: Could we comment all instances of deprecated why we do this, and if we plan on eventually replacing it add a link to a ticket?

crates/bitwarden-core/src/client/encryption_settings.rs
```diff
@@ -30,32 +31,9 @@ pub enum EncryptionSettingsError {
...
-    pub(crate) private_key: Option<AsymmetricCryptoKey>,
-    org_keys: HashMap<Uuid, SymmetricCryptoKey>,
-}
-
-impl std::fmt::Debug for EncryptionSettings {
-    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        f.debug_struct("EncryptionSettings").finish()
-    }
-}
+pub struct EncryptionSettings {}
```
Hinton, line 34: **question:** Should we maybe rethink the structure of enc settings in the future now that it doesn't contain any members?

crates/bitwarden-core/src/client/encryption_settings.rs
```diff
@@ -85,76 +66,58 @@ impl EncryptionSettings {
...
-            user_key: key,
-            private_key: None,
-            org_keys: HashMap::new(),
-        }
+    pub(crate) fn new_single_key(key: SymmetricCryptoKey, store: &KeyStore<KeyIds>) {
+        #[allow(deprecated)]
+        store
+            .context_mut()
+            .set_symmetric_key(SymmetricKeyId::User, key)
+            .expect("Mutable context");
```
Hinton: Maybe we can split `EncryptionSettings` into two, one for PM and one for SM?

crates/bitwarden-core/src/client/encryption_settings.rs
```diff
@@ -85,76 +66,58 @@ impl EncryptionSettings {
...
-        &mut self,
         org_enc_keys: Vec<(Uuid, AsymmetricEncString)>,
-    ) -> Result<&Self, EncryptionSettingsError> {
-        use bitwarden_crypto::KeyDecryptable;
+        store: &KeyStore<KeyIds>,
+    ) -> Result<(), EncryptionSettingsError> {
+        let mut ctx = store.context_mut();
+
+        if !ctx.has_asymmetric_key(AsymmetricKeyId::UserPrivateKey) {
+            return Err(VaultLockedError.into());
```
Hinton: **issue:** Technically this can also mean the user doesn't have a private key. I think we're breaking the later PM-11690 return. And we'd want to test this on such an account.

crates/bitwarden-core/src/client/internal.rs
```diff
@@ -223,30 +210,15 @@ impl InternalClient {
...
-            return Err(VaultLockedError.into());
-        };
-
-        let inner = Arc::make_mut(enc);
-        inner.set_org_keys(org_keys)?;
-
-        Ok(enc.clone())
+    ) -> Result<(), EncryptionSettingsError> {
+        EncryptionSettings::set_org_keys(org_keys, &self.key_store)?;
+        Ok(())
```
Hinton: We can just rely on the return type from `set_org_keys`.
```suggestion
        EncryptionSettings::set_org_keys(org_keys, &self.key_store)
```

crates/bitwarden-core/src/key_management/mod.rs
```diff
@@ -0,0 +1,55 @@
...
+    pub enum AsymmetricKeyId {
+        UserPrivateKey,
+        #[local]
+        Local(&'static str),
+    }
+
+    pub KeyIds => SymmetricKeyId, AsymmetricKeyId;
+}
+
+pub fn create_test_crypto_with_user_key(key: SymmetricCryptoKey) -> KeyStore<KeyIds> {
```
Hinton, line 35: **issue:** Can we add a comment why we don't mark these as `#[test]`?

crates/bitwarden-core/src/key_management/mod.rs
Hinton, line 1: **suggestion:** Can we add a quick module doc? And document the functions? I've been revisiting older areas and introduced more documentation where applicable.

crates/bitwarden-vault/src/cipher/cipher.rs
```diff
@@ -279,26 +291,38 @@ impl Cipher {
...
     /// in which case this will return Ok(None) and the key associated
     /// with this cipher's user or organization must be used instead
+    ///
+    /// # Arguments
+    ///
+    /// * `ctx` - The key store context where the cipher key will be decrypted, if it exists
+    /// * `key` - The key to use to decrypt the cipher key, this should be the user or organization
+    ///   key
+    /// * `ciphers_key` - The encrypted cipher key
     pub(super) fn get_cipher_key(
```
Hinton: **question:** Should this be renamed? We're actually adding the key to the context. Or at least updating the docs to reflect this.

crates/bitwarden-vault/src/cipher/cipher.rs
```diff
@@ -192,74 +194,84 @@ impl CipherListView {
...
-        let ciphers_key = Cipher::get_cipher_key(key, &self.key)?;
-        let key = ciphers_key.as_ref().unwrap_or(key);
+impl Decryptable<KeyIds, SymmetricKeyId, CipherView> for Cipher {
+    fn decrypt(
+        &self,
+        ctx: &mut KeyStoreContext<KeyIds>,
+        key: SymmetricKeyId,
+    ) -> Result<CipherView, CryptoError> {
+        let ciphers_key = Cipher::get_cipher_key(ctx, key, &self.key)?;
+        let key = ciphers_key.unwrap_or(key);
```
Hinton: **question:** We're doing this most of the time, maybe we can update `get_cipher_key` to default to the key passed in?

crates/bitwarden-vault/src/cipher/attachment.rs
```diff
@@ -56,22 +57,36 @@ pub struct AttachmentFileView<'a> {
...
 
         // Because this is a new attachment, we have to generate a key for it, encrypt the contents
         // with it, and then encrypt the key with the cipher key
-        let attachment_key = SymmetricCryptoKey::generate(rand::thread_rng());
-        let encrypted_contents = self.contents.encrypt_with_key(&attachment_key)?;
-        attachment.key = Some(attachment_key.to_vec().encrypt_with_key(ciphers_key)?);
+        let attachment_key = ctx.generate_symmetric_key(ATTACHMENT_KEY)?;
+        let encrypted_contents = self.contents.encrypt(ctx, attachment_key)?;
+        attachment.key =
+            Some(ctx.encrypt_symmetric_key_with_symmetric_key(ciphers_key, attachment_key)?);
```
Hinton, line 88: **thought:** It'd be nice if we could mark keys as exportable when we generate them. We'd probably need a way to bypass it but we could mark that as a dangerous operation to prevent accidental exfiltration of more sensitive keys?

crates/bitwarden-core/src/platform/generate_fingerprint.rs
```diff
@@ -61,11 +61,10 @@ pub(crate) fn generate_user_fingerprint(
...
 
-    let enc_settings = client.internal.get_encryption_settings()?;
-    let private_key = enc_settings
-        .private_key
-        .as_ref()
-        .ok_or(UserFingerprintError::MissingPrivateKey)?;
+    let key_store = client.internal.get_key_store();
+    let ctx = key_store.context();
+    #[allow(deprecated)]
+    let private_key = ctx.dangerous_get_asymmetric_key(AsymmetricKeyId::UserPrivateKey)?;
```
Hinton, line 69: **question:** Probably a follow up item, but it might be useful to have a way to get the public key of private keys which would be a non dangerous operation.

bitwarden_license/bitwarden-sm/src/projects/list.rs
dani-garcia: Yeah I don't like it very much either, this is kind of the reason I have `KeyIds` in the first place, previously you'd have to specify both key types individually.

I think the problem with that is that the KeyIds type also needs to be present on the impls for Decryptable:
```rust
impl Decryptable<KeyIds, SymmetricKeyId, CipherListView> for Cipher {
    fn decrypt(&self, ctx: &mut KeyStoreContext<KeyIds>, ...
```

And we can't alias a trait for implementation, I've tried this and it doesn't compile:
```rust
pub type KeyStore<Ids = KeyIds> = bitwarden_crypto::KeyStore<Ids>;
pub type KeyStoreContext<'a, Ids = KeyIds> = bitwarden_crypto::KeyStoreContext<'a, Ids>;
pub type Encryptable<Key, Output, Ids = KeyIds> =
    dyn bitwarden_crypto::Encryptable<Ids, Key, Output>;
pub type Decryptable<Key, Output, Ids = KeyIds> =
    dyn bitwarden_crypto::Decryptable<Ids, Key, Output>;

struct Test;

// expected trait, found type alias `Encryptable`
// type aliases cannot be used as traits
impl Encryptable<SymmetricKeyId, EncString> for Test {
    fn encrypt(
        &self,
        ctx: &mut KeyStoreContext,
        key: SymmetricKeyId,
    ) -> Result<bitwarden_crypto::EncString, bitwarden_crypto::CryptoError> {
        todo!()
    }
}
```

dani-garcia: Added https://bitwarden.atlassian.net/browse/PM-18106 to study this in the future

crates/bitwarden-core/src/client/encryption_settings.rs
dani-garcia: Yeah, I tried to keep it as is to make sure the PR was smaller but it's one of the first things I'd want to do after this PR.

I'd want to expose this initialization code using safe to use functions from bitwarden-crypto and remove EncryptionSettings entirely: https://bitwarden.atlassian.net/browse/PM-18098

crates/bitwarden-core/src/client/encryption_settings.rs
dani-garcia: Hmm that's weird, previously it was returning `EncryptionSettingsError::MissingPrivateKey` correctly, I have no idea why I'd change it.

I'll change it back to return the older error and skip if there are no org keys

crates/bitwarden-vault/src/cipher/cipher.rs
dani-garcia: Yeah sounds reasonable, we were doing this previously because we had to deal with lifetime issues because the parent key was passed as reference while the cipher key was owned. It's not needed anymore with key ids, changed it in https://github.com/bitwarden/sdk-internal/pull/8/commits/a1cc46c79a176f6c1e97aeba3bb93318fad733f9

crates/bitwarden-vault/src/cipher/cipher.rs
dani-garcia: Renamed to `decrypt_cipher_key` and expanded the comment to mention that:
```rust
/// Decrypt the individual encryption key for this cipher into the provided [KeyStoreContext]
/// and return it's identifier. Note that some ciphers do not have individual encryption
/// keys, in which case this will return the provided key identifier instead
```

crates/bitwarden-core/src/key_management/mod.rs
dani-garcia: Added in https://github.com/bitwarden/sdk-internal/pull/8/commits/45af28f9afb103e4cb798ff5cc5e1fa3bb17d97e

crates/bitwarden-core/src/key_management/mod.rs
dani-garcia: Done, it's because we're using them from other crates as well, and marking them as `#[cfg(test)]` won't make them available when testing the other crates. We could have some testing feature, but that's outside the scope of this PR

crates/bitwarden-core/src/client/encryption_settings.rs
dani-garcia: Ideally we can do as mentioned in https://github.com/bitwarden/sdk-internal/pull/8#discussion_r1946411661 and expose some safe to use functions for these from bitwarden-crypto and then remove encryptionSettings entirely.

crates/bitwarden-core/src/platform/generate_fingerprint.rs
dani-garcia: Good idea, noted https://bitwarden.atlassian.net/browse/PM-18110

crates/bitwarden-core/src/client/encryption_settings.rs
Hinton: I think we need to return this error after, we check if `org_enc_keys` is empty, https://github.com/bitwarden/sdk-internal/pull/8/files/4a6363b988a03d9393a44817993c701896cfae77#diff-673812a5c065e8c33981caaa0b1e1bedc9369bbf318782925769b5a9ae10cf73R109

crates/bitwarden-core/src/client/encryption_settings.rs
```diff
@@ -96,8 +96,13 @@ impl EncryptionSettings {
     ) -> Result<(), EncryptionSettingsError> {
         let mut ctx = store.context_mut();
 
+        // FIXME: [PM-11690] - Early abort to handle private key being corrupt
+        if org_enc_keys.is_empty() {
+            return Ok(());
+        }
```
Hinton: This is duplicated? Did we forget to remove the occurrence below?

crates/bitwarden-core/src/key_management/mod.rs
Hinton: Mind creating a ticket for it? We probably don't want to accidentally expose these code paths in production builds.

dani-garcia: Done: https://bitwarden.atlassian.net/browse/PM-18112

crates/bitwarden-core/src/client/encryption_settings.rs
dani-garcia: Oops, looks like I accidentally moved it during the rebase and then didn't notice it was already down there, fixed in 9b50890

crates/bitwarden-core/src/auth/auth_client.rs
dani-garcia: Added ticket references and short explanations to all the non-test uses, is this what you had in mind?
https://github.com/bitwarden/sdk-internal/pull/8/commits/36e8a51b05d26c536a66beb9d154e7afc2188ac0

crates/bitwarden-exporters/src/models.rs
dani-garcia: A fair amount of the uses of `context` are to access the raw key material and will dissappear once MasterKey/etc support KeyIds, but yeah we can soften the message a little bit, specially in cases like this where you directly pass the context to a function so a lot of the caveats of context don't apply.

Some of the needed contexts around passkey operations are because they need to decrypt the cipher key to deal with their data, there might be a better way to do it without requiring them to deal with contexts, I've made a ticket in case we can think of something: https://bitwarden.atlassian.net/browse/PM-18113

dani-garcia: Updated the context comment to remove advanced notice and reorganized it a bit, let me know what you think: 
https://github.com/bitwarden/sdk-internal/pull/8/commits/9c263f996a887781fe5c38998debcd1ef9f6c9c7

crates/bitwarden-vault/src/cipher/attachment.rs
dani-garcia: Sounds like a nice idea, added it to the list: https://bitwarden.atlassian.net/browse/PM-18114

Hinton (approved):

PR #31, opened 2024-11-12
[PM-13660] implement catchable non generic error types

coroiu:
## 📔 Objective

- Makes it easier to return errors from rust.
- Makes it easier to catch and handle errors in JS.

Every error outputs a function `is<type_name>()` that can be used from TS to check and cast errors.

Example:

```rs
#[bitwarden_error(flat)]
#[derive(Debug, Error)]
pub enum EncryptionSettingsError {
    #[error("Cryptography error, {0}")]
    Crypto(#[from] bitwarden_crypto::CryptoError),

    #[error(transparent)]
    InvalidBase64(#[from] base64::DecodeError),

    #[error(transparent)]
    VaultLocked(#[from] VaultLocked),

    #[error("Invalid private key")]
    InvalidPrivateKey,

    #[error("Missing private key")]
    MissingPrivateKey,
}
```

```ts
export interface EncryptionSettingsError extends Error {
  name: "EncryptionSettingsError";
  variant: "Crypto" | "InvalidBase64" | "VaultLocked" | "InvalidPrivateKey" | "MissingPrivateKey";
}

export function isEncryptionSettingsError(error: any): error is EncryptionSettingsError;
```

## ⏰ Reminders before review

- CI builds passed
  team

## 🦮 Reviewer guidelines


crates/bitwarden-error-macro/src/flat/attribute.rs
```diff
@@ -0,0 +1,117 @@
...
+    export_as_identifier: &proc_macro2::Ident,
+) -> proc_macro::TokenStream {
+    match &input.data {
+        Data::Enum(data) => {
+            let variant_names = data.variants.iter().map(|variant| &variant.ident);
+
+            let match_arms = data.variants.iter().map(|variant| match variant.fields {
+                syn::Fields::Unit => {
+                    let variant_ident = &variant.ident;
+                    let variant_name = format!("{}", variant_ident);
```
dani-garcia: There's a bit of repetition with these lines for all the branches, can they be moved outside the match?

Also small subjective choice, but `variant_ident.to_string()` is the same as `format!("{}", variant_ident)`, but I think it looks nicer

crates/bitwarden-error-macro/src/flat/attribute.rs
```diff
@@ -0,0 +1,117 @@
...
+                    &variant_names.collect::<Vec<_>>(),
+                )
+            });
+
+            quote! {
+                #input
+                #wasm
+
+                #[automatically_derived]
+                impl FlatError for #type_identifier {
```
dani-garcia: When referring to types in macro generated code, we want to always use the fully qualified names, otherwise we're requiring users to import the correct type. If you're importing`prelude::*` that's not an issue, but nothing keeps the user from directly importing just the macro.
```suggestion
                impl ::bitwarden_error::prelude::FlatError for #type_identifier {
```

Note that we also use `::` in front of the crate name, this is to avoid collisions with user defined modules using the same name.

crates/bitwarden-error-macro/src/args.rs
```diff
@@ -0,0 +1,25 @@
...
+pub(crate) struct BitwardenErrorArgs {
+    #[darling(flatten)]
+    pub error_type: BitwardenErrorType,
+
+    #[darling(default)]
+    pub export_as: Option<String>,
+}
+
+#[derive(FromMeta)]
+pub(crate) enum BitwardenErrorType {
```
dani-garcia, line 14: You should be able to use `#[darling(rename_all = "camelCase")]` once at the enum definition to avoid repeating the rename for each variant.

crates/bitwarden-error-macro/src/lib.rs
```diff
@@ -0,0 +1,13 @@
+mod args;
+mod attribute;
+mod basic;
+mod flat;
+mod full;
+
+#[proc_macro_attribute]
+pub fn bitwarden_error(
```
dani-garcia, line 115: We should add a doc comment here explaining what this macro does and what each of the supported variants mean, that way it should be visible in the user's IDE.

dani-garcia (commented):
Nice, looks pretty good to me, some initial comments on the macro implementation.

crates/bitwarden-error-macro/src/flat/attribute.rs
coroiu: I did this for types that we depend on (e.g. `wasm-bindgen`) but never considered our own types, good catch!

crates/bitwarden-error-macro/src/flat/attribute.rs
coroiu: Good catch, I'll fix both!

crates/bitwarden-error-macro/src/lib.rs
coroiu: Thanks for the reminder!

crates/bitwarden-error-macro/src/lib.rs
```diff
@@ -0,0 +1,120 @@
...
+
+/// A procedural macro for generating error types with customizable serialization behavior.
+///
+/// # Attributes
+///
+/// ## Error type
+
+/// - `basic`: The error is converted into a string using the `ToString` trait.
+/// - `flat`: The error is converted into a flat structure using the `FlatError` trait.
+/// - `full`: The entire error stack is made available using `serde`.
```
dani-garcia, line 15: Docs look pretty nice! It would be good to have some recommendations on when to use some variant over the others. As a new developer it might be hard to know when each is appropiate.

dani-garcia (approved):

crates/bitwarden-error-macro/src/lib.rs
coroiu: I was thinking about writing some contributing docs page about this, but honestly people will just look at how other's code looks and copy paste :D

PR #35, opened 2024-11-21
[PM-10942] Add bitwarden-ssh crate with keygeneration support

quexten:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-10942

## 📔 Objective

Adds a new bitwarden-ssh crate to the sdk. This should later include both import-parsing and key generation.
This PR also exposes this for the web clients (not mobile yet).

## ⏰ Reminders before review

- CI builds passed
  team

## 🦮 Reviewer guidelines


crates/bitwarden-ssh/Cargo.toml
```diff
@@ -0,0 +1,45 @@
...
+    "dep:js-sys",
+    "dep:tsify-next",
+] # WASM support
+
+[dependencies]
+bitwarden-error = { workspace = true }
+bitwarden-vault = { workspace = true }
+rand = "0.8.5"
+rand_chacha = "0.3.1"
+ssh-encoding = "0.2.0"
```
Hinton: Seems unused?

crates/bitwarden-ssh/Cargo.toml
```diff
@@ -0,0 +1,45 @@
...
+ssh-key = { version = "0.6.7", features = [
+    "ed25519",
+    "encryption",
+    "rsa",
+    "getrandom",
+] }
+thiserror = { workspace = true }
+uniffi = { workspace = true, optional = true }
+uuid = { workspace = true }
```
Hinton: Unused?

crates/bitwarden-ssh/Cargo.toml
```diff
@@ -0,0 +1,45 @@
...
+wasm-bindgen = { workspace = true, optional = true }
+tsify-next = { workspace = true, optional = true }
```
Hinton: Unused? Only needed for full errors.

crates/bitwarden-ssh/src/lib.rs
```diff
@@ -0,0 +1,51 @@
...
+    Rsa3072,
+    Rsa4096,
+}
+
+pub fn generate_keypair(
+    key_algorithm: KeyAlgorithm,
+) -> Result<models::SshKey, error::KeyGenerationError> {
+    // sourced from cryptographically secure entropy source, with sources for all targets: https://docs.rs/getrandom
+    // if it cannot be securely sourced, this will panic instead of leading to a weak key
+    let mut rng: ChaCha8Rng = ChaCha8Rng::from_entropy();
```
Hinton: **issue:** Please create a new private function which takes `mut rng: impl rand::RngCore,` as input. This way we can unit test it using chacha iniitalized from a seed. There are some examples in `bitwarden-crypto` on how to do this.

crates/bitwarden-ssh/src/models.rs
```diff
@@ -0,0 +1,5 @@
+pub struct SshKey {
+    pub private_key: String,
+    pub public_key: String,
+    pub key_fingerprint: String,
+}
```
Hinton: **suggestion**: I think the naming here can be confusing. We already have a concept of `SshKey`, and they are not the same type. We can either re-use `SshKeyView`, or change this to be a DTO, we generally call these the same as the function and append result, i.e. `GenerateKeypairResult`.

crates/bitwarden-ssh/src/lib.rs
```diff
@@ -0,0 +1,51 @@
...
+pub mod error;
+pub mod models;
+
+pub enum KeyAlgorithm {
+    Ed25519,
+    Rsa3072,
+    Rsa4096,
+}
+
+pub fn generate_keypair(
```
Hinton: **suggestion:** `keypair` generally means an RSA keypair. In this case we generate an ssh keypair and we may benefit from including ssh in the name, to avoid confusion on the calling side where we can depend on bitwarden-crypto and bitwarden-ssh.

crates/bitwarden-ssh/src/lib.rs
```diff
@@ -0,0 +1,51 @@
...
+                KeyAlgorithm::Rsa4096 => 4096,
+                _ => unreachable!(),
+            };
+
+            let rsa_keypair = ssh_key::private::RsaKeypair::random(&mut rng, bits)
+                .map_err(|e| error::KeyGenerationError::KeyGenerationError(e.to_string()))?;
+
+            let private_key = ssh_key::PrivateKey::new(
+                ssh_key::private::KeypairData::from(rsa_keypair),
+                "".to_string(),
```
Hinton: https://docs.rs/ssh-key/0.6.7/ssh_key/private/struct.PrivateKey.html#method.new accepts `Into<String>` which should work with string slices, i.e `""`.

crates/bitwarden-ssh/src/lib.rs
```diff
@@ -0,0 +1,51 @@
...
+        KeyAlgorithm::Ed25519 => ssh_key::PrivateKey::random(&mut rng, Algorithm::Ed25519),
+        KeyAlgorithm::Rsa3072 | KeyAlgorithm::Rsa4096 => {
+            let bits = match key_algorithm {
+                KeyAlgorithm::Rsa3072 => 3072,
+                KeyAlgorithm::Rsa4096 => 4096,
+                _ => unreachable!(),
+            };
+
+            let rsa_keypair = ssh_key::private::RsaKeypair::random(&mut rng, bits)
+                .map_err(|e| error::KeyGenerationError::KeyGenerationError(e.to_string()))?;
```
Hinton: **suggestion:** Please import the used error rather than always using the full path.
```suggestion
                .map_err(|e| KeyGenerationError::KeyGenerationError(e.to_string()))?;
```

crates/bitwarden-ssh/src/models.rs
Hinton: With a DTO we can also lift this into the lib.rs to keep it closer to the function producing it.

crates/bitwarden-wasm-internal/src/ssh.rs
```diff
@@ -0,0 +1,58 @@
...
+    }
+
+    pub fn public_key(&self) -> &str {
+        &self.public_key
+    }
+
+    pub fn key_fingerprint(&self) -> &str {
+        &self.key_fingerprint
+    }
+}
```
Hinton: @coroiu Do we need to re-define the types? We should look into a way to conditionally apply `wasm_bindgen` in the original crates, similar to our uniffi flow.

Hinton (commented):
Great start, let's get this function unit tested if possible, and remove some of the dependencies. I also started a discussion in slack about the divergence from our standard rand generator.

crates/bitwarden-wasm-internal/src/ssh.rs
coroiu: You should be able to use `#[cfg_attr(feature = "wasm", wasm_bindgen)]` at the source but ideally `bitwarden_ssh::models::SshKey` would implement `serialize` and `tsify` instead. See https://contributing.bitwarden.com/architecture/sdk/password-manager/web/interoperability#how-to-use-tsify

crates/bitwarden-ssh/Cargo.toml
```diff
@@ -0,0 +1,39 @@
...
+rand = "0.8.5"
+rand_chacha = "0.3.1"
+ssh-key = { version = "0.6.7", features = [
+    "ed25519",
+    "encryption",
+    "rsa",
+    "getrandom",
+] }
+thiserror = { workspace = true }
+uniffi = { workspace = true, optional = true }
```
Hinton: Are we using uniffi here?

crates/bitwarden-ssh/Cargo.toml
```diff
@@ -0,0 +1,39 @@
...
+homepage.workspace = true
+repository.workspace = true
+license-file.workspace = true
+keywords.workspace = true
+
+[features]
+uniffi = ["dep:uniffi", "bitwarden-vault/uniffi"]
+wasm = ["bitwarden-error/wasm", "dep:wasm-bindgen", "dep:js-sys"] # WASM support
+
+[dependencies]
```
Hinton, line 19: Please sort these alphabetically

crates/bitwarden-ssh/Cargo.toml
```diff
@@ -0,0 +1,39 @@
...
+bitwarden-error = { workspace = true }
+bitwarden-vault = { workspace = true }
+rand = "0.8.5"
+rand_chacha = "0.3.1"
```
Hinton: We can remove chacha now since we use thread rng. Or move to dev-dependency once tests are added.

crates/bitwarden-ssh/Cargo.toml
```diff
@@ -0,0 +1,39 @@
...
+    "ed25519",
+    "encryption",
+    "rsa",
+    "getrandom",
+] }
+thiserror = { workspace = true }
+uniffi = { workspace = true, optional = true }
+wasm-bindgen = { workspace = true, optional = true }
+js-sys = { workspace = true, optional = true }
+tsify-next.workspace = true
```
Hinton: Should be optional?

crates/bitwarden-wasm-internal/src/ssh.rs
```diff
@@ -0,0 +1,27 @@
...
+// impl conversion
+impl From<KeyAlgorithm> for bitwarden_ssh::KeyAlgorithm {
+    fn from(key_algorithm: KeyAlgorithm) -> Self {
+        match key_algorithm {
+            KeyAlgorithm::Ed25519 => bitwarden_ssh::KeyAlgorithm::Ed25519,
+            KeyAlgorithm::Rsa3072 => bitwarden_ssh::KeyAlgorithm::Rsa3072,
+            KeyAlgorithm::Rsa4096 => bitwarden_ssh::KeyAlgorithm::Rsa4096,
+        }
+    }
+}
```
Hinton: Now that `KeyAlgorithm` uses tsify we should be able to remove this.

crates/bitwarden-wasm-internal/src/lib.rs
```diff
@@ -1,6 +1,7 @@
 mod client;
 mod crypto;
 mod custom_types;
+pub mod ssh;
```
Hinton: Do we need pub here?

crates/bitwarden-ssh/src/lib.rs
```diff
@@ -0,0 +1,62 @@
...
+}
+
+pub fn generate_sshkey(
+    key_algorithm: KeyAlgorithm,
+) -> Result<GenerateSshKeyResult, error::KeyGenerationError> {
+    let rng = rand::thread_rng();
+    generate_sshkey_internal(key_algorithm, rng)
+}
+
+fn generate_sshkey_internal(
```
Hinton, line 25: Were we able to add some tests for this function?

quexten: Added tests with seeded keys now.

crates/bitwarden-ssh/src/error.rs
```diff
@@ -0,0 +1,11 @@
+use bitwarden_error::prelude::*;
+use thiserror::Error;
+
+#[bitwarden_error(flat)]
+#[derive(Error, Debug)]
+pub enum KeyGenerationError {
+    #[error("Failed to generate key: {0}")]
+    KeyGenerationError(String),
```
dani-garcia: You should be able to change these from `String` to `ssh_key::Error`, which would also simplify your `map_err` to:
```
let rsa_keypair = ssh_key::private::RsaKeypair::random(&mut rng, bits)
                .map_err(KeyGenerationError::KeyGenerationError)?;
```

dani-garcia (commented):
LGTM, looks like the tests are failing on windows because of different line endings 😭 

Probably need to update the tests to use something like `include_str!("../tests/rsa4096_key").replace("\r\n", "\n");` when loading the file, or configure Git to not do autocrlf on windows.

crates/bitwarden-ssh/src/error.rs
quexten: Nice, should both be fixed now I think

dani-garcia (approved):

PR #38, opened 2024-11-25
Replace ZeroizingAllocator

### LINKED ISSUE #37: Allocator wrapper implementation contains undefined behavior
Hey there,

While looking at Bitwarden's [implementation of zeroizing dealloc](https://github.com/bitwarden/sdk-internal/blob/main/crates/bitwarden-crypto/src/allocator.rs), I noticed that there is some undefined behavior present:
- [slice::from_raw_parts](https://github.com/bitwarden/sdk-internal/blob/main/crates/bitwarden-crypto/src/allocator.rs#L30) on an entire object is invalid. This is because structures and other types may contain padding bytes, which are _not_ initialized during the allocation process. By making a slice from it, you are essentially converting `MaybeUninit<u8>` to `u8` unsoundly. Rust is clear that integers may uninitialized.
- [All of the tests](https://github.com/bitwarden/sdk-internal/blob/main/crates/bitwarden-crypto/src/allocator.rs#L43), which are disabled on some platforms (out of curiosity, which ones?) are all full of UB. Even if its only in a test, reading memory from a freed heap object's pointer is undefined/unsound in basically every language, not just Rust. IMO the tests should be either deleted or replaced with a custom `GlobalAlloc` that keeps track of all allocations and can provide a sound API to check if "freed" memory was wiped with zeroes.

We (1Password) made a similar crate wrapper last year for our Core, and we just open sourced it today if you'd like to use it directly or as a reference to improve on this: https://github.com/1Password/zeroizing-alloc

dani-garcia:
## 📔 Objective

Replace our internal ZeroizingAllocator by the external crate `zeroizing-alloc`, which has the same functionality but removes some undefined behavior.

## ⏰ Reminders before review

- CI builds passed
  team

## 🦮 Reviewer guidelines


crates/bitwarden-crypto/src/lib.rs
```diff
@@ -80,8 +80,7 @@ mod util;
 pub use util::{generate_random_alphanumeric, generate_random_bytes, pbkdf2};
 mod wordlist;
 pub use wordlist::EFF_LONG_WORD_LIST;
-mod allocator;
-pub use allocator::ZeroizingAllocator;
+pub use zeroizing_alloc::ZeroAlloc as ZeroizingAllocator;
```
Hinton, line 83: Should we rename ours to `ZeroAlloc`?

Hinton (commented):
From a code perspective this looks good. I will spend some time digging into the produced code in a few environments to confirm it behaves as we expect.

crates/bitwarden-crypto/src/lib.rs
dani-garcia: I kept the old name for backwards compatibility, probably not really needed but it was simple enough so I didn't see a reason not to.

We can use `ZeroAlloc` and ditch the old naming, if we don't consider that useful.

Hinton: I guess there's no harm keeping the current name.

Hinton (approved):
Tested manually on the different environments. As far as I can tell it correctly zeroizes allocated memory.

PR #39, opened 2024-11-26
[PM-14229] Add more docs to `bitwarden_core`

justindbaur:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-14229

## 📔 Objective

Adds some more docs to `bitwarden_core` it's far from fully documented but I thought it was a good start to make sure I am on the right track about what should be expected.

## ⏰ Reminders before review

- CI builds passed
  team

## 🦮 Reviewer guidelines

### DIFF
--- a/crates/bitwarden-core/src/client/encryption_settings.rs
+++ b/crates/bitwarden-core/src/client/encryption_settings.rs
@@ -30,10 +30,20 @@ pub enum EncryptionSettingsError {
     MissingPrivateKey,
 }
 
+/// A struct containing the core keys of a Bitwarden user.
 #[derive(Clone)]
 pub struct EncryptionSettings {
+    /// The users symmetric key, stored on the server after being
+    /// encrypted with another key.
     user_key: SymmetricCryptoKey,
+    /// The users private key, stored on the server
+    /// encrypted with the users symmetric key.
     pub(crate) private_key: Option<AsymmetricCryptoKey>,
+    /// A map of the users organization keys with the key being
+    /// the ID of the organization and the value being the symmetric
+    /// key. This map may be empty if the user is not a part of
+    /// any organizations. These keys are stored on the server
+    /// encrypted with the users private key.
     org_keys: HashMap<Uuid, SymmetricCryptoKey>,
 }
 
--- a/crates/bitwarden-core/src/client/internal.rs
+++ b/crates/bitwarden-core/src/client/internal.rs
@@ -41,6 +41,9 @@ pub(crate) struct Tokens {
     pub(crate) refresh_token: Option<String>,
 }
 
+/// A struct contains various internal actions. Everything on this type
+/// should be considered unstable and subject to change at any time. Use
+/// with caution.
 #[derive(Debug)]
 pub struct InternalClient {
     pub(crate) tokens: RwLock<Tokens>,
--- a/crates/bitwarden-core/src/platform/client_platform.rs
+++ b/crates/bitwarden-core/src/platform/client_platform.rs
@@ -12,7 +12,7 @@ pub struct ClientPlatform<'a> {
 
 impl<'a> ClientPlatform<'a> {
     /// Will generate a fingerprint based on the `input`. Given the same `input` This
-    /// method will result in the exact same output.
+    /// method will always result in the exact same output.
     ///
     /// # Examples
     /// ```rust
@@ -40,7 +40,7 @@ impl<'a> ClientPlatform<'a> {
 
     /// Will generate a fingerprint based on the given `fingerprint_material`
     /// and the users public key. Given the same `fingerprint_material` and
-    /// the same user. This method will result in the exact same output.
+    /// the same user this method will always result in the exact same output.
     ///
     /// The returned fingerprint is a string of 5 words seperated by hyphens.
     ///
--- a/crates/bitwarden-core/src/platform/client_platform.rs
+++ b/crates/bitwarden-core/src/platform/client_platform.rs
@@ -16,19 +16,23 @@ impl<'a> ClientPlatform<'a> {
     ///
     /// # Examples
     /// ```rust
+    /// # use bitwarden_core::client::test_accounts::PUBLIC_KEY;
+    /// use base64::{engine::general_purpose::STANDARD, Engine};
     /// use bitwarden_core::{Client, platform::FingerprintRequest};
     ///
-    /// async fn test() {
-    ///     let client = Client::test_account().await;
+    /// fn test(client: Client) {
     ///     let fingerprint_response = client.platform()
     ///         .fingerprint(&FingerprintRequest {
     ///             fingerprint_material: "my_material".to_owned(),
-    ///             public_key: "...public key...".to_owned(),
+    ///             public_key: STANDARD.encode(PUBLIC_KEY),
     ///         })
     ///         .unwrap();
     ///
-    ///     println!("{}", fingerprint_response.fingerprint);
+    ///     assert_eq!(fingerprint_response.fingerprint, "unsure-unethical-bruising-semester-subscript");
     /// }
+    ///
+    /// # tokio::runtime::Builder::new_current_thread().enable_all().build().unwrap().block_on(
+    /// #   async { test(Client::test_account().await) });
     /// ```
     pub fn fingerprint(&self, input: &FingerprintRequest) -> Result<FingerprintResponse> {
         generate_fingerprint(input)
--- a/crates/bitwarden-core/src/platform/client_platform.rs
+++ b/crates/bitwarden-core/src/platform/client_platform.rs
@@ -5,15 +5,54 @@ use super::{
 };
 use crate::{error::Result, Client};
 
+/// A struct containing platform utilities.
 pub struct ClientPlatform<'a> {
     pub(crate) client: &'a Client,
 }
 
 impl<'a> ClientPlatform<'a> {
+    /// Will generate a fingerprint based on the `input`. Given the same `input` This
+    /// method will result in the exact same output.
+    ///
+    /// # Examples
+    /// ```rust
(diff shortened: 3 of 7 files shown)


crates/bitwarden-core/src/client/encryption_settings.rs
```diff
@@ -30,10 +30,20 @@ pub enum EncryptionSettingsError {
...
+    /// encrypted with another key.
     user_key: SymmetricCryptoKey,
+    /// The users private key, stored on the server
+    /// encrypted with the users symmetric key.
     pub(crate) private_key: Option<AsymmetricCryptoKey>,
+    /// A map of the users organization keys with the key being
+    /// the ID of the organization and the value being the symmetric
+    /// key. This map may be empty if the user is not a part of
+    /// any organizations. These keys are stored on the server
+    /// encrypted with the users private key.
```
coroiu, line 46: thought: I'm always confused on how to write this, technically they are encrypted with the public key and decrypted with the private key :shrug:

crates/bitwarden-core/src/client/internal.rs
```diff
@@ -41,6 +41,9 @@ pub(crate) struct Tokens {
     pub(crate) refresh_token: Option<String>,
 }
 
+/// A struct contains various internal actions. Everything on this type
+/// should be considered unstable and subject to change at any time. Use
+/// with caution.
```
coroiu, line 46: question/suggestion: Who are targeting with this documentation? For external consumers this makes sense, but for our internal products we're expected to use this and not have to "be afraid". Should we maybe clarify that?

crates/bitwarden-core/src/platform/client_platform.rs
```diff
@@ -5,15 +5,54 @@ use super::{
...
 use crate::{error::Result, Client};
 
+/// A struct containing platform utilities.
 pub struct ClientPlatform<'a> {
     pub(crate) client: &'a Client,
 }
 
 impl<'a> ClientPlatform<'a> {
+    /// Will generate a fingerprint based on the `input`. Given the same `input` This
+    /// method will result in the exact same output.
```
coroiu: suggestion (non-blocking): 
```suggestion
    /// method will always result in the exact same output.
```

crates/bitwarden-core/src/platform/client_platform.rs
```diff
@@ -5,15 +5,54 @@ use super::{
...
+    /// # Examples
+    /// ```rust
+    /// use bitwarden_core::{Client, platform::FingerprintRequest};
+    ///
+    /// async fn test() {
+    ///     let client = Client::test_account().await;
+    ///     let fingerprint_response = client.platform()
+    ///         .fingerprint(&FingerprintRequest {
+    ///             fingerprint_material: "my_material".to_owned(),
+    ///             public_key: "...public key...".to_owned(),
```
coroiu: question/suggestion: can these fields take any values, or do they expect the fields to be formatted in a certain way, e.g. base64? If yes, could we let the example clarify that somehow?

coroiu: Just realized that these docs are tested and so this must work 🙃

crates/bitwarden-core/src/platform/client_platform.rs
```diff
@@ -5,15 +5,54 @@ use super::{
...
+    ///         })
+    ///         .unwrap();
+    ///
+    ///     println!("{}", fingerprint_response.fingerprint);
```
coroiu: suggestion: give an example of the output
```suggestion
    ///     // Output: some-example
    ///     println!("{}", fingerprint_response.fingerprint);
```

crates/bitwarden-core/src/platform/client_platform.rs
```diff
@@ -5,15 +5,54 @@ use super::{
...
+    /// }
+    /// ```
     pub fn fingerprint(&self, input: &FingerprintRequest) -> Result<FingerprintResponse> {
         generate_fingerprint(input)
     }
 
+    /// Will generate a fingerprint based on the given `fingerprint_material`
+    /// and the users public key. Given the same `fingerprint_material` and
+    /// the same user. This method will result in the exact same output.
```
coroiu: typo/suggestion:
```suggestion
    /// the same user this method will always result in the exact same output.
```

crates/bitwarden-core/src/platform/client_platform.rs
```diff
@@ -5,15 +5,54 @@ use super::{
...
+    /// ```rust
+    /// use bitwarden_core::Client;
+    ///
+    /// async fn test() {
+    ///     let client = Client::test_account().await;
+    ///     let fingerprint = client.platform()
+    ///         .user_fingerprint("my_material".to_owned())
+    ///         .unwrap();
+    ///
+    ///     assert_eq!(fingerprint, "dreamland-desecrate-corrosive-ecard-retry");
```
coroiu, line 57: thought: huh, interesting, here we use `assert_eq` to give an example of the output, but in the doc above we use `println`. Should we choose one approach for consistency maybe? I like this one with `assert_eq`

crates/bitwarden-core/src/platform/client_platform.rs
dani-garcia: Well, the test runs the code at the top level, which means it sees the `test` function but doesn't actually call it.

If you want your test code to be called, it either needs to be outside of a function, or you need to call the function at the top level of the code block.

With sync functions you can just call them:
```rust
    /// Does a thing
    ///
    /// ```rust
    /// fn test() {
    ///     panic!("This test now fails")
    ///     Ok(())
    /// }
    /// # test();
    /// ```
```

With async functions for example you can do this:
```rust
    /// Does a thing
    ///
    /// ```rust
    /// async fn test() {
    ///     panic!("This test now fails")
    ///     Ok(())
    /// }
    /// # tokio_test::block_on(test());
    /// ```
```

Note the last line: `/// # tokio_test::block_on(test());` which runs the async test using `tokio_test`. It's at the top level of the code block, and not inside a function, so it gets run. The `#` also hides the line from the generated doc, while still running it during tests, which makes the docs nicer to look at:

image

Ps: the tokio-test crate is just doing `tokio::runtime::Builder::new_current_thread().enable_all().build().unwrap().block_on(test());` so we could make our own test utility function if we prefer.

crates/bitwarden-core/src/platform/client_platform.rs
dani-garcia: Ideally we'd just assert! as much as possible, that way we can have an explanatory code comment and test all in one. 

That might not be always acceptable, if the output of the function is too long we don't want the code sample to be dwarfed by a massive multi-line assert for example.

dani-garcia (commented):
It would be good to also have a doc at the top of `bitwarden-core/src/lib.rs` explaining a bit what the contents of the crate are:

```rust
//! This crate contains the core components that the rest of the crates use, 
//! the main entrypoint into this crate is [Client], etc, etc

#[cfg(feature = "uniffi")]
uniffi::setup_scaffolding!();
#[cfg(feature = "uniffi")]
mod uniffi_support;
```

Note that we're using the `//!` syntax for the comment, so the comment applies to the crate, as opposed to the next element on the file.

crates/bitwarden-core/src/platform/client_platform.rs
coroiu: I did _not_ notice that the code was inside of functions, good catch!

crates/bitwarden-core/src/platform/client_platform.rs
justindbaur: Yeah I also prefer assert but the one above didn't use it in the above doc because I couldn't come up with a way to concisely show off a public key that wouldn't be pretty ugly for a test. I could probably generate a new one in a simple one-liner but I still couldn't `assert_eq!` it because it would be different each time and `assert_ne(fingerprint, "")` doesn't seem like something we want to do.

crates/bitwarden-core/src/platform/client_platform.rs
justindbaur: Was able to rewrite it to actually work and always output the same fingerprint so I changed it to an assert.

crates/bitwarden-core/src/client/encryption_settings.rs
```diff
@@ -30,10 +30,20 @@ pub enum EncryptionSettingsError {
     MissingPrivateKey,
 }
 
+/// A struct containing the core keys of a Bitwarden user.
```
Hinton, line 33: **suggestion**: It's fairly self explanatory that this is a struct both from the code, but also in rust docs.

```suggestion
/// [EncryptionSettings] contains the encryption state related to a Bitwarden account.
```

crates/bitwarden-core/src/client/encryption_settings.rs
```diff
@@ -30,10 +30,20 @@ pub enum EncryptionSettingsError {
...
 #[derive(Clone)]
 pub struct EncryptionSettings {
+    /// The users symmetric key, stored on the server after being
+    /// encrypted with another key.
```
Hinton, line 37: Is these details relevant to the EncryptionSettings context? We could expand this to give the bigger picture in which case this information could be relevant. If not I think we should focus on the EncryptionSettings perspective which is not to share this anywhere.

```suggestion
    /// The user's primary encryption key.
    ///
    /// Every Bitwarden account has a symmetric user_key which acts as the encryption key to unlock everything else.
    /// .......
    /// This key is generally protected by a [MasterKey] derived from the users password, or .... and uploaded to the server.
    ///
    /// ## References:
    /// - https://bitwarden.com/help/account-encryption-key/
    /// <confluence link>?
```

Hinton (commented):
Some generic feedback on the documentation format.

PR #72, opened 2024-12-09
Update WASM to use MVP for transpilation

dani-garcia:
## 📔 Objective

Currently we're stuck on Rust 1.81 because 1.82 started enabling the `reference-types` feature by default, which breaks `wasm2js`. Ideally `wasm2js` would add support for it, but we don't know how long that is going to take.

To disable that feature, we need to use nightly or RUSTC_BOOTSTRAP to use the unstable `build-std` option. In this PR I'm building the WASM version normally first, and then with `build-std` for transpilation, though we could use `build-std` for both.

Local builds seem to work fine, but it'll need some more testing.

Note that webpack has support for `reference-types` in versions `>=5.97.0`, so we might want to make sure clients are updated first. Also the client support: https://caniuse.com/wasm-reference-types

## ⏰ Reminders before review

- CI builds passed
  team

## 🦮 Reviewer guidelines

### DIFF
--- a/.github/workflows/build-wasm-internal.yml
+++ b/.github/workflows/build-wasm-internal.yml
@@ -54,7 +54,7 @@ jobs:
       - name: Install rust
         uses: dtolnay/rust-toolchain@7b1c307e0dcbda6122208f10795a713336a9b35a # stable
         with:
-          toolchain: 1.81.0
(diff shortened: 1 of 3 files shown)


dani-garcia:
Replaced by #117

PR #82, opened 2024-12-16
[PM-10941] Add ssh import

quexten:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-10941

## 📔 Objective

Adds ssh import to the sdk so we can validate, decrypt, and convert keys on import on all clients, and unblock importing 1Password keys from pux exports.

This much cleans up the implementation of the `desktop_native` module, replacing a lot of mapping and matching code.

Further, this moves generator out of `bitwarden-ssh`'s `lib.rs`.

Note: importer will subsequently be removed from `desktop_native`

## ⏰ Reminders before review

- CI builds passed
  team

## 🦮 Reviewer guidelines


Hinton (commented):
In `bitwarden-exporters` we store test resources under `resources`. You probably want to copy the `excluded` instruction from the Cargo.toml to avoid the test resources being published to crates.io.

crates/bitwarden-ssh/Cargo.toml
```diff
@@ -22,14 +22,19 @@ wasm = [
...
+rsa = "0.9.7"
 serde.workspace = true
 ssh-key = { version = "0.6.7", features = [
     "ed25519",
     "encryption",
     "rsa",
     "getrandom",
-] }
+    "rand_core",
+    "std",
```
Hinton: Some of these features seems unecessary. `std` brings in several eliptic curve dependencies.

quexten: Removed the dep and the tests still run. Also moved the test keys to `resources` and excluded it.

crates/bitwarden-ssh/Cargo.toml
```diff
@@ -22,14 +23,17 @@ wasm = [
...
+ed25519 = { version = "2.2.3", features = ["pkcs8"] }
+pkcs8 = { version = "0.10.2", features = ["encryption"] }
 rand = "0.8.5"
+rsa = "0.9.7"
 serde.workspace = true
 ssh-key = { version = "0.6.7", features = [
     "ed25519",
     "encryption",
     "rsa",
     "getrandom",
```
Hinton: We can remove `getrandom`.

crates/bitwarden-ssh/Cargo.toml
```diff
@@ -22,14 +23,17 @@ wasm = [
...
-] }
+], default-features = false }
```
Hinton: Please use ranges here. This assists in preventing multiple versions of a single direct dependency from being dragged in. The ranges are based on existing ranges in other crates.

```toml
ed25519 = { version = ">=2.2.3, <3.0", features = ["pkcs8"] }
pkcs8 = { version = ">=0.10.2, <0.11", features = ["encryption"] }
rand = ">=0.8.5, <0.9"
rsa = ">=0.9.2, <0.10"
serde.workspace = true
ssh-key = { version = ">=0.6.7, <0.7", features = [
    "ed25519",
    "encryption",
    "rsa",
], default-features = false }
```

Hinton (commented):
There are some clippy errors that needs to be resolved. We can improve the documentation on most public functions and constants. Afterwards it should be easier to review the actual code.

crates/bitwarden-ssh/src/import.rs
```diff
@@ -0,0 +1,279 @@
...
+
+    let key_type: KeyType = match PrivateKeyInfo::from_der(decrypted_der.as_bytes())
+        .map_err(|_| SshKeyImportError::ParsingError)?
+        .algorithm
+        .oid
+    {
+        ed25519::pkcs8::ALGORITHM_OID => KeyType::Ed25519,
+        RSA_PKCS8_ALGORITHM_OID => KeyType::Rsa,
+        _ => KeyType::Unknown,
+    };
```
Hinton: **issue:** We seem to never expose `KeyType`? If so I suggest removing it and using `ObjectIdentifier` in the match case below. We can also benefit from having access to the `PrivateKeyInfo` directly.

crates/bitwarden-ssh/src/import.rs
```diff
@@ -0,0 +1,279 @@
...
+                &private_key.secret_key.into(),
+            ));
+            let private_key_openssh = private_key
+                .to_openssh(LineEnding::LF)
+                .map_err(|_| SshKeyImportError::ParsingError)?;
+            Ok(SshKey {
+                private_key: private_key_openssh.to_string(),
+                public_key: private_key.public_key().to_string(),
+                key_fingerprint: private_key.fingerprint(HashAlg::Sha256).to_string(),
+            })
```
Hinton: **suggestion:** This is re-used between all key types. We should extract it, possibly using `TryFrom`.

```rust
impl TryFrom<PrivateKey> for SshKey {
    type Error = SshKeyImportError;

    fn try_from(value: PrivateKey) -> Result<Self, Self::Error> {
        let private_key_openssh = value
            .to_openssh(LineEnding::LF)
            .map_err(|_| SshKeyImportError::ParsingError)?;

        Ok(SshKey {
            private_key: private_key_openssh.to_string(),
            public_key: value.public_key().to_string(),
            key_fingerprint: value.fingerprint(HashAlg::Sha256).to_string(),
        })
    }
}
```

```suggestion
            private_key.try_into()
```

crates/bitwarden-ssh/src/import.rs
```diff
@@ -0,0 +1,279 @@
...
+        }
+        _ => Err(SshKeyImportError::UnsupportedKeyType),
+    }
```
Hinton: **suggestion:** You can return the private_key from the match statement since that code is duplicated and do the final conversion outside.

```rust
let private_key = match ....

let private_key_openssh = private_key
    .to_openssh(LineEnding::LF)
    .map_err(|_| SshKeyImportError::ParsingError)?;
Ok(SshKey {
    private_key: private_key_openssh.to_string(),
    public_key: private_key.public_key().to_string(),
    key_fingerprint: private_key.fingerprint(HashAlg::Sha256).to_string(),
})
```

Of after moving the conversion to a `TryFrom`.

```
let private_key = match ....

private_key.try_into()
```

crates/bitwarden-ssh/src/import.rs
```diff
@@ -0,0 +1,279 @@
...
+                            }
+                            _ => SshKeyImportError::ParsingError,
+                        })?
+                }
+                None => ed25519::pkcs8::DecodePrivateKey::from_pkcs8_pem(&encoded_key)
+                    .map_err(|_| SshKeyImportError::ParsingError)?,
+            };
+            let private_key = ssh_key::private::PrivateKey::from(Ed25519Keypair::from(
+                &private_key.secret_key.into(),
+            ));
```
Hinton: **suggestion:** We've actually already decoded the private key to retrieve the algorithm. We can use https://docs.rs/ed25519/2.2.3/ed25519/pkcs8/struct.KeypairBytes.html#impl-TryFrom%3CPrivateKeyInfo%3C'_%3E%3E-for-KeypairBytes here.

 ```suggestion
            let private_key: ed25519::KeypairBytes = pki
                .try_into()
                .map_err(|_| SshKeyImportError::ParsingError)?;

            ssh_key::private::PrivateKey::from(Ed25519Keypair::from(&private_key.secret_key.into()))
```

File: crates/bitwarden-ssh/src/import.rs [hunk 8]
```diff
@@ -0,0 +1,279 @@
...
+                            _ => SshKeyImportError::ParsingError,
+                        })?
+                }
+                None => pkcs8::DecodePrivateKey::from_pkcs8_pem(&encoded_key)
+                    .map_err(|_| SshKeyImportError::ParsingError)?,
+            };
+
+            let private_key = ssh_key::private::PrivateKey::from(
+                RsaKeypair::try_from(private_key).map_err(|_| SshKeyImportError::ParsingError)?,
+            );
```
Hinton: **suggestion:** Similar here, since we already have the PKI we can use that to get the `RsaKeypair`.

```suggestion
            let private_key: rsa::RsaPrivateKey = pki
                .try_into()
                .map_err(|_| SshKeyImportError::ParsingError)?;

            ssh_key::private::PrivateKey::from(
                RsaKeypair::try_from(private_key).map_err(|_| SshKeyImportError::ParsingError)?,
            )
```

crates/bitwarden-ssh/src/import.rs
```diff
@@ -0,0 +1,279 @@
...
+pub fn import_key(encoded_key: String, password: String) -> Result<SshKey, SshKeyImportError> {
+    match encoded_key.lines().next() {
+        Some(PKCS1_HEADER) => Err(SshKeyImportError::UnsupportedKeyType),
+        Some(PKCS8_UNENCRYPTED_HEADER) => {
+            import_pkcs8_key(encoded_key, None).map_err(|_| SshKeyImportError::ParsingError)
+        }
+        Some(PKCS8_ENCRYPTED_HEADER) => import_pkcs8_key(encoded_key, Some(password)),
+        Some(OPENSSH_HEADER) => import_openssh_key(encoded_key, password),
+        _ => Err(SshKeyImportError::ParsingError),
+    }
```
Hinton: **thought:** Rather than defining our own constants what do you think about using https://docs.rs/pem-rfc7468/latest/pem_rfc7468/fn.decode_label.html here? The package is already a transient dependency, and this should allow us to use the `PEM_LABEL` constants from the various packages.

https://docs.rs/pkcs8/latest/pkcs8/struct.EncryptedPrivateKeyInfo.html#impl-PemLabel-for-EncryptedPrivateKeyInfo%3C'_%3E

crates/bitwarden-ssh/src/import.rs
```diff
@@ -0,0 +1,279 @@
...
+
+use crate::{error::SshKeyImportError, SshKey};
+
+const PKCS1_HEADER: &str = "-----BEGIN RSA PRIVATE KEY-----";
+const PKCS8_UNENCRYPTED_HEADER: &str = "-----BEGIN PRIVATE KEY-----";
+const PKCS8_ENCRYPTED_HEADER: &str = "-----BEGIN ENCRYPTED PRIVATE KEY-----";
+const OPENSSH_HEADER: &str = "-----BEGIN OPENSSH PRIVATE KEY-----";
+
+pub const RSA_PKCS8_ALGORITHM_OID: ObjectIdentifier =
+    ObjectIdentifier::new_unwrap("1.2.840.113549.1.1.1");
```
Hinton: **issue:** This constant is available through the pkcs1 package which is aliased in rsa, `rsa::pkcs1::ALGORITHM_OID`.

crates/bitwarden-ssh/src/import.rs
```diff
@@ -0,0 +1,279 @@
...
+        Some(OPENSSH_HEADER) => import_openssh_key(encoded_key, password),
+        _ => Err(SshKeyImportError::ParsingError),
+    }
+}
+
+fn import_pkcs8_key(
+    encoded_key: String,
+    password: Option<String>,
+) -> Result<SshKey, SshKeyImportError> {
+    let der = match SecretDocument::from_pem(&encoded_key) {
```
Hinton: **issue:** This variable is misleading. This is a `SecretDocument`.

Hinton: **issue:** Please use `map_err` here. We generally discourage using match unless you need to handle the separate cases.

```suggestion
    let (_, secret_doc) = SecretDocument::from_pem(&encoded_key).map_err ....
```

crates/bitwarden-ssh/src/import.rs
```diff
@@ -0,0 +1,279 @@
...
+        Err(_) => return Err(SshKeyImportError::ParsingError),
+    };
+
+    let decrypted_der = match password.clone() {
+        Some(password) => {
+            let encrypted_private_key_info = EncryptedPrivateKeyInfo::from_der(der.as_bytes())
+                .map_err(|_| SshKeyImportError::ParsingError)?;
+            encrypted_private_key_info
+                .decrypt(password.as_bytes())
+                .map_err(|_| SshKeyImportError::WrongPassword)?
```
Hinton: **thought:** `SecretDoc` defines a helper, `from_pkcs8_encrypted_pem` which could be used. `EncryptedPrivateKeyInfo` is somewhat lower level in my opinion?

You would need to map the error differently though since you get the generic pkcs8 error but it should expose a `DecryptFailed` case.

We can go a step further and lift the `SecretDocument::from_pem` into the else statement since we would be decoding it from scratch and not need the `der` representation.

crates/bitwarden-ssh/src/import.rs
```diff
@@ -0,0 +1,279 @@
...
+    let private_key = match ssh_key::private::PrivateKey::from_openssh(&encoded_key) {
+        Ok(k) => k,
+        Err(err) => match err {
+            ssh_key::Error::AlgorithmUnknown
+            | ssh_key::Error::AlgorithmUnsupported { algorithm: _ } => {
+                return Err(SshKeyImportError::UnsupportedKeyType)
+            }
+            _ => return Err(SshKeyImportError::ParsingError),
+        },
+    };
```
Hinton: **issue:** Please use `map_err´ here.

crates/bitwarden-ssh/src/import.rs
```diff
@@ -0,0 +1,279 @@
...
+        _ => Err(SshKeyImportError::UnsupportedKeyType),
+    }
+}
+
+fn import_openssh_key(encoded_key: String, password: String) -> Result<SshKey, SshKeyImportError> {
+    let private_key = match ssh_key::private::PrivateKey::from_openssh(&encoded_key) {
+        Ok(k) => k,
+        Err(err) => match err {
+            ssh_key::Error::AlgorithmUnknown
+            | ssh_key::Error::AlgorithmUnsupported { algorithm: _ } => {
```
Hinton: **issue:** If arguments are unused we prefer to use `..` since it's less fragile if the enum pair changes.
```suggestion
            | ssh_key::Error::AlgorithmUnsupported { .. } => {
```

crates/bitwarden-ssh/src/import.rs
```diff
@@ -0,0 +1,279 @@
...
+    };
+    let private_key_openssh = private_key
+        .to_openssh(LineEnding::LF)
+        .map_err(|_| SshKeyImportError::ParsingError)?;
+
+    Ok(SshKey {
+        private_key: private_key_openssh.to_string(),
+        public_key: private_key.public_key().to_string(),
+        key_fingerprint: private_key.fingerprint(HashAlg::Sha256).to_string(),
+    })
```
Hinton: **issue:** This is reused in `import_pkcs8_key`. We can replace this with `private_key.try_into()`.

Hinton (changes requested):
Alright, this is a bit of a long one 😅. Generally function wise it seems fine, but we have a few code cleanups we can do along the way to reduce code duplication, rely on slightly less lower level primitives in certain crates and better convey what's happening.

I've left a bunch of `suggestions` and `issues` along the way which should be fairly non controversial. But there are also a few `thoughts` which are more discussion topics.

crates/bitwarden-ssh/src/import.rs
```diff
@@ -0,0 +1,279 @@
...
+
+#[derive(Serialize, Deserialize, PartialEq, Debug)]
+#[cfg_attr(feature = "wasm", derive(Tsify), tsify(into_wasm_abi, from_wasm_abi))]
+enum KeyType {
+    Ed25519,
+    Rsa,
+    Unknown,
+}
+
+pub fn import_key(encoded_key: String, password: String) -> Result<SshKey, SshKeyImportError> {
```
Hinton: **question:** Should password be an `Option`? Relying on empty string is a bit less conventional.

crates/bitwarden-ssh/src/import.rs
quexten: Nice, this is much cleaner!

crates/bitwarden-ssh/src/import.rs
quexten:
```rust
    let decrypted_der = if let Some(password) = password {
        SecretDocument::from_pkcs8_encrypted_pem(&encoded_key, password.as_bytes()).map_err(
            |err| match err {
                pkcs8::Error::EncryptedPrivateKey(pkcs5::Error::DecryptFailed) => {
                    SshKeyImportError::WrongPassword
                }
                _ => SshKeyImportError::ParsingError,
            },
        )?
    } else {
        SecretDocument::from_pkcs8_pem(&encoded_key).map_err(|_| SshKeyImportError::ParsingError)?
    };
```

Like this?

Hinton: Yes, that seems nice. We could probably even rename it to `der` now that we don't have a regular `der`.

crates/bitwarden-ssh/src/error.rs
```diff
@@ -9,3 +9,16 @@ pub enum KeyGenerationError {
...
+pub enum SshKeyImportError {
+    #[error("Failed to parse key")]
+    ParsingError,
+    #[error("Password required")]
+    PasswordRequired,
+    #[error("Wrong password")]
+    WrongPassword,
+    #[error("Unsupported key type")]
+    UnsupportedKeyType,
+}
```
audreyality, line 24: ❓ @Hinton - Is there a path forward for i18n in the SDK? These seem like the kind of errors we'd like to forward to the user, which would mean communicating the i18n key we'd like the UI to use for localization along with the error.

crates/bitwarden-ssh/src/generator.rs
```diff
@@ -0,0 +1,88 @@
...
+    key_algorithm: KeyAlgorithm,
+    mut rng: impl CryptoRngCore,
+) -> Result<SshKey, error::KeyGenerationError> {
+    let key = match key_algorithm {
+        KeyAlgorithm::Ed25519 => ssh_key::PrivateKey::random(&mut rng, Algorithm::Ed25519),
+        KeyAlgorithm::Rsa3072 | KeyAlgorithm::Rsa4096 => {
+            let bits = match key_algorithm {
+                KeyAlgorithm::Rsa3072 => 3072,
+                KeyAlgorithm::Rsa4096 => 4096,
+                _ => unreachable!(),
```
audreyality: 🎨 Consider explaining what makes the `_` arm of the pattern match unreachable.

```
_ => unreachable!("Bitwarden does not support RSA keys with less than 3072 bits")
```

crates/bitwarden-ssh/src/generator.rs
```diff
@@ -0,0 +1,88 @@
...
+            };
+
+            let rsa_keypair = ssh_key::private::RsaKeypair::random(&mut rng, bits)
+                .map_err(KeyGenerationError::KeyGenerationError)?;
+
+            let private_key =
+                ssh_key::PrivateKey::new(ssh_key::private::KeypairData::from(rsa_keypair), "")
+                    .map_err(KeyGenerationError::KeyGenerationError)?;
+            Ok(private_key)
```
audreyality: 🎨 Seems like the duplication in the match arms could be eliminated by making these lines into a function.

```rust
// add the appropriate types here
fn create_rsa_key(&mut rng, bits) {
   let rsa_keypair = ssh_key::private::RsaKeypair::random(&mut rng, bits)
       .map_err(KeyGenerationError::KeyGenerationError)?;
    let private_key = ssh_key::PrivateKey::new(ssh_key::private::KeypairData::from(rsa_keypair), "")
        .map_err(KeyGenerationError::KeyGenerationError)?;
    Ok(private_key)
}
```

```rust
let key = match key_algorithm {
    KeyAlgorithm::Ed25519 => ssh_key::PrivateKey::random(&mut rng, Algorithm::Ed25519),
    KeyAlgorithm::Rsa3072 => create_rsa_key(3072),
    KeyAlgorithm::Rsa4096 => create_rsa_key(4096),
}
```

🎨 This could be simplified further by adding the key length as an algorithm parameter, though that might not be appropriate since only specific key lengths are allowed.

```rust
pub enum KeyAlgorithm {
    Ed25519,
    Rsa(usize),
}
```

crates/bitwarden-ssh/src/import.rs
```diff
@@ -0,0 +1,214 @@
...
+
+    match label {
+        PKCS1_LABEL => Err(SshKeyImportError::UnsupportedKeyType),
+        PKCS8_UNENCRYPTED_LABEL => import_pkcs8_key(encoded_key, None),
+        PKCS8_ENCRYPTED_LABEL => import_pkcs8_key(
+            encoded_key,
+            Some(password.ok_or(SshKeyImportError::PasswordRequired)?),
+        ),
+        OPENSSH_LABEL => import_openssh_key(encoded_key, password),
+        _ => Err(SshKeyImportError::ParsingError),
```
audreyality: ⚠️ Make sure you cover your error cases in your tests!

crates/bitwarden-ssh/src/import.rs
```diff
@@ -0,0 +1,214 @@
...
+use ssh_key::private::{Ed25519Keypair, RsaKeypair};
+
+use crate::{error::SshKeyImportError, SshKey};
+
+const PKCS1_LABEL: &str = "RSA PRIVATE KEY";
+const PKCS8_UNENCRYPTED_LABEL: &str = "PRIVATE KEY";
+const PKCS8_ENCRYPTED_LABEL: &str = "ENCRYPTED PRIVATE KEY";
+const OPENSSH_LABEL: &str = "OPENSSH PRIVATE KEY";
+
+pub fn import_key(
```
audreyality, line 16: ⛏️ I'd like to second Oscar's feedback about documentation. If something is `pub`, it should be documented.

🎨 Fun fact! [Tests provided in rust's documentation are executable!](https://doc.rust-lang.org/rustdoc/write-documentation/documentation-tests.html). This is of particular use when describing error cases that you expect the consuming code to handle.

crates/bitwarden-ssh/src/import.rs
```diff
@@ -0,0 +1,214 @@
...
+        PKCS8_ENCRYPTED_LABEL => import_pkcs8_key(
+            encoded_key,
+            Some(password.ok_or(SshKeyImportError::PasswordRequired)?),
+        ),
+        OPENSSH_LABEL => import_openssh_key(encoded_key, password),
+        _ => Err(SshKeyImportError::ParsingError),
+    }
+}
+
+fn import_pkcs8_key(
```
audreyality, line 34: 🤔 There's a lot of repetitive `map_err` calls in this method, which leads me to believe that there's a better way to represent this logic. 

For example, extracting the parsing out from the processing of the key would let the parser signal a wide number of error conditions. The importer can then map those parsing failures once as a failed parse.

crates/bitwarden-ssh/src/import.rs
```diff
@@ -0,0 +1,214 @@
...
+
+    let private_key_info =
+        PrivateKeyInfo::from_der(der.as_bytes()).map_err(|_| SshKeyImportError::ParsingError)?;
+
+    let private_key = match private_key_info.algorithm.oid {
+        ed25519::pkcs8::ALGORITHM_OID => {
+            let private_key: ed25519::KeypairBytes = private_key_info
+                .try_into()
+                .map_err(|_| SshKeyImportError::ParsingError)?;
+            ssh_key::private::PrivateKey::from(Ed25519Keypair::from(&private_key.secret_key.into()))
```
audreyality, line 60: 🤔 Three type conversions on a single line leads me to believe that there should be a more direct way to convert key formats. It looks like all of these types come from a library? If so, consider wrapping the logic in a [newtype to implement an external trait on an external type](https://doc.rust-lang.org/book/ch19-03-advanced-traits.html#using-the-newtype-pattern-to-implement-external-traits-on-external-types).

audreyality (changes requested):
Most of my feedback is advice that you can take or leave. Issues that block approval are marked with ⚠️.

crates/bitwarden-ssh/src/import.rs
quexten: I think this one specifically was covered by `import_non_key_error`, now that we switched to using pem_rfc7468 we need a new test though.

quexten: Added a test for this case

audreyality (dismissed):
My blocking concerns have been addressed. I'm leaving the non-blocking concerns and questions unresolved.

crates/bitwarden-ssh/src/import.rs
audreyality: `SecretDocument::from_pkcs8_encrypted_pem` and `SecretDocument::from_pkcs8_pem` have the same result type. That allows you to move the error handling out of the nested code block.

Something like:
```rust
let der = if let Some(password) = password {
    SecretDocument::from_pkcs8_encrypted_pem(&encoded_key, password.as_bytes())
} else {
    SecretDocument::from_pkcs8_pem(&encoded_key)
};

let der = match.map_err(|err| match err {
    pkcs8::Error::EncryptedPrivateKey(pkcs5::Error::DecryptFailed) =>
        SshKeyImportError::WrongPassword,
    _ => SshKeyImportError::ParsingError
})?;
```

It works as a one-liner if you want it to, but that feels a bit more cluttered. Splitting out the error handling from the control flow makes the purpose of each block clear and makes the error propagation stand out a bit more.

Rebinding the variable (`let der`) might seem a bit odd, but it's a common idiom when you're changing a type but not the underlying data. In this case, the first `der` is a `Result<T>` and it's unwrapped to a `T`, which fits the idiom exactly.

crates/bitwarden-ssh/src/error.rs
Hinton: No we leave that to the presentation layer for now. Using i18n in lower level services have always been sort of a crutch.

For WASM we support https://sdk-api-docs.bitwarden.com/bitwarden_error_macro/attr.bitwarden_error.html which exposes basic, flat and full. For non user facing errors it's safe to use `basic` or `flat` and log it. For user facing errors we generally encourage you to use `flat` or `full`, and properly handle the error in the presentation layer.

In this case if we want to show error messages to the user we should use `flat` or `full` and capture the different variants in the JS world and apply localization there.

Note: The story on mobile is slightly different due to some uniffi limitations. We're working on providing the same capabilities there.

crates/bitwarden-ssh/src/import.rs
```diff
@@ -0,0 +1,235 @@
...
+        .map_err(|_| SshKeyImportError::ParsingError)?;
+
+    match label {
+        PKCS1_LABEL => Err(SshKeyImportError::UnsupportedKeyType),
+        PKCS8_UNENCRYPTED_LABEL => import_pkcs8_key(encoded_key, None),
+        PKCS8_ENCRYPTED_LABEL => import_pkcs8_key(
+            encoded_key,
+            Some(password.ok_or(SshKeyImportError::PasswordRequired)?),
+        ),
+        OPENSSH_LABEL => import_openssh_key(encoded_key, password),
```
Hinton: **suggestion:** These are already defined in the packages.

```suggestion
        PKCS1_LABEL => Err(SshKeyImportError::UnsupportedKeyType),
        pkcs8::PrivateKeyInfo::PEM_LABEL => import_pkcs8_key(encoded_key, None),
        pkcs8::EncryptedPrivateKeyInfo::PEM_LABEL => import_pkcs8_key(
            encoded_key,
            Some(password.ok_or(SshKeyImportError::PasswordRequired)?),
        ),
        ssh_key::PrivateKey::PEM_LABEL => import_openssh_key(encoded_key, password),
```

crates/bitwarden-ssh/src/import.rs
```diff
@@ -0,0 +1,235 @@
...
+        _ => Err(SshKeyImportError::ParsingError),
```
Hinton: **question**: If we don't identify the label this should probably be a `UnsupportedKeyType` error? If so we can remove the initial case and return the generic error here.

```suggestion
        _ => Err(SshKeyImportError::UnsupportedKeyType),
```

crates/bitwarden-ssh/src/import.rs
```diff
@@ -0,0 +1,235 @@
...
+    }
+}
+
+fn import_pkcs8_key(
+    encoded_key: String,
+    password: Option<String>,
+) -> Result<SshKey, SshKeyImportError> {
+    let der = if let Some(password) = password {
```
Hinton: I think with the refactors this got lost, but since this is a `SecretDocument` the variable name should be updated to reflect it.

```suggestion
    let doc = if let Some(password) = password {
```

crates/bitwarden-ssh/src/import.rs
Hinton: Agree, that does look cleaner.

crates/bitwarden-ssh/src/import.rs
```diff
@@ -0,0 +1,235 @@
...
+    if private_key.is_encrypted() {
+        if let Some(password) = password {
+            private_key
+                .decrypt(password.as_bytes())
+                .map_err(|_| SshKeyImportError::WrongPassword)?
+                .try_into()
+                .map_err(|_| SshKeyImportError::ParsingError)
+        } else {
+            Err(SshKeyImportError::PasswordRequired)
+        }
```
Hinton, line 105: **suggestion**: Since we're just returning an error here if password doesn't exist we can use `ok_or` which maps it into a result.

```suggestion
        let password = password.ok_or(SshKeyImportError::PasswordRequired)?;
        private_key
            .decrypt(password.as_bytes())
            .map_err(|_| SshKeyImportError::WrongPassword)?
            .try_into()
            .map_err(|_| SshKeyImportError::ParsingError)
```

crates/bitwarden-ssh/src/import.rs
```diff
@@ -0,0 +1,235 @@
...
+    } else {
+        private_key
+            .try_into()
+            .map_err(|_| SshKeyImportError::ParsingError)
+    }
```
Hinton: **suggestion**: We can improve this code slightly.

- The if condition for password can be rewritten using `ok_or` which converts an `Option` to `Result`.
- We can remove some code duplication by assigning a temporary variable and lifting out some code from the if block.

```suggestion
    let private_key = if private_key.is_encrypted() {
        let password = password.ok_or(SshKeyImportError::PasswordRequired)?;
        private_key
            .decrypt(password.as_bytes())
            .map_err(|_| SshKeyImportError::WrongPassword)?
    } else {
        private_key
    }

    private_key
        .try_into()
        .map_err(|_| SshKeyImportError::ParsingError)
```

crates/bitwarden-wasm-internal/src/ssh.rs
Hinton, line 1: We might want to provide some documentation here as well, since this will be available in TS.

Hinton (commented):
Thanks for iterating on this! Looks much better, some minor suggestions and we should be good.

crates/bitwarden-ssh/src/import.rs
```diff
@@ -0,0 +1,229 @@
...
+/// - `Error::ParsingError` if the key is otherwise malformed and cannot be parsed
+pub fn import_key(
+    encoded_key: String,
+    password: Option<String>,
+) -> Result<SshKey, SshKeyImportError> {
+    let label = pem_rfc7468::decode_label(encoded_key.as_bytes())
+        .map_err(|_| SshKeyImportError::ParsingError)?;
+
+    match label {
+        pkcs1::RsaPrivateKey::PEM_LABEL => Err(SshKeyImportError::UnsupportedKeyType),
```
Hinton: **question:** Since this is the same as the default match should we remove this?

crates/bitwarden-ssh/src/import.rs
```diff
@@ -0,0 +1,229 @@
+use ed25519;
+use pem_rfc7468::PemLabel;
+use pkcs8::{der::Decode, pkcs5, DecodePrivateKey, PrivateKeyInfo, SecretDocument};
+use rsa::pkcs1;
+use ssh_key::private::{Ed25519Keypair, RsaKeypair};
+
+use crate::{error::SshKeyImportError, SshKey};
+
+/// Import a PKCS8 or OpenSSH encoded private key, and returns a decoded SshKey,
```
Hinton: **nit:** Might want to use a link here.
```suggestion
/// Import a PKCS8 or OpenSSH encoded private key, and returns a decoded [SshKey],
```

crates/bitwarden-ssh/src/import.rs
```diff
@@ -0,0 +1,229 @@
...
+/// with the public key and fingerprint, and the private key in OpenSSH format.
+/// A password can be provided for encrypted keys.
+/// # Returns
+/// - `Error::PasswordRequired` if the key is encrypted and no password is provided
+/// - `Error::WrongPassword` if the password provided is incorrect
+/// - `Error::UnsupportedKeyType` if the key type is not supported
+/// - `Error::ParsingError` if the key is otherwise malformed and cannot be parsed
```
Hinton: **nit:** Should probably be links?

crates/bitwarden-ssh/src/generator.rs
```diff
@@ -0,0 +1,87 @@
...
+#[cfg_attr(feature = "wasm", derive(Tsify), tsify(into_wasm_abi, from_wasm_abi))]
+pub enum KeyAlgorithm {
+    Ed25519,
+    Rsa3072,
+    Rsa4096,
+}
+
+/**
+ * Generate a new SSH key pair, for the provided key algorithm, returning
+ * an `SshKey` struct containing the private key, public key, and key fingerprint,
```
Hinton: **nit:** Link?
```suggestion
 * an [SshKey] struct containing the private key, public key, and key fingerprint,
```

Hinton (dismissed):

crates/bitwarden-ssh/src/generator.rs
quexten: Added, though clicking them just brings me to a not found page: https://docs.rs/bitwarden_ssh/1.0.0/bitwarden_ssh/struct.SshKey.html

Hinton (approved):

crates/bitwarden-ssh/src/generator.rs
Hinton: It should work when you generate docs locally or use rust analyzer in vscode.

crates/bitwarden-ssh/src/error.rs
audreyality: Thanks! I'm going to leave this item open for now, but it's not blocking the review.

PR #89, opened 2024-12-19
[PM-16221] uniffi bindings for ssh agent and keygen

quexten:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-16221

## 📔 Objective

Adds bindings so we can add keygen and import to the mobile clients.

## ⏰ Reminders before review

- CI builds passed
  team

## 🦮 Reviewer guidelines

### DIFF
--- a/crates/bitwarden-uniffi/src/lib.rs
+++ b/crates/bitwarden-uniffi/src/lib.rs
@@ -19,7 +19,7 @@ mod android_support;
 use crypto::CryptoClient;
 use error::Result;
 use platform::PlatformClient;
-use tool::{ExporterClient, GeneratorClients, SendClient};
+use tool::{ExporterClient, GeneratorClients, SendClient, SshClient};
 use vault::VaultClient;
 
 #[derive(uniffi::Object)]
@@ -67,6 +67,11 @@ impl Client {
         Arc::new(SendClient(self))
     }
 
+    /// SSH operations
+    pub fn ssh(self: Arc<Self>) -> Arc<SshClient> {
+        Arc::new(SshClient(self))
+    }
+
     /// Auth operations
     pub fn auth(self: Arc<Self>) -> Arc<AuthClient> {
         Arc::new(AuthClient(self))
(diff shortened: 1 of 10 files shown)


quexten:
Question for whoever reviews: On both the wasm bindings and here, I put the ssh methods on the global "namespace" directly, since things like cryptoclient seemed to be for holding state. They can also act as namespaces though. Should they be moved under something like "sshclient"?

crates/bitwarden-uniffi/src/lib.rs
```diff
@@ -88,6 +88,23 @@ impl Client {
...
+    }
+
+    pub fn import_ssh_key(
+        &self,
+        imported_key: String,
+        password: Option<String>,
+    ) -> Result<bitwarden_ssh::SshKey> {
+        bitwarden_ssh::import::import_key(imported_key, password)
+            .map_err(|e| error::BitwardenError::E(error::Error::SshImport(e)))
+    }
```
Hinton: **issue:** Please make a client for these, we don't want them in the root.

quexten: Updated and moved to a client, should we do the same for wasm in a follow-up pr?

crates/bitwarden-uniffi/src/tool/ssh.rs
```diff
@@ -0,0 +1,29 @@
...
+#[derive(uniffi::Object)]
+pub struct SshClient(pub Arc<Client>);
+
+#[uniffi::export]
+impl SshClient {
+    pub fn generate_ssh_key(
+        &self,
+        key_algorithm: bitwarden_ssh::generator::KeyAlgorithm,
+    ) -> Result<bitwarden_ssh::SshKey> {
+        bitwarden_ssh::generator::generate_sshkey(key_algorithm)
```
Hinton, line 17: **observation**: We would probably benefit from taking a look at the public interface of `bitwarden_ssh`. Exposing modules is a bit clunky and the library is small enough that re-publishing everything in the root would result in a cleaner interface.

Hinton (approved):
LGTM. Nit for a potential follow up.

PR #127, opened 2025-01-22
[PM-17370] Refactor symmetric keys to use enum

quexten:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-17370

## 📔 Objective

Symmetric crypto keys (in general) are just random arrays of data - in contrast to asymmetric keys which require specific structure. Still, it is cryptographic best practice to split these keys apart entirely, and not have them confused. An algorithm switch should include a change to a new key type.

For the upcoming xchacha cipher, we will introduce a symmetric key that is the same size as the old hmac-less cbc type 0 cipher. In the sdk we currently just represent this as one key with two byte arrays, "key" and "mac key" which is confusing and could lead to mis-usage, which can lead to bugs or even cryptographic flaws.

For this reason, it is paramount to enforce the key types in the type system. This PR refactors the current key usage to represent symmetric keys as an enum, with the two types being the current encstring type 0 and type 2, each of which has it's own struct with the data it needs.

In a later PR, the serialization format for symmetric keys will be adjusted to include the cipher type, so keys are not categorized based on the size alone.

## ⏰ Reminders before review

- CI builds passed
  team

## 🦮 Reviewer guidelines


crates/bitwarden-crypto/src/keys/symmetric_crypto_key.rs
```diff
@@ -9,22 +9,57 @@ use zeroize::Zeroize;
 use super::key_encryptable::CryptoKey;
 use crate::CryptoError;
 
+#[cfg_attr(test, derive(Debug))]
+pub(crate) struct KdfDerviedKeymaterial {
```
dani-garcia: Typo:
```suggestion
pub(crate) struct KdfDerivedKeyMaterial {
```

dani-garcia (commented):
I don't have a lot of time to review right now so I just gave it a quick look and just noticed a typo. Also I feel like `KdfDerivedKeyMaterial` should probably be moved somewhere else as it's fairly unrelated to `SymmetricCryptoKey`

crates/bitwarden-crypto/src/keys/symmetric_crypto_key.rs
quexten: Nice, fixed! Also moved the KdfDerivedKeyMaterial to `master_key.rs` since it seems more related to that.

crates/bitwarden-core/src/mobile/crypto.rs
```diff
@@ -166,7 +166,10 @@ pub async fn initialize_user_crypto(
             master_key,
             user_key,
         } => {
-            let master_key = MasterKey::new(SymmetricCryptoKey::try_from(master_key)?);
+            let mut master_key_bytes = STANDARD
+                .decode(master_key)
+                .map_err(|_| CryptoError::InvalidKey)?;
+            let master_key = MasterKey::try_from(master_key_bytes.as_mut_slice())?;
```
quexten, line 183: Do we need to additionally zeroize the master_key b64 encoded string here?

crates/bitwarden-crypto/src/keys/master_key.rs
```diff
@@ -118,39 +124,44 @@ impl MasterKey {
     }
 
     pub fn to_base64(&self) -> String {
-        self.0.to_base64()
+        STANDARD.encode(self.0.key_material.to_vec())
```
quexten: I think to_vec creates an additional copy on the heap, do we need to zeroize explicitly?

crates/bitwarden-crypto/src/keys/master_key.rs
```diff
@@ -118,39 +124,44 @@ impl MasterKey {
...
-
-    EncString::encrypt_aes256_hmac(
-        &user_key.to_vec(),
-        stretched_key
-            .mac_key
-            .as_ref()
-            .ok_or(CryptoError::InvalidMac)?,
-        &stretched_key.key,
-    )
+    EncString::encrypt_aes256_hmac(&user_key.to_vec(), &stretched_key)
```
quexten: To_vec was used here before my change, does this need to be zeroized?

crates/bitwarden-crypto/src/keys/symmetric_crypto_key.rs
```diff
@@ -153,8 +197,17 @@ mod tests {
...
+        match (&key, &key2) {
+            (SymmetricCryptoKey::Aes256CbcKey(k), SymmetricCryptoKey::Aes256CbcKey(k2)) => {
+                assert_eq!(k.encryption_key.to_vec(), k2.encryption_key.to_vec());
+            }
+            (SymmetricCryptoKey::Aes256CbcHmacKey(k), SymmetricCryptoKey::Aes256CbcHmacKey(k2)) => {
+                assert_eq!(k.encryption_key.to_vec(), k2.encryption_key.to_vec());
+                assert_eq!(k.mac_key.to_vec(), k2.mac_key.to_vec());
+            }
+            _ => panic!("Key types don't match"),
+        }
```
mzieniukbw: can't you just compare using `==`, having the `PartialEq` already implemented for `SymmetricCryptoKey` ?

quexten: Leftover, from before implementing the PartialEq. Cleaned it up now!

crates/bitwarden-crypto/src/keys/master_key.rs
quexten: @dani-garcia Does the zeroizing allocator we use in the SDK take care of this automatically, or do we manually have to call zeroize on it here (and in the other comments)? (I'm not sure if drop corresponds to deallocating always)

crates/bitwarden-crypto/src/keys/master_key.rs
quexten: Can be replaced by `as_slice`

crates/bitwarden-core/src/mobile/crypto.rs
quexten: Masterkey::try_from already zeroizes the parameter

crates/bitwarden-crypto/src/keys/master_key.rs
quexten: Added zeroizing now.

dani-garcia: As a general rule, anything that is heap allocated will be zeroized on drop by our custom allocator, so `Strings`, `Vecs`, etc, will be handled automatically for you. Even though we don't support it, users can technically disable this custom allocator with the `no-memory-hardening` feature in case they need a different one or it produces incompatibilities, so it doesn't hurt to explicitly call zeroize, but in this case it shouldn't be necessary.

The one case where we always need to zeroize manually is for stack allocated arrays, for example if you're using `[u8; 32]` or `GenericArray<>`. In those cases you either need to wrap them in a `Box` to move them to the heap or ensure you zeroize it every time.

crates/bitwarden-crypto/src/enc_string/symmetric.rs
```diff
@@ -231,40 +211,37 @@ impl EncString {
...
     }
 }
 
 impl KeyDecryptable<SymmetricCryptoKey, Vec<u8>> for EncString {
     fn decrypt_with_key(&self, key: &SymmetricCryptoKey) -> Result<Vec<u8>> {
         match self {
             EncString::AesCbc256_B64 { iv, data } => {
-                if key.mac_key.is_some() {
-                    return Err(CryptoError::MacNotProvided);
+                if let SymmetricCryptoKey::Aes256CbcKey(key) = key {
```
dani-garcia: Should we also match both value and key at the same time here, like it's done in `context.rs`?
```rust
        match (&self, &key) {
            (EncString::AesCbc256_B64 { iv, data }, SymmetricCryptoKey::Aes256CbcKey(key)) => {
             // AesCbc256_B64 encryption
            }
            (EncString::AesCbc256_HmacSha256_B64 { iv, mac, data }, SymmetricCryptoKey::Aes256CbcHmacKey(key)) => {
            // AesCbc256_HmacSha256_B64 encryption
            },
            _ => return Err(CryptoError::EncryptionTypeMismatch),
        }
```

crates/bitwarden-crypto/src/keys/symmetric_crypto_key.rs
```diff
@@ -7,24 +7,56 @@ use rand::Rng;
...
-    pub(crate) key: Pin<Box<GenericArray<u8, U32>>>,
-    pub(crate) mac_key: Option<Pin<Box<GenericArray<u8, U32>>>>,
+#[cfg_attr(test, derive(Debug))]
+pub enum SymmetricCryptoKey {
+    Aes256CbcKey(Aes256CbcKey),
+    Aes256CbcHmacKey(Aes256CbcHmacKey),
+}
+
+#[cfg(test)]
+impl PartialEq for SymmetricCryptoKey {
```
dani-garcia: I think this impl can be autogenerated by just adding it to the current cfg_attr: 
`#[cfg_attr(test, derive(Debug, PartialEq))]`


Bit of a side thought, but maybe in the future we could have our own `sensitive_derive` macro or find a crate that will do it for us, so we don't have to derive one Debug implementation for prod and implementing one for tests.

crates/bitwarden-crypto/src/keys/symmetric_crypto_key.rs
```diff
@@ -7,24 +7,56 @@ use rand::Rng;
...
+        }
+    }
 }
 
 impl Drop for SymmetricCryptoKey {
     fn drop(&mut self) {
-        self.key.zeroize();
-        if let Some(mac_key) = &mut self.mac_key {
-            mac_key.zeroize();
+        match self {
```
dani-garcia: If you add `Zeroize` to the derive of `SymmetricCryptoKey`, you can probably just call `self.zeroize()` here and avoid matching every branch

crates/bitwarden-crypto/src/keys/master_key.rs
```diff
@@ -342,16 +359,21 @@ mod tests {
         let user_key: EncString = "0.8UClLa8IPE1iZT7chy5wzQ==|6PVfHnVk5S3XqEtQemnM5yb4JodxmPkkWzmDRdfyHtjORmvxqlLX40tBJZ+CKxQWmS8tpEB5w39rbgHg/gqs0haGdZG4cPbywsgGzxZ7uNI=".parse().unwrap();
 
         let decrypted = master_key.decrypt_user_key(user_key).unwrap();
+        let decrypted = if let SymmetricCryptoKey::Aes256CbcHmacKey(k) = &decrypted {
+            k
+        } else {
+            panic!("Decrypted key is not an Aes256CbcHmacKey");
+        };
```
dani-garcia: Using let-else should simplify this a bit:
```suggestion
        let SymmetricCryptoKey::Aes256CbcHmacKey(decrypted) = &decrypted else {
            panic!("Decrypted key is not an Aes256CbcHmacKey");
        };
```

dani-garcia (commented):
Changes look pretty good to me, I've left a couple of small comments.

crates/bitwarden-crypto/src/keys/symmetric_crypto_key.rs
quexten: You're right. I think initially I implemented an actual partialeq with constant time comparison, but then scrapped the constant time since it was only for test, and never changed it to deriving it.

crates/bitwarden-crypto/src/keys/master_key.rs
quexten: Nice, syntax I definitely was not aware of. Applied!

Hinton:
As this PR is performing two separate things, is there a reason this is a single PR vs two separate?

quexten:
> As this PR is performing two separate things, is there a reason this is a single PR vs two separate?

If it helps we can do the "drop AES_CBC_128_HMAC" first, separately, but it is a requirement for this PR.

Hinton:
I would suggest doing that. It allows the merge commit to contain clear reasons for the whole change vs two separate units of work. The refactor can be dependent on the removal of `AES_CBC_128_HMAC`.

quexten:
> I would suggest doing that. It allows the merge commit to contain clear reasons for the whole change vs two separate units of work. The refactor can be dependent on the removal of `AES_CBC_128_HMAC`.

Moved those changes to here: https://github.com/bitwarden/sdk-internal/pull/147

crates/bitwarden-crypto/src/keys/master_key.rs
```diff
@@ -269,28 +284,32 @@ mod tests {
...
-            None,
-        ));
+        });
 
         let (user_key, protected) = make_user_key(&mut rng, &master_key).unwrap();
+        let user_key_unwrapped = if let SymmetricCryptoKey::Aes256CbcHmacKey(k) = &user_key.0 {
+            k
+        } else {
+            panic!("User key is not an Aes256CbcHmacKey");
+        };
```
dani-garcia: You should also be able to do a let-else here as well, I think:

```suggestion
        let SymmetricCryptoKey::Aes256CbcHmacKey(user_key_unwrapped) = &user_key.0 else {
            panic!("User key is not an Aes256CbcHmacKey");
        };
```

crates/bitwarden-crypto/src/keys/symmetric_crypto_key.rs
```diff
@@ -43,21 +56,17 @@ impl SymmetricCryptoKey {
...
-        Self { key, mac_key }
+        SymmetricCryptoKey::Aes256CbcHmacKey(Aes256CbcHmacKey {
+            encryption_key: key,
+            mac_key,
+        })
     }
 
     fn total_len(&self) -> usize {
-        self.key.len() + self.mac_key.as_ref().map_or(0, |mac| mac.len())
+        match &self {
```
dani-garcia: I'm quite surprised clippy didn't catch this, but you don't need a `&` on the match value, as self is already received by reference (so you're matching against `&&SymmetricCryptoKey`, which then gets auto dereferenced):
```suggestion
        match self {
```

Same thing below in `to_vec`

dani-garcia (dismissed):
Nice, I've left a couple of non-blocking nits, otherwise this LGTM

crates/bitwarden-crypto/src/store/context.rs
```diff
@@ -396,11 +400,11 @@ impl<Ids: KeyIds> KeyStoreContext<'_, Ids> {
...
-        EncString::encrypt_aes256_hmac(
-            data,
-            key.mac_key.as_ref().ok_or(CryptoError::InvalidMac)?,
-            &key.key,
-        )
+        if let SymmetricCryptoKey::Aes256CbcHmacKey(key) = key {
+            EncString::encrypt_aes256_hmac(data, key)
+        } else {
+            Err(CryptoError::InvalidKey)
+        }
```
Hinton: **suggestion:** I think it's nicer to use pattern matching here. You could consider to be explicit and use `Aes256CbcKey` for the other instance, which would give us a compile error if we add a new type.

```suggestion
        match key {
            SymmetricCryptoKey::Aes256CbcHmacKey(key) => EncString::encrypt_aes256_hmac(data, key),
            _ => Err(CryptoError::InvalidKey),
        }
```

crates/bitwarden-crypto/src/rsa.rs
```diff
@@ -37,11 +37,12 @@ pub(crate) fn make_key_pair(key: &SymmetricCryptoKey) -> Result<RsaKeyPair> {
...
-    let protected = EncString::encrypt_aes256_hmac(
-        pkcs.as_bytes(),
-        key.mac_key.as_ref().ok_or(CryptoError::InvalidMac)?,
-        &key.key,
-    )?;
+    let protected = match key {
+        SymmetricCryptoKey::Aes256CbcHmacKey(key) => {
+            EncString::encrypt_aes256_hmac(pkcs.as_bytes(), key)
+        }
+        _ => return Err(CryptoError::UnsupportedCipher),
```
Hinton: **question:** Is it understandable to get a `Encryption unsupported for cipher` error when calling `make_key_pair`?

crates/bitwarden-crypto/src/keys/symmetric_crypto_key.rs
```diff
@@ -129,12 +146,20 @@ impl TryFrom<&mut [u8]> for SymmetricCryptoKey {
...
         f.debug_struct("SymmetricCryptoKey").finish()
     }
 }
 
+#[cfg(not(test))]
+impl std::fmt::Debug for KdfDerivedKeymaterial {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("KdfDerviedKeymaterial").finish()
+    }
+}
```
Hinton: **issue:** This is not the appropriate location to define this. It belongs to the `KdfDerivedKeymaterial` struct.

crates/bitwarden-crypto/src/keys/master_key.rs
```diff
@@ -1,15 +1,21 @@
...
+use zeroize::Zeroize;
 #[cfg(feature = "wasm")]
 use {tsify_next::Tsify, wasm_bindgen::prelude::*};
 
 use super::utils::{derive_kdf_key, stretch_kdf_key};
-use crate::{util, CryptoError, EncString, KeyDecryptable, Result, SymmetricCryptoKey, UserKey};
+use crate::{util, EncString, KeyDecryptable, Result, SymmetricCryptoKey, UserKey};
+
+#[cfg_attr(test, derive(Debug))]
+pub(crate) struct KdfDerivedKeymaterial {
```
Hinton: **issue:** This struct violates naming conventions.

```suggestion
pub(crate) struct KdfDerivedKeyMaterial {
```

crates/bitwarden-crypto/src/enc_string/symmetric.rs
```diff
@@ -213,31 +211,30 @@ impl EncString {
...
 impl KeyEncryptable<SymmetricCryptoKey, EncString> for &[u8] {
     fn encrypt_with_key(self, key: &SymmetricCryptoKey) -> Result<EncString> {
-        EncString::encrypt_aes256_hmac(
-            self,
-            key.mac_key.as_ref().ok_or(CryptoError::InvalidMac)?,
-            &key.key,
-        )
+        match key {
+            SymmetricCryptoKey::Aes256CbcHmacKey(key) => EncString::encrypt_aes256_hmac(self, key),
+            _ => Err(CryptoError::UnsupportedCipher),
```
Hinton: **question:** When adding a new `SymmetricCryptoKey` variant is the expected behaviour to be to throw `UnsupportedCipher`? Or do we want the implementer to make an explicit decision about this?

If it's the latter we should not use the catch all since that will still compile after adding a new variant.

crates/bitwarden-crypto/src/error.rs
```diff
@@ -47,6 +47,12 @@ pub enum CryptoError {
 
     #[error("Number is zero")]
     ZeroNumber,
+
+    #[error("Encryption unsupported for cipher")]
+    UnsupportedCipher,
```
Hinton: **question:** What does `Encryption unsupported for cipher` mean? What is `cipher`?

crates/bitwarden-crypto/src/error.rs
```diff
@@ -47,6 +47,12 @@ pub enum CryptoError {
...
+
+    #[error("Encryption type mismatch")]
+    EncryptionTypeMismatch,
```
Hinton: **question:** Is `EncryptionTypeMismatch` an appropriate name for this? You are using the wrong key for the wrong purpose but `Encryption type mismatch` is not particularly helpful.

I would argue this is more a key misuse or wrong key type error.

crates/bitwarden-crypto/src/keys/master_key.rs
```diff
@@ -1,15 +1,21 @@
...
 use serde::{Deserialize, Serialize};
+use zeroize::Zeroize;
 #[cfg(feature = "wasm")]
 use {tsify_next::Tsify, wasm_bindgen::prelude::*};
 
 use super::utils::{derive_kdf_key, stretch_kdf_key};
-use crate::{util, CryptoError, EncString, KeyDecryptable, Result, SymmetricCryptoKey, UserKey};
+use crate::{util, EncString, KeyDecryptable, Result, SymmetricCryptoKey, UserKey};
+
+#[cfg_attr(test, derive(Debug))]
```
Hinton: **question:** I'm not huge fan of conditional compilation for test, it would be fairly trivial to remove this and expose the key material by removing the conditional compilation.

Is this only for improving test ergonomics? If so I would argue to just continue checking the inner type directly.

crates/bitwarden-crypto/src/keys/master_key.rs
Hinton: **issue:** Please document this struct.

crates/bitwarden-crypto/src/keys/master_key.rs
```diff
@@ -1,15 +1,21 @@
...
+pub(crate) struct KdfDerivedKeymaterial {
+    pub(crate) key_material: Pin<Box<GenericArray<u8, U32>>>,
```
Hinton: @dani-garcia We will eventually want this to also be stored in the memory protected region. Do you foresee any complexity around this? And if so is this making it harder than before?

crates/bitwarden-crypto/src/keys/master_key.rs
```diff
@@ -118,39 +124,47 @@ impl MasterKey {
...
-        self.0.to_base64()
+        STANDARD.encode(self.0.key_material.as_slice())
+    }
+
+    pub fn try_from(value: &mut [u8]) -> Result<Self> {
+        let material = KdfDerivedKeymaterial {
+            key_material: Box::pin(GenericArray::<u8, U32>::clone_from_slice(value)),
+        };
+        value.zeroize();
+        Ok(Self::new(material))
```
Hinton: **issue:** This can never fail?

**suggestion:** Implement using the From trait, https://doc.rust-lang.org/std/convert/trait.From.html.

crates/bitwarden-crypto/src/keys/master_key.rs
```diff
@@ -118,39 +124,47 @@ impl MasterKey {
...
     }
 }
 
 /// Helper function to encrypt a user key with a master or pin key.
 pub(super) fn encrypt_user_key(
-    key: &SymmetricCryptoKey,
+    key: &KdfDerivedKeymaterial,
     user_key: &SymmetricCryptoKey,
 ) -> Result<EncString> {
+    let mut userkey_bytes = user_key.to_vec();
```
Hinton: **issue:** Currently you can fail on `stretch_kdf_key` which would not immediately drop the memory. However since we use a custom allocator it would probably still be cleaned up.

**suggestion:** I think this would look cleaner.

```rust
pub(super) fn encrypt_user_key(
    key: &KdfDerivedKeymaterial,
    user_key: &SymmetricCryptoKey,
) -> Result<EncString> {
  let stretched_key = stretch_kdf_key(key)?;

  let bytes = Zeroizing::new(user_key.to_vec());
  EncString::encrypt_aes256_hmac(&bytes, &stretched_key)
}
```

crates/bitwarden-crypto/src/keys/master_key.rs
```diff
@@ -301,32 +320,30 @@ mod tests {
...
-        let master_key = MasterKey(derive_symmetric_key("test1"));
+        let kdf_material = KdfDerivedKeymaterial {
+            key_material: if let SymmetricCryptoKey::Aes256CbcHmacKey(k) =
+                &derive_symmetric_key("test1")
+            {
+                k.encryption_key.clone()
+            } else {
+                panic!("Key is not an Aes256CbcHmacKey");
+            },
+        };
```
Hinton: **suggestion:** Update `derive_symmetric_key` to return `Aes256CbcHmacKey`?

crates/bitwarden-crypto/src/keys/mod.rs
```diff
@@ -11,6 +11,7 @@ mod symmetric_crypto_key;
 #[cfg(test)]
 pub use symmetric_crypto_key::derive_symmetric_key;
 pub use symmetric_crypto_key::SymmetricCryptoKey;
+pub(crate) use symmetric_crypto_key::{Aes256CbcHmacKey, Aes256CbcKey};
```
Hinton: **issue:** `SymmetricCryptoKey` is public, it's members `Aes256CbcHmacKey` and `Aes256CbcKey` are also public but not publicly accessible? Any reason to limit these to the bitwarden_crypto crate? And if so should they be public?

crates/bitwarden-crypto/src/keys/symmetric_crypto_key.rs
```diff
@@ -7,25 +7,38 @@ use rand::Rng;
 use zeroize::Zeroize;
 
 use super::key_encryptable::CryptoKey;
+#[cfg(not(test))]
+use super::master_key::KdfDerivedKeymaterial;
```
Hinton: **observation:** Having test only imports in the file header is generally a sign of something weird going on.

crates/bitwarden-crypto/src/keys/symmetric_crypto_key.rs
```diff
@@ -7,25 +7,38 @@ use rand::Rng;
...
 use crate::CryptoError;
 
+// GenericArray is equivalent to [u8; N], which is a Copy type placed on the stack.
+// To keep the compiler from making stack copies when moving this struct around,
+// we use a Box to keep the values on the heap. We also pin the box to make sure
+// that the contents can't be pulled out of the box and moved
```
Hinton: **issue:** This comment also applies to `Aes256CbcHmacKey`.

crates/bitwarden-crypto/src/keys/symmetric_crypto_key.rs
```diff
@@ -7,25 +7,38 @@ use rand::Rng;
...
+#[derive(Zeroize, Clone)]
+#[cfg_attr(test, derive(Debug, PartialEq))]
+pub struct Aes256CbcKey {
```
Hinton, line 17: **issue:** This is a public struct, and should be documented. (We prefer to document internal types to but it's more important for public)

crates/bitwarden-crypto/src/keys/symmetric_crypto_key.rs
```diff
@@ -7,25 +7,38 @@ use rand::Rng;
...
-#[derive(Clone)]
-pub struct SymmetricCryptoKey {
-    // GenericArray is equivalent to [u8; N], which is a Copy type placed on the stack.
-    // To keep the compiler from making stack copies when moving this struct around,
-    // we use a Box to keep the values on the heap. We also pin the box to make sure
-    // that the contents can't be pulled out of the box and moved
-    pub(crate) key: Pin<Box<GenericArray<u8, U32>>>,
-    pub(crate) mac_key: Option<Pin<Box<GenericArray<u8, U32>>>>,
+#[derive(Zeroize, Clone)]
+#[cfg_attr(test, derive(Debug, PartialEq))]
```
Hinton: **thought**: Consider if we actually need to derive Debug.

crates/bitwarden-crypto/src/keys/symmetric_crypto_key.rs
```diff
@@ -7,25 +7,38 @@ use rand::Rng;
...
+    Aes256CbcHmacKey(Aes256CbcHmacKey),
 }
 
 impl Drop for SymmetricCryptoKey {
     fn drop(&mut self) {
-        self.key.zeroize();
-        if let Some(mac_key) = &mut self.mac_key {
-            mac_key.zeroize();
-        }
+        self.zeroize();
```
Hinton: **suggestion:** Now that we don't need a custom implementation of Drop we can just derive `ZeroizeOnDrop` and remove this + the marker.

crates/bitwarden-crypto/src/keys/symmetric_crypto_key.rs
```diff
@@ -67,10 +76,16 @@ impl SymmetricCryptoKey {
     pub fn to_vec(&self) -> Vec<u8> {
         let mut buf = Vec::with_capacity(self.total_len());
 
-        buf.extend_from_slice(&self.key);
-        if let Some(mac) = &self.mac_key {
-            buf.extend_from_slice(mac);
+        match &self {
```
Hinton: Same comment as Dani made on `total_len`, we can remove `&`.

crates/bitwarden-crypto/src/keys/symmetric_crypto_key.rs
```diff
@@ -7,25 +7,38 @@ use rand::Rng;
...
 use crate::CryptoError;
 
+// GenericArray is equivalent to [u8; N], which is a Copy type placed on the stack.
+// To keep the compiler from making stack copies when moving this struct around,
+// we use a Box to keep the values on the heap. We also pin the box to make sure
+// that the contents can't be pulled out of the box and moved
+#[derive(Zeroize, Clone)]
+#[cfg_attr(test, derive(Debug, PartialEq))]
+pub struct Aes256CbcKey {
+    pub(crate) encryption_key: Pin<Box<GenericArray<u8, U32>>>,
```
Hinton: **thought:** Should we shorten this to `enc_key` since we don't call `mac` for `message_authentication_code_key`?

crates/bitwarden-crypto/src/keys/utils.rs
```diff
@@ -59,15 +62,17 @@ pub(super) fn derive_kdf_key(secret: &[u8], salt: &[u8], kdf: &Kdf) -> Result<Sy
             hash
         }
     };
-    SymmetricCryptoKey::try_from(hash.as_mut_slice())
+    Ok(KdfDerivedKeymaterial {
+        key_material: Box::pin(GenericArray::clone_from_slice(&hash)),
```
Hinton: **issue:** We're not zeroizing hash any longer.

@dani-garcia I don't think we actually need to with the zeroizing allocator?

crates/bitwarden-crypto/src/keys/symmetric_crypto_key.rs
```diff
@@ -7,25 +7,38 @@ use rand::Rng;
...
 use super::key_encryptable::CryptoKey;
+#[cfg(not(test))]
+use super::master_key::KdfDerivedKeymaterial;
 use crate::CryptoError;
 
+// GenericArray is equivalent to [u8; N], which is a Copy type placed on the stack.
+// To keep the compiler from making stack copies when moving this struct around,
+// we use a Box to keep the values on the heap. We also pin the box to make sure
+// that the contents can't be pulled out of the box and moved
+#[derive(Zeroize, Clone)]
```
Hinton: **question:** Should these two structs derive `ZeroizeOnDrop`?

Hinton (changes requested):
Can we remove `encrypt_aes256` from `aes.rs`? Since we don't ever plan on supporting that type.

crates/bitwarden-crypto/src/rsa.rs
quexten: Hmm, `UnsupportedCipher` is indeed not a great error. Maybe `EncryptionNotSupported` or `LegacyEncryptionNotSupported` (to convey that this is on purpose for a decrypt-only legacy key?).

```rust
    let protected = match key {
        SymmetricCryptoKey::Aes256CbcHmacKey(key) => {
            EncString::encrypt_aes256_hmac(pkcs.as_bytes(), key)
        }
        SymmetricCryptoKey::Aes256CbcKey(_) => Err(CryptoError::EncryptionOperationNotSupported(
            "Aes256Cbc".to_string(),
        )),
    }?;
```
and
```rust
    #[error("Encryption is not implemented for key of type {0}")]
    EncryptionOperationNotSupported(String),
```

crates/bitwarden-crypto/src/enc_string/symmetric.rs
quexten: Yeah, I think no more catch alls here.

crates/bitwarden-crypto/src/error.rs
quexten: Renaming to `WrongKeyType`

crates/bitwarden-crypto/src/keys/master_key.rs
quexten: Just noting: This material is not meant to stay around in memory longer than needed (i.e until used to decapsulate/decrypt the userkey), which is only during unlock. So if this needs to be protected, then the memory that we give the KDF (argon/pbkdf) to work on probably also needs to be protected (the lifetimes are, sparring a few milliseconds, the same)? Or am I missing something.

crates/bitwarden-crypto/src/keys/master_key.rs
quexten: Hmm, looking at it, this should check the length of value and error, such that `clone_from_slice` can never panic, so `try_from` is still correct, but is missing a check?

```rust
    pub fn try_from(value: &mut [u8]) -> Result<Self> {
        if value.len() != 32 {
            value.zeroize();
            return Err(CryptoError::InvalidKey);
        }

        let material = KdfDerivedKeyMaterial {
            key_material: Box::pin(GenericArray::<u8, U32>::clone_from_slice(value)),
        };
        value.zeroize();
        Ok(Self::new(material))
    }
```

(though I'm not sure yet invalidkey is the correct error)

crates/bitwarden-crypto/src/keys/symmetric_crypto_key.rs
quexten: That's fine with me. I'd probably go for "authentication_key" if we were to not use shortened versions though.

crates/bitwarden-crypto/src/keys/mod.rs
quexten: Made them public now. It's just important not to make the contained key arrays public. Ideally, non-bitwarden-crypto consumers  would not interact with the key types directly but just use the key references and not care about actual encryption types (unless there is an error).

crates/bitwarden-crypto/src/keys/utils.rs
quexten: Wrapped with `Zeroizing` now

crates/bitwarden-crypto/src/keys/symmetric_crypto_key.rs
quexten: This import is removed now

crates/bitwarden-crypto/src/keys/symmetric_crypto_key.rs
quexten: Without deriving debug, I don't think we can just assert_eq the keys during tests, so removing this would make testing ergonomics significantly worse (unless I'm missing something).

crates/bitwarden-crypto/src/keys/master_key.rs
quexten: This case was fixed by replacing an `unwrap_err` with `.err().unwrap()` in a test. So removed it now.

crates/bitwarden-crypto/src/rsa.rs
Hinton: I presume we need to support this for the new `SymmetricCryptoKey` type? If not the idiomatic rust approach would have been to set the type of `key` to `Aes256CbcHmacKey` and let the caller unpack it.

We could use `WrongKeyType` error here since you're literally sending in a key of the wrong type. But `OperationNotSupported` would also not be bad, (although I would use an enum instead of string).

I think this also highlights that the `CryptoError` struct is getting a bit big and we should consider returning more specialized error enums for some operations. (Should probably not be resolved in this PR)

crates/bitwarden-crypto/src/keys/master_key.rs
```diff
@@ -5,18 +5,28 @@ use generic_array::{typenum::U32, GenericArray};
...
-pub(crate) struct KdfDerivedKeymaterial {
+/// The entropy that gets produced after calling a KDF on a password or PIN.
+/// This should not be directly used to encrypt, but instead be explicitly
+/// converted to a SymmetricCryptoKey of a specific encryption type, by re-using
+/// the key-material, or by stretching it with HKDF
+pub(crate) struct KdfDerivedKeyMaterial {
     pub(crate) key_material: Pin<Box<GenericArray<u8, U32>>>,
 }
 
+#[cfg(not(test))]
```
Hinton: **issue:** Since we don't have a specific derive for tests we should just remove this.

crates/bitwarden-crypto/src/keys/master_key.rs
```diff
@@ -138,19 +153,17 @@ impl MasterKey {
...
     user_key: &SymmetricCryptoKey,
 ) -> Result<EncString> {
-    let mut userkey_bytes = user_key.to_vec();
-    let stretched_key = stretch_kdf_key(key)?;
-    let encrypted_userkey = EncString::encrypt_aes256_hmac(&userkey_bytes, &stretched_key);
-    userkey_bytes.zeroize();
-    encrypted_userkey
+    let stretched_master_key = stretch_kdf_key(master_key)?;
+    let userkey_bytes = Zeroizing::new(user_key.to_vec());
+    EncString::encrypt_aes256_hmac(&userkey_bytes, &stretched_master_key)
```
Hinton: We generally refer to this as `user_key` not `userkey`.
```suggestion
    let user_key_bytes = Zeroizing::new(user_key.to_vec());
    EncString::encrypt_aes256_hmac(&user_key_bytes, &stretched_master_key)
```

crates/bitwarden-crypto/src/keys/master_key.rs
Hinton: Yes that does seem more reasonable. It should use the `TryFrom` trait, https://doc.rust-lang.org/std/convert/trait.TryFrom.html.

mzieniukbw (dismissed):

crates/bitwarden-crypto/src/rsa.rs
quexten: Replaced with enum!

crates/bitwarden-crypto/src/keys/symmetric_crypto_key.rs
```diff
@@ -4,32 +4,46 @@ use aes::cipher::typenum::U32;
...
+/// A symmetric encryption key. Used to encrypt and decrypt [`EncString`](crate::EncString)
+#[derive(Clone)]
+#[cfg_attr(test, derive(Debug, PartialEq))]
+pub enum SymmetricCryptoKey {
+    Aes256CbcKey(Aes256CbcKey),
+    Aes256CbcHmacKey(Aes256CbcHmacKey),
 }
 
-impl zeroize::ZeroizeOnDrop for SymmetricCryptoKey {}
+impl ZeroizeOnDrop for SymmetricCryptoKey {}
```
Hinton: **issue:** Please derive this instead.

crates/bitwarden-crypto/src/keys/utils.rs
```diff
@@ -12,15 +11,19 @@ const ARGON2ID_MIN_ITERATIONS: u32 = 2;
...
+) -> Result<KdfDerivedKeyMaterial> {
+    let hash = match kdf {
         Kdf::PBKDF2 { iterations } => {
             let iterations = iterations.get();
             if iterations < PBKDF2_MIN_ITERATIONS {
                 return Err(CryptoError::InsufficientKdfParameters);
             }
 
-            crate::util::pbkdf2(secret, salt, iterations)
+            zeroize::Zeroizing::new(crate::util::pbkdf2(secret, salt, iterations))
```
Hinton:
```suggestion
           Zeroizing::new(crate::util::pbkdf2(secret, salt, iterations))
```

crates/bitwarden-crypto/src/keys/utils.rs
```diff
@@ -56,18 +59,20 @@ pub(super) fn derive_kdf_key(secret: &[u8], salt: &[u8], kdf: &Kdf) -> Result<Sy
             }
             clear_stack();
 
-            hash
+            zeroize::Zeroizing::new(hash)
```
Hinton:
```suggestion
            Zeroizing::new(hash)
```

crates/bitwarden-crypto/src/keys/master_key.rs
Hinton: Seems reasonable as long as they actually don't stick around.

crates/bitwarden-crypto/src/keys/master_key.rs
Hinton: This doesn't seem to be updated?

crates/bitwarden-crypto/src/keys/symmetric_crypto_key.rs
Hinton: When running the tests locally I only need to derive `PartialEq`.

crates/bitwarden-crypto/src/enc_string/symmetric.rs
```diff
@@ -213,31 +211,26 @@ impl EncString {
...
     fn encrypt_with_key(self, key: &SymmetricCryptoKey) -> Result<EncString> {
-        EncString::encrypt_aes256_hmac(
-            self,
-            key.mac_key.as_ref().ok_or(CryptoError::InvalidMac)?,
-            &key.key,
-        )
+        match key {
+            SymmetricCryptoKey::Aes256CbcHmacKey(key) => EncString::encrypt_aes256_hmac(self, key),
+            SymmetricCryptoKey::Aes256CbcKey(_) => Err(
+                CryptoError::EncryptionOperationNotSupported(UnsupportedKeyType::Aes256Cbc),
```
Hinton: Looking at this more I think we can generalize the error into `OperationNotSupported`.

```suggestion
                CryptoError::OperationNotSupported(UnsupportedKeyType::Aes256Cbc),
```

crates/bitwarden-crypto/src/keys/master_key.rs
quexten: Seems github ate my reply. I didn't update this because I don't think we want to change the signature of `derive_shareable_key` to always return a Aes256CbcHmacKey, because in the future in might return a XChaCha20Poly1305Key. 

So, I can change it here, but it might mean that we need to change it back in the future? Or pull in the logic in `derive_shareable_key` to the `derive_symmetric_key` to make sure the test keys always stay the same?

quexten: Actually, thinking about it this would need to be handled once those keys are introduced anyways, so let's care about it then. Fixed (39adb185)

quexten: Ok, leaving this issue up for whenever we actually add the new keys; Updated so it returns the `Aes256CbcHmacKey`. It does force wrapping with `SymmetricCryptoKey::Aes256CbcHmacKey()` in a few other places though.

crates/bitwarden-crypto/src/error.rs
```diff
@@ -47,6 +47,18 @@ pub enum CryptoError {
 
     #[error("Number is zero")]
     ZeroNumber,
+
+    #[error("Encryption is not implemented for {0}")]
+    EncryptionOperationNotSupported(UnsupportedKeyType),
```
Hinton: After pondering on this for a bit longer I wonder if we need to pass in the key type. Having `Encryption is not implemented for provided key` might be sufficient?

crates/bitwarden-crypto/src/error.rs
```diff
@@ -47,6 +47,18 @@ pub enum CryptoError {
...
+
+    #[error("Key encryption algorithm does not match encrypted data type")]
```
Hinton: **nit:** It's not technically encryption since we're using it for decryption. crypto algorithm or just omitting it completely?

```suggestion
    #[error("Key algorithm does not match encrypted data type")]
```

crates/bitwarden-crypto/src/keys/master_key.rs
```diff
@@ -1,16 +1,25 @@
...
 use super::utils::{derive_kdf_key, stretch_kdf_key};
 use crate::{util, CryptoError, EncString, KeyDecryptable, Result, SymmetricCryptoKey, UserKey};
 
+/// The entropy that gets produced after calling a KDF on a password or PIN.
+/// This should not be directly used to encrypt, but instead be explicitly
+/// converted to a SymmetricCryptoKey of a specific encryption type, by re-using
+/// the key-material, or by stretching it with HKDF
+pub(crate) struct KdfDerivedKeyMaterial {
+    pub(crate) key_material: Pin<Box<GenericArray<u8, U32>>>,
+}
```
Hinton: **suggestion:** I think we can make the key material internal, and use the NewType pattern instead. We should add a note about the `Pin<Box>` pattern. It's probably worthwile to move the description to a linkable section and just add a note like `We use the Pin<Box> pattern as described in <link> to ensure it's stored on the heap and not moved around`.

```suggestion
/// Holding struct for key material derived from a KDF.
///
/// The internal key material should not be used directly for cryptographic operations. Instead it
/// MUST be converted to the appropriate type such as `SymmetricCryptoKey`, `MasterKey` or any other
/// key type. This can be done by either directly consuming the key material or by stretching it
/// further using HKDF (HMAC-based Key Derivation Function).
pub(crate) struct KdfDerivedKeyMaterial(Pin<Box<GenericArray<u8, U32>>>);
```

crates/bitwarden-crypto/src/keys/master_key.rs
```diff
@@ -118,39 +127,54 @@ impl MasterKey {
...
+        };
+        value.zeroize();
+        Ok(Self::new(material))
     }
 }
 
 /// Helper function to encrypt a user key with a master or pin key.
 pub(super) fn encrypt_user_key(
-    key: &SymmetricCryptoKey,
+    master_key: &KdfDerivedKeyMaterial,
```
Hinton: **issue:** This is not a `MasterKey`.

crates/bitwarden-crypto/src/keys/master_key.rs
Hinton: I would also move this to a new file and lift up functions like encrypt/decrypt key into the `KdfDerivedKeyMaterial` struct.

crates/bitwarden-crypto/src/store/context.rs
```diff
@@ -376,17 +376,15 @@ impl<Ids: KeyIds> KeyStoreContext<'_, Ids> {
...
-                Ok(dec)
+        match (data, key) {
+            (EncString::AesCbc256_B64 { iv, data }, SymmetricCryptoKey::Aes256CbcKey(key)) => {
+                crate::aes::decrypt_aes256(iv, data.clone(), &key.enc_key)
             }
+            (
+                EncString::AesCbc256_HmacSha256_B64 { iv, mac, data },
+                SymmetricCryptoKey::Aes256CbcHmacKey(key),
+            ) => crate::aes::decrypt_aes256_hmac(iv, mac, data.clone(), &key.mac_key, &key.enc_key),
+            _ => Err(CryptoError::InvalidKey),
```
Hinton, line 341: **question:** Do we want this catch all or be required to update usages when adding a new type?

crates/bitwarden-crypto/src/store/context.rs
```diff
@@ -396,11 +394,10 @@ impl<Ids: KeyIds> KeyStoreContext<'_, Ids> {
...
     ) -> Result<EncString> {
         let key = self.get_symmetric_key(key)?;
-        EncString::encrypt_aes256_hmac(
-            data,
-            key.mac_key.as_ref().ok_or(CryptoError::InvalidMac)?,
-            &key.key,
-        )
+        match key {
+            SymmetricCryptoKey::Aes256CbcHmacKey(key) => EncString::encrypt_aes256_hmac(data, key),
+            _ => Err(CryptoError::InvalidKey),
```
Hinton: **question:** Do we want this catch all or be required to update usages when adding a new type?

crates/bitwarden-crypto/src/keys/symmetric_crypto_key.rs
```diff
@@ -43,21 +55,17 @@ impl SymmetricCryptoKey {
...
-        }
-    }
-
-    pub(crate) fn new(
-        key: Pin<Box<GenericArray<u8, U32>>>,
-        mac_key: Option<Pin<Box<GenericArray<u8, U32>>>>,
-    ) -> Self {
-        Self { key, mac_key }
+        SymmetricCryptoKey::Aes256CbcHmacKey(Aes256CbcHmacKey {
+            enc_key: key,
```
Hinton: **suggestion (nit):** Rename key to `enc_key`.

crates/bitwarden-crypto/src/keys/symmetric_crypto_key.rs
```diff
@@ -107,16 +121,18 @@ impl TryFrom<&mut [u8]> for SymmetricCryptoKey {
             key.copy_from_slice(&value[..Self::KEY_LEN]);
             mac_key.copy_from_slice(&value[Self::KEY_LEN..]);
 
-            Ok(SymmetricCryptoKey {
-                key,
-                mac_key: Some(mac_key),
-            })
+            Ok(SymmetricCryptoKey::Aes256CbcHmacKey(Aes256CbcHmacKey {
+                enc_key: key,
```
Hinton: **suggestion (nit):** Rename key to `enc_key`.

crates/bitwarden-crypto/src/keys/symmetric_crypto_key.rs
```diff
@@ -129,6 +145,7 @@ impl TryFrom<&mut [u8]> for SymmetricCryptoKey {
 impl CryptoKey for SymmetricCryptoKey {}
 
 // We manually implement these to make sure we don't print any sensitive data
+#[cfg(not(test))]
```
Hinton: **suggestion:** Remove `not(test)` since we don't have to derive debug for tests.

crates/bitwarden-crypto/src/keys/utils.rs
```diff
@@ -12,15 +12,19 @@ const ARGON2ID_MIN_ITERATIONS: u32 = 2;
 const ARGON2ID_MIN_PARALLELISM: u32 = 1;
 
 /// Derive a generic key from a secret and salt using the provided KDF.
-pub(super) fn derive_kdf_key(secret: &[u8], salt: &[u8], kdf: &Kdf) -> Result<SymmetricCryptoKey> {
-    let mut hash = match kdf {
+pub(super) fn derive_kdf_key(
```
Hinton: **suggestion:** Move to `KdfDerivedKeyMaterial`.

crates/bitwarden-crypto/src/keys/utils.rs
```diff
@@ -56,18 +60,20 @@ pub(super) fn derive_kdf_key(secret: &[u8], salt: &[u8], kdf: &Kdf) -> Result<Sy
...
+    })
 }
 
 /// Stretch the given key using HKDF.
-pub(super) fn stretch_kdf_key(k: &SymmetricCryptoKey) -> Result<SymmetricCryptoKey> {
-    let key: Pin<Box<GenericArray<u8, U32>>> = hkdf_expand(&k.key, Some("enc"))?;
-    let mac_key: Pin<Box<GenericArray<u8, U32>>> = hkdf_expand(&k.key, Some("mac"))?;
-
-    Ok(SymmetricCryptoKey::new(key, Some(mac_key)))
+pub(super) fn stretch_kdf_key(k: &KdfDerivedKeyMaterial) -> Result<Aes256CbcHmacKey> {
```
Hinton: **suggestion:** Move to `KdfDerivedKeyMaterial`.

crates/bitwarden-crypto/src/keys/symmetric_crypto_key.rs
```diff
@@ -153,13 +153,18 @@ impl std::fmt::Debug for SymmetricCryptoKey {
...
-pub fn derive_symmetric_key(name: &str) -> SymmetricCryptoKey {
+pub fn derive_symmetric_key(name: &str) -> Aes256CbcHmacKey {
     use zeroize::Zeroizing;
 
     use crate::{derive_shareable_key, generate_random_bytes};
 
     let secret: Zeroizing<[u8; 16]> = generate_random_bytes();
-    derive_shareable_key(secret, name, None)
+    let SymmetricCryptoKey::Aes256CbcHmacKey(key) =
+        &derive_shareable_key(secret, name, None).clone()
```
Hinton: **question:** Shareable keys are also always `Aes256CbcHmacKey` right?

crates/bitwarden-crypto/src/keys/master_key.rs
Hinton: I think that's actually a benefit if we do end up changing it to ChaCha? It forces us to ensure the usages are compatible vs it potentially silently failing?

That said I'm unsure if we can actually change shareable keys as it would be a breaking change for things like send.

crates/bitwarden-crypto/src/keys/symmetric_crypto_key.rs
quexten: Currently yes, and even for the initial changes implementing XChacha20Poly1305, yes. But in the medium to long-term we'll probably want to replace these too?

Do we change derive_sharable_key to return Aes256CbcHmacKey until then?

crates/bitwarden-crypto/src/error.rs
quexten: Yeah, makes sense to me.

crates/bitwarden-crypto/src/error.rs
quexten: To be honest, to me encryption algorithm still makes sense. The encryption algorithm necessarily always describes both encryption and decryption operations. But I'm still happy to remove it for brevity.

crates/bitwarden-crypto/src/keys/master_key.rs
quexten: Tbh, this makes me think the key-connector-key (also called masterkey) should also be split out to a new type, called "KeyConnectorKey". Using a "KdfDerivedKeyMaterial" holding struct does not really make sense semantically since the key connector just holds randomly generated data..

crates/bitwarden-crypto/src/store/context.rs
quexten: I think we need the catch all. Otherwise we need to create NxN entries, for all left out ones? Since the default behavior is to return an error, I do think that this is fine.

crates/bitwarden-crypto/src/store/context.rs
quexten: In this case, I think the catchall is not wanted. Removing it.

crates/bitwarden-crypto/src/store/context.rs
Hinton: True, I was a bit trigger happy and missed that we're using a tuple here.

crates/bitwarden-crypto/src/keys/symmetric_crypto_key.rs
Hinton: I think we can, it's a violation of backwards compatibility for `derive_sharable_key` to return anything other than the same key indefinitely.

We will need to figure out how we can solve this in the future as this low level function can't accurately track what algorithm should be used for the shareable key.

crates/bitwarden-crypto/src/keys/master_key.rs
Hinton: We could rename `KdfDerivedKeyMaterial` to `KeyMaterial` 🤔. Or have `KdfDerivedKeyMaterial` and `KeyConnectorKey` both implement `KeyMaterial` trait which exposes the `encrypt_user_key` functions.

But I would not mind holding of on this for a future refactor. Would be nice with a ticket to capture the work?

crates/bitwarden-crypto/src/keys/utils.rs
quexten: Left in utils, but now takes a `&Pin<Box<GenericArray<u8, U32>>>` because masterkeys that are random keyconnectorkeys can also be stretched. (So in total: KdfKeyMaterials (from MasterKey or PinKey, and KeyConnectorKey from MasterKey).

crates/bitwarden-crypto/src/keys/master_key.rs
quexten: While I feel this is not optimal yet, for now I've made the `MasterKey` into an enum, holding either a `KdfKey(KdfDerivedKeyMaterial` or a `KeyConnectorKey(Pin<Box<GenericArray<u8, U32>>>`, which should somewhat better reflect the semantics of these keys in the type system.

KdfDerivedKeyMaterial is also used by PinKey, and all three use some shared methods (wrapping the userkey), so that could be improved in the future.

crates/bitwarden-crypto/src/keys/symmetric_crypto_key.rs
quexten: Ok, sharable key returns the `Aes256CBCHmacKey` now.

crates/bitwarden-crypto/src/keys/utils.rs
```diff
@@ -1,100 +1,35 @@
...
-        enc_key: hkdf_expand(&k.key_material, Some("enc"))?,
-        mac_key: hkdf_expand(&k.key_material, Some("mac"))?,
-    })
+/// This can be either a kdf-derived key (PIN/Master password) or
+/// a random key from key connector
+pub(super) fn stretch_key(key: &Pin<Box<GenericArray<u8, U32>>>) -> Result<Aes256CbcHmacKey> {
+    let enc_key: Pin<Box<GenericArray<u8, U32>>> = hkdf_expand(key, Some("enc"))?;
+    let mac_key: Pin<Box<GenericArray<u8, U32>>> = hkdf_expand(key, Some("mac"))?;
+
+    Ok(Aes256CbcHmacKey { enc_key, mac_key })
```
Hinton: **question (nit):** Did we need to change the syntax here?

crates/bitwarden-crypto/src/keys/master_key.rs
Hinton: Can we please create a jira ticket for this work and add a comment in the code next to `MasterKey` about this upcoming work?

crates/bitwarden-crypto/src/keys/kdf.rs
```diff
@@ -0,0 +1,232 @@
...
+/// Holding struct for key material derived from a KDF.
+///
+/// The internal key material should not be used directly for cryptographic operations. Instead it
+/// MUST be converted to the appropriate type such as `SymmetricCryptoKey`, `MasterKey` or any other
+/// key type. This can be done by either directly consuming the key material or by stretching it
+/// further using HKDF (HMAC-based Key Derivation Function).
+///
+/// Uses pinned heap data as described in [Pinned heap
+/// data](super::symmetric_crypto_key#pinned-heap-data).
+pub struct KdfDerivedKeyMaterial(pub(crate) Pin<Box<GenericArray<u8, U32>>>);
```
Hinton: **suggestion:** We can change this to `super` which only exposes it for files in `keys`.

```suggestion
pub struct KdfDerivedKeyMaterial(pub(super) Pin<Box<GenericArray<u8, U32>>>);
```

crates/bitwarden-crypto/src/keys/kdf.rs
```diff
@@ -0,0 +1,232 @@
...
+        };
+        let key_material = Box::pin(GenericArray::clone_from_slice(&hash));
+        hash.zeroize();
+        Ok(KdfDerivedKeyMaterial(key_material))
+    }
+
+    /// Derives a users master key from their password, email and KDF.
+    ///
+    /// Note: the email is trimmed and converted to lowercase before being used.
+    pub fn derive(password: &str, email: &str, kdf: &Kdf) -> Result<Self, CryptoError> {
```
Hinton: **question:** Does this need to be pub or is `pub(super)` sufficient like `derive_kdf_key`?

crates/bitwarden-crypto/src/keys/master_key.rs
```diff
@@ -75,38 +25,41 @@ pub enum HashPurpose {
...
-        Self(key)
+        Self::KdfKey(key)
     }
 
     /// Generate a new random master key. Primarily used for KeyConnector.
     pub fn generate(mut rng: impl rand::RngCore) -> Self {
         let mut key = Box::pin(GenericArray::<u8, U32>::default());
 
         rng.fill(key.as_mut_slice());
+        Self::KeyConnectorKey(key)
```
Hinton, line 44: **issue:** The comment above should be updated to indicate it's only for key connector?

crates/bitwarden-crypto/src/keys/symmetric_crypto_key.rs
```diff
@@ -27,17 +28,14 @@ pub struct Aes256CbcKey {
 #[derive(ZeroizeOnDrop, Clone)]
 #[cfg_attr(test, derive(Debug, PartialEq))]
 pub struct Aes256CbcHmacKey {
-    // GenericArray is equivalent to [u8; N], which is a Copy type placed on the stack.
-    // To keep the compiler from making stack copies when moving this struct around,
-    // we use a Box to keep the values on the heap. We also pin the box to make sure
-    // that the contents can't be pulled out of the box and moved
+    /// Uses a pinned heap data structure, as noted in [Pinned heap data]
```
Hinton: **nit**: Since this is a field rustdoc it should be copied to both fields, or changed to non rustdoc comment.

crates/bitwarden-crypto/src/keys/kdf.rs
```diff
@@ -0,0 +1,232 @@
...
+
+/// Holding struct for key material derived from a KDF.
+///
+/// The internal key material should not be used directly for cryptographic operations. Instead it
+/// MUST be converted to the appropriate type such as `SymmetricCryptoKey`, `MasterKey` or any other
+/// key type. This can be done by either directly consuming the key material or by stretching it
+/// further using HKDF (HMAC-based Key Derivation Function).
+///
+/// Uses pinned heap data as described in [Pinned heap
+/// data](super::symmetric_crypto_key#pinned-heap-data).
```
Hinton: **issue:** This link doesn't seem to work.

I suggest adding the section to the crate root. This also ensures it's part of the public documentation since `keys` is an internal module.

```markdown
//! # Pinned heap data
//!
//! This crate uses a `Pin<Box<>>` strategy to ensure data is stored on the heap and not moved
//! around. This pattern is commonly used for `GenericArray` since it's equivalent to `[u8; N]`
//! which is a Copy type placed on the stack. To keep the compiler from making stack copies when
//! moving this struct around, we use a Box to keep the values on the heap. We also pin the box to
//! make sure that the contents can't be pulled out of the box and moved.
```

And then linking to it as `[Pinned heap data][crate#pinned-heap-data]`

Hinton (commented):
Some small tweaks and we should be good.

crates/bitwarden-crypto/src/keys/utils.rs
quexten: Reverted

crates/bitwarden-crypto/src/keys/kdf.rs
quexten: Seems pub(super) is sufficient.

crates/bitwarden-crypto/src/keys/kdf.rs
quexten: Moved.

Hinton (approved):

PR #129, opened 2025-01-24
[PM-15389] Parse account and issuer from otpauth URLs

KatherineInCode:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-15389

## 📔 Objective

This adds optional `account` and `issuer` properties to the `Totp` object, and parses them out of `otpauth://` URLs. This attempts to follow a similar logic as to how [the iOS PM app currently is doing it](https://github.com/bitwarden/ios/blob/main/BitwardenShared/Core/Vault/Services/TOTP/OTPAuthModel.swift#L65), though compare also [how the iOS BWA app is doing it](https://github.com/bitwarden/authenticator-ios/blob/main/AuthenticatorShared/Core/Vault/Services/TOTP/OTPAuthModel.swift#L88). Ultimately because both the [Yubico spec](https://docs.yubico.com/yesdk/users-manual/application-oath/uri-string-format.html) and [Google spec](https://github.com/google/google-authenticator/wiki/Key-Uri-Format) indicate that colons should not be in issuers or account names, I went with the naïve approach of splitting on a colon. As well, neither spec defines appropriate behavior if the label's issuer and the parameter issuer are different, and I went with using the parameter; it's possible we should instead treat this as an error case?

This is my first foray into Rust, so I fully expect I am not as idiomatic about things as I could be, or that I likely missed some crucial detail.

## ⏰ Reminders before review

- CI builds passed
  team

## 🦮 Reviewer guidelines

### DIFF
--- a/crates/bitwarden-vault/Cargo.toml
+++ b/crates/bitwarden-vault/Cargo.toml
@@ -28,6 +28,7 @@ bitwarden-core = { workspace = true, features = ["internal"] }
 bitwarden-crypto = { workspace = true }
 chrono = { workspace = true }
 hmac = ">=0.12.1, <0.13"
+percent-encoding = ">=2.1, <3.0"
 rand = ">=0.8.5, <0.9"
 reqwest = { workspace = true }
 schemars = { workspace = true }
--- a/crates/bitwarden-vault/src/totp.rs
+++ b/crates/bitwarden-vault/src/totp.rs
@@ -168,28 +168,23 @@ impl FromStr for Totp {
 
         let params = if key.starts_with("otpauth://") {
             let url = Url::parse(&key).map_err(|_| TotpError::InvalidOtpauth)?;
-            let label = remove_first(url.path());
-            let label_string = label.map(|v| v.to_string());
-            let pieces: Vec<String> = label
-                .unwrap_or("")
-                .split(":")
-                .map(|v| v.to_string())
-                .collect();
-            let label_issuer = if pieces.len() > 1 {
-                pieces.first()
+            let label = url.path().strip_prefix("/");
+            let (issuer, account) = if let Some(label) = label {
+                if let Some((issuer, account)) = label.split_once(":") {
+                    (Some(issuer.trim()), Some(account.trim()))
+                } else if let Some((issuer, account)) = label.split_once("%3a") {
+                    (Some(issuer.trim()), Some(account.trim()))
+                } else {
+                    (None, Some(label.trim()))
+                }
             } else {
-                None
-            };
-            let label_account = if pieces.len() > 1 {
-                pieces.get(1)
-            } else {
-                label_string.as_ref()
+                (None, None)
             };
 
             let parts: HashMap<_, _> = url.query_pairs().collect();
 
             Totp {
-                account: label_account.cloned(),
+                account: account.map(|s| s.to_string()),
                 algorithm: parts
                     .get("algorithm")
                     .and_then(|v| match v.as_ref() {
@@ -207,7 +202,7 @@ impl FromStr for Totp {
                 issuer: parts
                     .get("issuer")
                     .map(|v| v.to_string())
-                    .or(label_issuer.cloned()),
+                    .or(issuer.map(|s| s.to_string())),
                 period: parts
                     .get("period")
                     .and_then(|v| v.parse().ok())
@@ -242,11 +237,6 @@ impl FromStr for Totp {
 
         Ok(params)
     }
-}
-
-/// Remove the first character from a string; returns None if the string is one character long.
-fn remove_first(s: &str) -> Option<&str> {
-    s.chars().next().map(|c| &s[c.len_utf8()..])
 }
 
 /// Derive the Steam OTP from the hash with the given number of digits.
@@ -453,7 +443,7 @@ mod tests {
         let totp = Totp::from_str(key).unwrap();
 
         assert_eq!(totp.account, Some("test-account@example.com".to_string()));
-        assert_eq!(totp.issuer, Some("other-test-issuer".to_string()));
+        assert_eq!(totp.issuer, Some("test-issuer".to_string()));
     }
 
     #[test]
--- a/crates/bitwarden-vault/src/totp.rs
+++ b/crates/bitwarden-vault/src/totp.rs
@@ -4,6 +4,7 @@ use bitwarden_core::VaultLocked;
 use bitwarden_crypto::{CryptoError, KeyContainer};
 use chrono::{DateTime, Utc};
 use hmac::{Hmac, Mac};
+use percent_encoding::{percent_decode, percent_decode_str};
 use reqwest::Url;
 use schemars::JsonSchema;
 use serde::{Deserialize, Serialize};
@@ -168,7 +169,8 @@ impl FromStr for Totp {
 
         let params = if key.starts_with("otpauth://") {
             let url = Url::parse(&key).map_err(|_| TotpError::InvalidOtpauth)?;
-            let label = url.path().strip_prefix("/");
+            let decoded_path = percent_decode_str(url.path()).decode_utf8_lossy();
+            let label = decoded_path.strip_prefix("/");
             let (issuer, account) = if let Some(label) = label {
                 if let Some((issuer, account)) = label.split_once(":") {
                     (Some(issuer.trim()), Some(account.trim()))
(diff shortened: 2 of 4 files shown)


crates/bitwarden-vault/src/totp.rs
```diff
@@ -372,6 +406,37 @@ mod tests {
         assert_eq!(response.period, 30);
     }
 
+    #[test]
```
Hinton, line 408: **suggestion:** Can we add another test for checking the behaviour with no label? It's technically not valid but I believe we handle it gracefully currently.

crates/bitwarden-vault/src/totp.rs
```diff
@@ -166,9 +168,28 @@ impl FromStr for Totp {
 
         let params = if key.starts_with("otpauth://") {
             let url = Url::parse(&key).map_err(|_| TotpError::InvalidOtpauth)?;
+            let label = remove_first(url.path());
```
Hinton: **suggestion:** There is a nice function on `Url` called [`path_segments`](https://docs.rs/url/latest/url/struct.Url.html#method.path_segments) which returns an iterator over path segments.

We can get the first value of the iterator using [`and_then`](https://doc.rust-lang.org/rust-by-example/error/option_unwrap/and_then.html)

```suggestion
            let label = url.path_segments().and_then(|mut s| s.next());
```

Alternatively using `url.path().strip_prefix("/")` would return the whole path without the initial `/` prefix which would be a more idiomatic way to write `remove_first`.

crates/bitwarden-vault/src/totp.rs
```diff
@@ -166,9 +168,28 @@ impl FromStr for Totp {
...
+            let label_issuer = if pieces.len() > 1 {
+                pieces.first()
+            } else {
+                None
+            };
+            let label_account = if pieces.len() > 1 {
+                pieces.get(1)
+            } else {
+                label_string.as_ref()
+            };
```
Hinton: There's a bit to unpack here. We generally don't need to take ownership of strings unless we use them as a return value. Meaning it's not until line 192 and 210 that we really have to call `clone`, or `to_string` on the inner value of the option, i.e. `opt.map(|v| v.to_string())`.

Removing those simplifies the code a bit.

```suggestion
          let pieces: Vec<&str> = label.unwrap_or("").split(":").collect();
            let label_issuer = if pieces.len() > 1 {
                pieces.first()
            } else {
                None
            };
            let label_account = if pieces.len() > 1 {
                pieces.get(1)
            } else {
                label.as_ref()
            };
```

Hinton: **suggestion:** Since we're repeating the same if statement twice this would be a good opportunity to use a tuple as return value from the if statement, https://doc.rust-lang.org/rust-by-example/primitives/tuples.html.

```rust
let (issuer, account) = if .... > 1 {
  (issuer, account)
} else {
  (None, account)
}
```

crates/bitwarden-vault/src/totp.rs
```diff
@@ -166,9 +168,28 @@ impl FromStr for Totp {
 
         let params = if key.starts_with("otpauth://") {
             let url = Url::parse(&key).map_err(|_| TotpError::InvalidOtpauth)?;
+            let label = remove_first(url.path());
+            let label_string = label.map(|v| v.to_string());
+            let pieces: Vec<String> = label
+                .unwrap_or("")
+                .split(":")
```
Hinton: **thought:** `unwrap` is generally discouraged unless we actually need a value. 

I think it would be more idiomatic if there is a way to return:
- `None` if we can't split the label
- `Some()` if we can.

We could use `map(|v| v.split(":"))` but we would need to continue checking length which to some degree is worse since we have to check both the option and the length of the content.

There is a [`split_once`](https://doc.rust-lang.org/std/string/struct.String.html#method.split_once) function we could maybe use instead, which returns the first occurrence and remainder after the deliminator. Which could be used with `and_then`.

Hinton (changes requested):
Thanks for taking this work.

I wasn't aware the iOS app currently supported parsing otp. Is it part of scanning totp tokens or something else? If we need the capability to split an otpauth url into it's portions we could expose a function from the SDK now that we support issuer and account?

Function wise this seems great. I've suggested adding a new test case for when there is no label. And given some suggestions on how we can rework some portions to be more rust idiomatic.

KatherineInCode:
> I wasn't aware the iOS app currently supported parsing otp. Is it part of scanning totp tokens or something else? If we need the capability to split an otpauth url into it's portions we could expose a function from the SDK now that we support issuer and account?

On the Authenticator side, manual entry/editing allows you to specify the issuer/account individually, and then we normalize it to an `otpauth` URL when saving to the database; we also would want to have those for sorting/searching purposes.

It seems reasonable that we could probably move that behavior into the SDK, to centralize the business logic, though it would take a bit of work on both the PM and BWA side to incorporate it. Likewise with taking advantage of the companion ticket to output a token as an `otpauth` URL (which I'll be grabbing after this one if no one else has yet).

The suggestions are all reasonable, and I'll get on making those changes.

crates/bitwarden-vault/Cargo.toml
```diff
@@ -29,6 +29,7 @@ bitwarden-crypto = { workspace = true }
 bitwarden-error = { workspace = true }
 chrono = { workspace = true }
 hmac = ">=0.12.1, <0.13"
+percent-encoding = ">=2.1, <3.0"
```
Hinton, line 32: 😢 it's unfortunate this isn't exposed through reqwest or url.

crates/bitwarden-vault/src/totp.rs
```diff
@@ -171,9 +174,24 @@ impl FromStr for Totp {
 
         let params = if key.starts_with("otpauth://") {
             let url = Url::parse(&key).map_err(|_| TotpError::InvalidOtpauth)?;
+            let decoded_path = percent_decode_str(url.path()).decode_utf8_lossy();
+            let label = decoded_path.strip_prefix("/");
+            let (issuer, account) = if let Some(label) = label {
+                if let Some((issuer, account)) = label.split_once(":") {
+                    (Some(issuer.trim()), Some(account.trim()))
+                } else if let Some((issuer, account)) = label.split_once("%3a") {
+                    (Some(issuer.trim()), Some(account.trim()))
```
Hinton: **question:** Is this needed? I think this would only catch double escaped characters as we've decoded them already.

Hinton: If it's not needed we can use pattern matching and rewrite the remaining logic into

```rust
let (issuer, account) = match label.and_then(|v| v.split_once(':')) {
    Some((issuer, account)) => (Some(issuer.trim()), Some(account.trim())),
    None => (None, label),
};
```

KatherineInCode: It's not; I realized the issue a couple hours later. I'd introduced it originally, because I also dealt with the percent-encoding properly 😅

I thought we were trying to avoid doing `match`? Or is this just one of those cases where it is, in fact, the best way of handling it?

crates/bitwarden-vault/Cargo.toml
KatherineInCode: Agreed; it took me an inordinate amount of time to figure out how to do the percent decoding because of that, especially since `Url` already decodes some parts for you, but not other parts.

crates/bitwarden-vault/src/totp.rs
Hinton: Avoid doing `match`? Clippy discourages using match where you only use one of the two cases, and I generally don't like using match for booleans.

In this case we would have logic in both branches and the alternative code would look something like this:

```rust
let (issuer, account) = label
  .and_then(|v| v.split_once(':'))
  .map_or((None, label), |(issuer, account)| {
    (Some(issuer.trim()), Some(account.trim()))
  });
```
I consider the match to be more readable.

KatherineInCode: That makes sense. It's probably just my relative rust inexperience while reading our [contributing guidelines](https://contributing.bitwarden.com/contributing/code-style/rust/) and not knowing enough to understand the nuance.

Hinton (approved):
I think we're good. Thanks for iterating on this!

KatherineInCode:
> I think we're good. Thanks for iterating on this!

Thank you for the help! :)

PR #142, opened 2025-02-04
Encoding trait

MGibson1:
## 📔 Objective

Add encoding trait. This trait will reduce code as more types are convertable to/from `EncString`. Additionally, this common crate will be useful to define things like type-safe base64 encoded string types, which can then be used to explicitly define API expectations.

## ⏰ Reminders before review

- CI builds passed
  team

## 🦮 Reviewer guidelines

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -429,6 +429,7 @@ dependencies = [
  "aes",
  "argon2",
  "base64",
+ "bitwarden-common",
  "cbc",
  "criterion",
  "generic-array",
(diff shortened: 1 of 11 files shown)


Hinton:
Common is a really bad name and tells nothing about what it contains. Can this be a part of crypto or is it more generic than that?

MGibson1:
@Hinton it's more generic. For this improvement, we could probably move to crypto, but eventually we should create a `Base64` struct that explicitly marks a base64 encoded string. The same would be true of any non-utf8 encoded string data. That kind of thing would live in a common crate. However, if we're interested in a many-small-crates model, I could get behind something like `bitwarden-encoding` 🤷

PR #143, opened 2025-02-05
expose cryptography primitives to wasm

MGibson1:
## 📔 Objective

This is a stop-gap to allow encryption changes to be developed only in the rust SDK and used from the javascript clients.

## ⏰ Reminders before review

- CI builds passed
  team

## 🦮 Reviewer guidelines


crates/bitwarden-wasm-internal/src/client.rs
```diff
@@ -27,6 +27,56 @@ fn convert_level(level: LogLevel) -> Level {
     }
 }
 
+#[wasm_bindgen]
+pub struct BitwardenPure;
+
+#[wasm_bindgen]
+impl BitwardenPure {
```
quexten: Since this is pub, could we get some docs on it so it's clear that teams should not be consuming this directly?

crates/bitwarden-wasm-internal/src/client.rs
```diff
@@ -27,6 +27,56 @@ fn convert_level(level: LogLevel) -> Level {
...
+        Self::setup_once();
+        env!("SDK_VERSION").to_owned()
+    }
+
+    pub fn echo(msg: String) -> String {
+        Self::setup_once();
+        msg
+    }
+
+    pub fn throw(msg: String) -> Result<(), TestError> {
```
quexten:
```suggestion
```

crates/bitwarden-wasm-internal/src/client.rs
```diff
@@ -27,6 +27,56 @@ fn convert_level(level: LogLevel) -> Level {
...
+pub struct BitwardenPure;
+
+#[wasm_bindgen]
+impl BitwardenPure {
+    pub fn version() -> String {
+        Self::setup_once();
+        env!("SDK_VERSION").to_owned()
+    }
+
+    pub fn echo(msg: String) -> String {
```
quexten:
```suggestion
```

Seems like left over debugging code?

quexten (changes requested):

crates/bitwarden-wasm-internal/src/client.rs
```diff
@@ -27,6 +31,56 @@ fn convert_level(level: LogLevel) -> Level {
...
+
+    pub fn symmetric_decrypt(
+        enc_string: String,
+        key_b64: String,
+    ) -> Result<String, PureCryptoError> {
+        Self::setup_once();
+        pure_crypto::symmetric_decrypt(enc_string, key_b64)
+    }
+
+    pub fn symmetric_decrypt_to_bytes(
```
quexten: I think we want an additional function here that decrypts the enc buffers (attachments)

crates/bitwarden-wasm-internal/src/client.rs
MGibson1: No I included those because they are present in the `BitwardenClient`, but more appropriately part of a pure function library.

MGibson1:
I decided to take your recommendation on the helper functions and separated the `BitwardenPure` namespace into different purposes.

this left `bitwarden_pure` as a reasonable place to put the required startup code to link panics and errors to console.

crates/bitwarden-wasm-internal/src/crypto.rs
```diff
@@ -8,8 +8,113 @@ use bitwarden_core::{
...
+
+        Ok(enc_string.decrypt_with_key(&key)?)
+    }
+
+    /// Stopgap method providing access to decryption through the SDKs handling of
+    /// [bitwarden_crypto::EncString] and [bitwarden_crypto::SymmetricCryptoKey]
+    ///
+    /// This method is intended for use in the javascript clients at the EncryptService layer and
+    /// should not be used elsewhere.
+    pub fn symmetric_decrypt_to_bytes(
```
quexten: Nice, change look good. One more thing: How would cases where the encrypted version is an encstring but the decrypted version is a buffer be handeled here? So f.e a userkey is an encstring, but would not necessarily be valid utf8. Would the caller need to convert their encstring to the buffer representation first? Do we want a separate function here?

quexten (dismissed):

crates/bitwarden-wasm-internal/src/crypto.rs
```diff
@@ -8,8 +8,113 @@ use bitwarden_core::{
...
+    }
+
+    /// Stopgap method providing access to encryption through the SDKs handling of
+    /// [bitwarden_crypto::EncString]
+    ///
+    /// Encrypts cleartext strings to string-serialized (base64) [EncString]s.
+    ///
+    /// This method is intended for use in the javascript clients at the EncryptService layer and
+    /// should not be used elsewhere.
+    pub fn symmetric_encrypt(plain: String, key_b64: String) -> Result<String, PureCryptoError> {
```
quexten: Optional: encrypt_bytes? (So a userkey can be encrypted)

Hinton:
Requesting a review from platform since this is named `PureClient` implying usages beyond crypto and therefore platform ownership.

crates/bitwarden-wasm-internal/src/crypto.rs
Hinton, line 1: This diverges from the established convention of just having the wasm-internal crate be a wasm wrapper.

coroiu: I agree with Oscar here, I suggest creating a new crate `bitwarden-pure-crypto` or moving `PureCrypto` to the existing `bitwarden-crypto`. You can also use conditional compilation for the functions if you only want these functions to be available in wasm

MGibson1:
> Requesting a review from platform since this is named PureClient implying usages beyond crypto and therefore platform ownership.

There's no `PureClient`, just the crypto one for now. I do think that the [helper functions present in the `BitwardenClient` ](https://github.com/bitwarden/sdk-internal/blob/b24c9736c10b8b56857b1faded309faf363d5e76/crates/bitwarden-wasm-internal/src/client.rs#L62-L74) make more sense to be part of exposed pure functions.

crates/bitwarden-wasm-internal/src/crypto.rs
MGibson1: I don't understand. How is this not a wrapper? The only thing it does is convert a few types because we don't necessarily want to write that conversion in javascript.

A crate doing that seems entirely out of proportion to what it would be handling. In that case we'd have a crate that calls a method that literally just converts a string to a byte array to call the actual cryptographic implementation. What's more, we would explicitly not want to use that crate for anything other than WASM bidings. The `bitwarden-crypto` crate also shouldn't be polluted with the weak typings necessary for javascript interoperability.

Are you disagreeing to this being the crypto file?

crates/bitwarden-wasm-internal/src/crypto.rs
```diff
@@ -8,8 +8,113 @@ use bitwarden_core::{
...
+use bitwarden_error::bitwarden_error;
+use thiserror::Error;
 use wasm_bindgen::prelude::*;
 
+#[bitwarden_error(flat)]
+#[derive(Debug, Error)]
+pub enum PureCryptoError {
+    #[error("Cryptography error, {0}")]
+    Crypto(#[from] bitwarden_crypto::CryptoError),
+}
```
coroiu: suggestion: just add `bitwarden_error` to `bitwarden_crypto::CryptoError` and return that directly

crates/bitwarden-wasm-internal/src/crypto.rs
coroiu: It uses multiple types to create a utility function which might be useful outside of WASM. I just think that we shouldn't default to putting these in the wasm crate and instead opt for putting them in a common place such as the `bitwarden-crypto` crate.

> The bitwarden-crypto crate also shouldn't be polluted with the weak typings necessary for javascript interoperability.

We're actually actively trying to get rid of as much as possible of the wasm crate and instead try to keep everything related to the same feature closer together.

crates/bitwarden-wasm-internal/src/client.rs
```diff
@@ -27,6 +27,19 @@ fn convert_level(level: LogLevel) -> Level {
...
+pub(crate) mod bitwarden_pure {
+    use super::*;
+
+    pub(crate) fn setup_once() {
+        console_error_panic_hook::set_once();
+        let log_level = convert_level(LogLevel::Info);
+        if let Err(_e) = console_log::init_with_level(log_level) {
+            set_max_level(log_level.to_level_filter())
+        }
+    }
```
coroiu: issue: Can we move/remove this? We want to eventually add more control over the initialization process and this will make it much harder to coordinate that. I don't think calling `symmetric_decrypt()` should implicitly mean setting the logging level to `Info`. For context I don't think setting it in `BitwardenClient::new` is the right way either. I like what you did with the `SdkLoadService` and I think we should try to move the `init` there instead. I also think that consumers of the SDK in the clients might need to depend on the SDK initialization somehow and not just assume that WASM has been initialized by the time their function runs

coroiu:
@MGibson1 Just for some context, why not just put these functions in their own `Client` and expose them through the `BitwardenClient`? They could still be "pure" in the sense that they don't use any state and it would follow existing patterns. Is it because the call would become `async` in js? Or is it because you think this has static typing guarantees about not needing a logged in user? We are working together with Auth on solving that btw

crates/bitwarden-wasm-internal/src/crypto.rs
quexten: > It uses multiple types to create a utility function which might be useful outside of WASM 

But the explicit purpose of the functions introduced in this PR is a stop-gap for javascript clients that can't use the SDK yet for encrypting / decrypting all domains, but need support for e.g. encstrings for new ciphertypes not available in typescript.
It really has no use outside of WASM/typescript; the intent is to remove it again as soon as *all* domains have been moved.

coroiu: Ok so you just want to expose the existing `bitwarden-crypto` crate to WASM? Just add the methods to the existing `CryptoClient` :) If you want to separate it from the existing stuff maybe create a sub-client called `TemporaryCryptoClient`, `RawCryptoClient`, `SubtleCryptoClient` or maybe even `PureCryptoClient`. I'm not gonna block the addition of a separate `PureCrypto` struct (I'm adding a similar standalone `IpcManager`) but if you go that route then I don't think it should be calling `setup_once()` everywhere (see comment further down)

crates/bitwarden-wasm-internal/src/client.rs
MGibson1: This is kind of tricky.

The issue is that these methods are otherwise pure and don't need to depend on any previous calls or state. In the case of the panic hook, I think what is here is fine. The documentation specifically calls out subsequent invocations doing nothing.

All that can be done with your suggesting is taking an optional `LogLevel` in the signature of the pure functions and passing it along here. I didn't think that was worthwhile.

MGibson1:
The logged in user guarantees are worrying in general, but aren't really motivations here.

These aren't in the `BitwardenClient` because they don't need the added complexity associated with initializing a user-bound client. The intent is to temporarily expose cryptographic primitives to the JS clients while those clients are still in control of keys. The situation is distinct from the mobile clients, because the SDK is in control of user keys in those applications.

As a background, a new `EncString` type is coming that will be implemented only in the SDK. However, we want to meter transition to this type slowly and allow for clients to decrypt these values long before we actually turn on their use. These methods will be used to tie into the SDKs `EncString` processing while allowing the JS clients to maintain control of user keys. By exposing primitive methods that don't need any external state or setup, we can enable JS clients to handle the new crypto type, and use a feature flag to turn on creating them in JS.

crates/bitwarden-wasm-internal/src/crypto.rs
MGibson1: This is not an error we want to expose long term. I don't think there's a need to compile this macro's expansion for all crates depending on crypto

coroiu (dismissed):

crates/bitwarden-wasm-internal/src/crypto.rs
Hinton, line 1: **suggestion:** We can move all of the pure-crypto content into a separate file, which will make it easier to remove later. And add a file level comment with disclaimers to not re-use this pattern which should make it explicitly clear this is not a sanctioned pattern.

Problems with this pattern:
- Logic in bitwarden-wasm-internal
- untested code.

crates/bitwarden-wasm-internal/Cargo.toml
```diff
@@ -26,6 +26,7 @@ console_log = { version = "1.0.0", features = ["color"] }
 js-sys = "0.3.68"
 log = "0.4.20"
 serde_json = ">=1.0.96, <2.0"
+thiserror.workspace = true
```
Hinton: **nit:** We generally use this syntax everywhere else.
```suggestion
thiserror = { workspace = true }
```

crates/bitwarden-wasm-internal/src/crypto.rs
Hinton: We've already added `bitwarden_error` to `CryptoError` since we now explicitly expose the underlying errors instead of the grouped error struct.

I don't see a reason to keep the `PureCryptoError` any longer.

Hinton (changes requested):
I want a more clear disclaimer that this pattern is forbidden from propagating to make it clear to not copy this behaviour in the future.

coroiu (dismissed):

crates/bitwarden-wasm-internal/src/pure_crypto.rs
```diff
@@ -0,0 +1,116 @@
+
```
Hinton: Linting seems to be failing.

```suggestion
```

Hinton (approved):

PR #149, opened 2025-02-06
[PM-15556] Build TOTP toString

gbubemismith:
## 🎟️ Tracking
https://bitwarden.atlassian.net/browse/PM-15556

## 📔 Objective
The SDK currently supports going from an otpauth string to a Totp struct. However the Credential Exchange protocol provides otp components, and in order to save the items it must be converted to an otpauth string.

This should be done by implement the ToString trait for Totp.

The format is defined in [URI string format](https://docs.yubico.com/yesdk/users-manual/application-oath/uri-string-format.html) .

## ⏰ Reminders before review

- CI builds passed
  team

## 🦮 Reviewer guidelines

### DIFF
--- a/crates/bitwarden-vault/Cargo.toml
+++ b/crates/bitwarden-vault/Cargo.toml
@@ -28,6 +28,7 @@ bitwarden-core = { workspace = true, features = ["internal"] }
 bitwarden-crypto = { workspace = true }
 bitwarden-error = { workspace = true }
 chrono = { workspace = true }
+data-encoding = "2.7.0"
 hmac = ">=0.12.1, <0.13"
 rand = ">=0.8.5, <0.9"
 reqwest = { workspace = true }
--- a/crates/bitwarden-vault/Cargo.toml
+++ b/crates/bitwarden-vault/Cargo.toml
@@ -17,7 +17,7 @@ keywords.workspace = true
 uniffi = [
     "bitwarden-core/uniffi",
     "bitwarden-crypto/uniffi",
-    "dep:uniffi",
+    "dep:uniffi"
 ] # Uniffi bindings
 wasm = ["dep:tsify-next", "dep:wasm-bindgen"] # WASM support
 
--- a/crates/bitwarden-vault/Cargo.toml
+++ b/crates/bitwarden-vault/Cargo.toml
@@ -17,7 +17,7 @@ keywords.workspace = true
 uniffi = [
     "bitwarden-core/uniffi",
     "bitwarden-crypto/uniffi",
-    "dep:uniffi"
+    "dep:uniffi",
 ] # Uniffi bindings
 wasm = ["dep:tsify-next", "dep:wasm-bindgen"] # WASM support
 
@@ -28,7 +28,7 @@ bitwarden-core = { workspace = true, features = ["internal"] }
 bitwarden-crypto = { workspace = true }
 bitwarden-error = { workspace = true }
 chrono = { workspace = true }
-data-encoding = "2.7.0"
+data-encoding = ">=2.0, <3"
 hmac = ">=0.12.1, <0.13"
 rand = ">=0.8.5, <0.9"
 reqwest = { workspace = true }
--- a/crates/bitwarden-vault/src/totp.rs
+++ b/crates/bitwarden-vault/src/totp.rs
@@ -91,7 +91,7 @@ pub fn generate_totp_cipher_view(
     generate_totp(key, time)
 }
 
-#[derive(Clone, Copy, Debug, PartialEq)]
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
 pub enum TotpAlgorithm {
     Sha1,
     Sha256,
--- a/crates/bitwarden-vault/src/totp.rs
+++ b/crates/bitwarden-vault/src/totp.rs
@@ -589,7 +589,7 @@ mod tests {
     }
 
     #[test]
-    fn test_totp_to_string_period() {
+    fn test_totp_to_string_with_custom_period() {
         let totp = Totp {
             account: Some("test@bitwarden.com".to_string()),
             algorithm: DEFAULT_ALGORITHM,
@@ -623,7 +623,7 @@ mod tests {
     }
 
     #[test]
-    fn test_totp_to_string_issuer_with_spaces() {
+    fn test_totp_to_string_encodes_spaces_in_issuer() {
         let totp = Totp {
             account: Some("test@bitwarden.com".to_string()),
             algorithm: DEFAULT_ALGORITHM,
@@ -640,7 +640,7 @@ mod tests {
     }
 
     #[test]
-    fn test_totp_to_string_issuer_special_characters() {
+    fn test_totp_to_string_encodes_special_characters_in_issuer() {
         let totp = Totp {
             account: Some("test@bitwarden.com".to_string()),
             algorithm: DEFAULT_ALGORITHM,
--- a/crates/bitwarden-vault/src/totp.rs
+++ b/crates/bitwarden-vault/src/totp.rs
@@ -260,7 +260,8 @@ impl fmt::Display for Totp {
     /// Formats the TOTP as an OTP Auth URI.
     ///
     /// Returns a steam::// URI if the algorithm is Steam.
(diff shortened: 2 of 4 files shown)


crates/bitwarden-vault/Cargo.toml
```diff
@@ -28,6 +28,7 @@ bitwarden-core = { workspace = true, features = ["internal"] }
 bitwarden-crypto = { workspace = true }
 bitwarden-error = { workspace = true }
 chrono = { workspace = true }
+data-encoding = "2.7.0"
```
Hinton: **issue:** Please use a range here as per https://contributing.bitwarden.com/architecture/sdk/dependencies#explicit-ranges.

`>=2.0, <3.0` is probably good.

crates/bitwarden-vault/src/totp.rs
```diff
@@ -90,7 +91,7 @@ pub fn generate_totp_cipher_view(
     generate_totp(key, time)
 }
 
-#[derive(Clone, Copy, Debug)]
+#[derive(Clone, Copy, Debug, PartialEq)]
```
Hinton: It's generally a good idea to also derive `Eq` when deriving `PartialEq` https://rust-lang.github.io/rust-clippy/master/index.html#derive_partial_eq_without_eq

crates/bitwarden-vault/src/totp.rs
```diff
@@ -174,6 +177,7 @@ impl FromStr for Totp {
             let parts: HashMap<_, _> = url.query_pairs().collect();
 
             Totp {
+                account: None,
```
Hinton: **suggestion:** Can we add some TODO's here linking to PM-15389

crates/bitwarden-vault/src/totp.rs
```diff
@@ -377,6 +435,71 @@ mod tests {
...
 
+    #[test]
+    fn test_totp_to_string_with_defaults() {
+        let totp = Totp {
+            account: Some("test@bitwarden.com".to_string()),
+            algorithm: DEFAULT_ALGORITHM,
+            digits: DEFAULT_DIGITS,
+            issuer: Some("Example".to_string()),
+            period: DEFAULT_PERIOD,
+            secret: vec![180, 17, 13, 116, 49, 86, 112, 36, 215, 15],
```
Hinton: **suggestion:** Can we hard-code the values here? They generally shouldn't change and relying on the test breaking should make this slightly more obvious.

crates/bitwarden-vault/src/totp.rs
```diff
@@ -377,6 +435,71 @@ mod tests {
...
+            algorithm: TotpAlgorithm::Steam,
+            digits: 5,
+            issuer: None,
+            period: DEFAULT_PERIOD,
+            secret: vec![1, 2, 3, 4],
+        };
+        let secret_b32 = BASE32_NOPAD.encode(&totp.secret);
+        assert_eq!(totp.to_string(), format!("steam://{}", secret_b32));
+    }
+
```
Hinton, line 718: **issue:** Can we add some test for the more complex scenarios described in https://docs.yubico.com/yesdk/users-manual/application-oath/uri-string-format.html? Such as ensuring the label is url encoded.

crates/bitwarden-vault/src/totp.rs
```diff
@@ -220,6 +229,55 @@ impl FromStr for Totp {
...
+            }
+
+            if self.digits != DEFAULT_DIGITS {
+                pairs.append_pair("digits", &self.digits.to_string());
+            }
+
+            pairs.finish();
+        }
+
+        write!(f, "{}", url)
```
Hinton: **suggestion:** We can use https://docs.rs/url/latest/url/struct.Url.html#method.fmt here to avoid needing a macro.

```suggestion
        url.fmt(f)
```

crates/bitwarden-vault/src/totp.rs
```diff
@@ -220,6 +229,55 @@ impl FromStr for Totp {
...
+
+            if self.period != DEFAULT_PERIOD {
+                pairs.append_pair("period", &self.period.to_string());
+            }
+
+            if self.algorithm == TotpAlgorithm::Sha256 {
+                pairs.append_pair("algorithm", "SHA256");
+            } else if self.algorithm == TotpAlgorithm::Sha512 {
+                pairs.append_pair("algorithm", "SHA512");
+            }
```
Hinton: **thought (non blocking):** We could implement `Display` on `TotpAlgorithm` and use that to generate the algorithm. Which would result in something like 

```rust
if self.algorithm != DEFAULT_ALGORITHM {
    pairs.append_pair("algorithm", &self.algorithm.to_string());
}
```

crates/bitwarden-vault/src/totp.rs
```diff
@@ -220,6 +229,55 @@ impl FromStr for Totp {
...
+
+            if self.digits != DEFAULT_DIGITS {
+                pairs.append_pair("digits", &self.digits.to_string());
+            }
+
+            pairs.finish();
+        }
```
Hinton: **suggestion:** We could define a local function `url_params`

```rust
fn url_params(totp: &Totp) -> Vec<(&str, Cow<str>)> {
    let secret_b32 = BASE32_NOPAD.encode(&totp.secret);

    let mut params: Vec<(&str, Cow<str>)> = vec![("secret", secret_b32.into())];

    if let Some(issuer) = &totp.issuer {
        params.push(("issuer", issuer.into()));
    }

    if totp.period != DEFAULT_PERIOD {
        params.push(("period", totp.period.to_string().into()));
    }

    if totp.algorithm == TotpAlgorithm::Sha256 {
        params.push(("algorithm", "SHA256".into()));
    } else if totp.algorithm == TotpAlgorithm::Sha512 {
        params.push(("algorithm", "SHA512".into()));
    }

    if totp.digits != DEFAULT_DIGITS {
        params.push(("digits", totp.digits.to_string().into()));
    }

    params
}
```

And then use with `parse_with_params` which removes the need for the manual scope.
```rust
let mut url =
  Url::parse_with_params("otpauth://totp", url_params(self)).map_err(|_| fmt::Error)?;
```

Hinton (commented):
Looks like a solid start. I've noted a few things I would like resolved.

crates/bitwarden-vault/src/totp.rs
Hinton: Is there a particular reason we started building the query string manually? I don't believe that will properly escape url characters.

gbubemismith: I had to switch to manually building the query cause reqwest's URl implementation doesn't expose an API to control how spaces are encoded in query params. query_pairs_mut always uses '+' for encoding spaces

Hinton:
I think we should add some tests confirming we can re-parse it after converting to string. To ensure we're not loosing any data along the way.

We also probably also need to
- Escape account name as well 😩.
- Escape `:` values in issuer or account name?

gbubemismith:
> I think we should add some tests confirming we can re-parse it after converting to string. To ensure we're not loosing any data along the way.
> 
> We also probably also need to
> 
> * Escape account name as well 😩.
> * Escape `:` values in issuer or account name?

I think reparsing shouldn't be an issue cause we use `percent_decode_str`, but tests would add extra confidence and `:` doesn't need to be encoded

crates/bitwarden-vault/src/totp.rs
```diff
@@ -239,6 +256,67 @@ impl FromStr for Totp {
     }
 }
 
+impl fmt::Display for Totp {
+    /// Formats the TOTP as an OTP Auth URI.
+    ///
+    /// Returns a steam::// URI if the algorithm is Steam.
+    /// Otherwise returns an otpauth:// URI according to the Key Uri Format Specification:  <https://docs.yubico.com/yesdk/users-manual/application-oath/uri-string-format.html>
```
Hinton: **nit:** We attempt to keep the line width to ~ 100 chars. Fmt doesn't automatically wrap comments with links in it.
```suggestion
    /// Otherwise returns an otpauth:// URI according to the Key Uri Format Specification:
    /// <https://docs.yubico.com/yesdk/users-manual/application-oath/uri-string-format.html>
```

Hinton (commented):
A documentation nit and I think we're good.

Hinton (approved):

PR #150, opened 2025-02-07
[PM-15096] Implement xchacha20-poly1305 cipher functions

quexten:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-15096

## 📔 Objective

Implements xchacha20-poly1305 (https://libsodium.gitbook.io/doc/secret-key_cryptography/aead/chacha20-poly1305/xchacha20-poly1305_construction) as a new encryption type.

Further, this introduces a dependency patch override (to https://github.com/bitwarden/rustcrypto-formats.git at revision 2b27c63034217dd126bbf5ed874da51b84f8c705) for rustcrypto formats, until this gets updated upstream to fix a type incompatibility when building on WASM and with chacha as a dependency.

These functions are unused until a follow-up PR (soon).

## ⏰ Reminders before review

- CI builds passed
  team

## 🦮 Reviewer guidelines


crates/bitwarden-crypto/Cargo.toml
```diff
@@ -29,12 +29,14 @@ argon2 = { version = ">=0.5.0, <0.6", features = [
...
 bitwarden-error = { workspace = true }
 cbc = { version = ">=0.1.2, <0.2", features = ["alloc", "zeroize"] }
+chacha20poly1305 = { version = "0.10.1" }
 generic-array = { version = ">=0.14.7, <1.0", features = ["zeroize"] }
 hkdf = ">=0.12.3, <0.13"
 hmac = ">=0.12.1, <0.13"
 num-bigint = ">=0.4, <0.5"
 num-traits = ">=0.2.15, <0.3"
 pbkdf2 = { version = ">=0.12.1, <0.13", default-features = false }
+poly1305 = { version = "0.8.0", features = ["zeroize"] }
```
dani-garcia: This crate doesn't seem to be used at the moment, I assume it's used in your other PR with the key-commiting implementation? If so we can probably remove it for now.

crates/bitwarden-crypto/src/xchacha20.rs
```diff
@@ -0,0 +1,137 @@
...
+ *
+ * If multiple keys are possible, a key-committing cipher such as
+ * XChaCha20Poly1305Blake3CTX should be used: `https://github.com/bitwarden/sdk-internal/pull/41` to prevent invisible-salamander style attacks.
+ * `https://eprint.iacr.org/2019/016.pdf`
+ * `https://soatok.blog/2024/09/10/invisible-salamanders-are-not-what-you-think/`
+ */
+use crate::CryptoError;
+
+#[allow(unused)]
+pub(crate) fn generate_nonce() -> GenericArray<u8, U24> {
```
dani-garcia: Is this function only meant to be used for tests? If so, it should be moved inside `mod tests {`

crates/bitwarden-crypto/src/xchacha20.rs
```diff
@@ -0,0 +1,137 @@
...
+//!
+//! Contains low level XChaCha20Poly1305 operations used by the rest of the crate.
+//!
+//! In most cases you should use the [EncString][crate::EncString] with
+//! [KeyEncryptable][crate::KeyEncryptable] & [KeyDecryptable][crate::KeyDecryptable] instead.
+
+use chacha20poly1305::{AeadCore, AeadInPlace, KeyInit, XChaCha20Poly1305};
+use generic_array::{typenum::U24, GenericArray};
+
+/**
```
dani-garcia: Small nit, but I would move this comment to the top of the file and use `//!` like on the other one. As it is right now, we're technically adding a doc comment to `use crate::CryptoError`

crates/bitwarden-crypto/src/lib.rs
```diff
@@ -91,6 +91,7 @@ pub use wordlist::EFF_LONG_WORD_LIST;
 mod store;
 pub use store::{KeyStore, KeyStoreContext};
 mod traits;
+mod xchacha20;
```
dani-garcia, line 94: This is more of an idea for the future, but it would be nice to think about moving the `aes`, `rsa` and the new `xchacha20` modules into their own separate low level algorithms module so they're not all exposed in the root.

crates/bitwarden-crypto/src/xchacha20.rs
```diff
@@ -0,0 +1,137 @@
...
+use crate::CryptoError;
+
+#[allow(unused)]
+pub(crate) fn generate_nonce() -> GenericArray<u8, U24> {
+    XChaCha20Poly1305::generate_nonce(rand::thread_rng())
+}
+
+#[allow(unused)]
+pub(crate) fn encrypt_xchacha20_poly1305(
+    nonce: &[u8; 24],
```
dani-garcia: On the `aes` module we generate the `IV` internally to ensure we always generate a random one each time for encryption, assuming that the `nonce` works the same in `chacha20`, have you considered not exposing this as a parameter for encryption and just generating it internally?

crates/bitwarden-crypto/src/xchacha20.rs
```diff
@@ -0,0 +1,137 @@
...
+    key: &[u8; 32],
+    plaintext_secret_data: &[u8],
+    associated_data: &[u8],
+) -> Result<Vec<u8>, CryptoError> {
+    // This buffer contains the plaintext, that will be encrypted in-place
+    let mut buffer = Vec::from(plaintext_secret_data);
```
dani-garcia: Minor nit: In `decrypt_xchacha20_poly1305` we're using `slice.to_vec()`, while here we're using `Vec::from(slice)`. We should try to be consistent and use one or the other.

crates/bitwarden-crypto/src/xchacha20.rs
```diff
@@ -0,0 +1,137 @@
...
+    associated_data: &[u8],
+) -> Result<Vec<u8>, CryptoError> {
+    let mut buffer = ciphertext.to_vec();
+    XChaCha20Poly1305::new(GenericArray::from_slice(key))
+        .decrypt_in_place(
+            GenericArray::from_slice(nonce),
+            associated_data,
+            &mut buffer,
+        )
+        .map_err(|_| CryptoError::InvalidKey)?;
```
dani-garcia: I think for aes and rsa we use the `CryptoError::KeyDecrypt` error when decrypting. If so we should do the same here as well

crates/bitwarden-crypto/src/xchacha20.rs
```diff
@@ -0,0 +1,137 @@
...
+) -> Result<Vec<u8>, CryptoError> {
+    // This buffer contains the plaintext, that will be encrypted in-place
+    let mut buffer = Vec::from(plaintext_secret_data);
+    XChaCha20Poly1305::new(GenericArray::from_slice(key))
+        .encrypt_in_place(
+            GenericArray::from_slice(nonce),
+            associated_data,
+            &mut buffer,
+        )
+        .map_err(|_| CryptoError::InvalidKey)?;
```
dani-garcia: What's the error case for chacha20 encryption? Key and nonce sizes are checked at compile time, and as far as I know, chacha20 doesn't have any specific key format requirements.

If the only error case is the buffer not having enough space, I think we should be fine calling `.expect("")` and not returning a Result.

crates/bitwarden-crypto/src/xchacha20.rs
quexten: Hmm, initially I thought we needed the IV to be included in the authenticated data, but I checked the spec of the format we want to follow for the new keys (COSE https://datatracker.ietf.org/doc/html/rfc8152), and the IV/nonce does *not* need to be authenticated, so it could just be generated during encrypt. We do however need to serialize it onto the encstring. So What would the recommendation here be, return a tuple / struct of Nonce + Ciphertext, and generate the nonce in the encrypt function?

crates/bitwarden-crypto/src/xchacha20.rs
quexten: As mentioned for the nonce generating function, initially I thought the IV needed to be authenticated, but this is not the case, so I'll move this to generate during encrypt.

crates/bitwarden-crypto/Cargo.toml
quexten: Weird, for chacha the CI pipeline complained about an unused dependency, for poly1305 it did not.

crates/bitwarden-crypto/src/xchacha20.rs
dani-garcia: That's more or less what we do for aes now yeah, so it seems reasonable to do the same for chacha

```rust
pub(crate) fn encrypt_aes256_hmac(
    data_dec: &[u8],
    mac_key: &GenericArray<u8, U32>,
    key: &GenericArray<u8, U32>,

) -> Result<([u8; 16], [u8; 32], Vec<u8>)> {
    /// IV is generated and returned with the encrypted data and the MAC as a tuple
    Ok((iv, mac, data))
}
```

quexten: Updated!

dani-garcia (approved):

MGibson1 (commented):
My only note is that I kind of want to see the interface one level up, as well.

I say this because we _could_ design certain aspects of the AAD set into this level. I'm not set definitively in either direction, but there's an elegance to tying the cipherText to a key and cipher type at this level, rather than in the `Encryptable`

quexten:
@MGibson1 Let's revisit this on the cose PR. I do at this point believe that COSE is better suited though, if we want a structured format to add our own additional data.

PR #155, opened 2025-02-10
Fix rustls-platform-verifier initialization

dani-garcia:
## 📔 Objective

Update rustls-platform-verifier initialization on Android to support being manually loaded using `System.loadLibrary()`. This bypasses the need to dynamically load `JNI_GetCreatedJavaVMs`, which is not available for APIs <31.

I've still left the original code that calls `JNI_GetCreatedJavaVMs` as a fallback, but once the Android APP is on minAPI >31, we should just statically call `JNI_GetCreatedJavaVMs` using `jni::sys::JNI_GetCreatedJavaVMs`. 

## ⏰ Reminders before review

- CI builds passed
  team

## 🦮 Reviewer guidelines


crates/bitwarden-uniffi/kotlin/publish-local.sh
```diff
@@ -12,7 +12,7 @@ cross build -p bitwarden-uniffi --release --target=aarch64-linux-android
 mv $SDK_REPO_ROOT/target/aarch64-linux-android/release/libbitwarden_uniffi.so ./sdk/src/main/jniLibs/arm64-v8a/libbitwarden_uniffi.so
 
 # Build other archs
-if [ $1 = "all" ]; then
+if [ "$1" = "all" ]; then
```
dani-garcia, line 15: I was getting a warning in the console when running this, so I fixed it here as well, I can split to another PR if preferred

crates/bitwarden-uniffi/src/lib.rs
```diff
@@ -107,6 +107,8 @@ fn init_logger() {
 
     #[cfg(target_os = "android")]
     android_logger::init_once(
-        android_logger::Config::default().with_max_level(uniffi::deps::log::LevelFilter::Info),
+        android_logger::Config::default()
+            .with_tag("com.bitwarden.sdk")
```
dani-garcia, line 111: Added this to make sure all the SDK logs consistently use the same tag, for easier filtering. Not necessary for this fix

crates/bitwarden-uniffi/kotlin/publish-local.sh
Hinton: Should be fine.

crates/bitwarden-uniffi/src/android_support.rs
```diff
@@ -1,22 +1,43 @@
...
         Ok(())
     }
 
-    ANDROID_INIT.call_once(|| {
-        if let Err(e) = init_inner() {
-            log::error!("Failed to initialize Android support: {}", e);
-        }
-    });
+    if let Err(e) = init_inner() {
+        log::error!("Failed to initialize Android support: {:#?}", e);
```
Hinton, line 39: **question:** I don't suppose we can somehow expose this error to the native apps?

Hinton (approved):

crates/bitwarden-uniffi/src/android_support.rs
dani-garcia: Not easily, if we changed `Client::new` to return a result that will throw an exception in the apps and probably crash them, so it's probably a bad idea. We could store the result in the client and provide a `Client::check_android_init() -> Result<(), Error>` that just returns the stored error, but that would require support from the apps.

PR #157, opened 2025-02-11
[PM-17430] Implement new key format

quexten:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-17430

## 📔 Objective

Adds a new serialization format for symmetric keys. _**Read only in this PR**._

Historically, keys have been encoded directly as byte arrays, without describing their type. They can be categorized by size only; a 32 byte key is for encrypting AES256-CBC (No HMAC), a 64 byte key is for encrypting AES256-CBC-HMAC.

To distinguish these, the new key format now self-describes the key type. The struct contains the key type, and the key data, and is serialized into the final byte array using cbor. Cbor serialization is also used in e.g. passkeys.

Since it *could* be the case that a 32 byte cbor serialized key with type information ends up being 64 bytes, padding is introduced, ensuring the keys are always 65 or more bytes.

Specifically, the padding looks as follows:
```
N|0x00...0x00|KEY_BYTES
  ^ N bytes
```
where N is chosen such that the padded key is at least 65 bytes, ensuring that it cannot be confused with other keys.

The key itself looks as follows:
```
#[derive(Clone, Serialize, Deserialize)]
#[serde(tag = "alg")]
enum SerializedSymmetricCryptoKey {
    #[serde(rename = "A256C")]
    Aes256Cbc {
        #[serde(with = "serde_bytes", rename = "ek")]
        encryption_key: Vec<u8>,
    },
    #[serde(rename = "A256C-H")]
    Aes256CbcHmac {
        #[serde(with = "serde_bytes", rename = "ek")]
        encryption_key: Vec<u8>,
        #[serde(with = "serde_bytes", rename = "ak")]
        authentication_key: Vec<u8>,
    },
}

```

The considerations here are as follows:

https://bitwarden.atlassian.net/wiki/spaces/EN/pages/1557561420/Key+Crypto+Serialization+Format

- The encoding size matters, thus a format that serializes cryptographic ciphertexts as byte arrays well is chosen, rather than a format that needs text-based (double) encoding to b64. In this case we already use cbor, and it is a well-established format. It self-describes the structure in contrast to e.g. protobuf, which means there are no implicit assumptions in the format made by the encoding side, but explicitly the format is described.
for encoding cryptographic ciphertexts or keys
- The algorithm choices are abbreviated to save space during serialization
- The key (algorithm) field is shortened to "alg" in the serialization

The "algorithm" field allows adding as many encryption algorithms (or formats) as needed in the future, since it would just be another entry in the enum.

> [!NOTE]  
> This implements PartialEq for symmetric keys and makes it constant time. This is somewhat out of scope of this PR but was required to fix a section of code that otherwise could break later, when we introduce new keys that *only* use the new serialization format.

## ⏰ Reminders before review

- CI builds passed
  team

## 🦮 Reviewer guidelines

### DIFF
--- a/crates/bitwarden-crypto/src/keys/symmetric_crypto_key.rs
+++ b/crates/bitwarden-crypto/src/keys/symmetric_crypto_key.rs
@@ -167,8 +167,7 @@ impl TryFrom<&mut [u8]> for SymmetricCryptoKey {
         } else if value.len() > Self::AES256_CBC_HMAC_KEY_LEN {
             let unpadded_value = unpad_key(value);
             let decoded_key: SerializedSymmetricCryptoKey =
-                ciborium::from_reader(unpadded_value)
-                    .map_err(|_| CryptoError::EncodingError)?;
+                ciborium::from_reader(unpadded_value).map_err(|_| CryptoError::EncodingError)?;
             value.zeroize();
             SymmetricCryptoKey::try_from(decoded_key)
         } else {
@@ -201,9 +200,11 @@ impl TryFrom<SerializedSymmetricCryptoKey> for SymmetricCryptoKey {
                 enc_key: Box::pin(*GenericArray::from_slice(encryption_key.as_slice())),
                 mac_key: Box::pin(*GenericArray::from_slice(authentication_key.as_slice())),
             })),
-            SerializedSymmetricCryptoKey::Aes256Cbc { encryption_key } => Ok(SymmetricCryptoKey::Aes256CbcKey(Aes256CbcKey {
-                enc_key: Box::pin(*GenericArray::from_slice(encryption_key.as_slice())),
-            })),
+            SerializedSymmetricCryptoKey::Aes256Cbc { encryption_key } => {
+                Ok(SymmetricCryptoKey::Aes256CbcKey(Aes256CbcKey {
+                    enc_key: Box::pin(*GenericArray::from_slice(encryption_key.as_slice())),
+                }))
+            }
         }
     }
 }
--- a/crates/bitwarden-crypto/src/keys/symmetric_crypto_key.rs
+++ b/crates/bitwarden-crypto/src/keys/symmetric_crypto_key.rs
@@ -10,20 +10,32 @@ use zeroize::{Zeroize, ZeroizeOnDrop};
 use super::key_encryptable::CryptoKey;
 use crate::CryptoError;
 
+use subtle::{Choice, ConstantTimeEq};
+
 /// Aes256CbcKey is a symmetric encryption key, consisting of one 256-bit key,
 /// used to decrypt legacy type 0 encstrings. The data is not autenticated
 /// so this should be used with caution, and removed where possible.
 #[derive(ZeroizeOnDrop, Clone)]
-#[cfg_attr(test, derive(Debug, PartialEq))]
 pub struct Aes256CbcKey {
     /// Uses a pinned heap data structure, as noted in [Pinned heap data][crate#pinned-heap-data]
     pub(crate) enc_key: Pin<Box<GenericArray<u8, U32>>>,
 }
 
+impl ConstantTimeEq for Aes256CbcKey {
+    fn ct_eq(&self, other: &Self) -> Choice {
+        self.enc_key.ct_eq(&other.enc_key)
+    }
+}
+
+impl PartialEq for Aes256CbcKey {
+    fn eq(&self, other: &Self) -> bool {
+        self.ct_eq(other).into()
+    }
+}
+
 /// Aes256CbcHmacKey is a symmetric encryption key consisting
 /// of two 256-bit keys, one for encryption and one for MAC
 #[derive(ZeroizeOnDrop, Clone)]
-#[cfg_attr(test, derive(Debug, PartialEq))]
 pub struct Aes256CbcHmacKey {
     /// Uses a pinned heap data structure, as noted in [Pinned heap data][crate#pinned-heap-data]
     pub(crate) enc_key: Pin<Box<GenericArray<u8, U32>>>,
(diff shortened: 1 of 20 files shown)


crates/bitwarden-crypto/src/keys/symmetric_crypto_key.rs
```diff
@@ -53,31 +78,48 @@ impl SymmetricCryptoKey {
...
+            (SymmetricCryptoKey::Aes256CbcHmacKey(key), false) => {
+                let mut buf = Vec::with_capacity(64);
+                buf.extend_from_slice(&key.enc_key);
+                buf.extend_from_slice(&key.mac_key);
+                Ok(buf)
+            }
+            (_, true) => {
+                let serialized_key = SerializedSymmetricCryptoKey::from(self.clone());
+                let mut encoded_key = Vec::with_capacity(1024);
+                ciborium::into_writer(&serialized_key, &mut encoded_key)
```
dani-garcia: Two things:
I don't think we need to preallocate the Vec, it seems like ciborium already does it for you according to some tests: https://github.com/enarx/ciborium/blob/f3671472f80a67008f2f90ef33b185d8d81d703d/ciborium/tests/canonical.rs#L57-L58. If we decide to preallocate, we should probably choose a smaller size, otherwise we're wasting a fair amount of space.

What are the error conditions for `ciborium::into_writer`? Having to expose a result type if it can never happen seems very annoying, specially if it's something that can't ever happen in practice. At least with `serde_json::to_string`, it can only really fail in very specific circumstances (the only case I've seen were hashmaps with non-string keys, which are invalid JSON). If the errors can only appear when `SerializedSymmetricCryptoKey` has an invalid CBOR representation, I think we should just unwrap the error and add a test to make sure it's not broken in the future.

crates/bitwarden-crypto/src/keys/symmetric_crypto_key.rs
```diff
@@ -141,6 +200,58 @@ impl std::fmt::Debug for SymmetricCryptoKey {
...
+            }
+        }
+    }
+}
+
+/// Pad a key to a minimum length;
+/// The first byte describes the number (N) of subsequently following null bytes
+/// Next, there are N null bytes
+/// Finally, the key bytes are appended
+fn pad_key(key_bytes: &[u8], min_length: usize) -> Vec<u8> {
```
dani-garcia: Does the padding need to be at the start? If we put it at the end we can modify the vector in place and avoid making new allocations.

crates/bitwarden-crypto/src/keys/symmetric_crypto_key.rs
```diff
@@ -106,23 +148,40 @@ impl TryFrom<&mut [u8]> for SymmetricCryptoKey {
...
 
             Ok(SymmetricCryptoKey::Aes256CbcKey(Aes256CbcKey { enc_key }))
+        } else if value.len() > Self::AES256_CBC_HMAC_KEY_LEN {
+            let unpadded_value = zeroize::Zeroizing::new(unpad_key(value));
+            let decoded_key: SerializedSymmetricCryptoKey =
+                ciborium::from_reader(unpadded_value.as_slice())
+                    .map_err(|_| CryptoError::EncodingError)?;
+            value.zeroize();
+
+            match decoded_key {
```
dani-garcia: This should probably be moved into an `impl TryFrom<SerializedSymmetricCryptoKey> for SymmetricCryptoKey`

crates/bitwarden-crypto/src/keys/symmetric_crypto_key.rs
```diff
@@ -141,6 +200,58 @@ impl std::fmt::Debug for SymmetricCryptoKey {
...
+    let null_bytes = min_length.saturating_sub(1).saturating_sub(key_bytes.len());
+    let mut padded_key = Vec::with_capacity(min_length);
+    padded_key.extend_from_slice(&[null_bytes as u8]);
+    padded_key.extend_from_slice(vec![0u8; null_bytes].as_slice());
+    padded_key.extend_from_slice(key_bytes);
+    padded_key
+}
+
+// Unpad a key that was padded with pad_key
+fn unpad_key(key_bytes: &[u8]) -> Vec<u8> {
```
dani-garcia: We could change the interface here to be `fn unpad_key(key_bytes: &[u8]) -> &[u8]` and avoid an extra allocation as well. This would also save us from having to wrap it on Zeroizing when using it.

crates/bitwarden-crypto/src/keys/symmetric_crypto_key.rs
quexten: That is fine. So do I understand correctly that the requested format would be:
```
payload|0x00...0x00|pad_len
             ^ pad_len times 0x00 bytes
```

dani-garcia: Yeah, that was my thinking. We can also do like `PKCS7` padding and set the pad values to `pad_len`, which simplifies the implementation a bit:
```rust
fn pad_key(key_bytes: &mut Vec<u8>, min_length: usize) {
    let pad_bytes = min_length.saturating_sub(key_bytes.len());
    if pad_bytes > 0 {
        key_bytes.resize(min_length, pad_bytes as u8);
    }
}
```

quexten: This didn't seem to work quite right. Padding should happen in all cases, for at least one byte, otherwise we cannot reliably tell whether something was padded or not.

```
/// Pad a key to a minimum length using PKCS7-like padding
fn pad_key(key_bytes: &mut Vec<u8>, min_length: usize) {
    // at least 1 byte of padding is required
    let pad_bytes = min_length.saturating_sub(key_bytes.len()).max(1);
    let padded_length = max(min_length, key_bytes.len() + 1);
    key_bytes.resize(padded_length, pad_bytes as u8);
}
```
seems to work though, with added tests for the boundary values.

crates/bitwarden-crypto/src/keys/symmetric_crypto_key.rs
quexten: Question: If we use `TryFrom` we expect error cases. What would be the error cases in this case? I can imagine `GenericArray::from_slice` failing because for whatever reason the `Vec` from the SerializedCryptoKey has the wrong size? Is there anything else?

crates/bitwarden-crypto/src/keys/symmetric_crypto_key.rs
quexten: The two conditions I found are: https://github.com/enarx/ciborium/blob/f3671472f80a67008f2f90ef33b185d8d81d703d/ciborium/src/ser/mod.rs#L417 and if the underlying writer errors (which shouldn't happen for a vec?).
So, unwrap seems fine to me.

crates/bitwarden-crypto/src/keys/symmetric_crypto_key.rs
dani-garcia: I was thinking for the `_ => Err(CryptoError::InvalidKey),` variant, but we're probably better off implementing `From` and just implementing the missing `Aes256Cbc` variant?

crates/bitwarden-crypto/src/keys/symmetric_crypto_key.rs
quexten: (Updated to unwrap and not use `Result`)

crates/bitwarden-crypto/src/keys/symmetric_crypto_key.rs
dani-garcia: Oh yep, I forgot that we always want to pad at least once, your example looks good to me.

PR #179, opened 2025-03-03
Fix MSRV build failures during CI

dani-garcia:
## 📔 Objective

I've removed the MSRV from the clippy workflow. It wasn't working correctly as the toolchain file was overriding the manually provided rust version, and with the newest rustup it was erroring as the toolchain wasn't installed (and the latest rustup won't do it automatically like before).

MSRV is still being checked by the `minimum-rust-version.yml` workflow, which I've also updated to use an explicit rust version in case the toolchain was also overriding it.

## ⏰ Reminders before review

- CI builds passed
  team

## 🦮 Reviewer guidelines

### DIFF
--- a/.github/workflows/minimum-rust-version.yml
+++ b/.github/workflows/minimum-rust-version.yml
@@ -35,13 +35,6 @@ jobs:
           RUST_TOOLCHAIN="$(grep -oP '^rust-version.*"(\K.*?)(?=")' Cargo.toml)"
           echo "RUST_TOOLCHAIN=${RUST_TOOLCHAIN}" | tee -a "${GITHUB_OUTPUT}"
 
-      - name: Set Rust Nightly Toolchain
-        id: nightly-toolchain
-        shell: bash
-        run: |
-          RUST_NIGHTLY_TOOLCHAIN="$(grep -oP '^nightly-channel.*"(\K.*?)(?=")' rust-toolchain.toml)"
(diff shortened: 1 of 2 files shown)


.github/workflows/minimum-rust-version.yml
```diff
@@ -54,4 +47,4 @@ jobs:
           key: msrv-${{ matrix.settings.target }}-cargo-${{ matrix.settings.os }}
 
       - name: Cargo check MSRV
-        run: cargo check --all-features
+        run: cargo +"${{ steps.toolchain.outputs.RUST_TOOLCHAIN }}" check --all-features
```
Hinton, line 50: Are we not using the MSRV version here?

dani-garcia: On the lint workflow the `rust-toolchain.toml` file was overriding the installed version, which was causing the errors we saw (we installed 1.75 but cargo tried to use 1.85, which wasn't installed)

I can't say for sure for sure if the same thing is happening here, but since this is the only other place were we want to override the rust version used, I figured might as well be safe and explicitly select the version.

Hinton (approved):

.github/workflows/minimum-rust-version.yml
Hinton: Ah, we read toolchain version from cargo.toml. I naively assumed it was reading rust-toolchain.toml

PR #185, opened 2025-03-11
[Innovation-Sprint/PM-19065] Add opaque authentication / unlock

quexten:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-19065

## 📔 Objective

Adds functions to use OPAQUE-key exchange from the clients. Additionally, adds rotateable keyset to the SDK.

Tests are provided against a local "mockserver" implementation of the server.

## ⏰ Reminders before review

- CI builds passed
  team

## 🦮 Reviewer guidelines


crates/bitwarden-crypto/src/opaque_ke/opaque.rs
```diff
@@ -0,0 +1,374 @@
...
+// This implementation will be identical between any cipher suite, but we can't simply reuse it
+// because of all the generic bounds on the CipherSuite trait. If we need to add more cipher suites,
+// we will need to copy this implementation over, or ideally use a macro to generate it.
+impl OpaqueImpl for RistrettoTripleDhArgonSuite {
+    fn start_client_registration(
+        &self,
+        password: &str,
+    ) -> Result<types::ClientRegistrationStartResult, Error> {
+        let result =
+            ClientRegistration::<Self>::start(&mut rand::thread_rng(), password.as_bytes())?;
```
justindbaur, line 209: `rand::thread_rng()` is correct? Most examples I had seen were using `opaque_ke::rand::rngs::OsRng`.

crates/bitwarden-crypto/src/opaque_ke/opaque.rs
```diff
@@ -0,0 +1,374 @@
...
+            credential_finalization,
+        )?)?;
+        Ok(types::ServerLoginFinishResult {
+            session_key: result.session_key.to_vec(),
+        })
+    }
+}
+
+#[cfg(test)]
+mod test {
```
justindbaur, line 330: Should we add any unhappy path tests? My thinking is the error that we return on a 
wrong password should be well documented and tested so that we can handle that on the client side and show a nice error.

crates/bitwarden-crypto/src/opaque_ke/types.rs
```diff
@@ -0,0 +1,168 @@
...
+    type Ksf = Argon2Id;
+}
+
+#[derive(Serialize, Deserialize)]
+#[cfg_attr(feature = "wasm", derive(Tsify), tsify(into_wasm_abi, from_wasm_abi))]
+pub struct ClientRegistrationStartResult {
+    /// The client side, serialized state of the registration process.
+    /// MUST NOT BE SHARED WITH THE SERVER and must be cleared from memory after registration
+    /// finish.
+    pub state: ByteBuf,
```
justindbaur, line 103: Just curious, what does `ByteBuf` become on the TS side? Is it special cased to become `ArrayBuffer`, `UInt8Array`, or does it copy the layout of this type? Is it actually easy to clear from memory?

justindbaur (commented):
Just a few questions.

crates/bitwarden-crypto/src/opaque_ke/types.rs
quexten: https://github.com/AmbientRun/tsify-next/blob/2e5e0a174fb4dd3b87bddec93286250001f57c46/tsify-macros/src/typescript.rs#L363

As far as I understand, this is the only way to get `Uint8array` in TS directly in `tsify-next`. Otherwise you get `number[]`, which has to be manually converted again.

crates/bitwarden-crypto/src/opaque_ke/opaque.rs
quexten: I think this is OK because: 
- We use thread_rng everywhere in the SDK, we don't use OsRng
- ThreadRng implements the CryptoRng marker trait, and is seeded via OsRng
- opaque-ke expects a `CryptoRng + RngCore`, which both `OsRng` but also `ThreadRng` fulfill.

I think the main difference to `OsRng` is speed, but I'll leave this open to @dani-garcia to comment on.

crates/bitwarden-crypto/Cargo.toml
Hinton, line 1: **issue:** These does not follow https://contributing.bitwarden.com/architecture/sdk/dependencies/#explicit-ranges

crates/bitwarden-core/src/mobile/crypto_client.rs
Hinton, line 1: **question:** There seems to be little reason to use the client at all here since it's just obfuscating direct calls to types?

This is related to my comment about `CipherConfiguration` leaking sensitive values.

crates/bitwarden-crypto/src/opaque_ke/types.rs
```diff
@@ -0,0 +1,168 @@
+use argon2::Params;
+use generic_array::{ArrayLength, GenericArray};
+use opaque_ke::{errors::InternalError, CipherSuite};
+use serde::{Deserialize, Serialize};
+use serde_bytes::ByteBuf;
+#[cfg(feature = "wasm")]
+use tsify_next::Tsify;
+
+#[derive(Serialize, Deserialize, Clone)]
+#[cfg_attr(feature = "wasm", derive(Tsify), tsify(into_wasm_abi, from_wasm_abi))]
```
Hinton, line 10: **question:** This leaks the sensitive values to the presentation layer. Would we prefer that these stay inside the rust world or that we pass a object back and forth?

crates/bitwarden-crypto/src/opaque_ke/types.rs
Hinton: We generally encode these using b64 in other places. @dani-garcia not sure if that is more appropriate or if we revisit our practice. (I believe we b64 encode a few things for mobile at least)

crates/bitwarden-crypto/src/opaque_ke/error.rs
```diff
@@ -0,0 +1,23 @@
...
+#[cfg(feature = "wasm")]
+use tsify_next::Tsify;
+
+#[derive(Debug, Error, Deserialize, Serialize)]
+#[cfg_attr(feature = "wasm", derive(Tsify), tsify(into_wasm_abi, from_wasm_abi))]
+pub enum OpaqueError {
+    #[error("Error creating message {0}")]
+    Message(String),
+    #[error("Error deserializing message")]
+    Deserialize,
```
Hinton, line 12: **issue:** Unused?

crates/bitwarden-crypto/src/opaque_ke/mod.rs
Hinton, line 1: **issue:** Lacks documentation about what this is.

`bitwarden-crypto` is a generally well documented crate and we want to maintain that high standard.

crates/bitwarden-crypto/src/opaque_ke/types.rs
```diff
@@ -0,0 +1,168 @@
...
+#[allow(dead_code)]
+pub struct ServerLoginStartResult {
+    pub(crate) credential_response: Vec<u8>,
+    pub(crate) state: Vec<u8>,
+}
+
+#[allow(dead_code)]
+pub struct ServerLoginFinishResult {
+    pub(crate) session_key: Vec<u8>,
+}
```
Hinton, line 168: **question:** Why do we need allow dead here?

crates/bitwarden-crypto/src/opaque_ke/types.rs
Hinton, line 1: **issue:** These types are public and should be properly documented what they do. Same for any public fields inside.

Private fields should ideally also be appropriate documented but it's especially important for public since they are actively expected to be used.

crates/bitwarden-crypto/src/opaque_ke/types.rs
```diff
@@ -0,0 +1,168 @@
...
+#[derive(Serialize, Deserialize, Default, Clone)]
+#[cfg_attr(feature = "wasm", derive(Tsify), tsify(into_wasm_abi, from_wasm_abi))]
+pub struct Argon2Id {
+    /// Memory cost in kibibytes
+    pub memory_kib: u32,
+    /// Time cost in number of iterations
+    pub iterations: u32,
+    /// Number of threads that may be used
+    pub parallelism: u32,
+}
```
Hinton, line 59: **question:** We already have an Argon2Id representation in `Kdf`. Should we refactor this so `Kdf` consumes the `Argon2Id` struct?

They also differ since this uses u32 while `Kdf` uses `NonZeroU32`.

crates/bitwarden-crypto/src/opaque_ke/types.rs
```diff
@@ -0,0 +1,168 @@
...
+    TripleDh,
+}
+
+#[derive(Serialize, Deserialize, Clone)]
+#[cfg_attr(feature = "wasm", derive(Tsify), tsify(into_wasm_abi, from_wasm_abi))]
+#[cfg_attr(feature = "uniffi", derive(uniffi::Enum))]
+pub enum Ksf {
+    #[serde(rename = "argon2id")]
+    Argon2id(u32, u32, u32),
+}
```
Hinton, line 48: **question:** How does this compare to `Kdf`? And any reason this does not consume `Argon2Id` defined a few lines below.

Hinton, line 48: **question:** Is it more appropriate to use https://serde.rs/container-attrs.html#rename_all?

crates/bitwarden-crypto/src/opaque_ke/opaque.rs
```diff
@@ -0,0 +1,374 @@
...
+use super::types::{self, Argon2Id, CipherConfiguration, KeGroup, KeyExchange, Ksf, OprfCs};
+
+type Error = super::error::OpaqueError;
+
+pub trait OpaqueImpl {
+    fn start_client_registration(
+        &self,
+        password: &str,
+    ) -> Result<types::ClientRegistrationStartResult, Error>;
+    fn start_server_registration(
```
Hinton, line 17: **question:** Half of these seems to only be used for the mock server?

Is this actually two implementations masquerading as one? I.e. would it be appropriate to split this in `OpaqueServer` and `OpaqueClient`. Server would only be used for testing and should be appropriate behind a `cfg` check.

crates/bitwarden-crypto/src/opaque_ke/opaque.rs
```diff
@@ -0,0 +1,374 @@
...
+    ClientRegistrationFinishParameters, CredentialFinalization, CredentialRequest,
+    CredentialResponse, Identifiers, RegistrationRequest, RegistrationResponse, RegistrationUpload,
+    ServerLogin, ServerLoginStartParameters, ServerRegistration, ServerSetup,
+};
+
+use super::types::{self, Argon2Id, CipherConfiguration, KeGroup, KeyExchange, Ksf, OprfCs};
+
+type Error = super::error::OpaqueError;
+
+pub trait OpaqueImpl {
```
Hinton, line 12: **issue:** Can we document what `OpaqueImpl` is? The naming is also a bit unfortunate as it's not an implementation but a trait.

crates/bitwarden-crypto/src/opaque_ke/opaque.rs
```diff
@@ -0,0 +1,374 @@
...
+    ) -> Result<types::ServerLoginFinishResult, Error>;
+}
+
+// This trait exists to extract the differences between all the OpaqueImpl implementations.
+// This would allow replacing those impls by a macro in the future.
+pub trait OpaqueUtil: Sized {
+    type Output;
+    fn as_variant(config: &CipherConfiguration) -> Option<Self>;
+    fn get_ksf(&self) -> Result<Self::Output, Error>;
+}
```
Hinton, line 61: **question:** This seems like an internal trait, should it be `pub`?

crates/bitwarden-crypto/src/rotateable_keyset.rs
```diff
@@ -0,0 +1,90 @@
...
+/// is the device key, and the downstream encapsulated key is the user key. The keyset
+/// can be rotated by having just access to the downstream key, and the downstream key can
+/// be accessed by having access to the upstream key.
+#[derive(Deserialize, Serialize)]
+#[cfg_attr(feature = "wasm", derive(Tsify), tsify(into_wasm_abi, from_wasm_abi))]
+pub struct RotateableKeyset {
+    /// External key encrypted private key
+    private_key: EncString,
+
+    /// Userkey encrypted private key
```
Hinton, line 22:
```suggestion
    /// [UserKey] encrypted private key
```

crates/bitwarden-crypto/src/rotateable_keyset.rs
```diff
@@ -0,0 +1,90 @@
...
+};
+
+/// A rotateable keyset is a structure that takes an upstream external key
+/// and a downstream encapsulated key. As an example, in TDE, the upstream, external key
+/// is the device key, and the downstream encapsulated key is the user key. The keyset
+/// can be rotated by having just access to the downstream key, and the downstream key can
+/// be accessed by having access to the upstream key.
+#[derive(Deserialize, Serialize)]
+#[cfg_attr(feature = "wasm", derive(Tsify), tsify(into_wasm_abi, from_wasm_abi))]
+pub struct RotateableKeyset {
```
Hinton, line 18: **issue:** Should probably be called `RotatableKeyset`?

Consider installing code spell checker and resolving any unknown words. And adding any appropriate to the project word list.

crates/bitwarden-crypto/src/rotateable_keyset.rs
```diff
@@ -0,0 +1,90 @@
...
+    /// Publickey encrypted user key
+    encapsulated_key: AsymmetricEncString,
+}
+
+impl RotateableKeyset {
+    pub fn new(
+        encapsulating_key_material: &[u8],
+        encapsulated_key: &SymmetricCryptoKey,
+    ) -> Result<Self, CryptoError> {
+        let key = if encapsulating_key_material.len() == 64 {
```
Hinton, line 34: **question:** Is there a reason we verify the length here? What are we attempting to protect against? (add a comment?)

I think it's more appropriate to use a guard statement here

```rust
if encapsulating_key_material.len() != 64 {
    return Err(CryptoError::InvalidKey);
}
let key = SymmetricCryptoKey::try_from(encapsulating_key_material.to_vec())?;
```

crates/bitwarden-core/src/mobile/crypto_client.rs
```diff
@@ -81,6 +81,72 @@ impl CryptoClient<'_> {
...
+        bitwarden_crypto::rotateable_keyset::RotateableKeyset::new(export_key, encapsulated_key)
+    }
+
+    pub fn decapsulate_key_from_rotateablekeyset(
+        &self,
+        rotateable_keyset: bitwarden_crypto::rotateable_keyset::RotateableKeyset,
+        encapsulating_key: &[u8],
+    ) -> Result<SymmetricCryptoKey, CryptoError> {
+        rotateable_keyset.decapsulate_key(encapsulating_key)
+    }
```
Hinton, line 149: @dani-garcia will these APIs work in the more limited mobile environment? I don't believe mobile can create a `SymmetricCryptoKey` for example.

crates/bitwarden-crypto/src/opaque_ke/types.rs
```diff
@@ -0,0 +1,168 @@
...
+    pub export_key: ByteBuf,
+}
+
+#[derive(Serialize, Deserialize)]
+#[cfg_attr(feature = "wasm", derive(Tsify), tsify(into_wasm_abi, from_wasm_abi))]
+pub struct ClientLoginStartResult {
+    /// The client side, serialized state of the login process.
+    /// MUST NOT BE SHARED WITH THE SERVER and must be cleared from memory after registration
+    /// finish.
+    pub state: ByteBuf,
```
Hinton, line 125: Sounds like this state should be encapsulated by the SDK and not leave it's boundary?

crates/bitwarden-crypto/src/opaque_ke/types.rs
```diff
@@ -0,0 +1,168 @@
...
+#[derive(Serialize, Deserialize)]
+#[cfg_attr(feature = "wasm", derive(Tsify), tsify(into_wasm_abi, from_wasm_abi))]
+pub struct ClientLoginFinishResult {
+    /// The serialized LoginFinish message from the client to be sent to the server
+    pub credential_finalization: ByteBuf,
+    /// The client-side only export key result from the AKE. Note: This is limited to 256 bits,
+    /// if the cipher parameters have a larger key size, the key will be truncated.
+    /// This key is consistent between authentications.
+    pub export_key: ByteBuf,
+
```
Hinton, line 139: **nit:** Inconsistent use of newline.

Hinton (changes requested):
I assume this initiative has been green lit and will move forward hence the requests for reviews. And the code should be considered "production ready".

I don't have any knowledge of the OPAQUE protocol so this review is purely from a rust and SDK perspective.

---

The pull request description is fairly lightweight and contains little information about what this work entails. Could we expand it a bit or at least link to the appropriate confluence documents.

As far as I can tell this has little to do with cryptographic primitives and is more a new registration / login flow. If that's accurate we should consider lifting it out of the crypto crate which would also simplify ownership if we suspect everything to not be owned by KM.

There are a lot of public types and fields in this PR. Do all of them need `pub` visibility or can we restrict it? It's always easier to make more things public than to decrease visibility in the future.

quexten:
> As far as I can tell this has little to do with cryptographic primitives and is more a new registration / login flow. If that's accurate we should consider lifting it out of the crypto crate which would also simplify ownership if we suspect everything to not be owned by KM.

Cryptographic primitives here refer to the most basic building blocks of Opaque, as exposed to the outside - I.e entire, high level messages (from OPAQUES view). I realize that is leading to some confusion here, so I'll refrain from the word from now on.

crates/bitwarden-crypto/src/opaque_ke/opaque.rs
quexten: cc @dani-garcia Since we do the same in the dotnet bindings, and we planned to consolidate them.

crates/bitwarden-crypto/src/opaque_ke/types.rs
quexten: Note: These are intentionally different. What we call "KdfConfig" really is more of a "BitwardenAccountKdfConfig". Specifically, it maintains a mapping of the argon2id memory value to a different unit (BitwardenAccountKdfConfig uses MiB, while Argon2 itself uses KiB). We currently map these to argon2's inputs in e.g. cryptofunctions, which implies that our config is not actually a config for the argon2 primitive but the function: `bitwardenArgon2(email, password, memory, iterations, prallelism) = argon2(password, sha256(email), memory*1024, iterations, parallelism)`.

This has led to confusion before, but with Opaque, it was important that we maintain a config that does not need any logical mapping. So the KSF parameters are the actual parameters passed into the KSF, and thus also have a different type. You can convert between them, but only explicitly.

This also might mean that the place we perform the conversion should eventually be moved, to be much closer to the serialization+UI.

crates/bitwarden-crypto/src/opaque_ke/opaque.rs
dani-garcia: Yeah, `OsRng` is reading directly from the OS entropy source, while `ThreadRng` just seeds a PRNG with the OS entropy.

The rand docs qualify it as `... good compromise between cryptographic security, fast generation ...` https://github.com/rust-random/rand/blob/master/SECURITY.md#specific-generators

crates/bitwarden-crypto/src/opaque_ke/types.rs
dani-garcia: Yeah we've had complaints in the swift native apps when using bytes as they were hard to work with, something about equality not being implemented for the swift bytes type I think.

That said, using b64 is probably just as annoying and pushing the conversion work to the clients. I wonder if this just means we need to consider having our own `Bytes` type that does the right conversion for each platform.

crates/bitwarden-crypto/src/opaque_ke/opaque.rs
dani-garcia: On the C# bindings I'm using both implementations from C# to be able to test them, so in that case I couldn't hide the client under a `cfg(test)`. But yeah, splitting into two traits should be pretty easy.

crates/bitwarden-core/src/mobile/crypto_client.rs
dani-garcia: The exact API won't, but we can adapt them slightly in `bitwarden_uniffi`

The `SymmetricCryptoKey` in `create_rotateablekeyset_from_exportkey` would be coming from the `ClientRegistrationFinishResult ` struct in the `opaque_register_finish` function. I imagine in the UniFFI layer we could pass the entire struct as an opaque pointer directly so the mobile apps won't have to deal with the key themselves.

For `decapsulate_key_from_rotateablekeyset`, the UniFFI function would just set the result key as the user key and return `Ok(())`, I think

quexten:
Setting back to draft for now.

PR #187, opened 2025-03-12
[PM-18043] Add serialization/deserialization support

coroiu:
## 📔 Objective

I did my best to anticipate usage patterns but I'm open to feedback from other teams 

Relevant clients PR: https://github.com/bitwarden/clients/pull/13804

## ⏰ Reminders before review

- CI builds passed
  team

## 🦮 Reviewer guidelines


crates/bitwarden-ipc/src/ipc_client.rs
```diff
@@ -38,66 +39,81 @@ where
...
+
+    /// Receive a message, skipping any messages that cannot be deserialized into the expected
+    /// payload type.
+    pub async fn receive_typed<Payload>(
+        &self,
+    ) -> Result<TypedIncomingMessage<Payload>, ReceiveError<Crypto::ReceiveError, Com::ReceiveError>>
+    where
+        Payload: TryFrom<Vec<u8>>,
+    {
+        loop {
```
dani-garcia: Suggestion: It might be a nice idea to have some kind of timeout here, so we don't accidentally end up stuck on an infinite loop. Not sure how that would look like, though.

crates/bitwarden-ipc/src/ipc_client.rs
```diff
@@ -38,66 +39,81 @@ where
...
+            let received = self
+                .crypto
+                .receive(&self.communication, &self.sessions)
+                .await?;
+            if let Ok(typed) = received.try_into() {
+                return Ok(typed);
+            }
```
dani-garcia: I wonder if we can come up with a way of validating the type other than just trying to parse it. As it is, we can't know for sure if we haven't received any message, or the message failed to parse correctly. Plus if we're passing keys as Vec<u8>, we might find ourselves accepting the wrong messages if they just happen to deserialize okay.

Only way to improve that I can think of would be to including the type as a string in the message, which probably would require also adding a `send_typed` equivalent.

crates/bitwarden-ipc/src/message.rs
```diff
@@ -1,20 +1,97 @@
...
 }
+
+#[derive(Clone, Debug)]
+#[cfg_attr(test, derive(PartialEq))]
+pub struct TypedOutgoingMessage<Payload> {
+    pub payload: Payload,
+    pub destination: Endpoint,
+}
+
+impl<Payload> TryFrom<OutgoingMessage> for TypedOutgoingMessage<Payload>
```
dani-garcia, line 37: Would be nice to implement this for both `Payload: TryFrom<Vec<u8>>`  and `Payload: Deserialize`, but I'm pretty sure we'd be hitting conflicts with trait impls

crates/bitwarden-ipc/src/ipc_client.rs
coroiu: Hmm, very good point.

What about something like this? https://github.com/bitwarden/sdk-internal/pull/187/commits/1dd01642843290979fe9b6af89c3097ef702e7f2 We need to add Topic support for request/response anyways so

crates/bitwarden-ipc/src/ipc_client.rs
coroiu: Added timeout support :) https://github.com/bitwarden/sdk-internal/pull/187/commits/941bb068507d8c2f8b90bc43e60d45c00afcbd27

crates/bitwarden-ipc/src/message.rs
coroiu: To simplify conversions? Which types would you want to move between?

dani-garcia: Yeah, my thinking was messages are either going to be structured JSON data, or unstructured key bytes and crypto stuff. The current From<Vec<u8>> solves the second case, but the first case needs a little bit of boilerplate, like this test code:
```
        impl TryFrom<Vec<u8>> for TestPayload {
            type Error = serde_json::Error;

            fn try_from(value: Vec<u8>) -> Result<Self, Self::Error> {
                serde_json::from_slice(&value)
            }
        }

        impl TryFrom<TestPayload> for Vec<u8> {
            type Error = serde_json::Error;

            fn try_from(value: TestPayload) -> Result<Self, Self::Error> {
                serde_json::to_vec(&value)
            }
        }
``` 

Would be nice to have a way to do both automatically, though I'm not sure it's possible.

crates/bitwarden-ipc/src/ipc_client.rs
dani-garcia: Looks pretty good, though I would add some more documentation about what the `None` case does:
```
    /// Receive a message, filtering by topic.
    pub async fn receive_topic(
        &self,
        topic: Option<String>,
```

Reading that, I would assume that if I passed `None` to topic, no filtering would be done (so i would get all messages). What really happens is that I only get the messages without a topic.

crates/bitwarden-ipc/src/message.rs
coroiu: Aha, yeah I also thought about this a little bit and concluded that a macro is probably the simplest solution, something like 
```
#derive[SerdeJsonIpcPayload]
```
but I didn't wanna complicate the PR before we have some real use-cases

crates/bitwarden-ipc/src/ipc_client.rs
coroiu: Good point, we don't need two functions for this. Merged both into one and implemented the logic you expected

dani-garcia (approved):

PR #193, opened 2025-03-17
[PM-18044] Ensure all calls to receive should get all messages broadcast

coroiu:
## ⏰ Reminders before review

- CI builds passed
  team

## 🦮 Reviewer guidelines

### DIFF
--- a/crates/bitwarden-ipc/src/ipc_client.rs
+++ b/crates/bitwarden-ipc/src/ipc_client.rs
@@ -19,6 +19,13 @@ where
     sessions: Ses,
 }
 
+pub struct IpcClientSubscription<Receiver>
+where
+    Receiver: CommunicationBackendReceiver,
+{
+    receiver: Receiver,
+}
+
 impl<Crypto, Com, Ses> IpcClient<Crypto, Com, Ses>
 where
     Crypto: CryptoProvider<Com, Ses>,
@@ -43,11 +50,18 @@ where
             .await
     }
 
+    pub async fn subscribe(&self) -> IpcClientSubscription<Com::Receiver> {
+        IpcClientSubscription {
+            receiver: self.communication.subscribe(),
+        }
+    }
+
     /// Receive a message, optionally filtering by topic.
     /// Setting the topic to `None` will receive all messages.
     /// Setting the timeout to `None` will wait indefinitely.
     pub async fn receive(
         &self,
+        subscription: &IpcClientSubscription<Com::Receiver>,
         topic: Option<String>,
         timeout: Option<Duration>,
     ) -> Result<
@@ -61,11 +75,7 @@ where
             loop {
                 let received = self
                     .crypto
-                    .receive(
-                        &self.communication,
-                        &self.sessions,
-                        &self.communication.subscribe(),
-                    )
+                    .receive(&subscription.receiver, &self.communication, &self.sessions)
                     .await?;
                 if topic.is_none() || received.topic == topic {
                     return Ok(received);
@@ -86,6 +96,7 @@ where
     /// payload type.
     pub async fn receive_typed<Payload>(
         &self,
+        subscription: &IpcClientSubscription<Com::Receiver>,
         timeout: Option<Duration>,
     ) -> Result<
         TypedIncomingMessage<Payload>,
@@ -99,7 +110,7 @@ where
         Payload: TryFrom<Vec<u8>> + PayloadTypeName,
     {
         let topic = Some(Payload::name());
-        let received = self.receive(topic, timeout).await?;
+        let received = self.receive(&subscription, topic, timeout).await?;
         received.try_into().map_err(TypedReceiveError::Typing)
     }
 }
@@ -133,9 +144,9 @@ mod tests {
 
         async fn receive(
             &self,
+            _receiver: &<TestCommunicationBackend as CommunicationBackend>::Receiver,
             _communication: &TestCommunicationBackend,
             _sessions: &TestSessionRepository,
-            _receiver: &<TestCommunicationBackend as CommunicationBackend>::Receiver,
         ) -> Result<IncomingMessage, ReceiveError<String, TestCommunicationBackendReceiveError>>
         {
             self.receive_result.clone()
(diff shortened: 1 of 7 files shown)


crates/bitwarden-ipc/src/traits/communication_backend.rs
```diff
@@ -20,7 +20,27 @@ pub trait CommunicationBackend {
...
         &self,
     ) -> impl std::future::Future<Output = Result<IncomingMessage, Self::ReceiveError>>;
+
+    /// Subscribe to receive messages. This function will return a receiver that can be used to
+    /// receive messages asynchronously. Multiple receivers may be created, and all receivers will
+    /// receive the same messages.
+    ///
+    /// Use this function if you are expecting to receive a message as a response to a message you
+    /// sent. Subscribing will start buffering messages and help you avoid missing messages.
+    fn subscribe(&self) -> impl CommunicationBackendReceiver<ReceiveError = Self::ReceiveError>;
```
dani-garcia: As far as I understand it, the invariant that all the subscribers will receive all the same messages is something that the trait implementations need to enforce right? If so it might be a good idea to mention it in here. Otherwise someome might think that it's the `IpcClient` that handles that for you, leading to inconsistent implementations.

The same also probably applies to `CommunicationBackendReceiver::receive`

coroiu: Good point, I've updated the doc, thoughts?

crates/bitwarden-ipc/src/wasm/communication_backend.rs
```diff
@@ -1,53 +1,102 @@
...
+        self.receive_rx.resubscribe()
+    }
+}
+
+impl CommunicationBackendReceiver for tokio::sync::broadcast::Receiver<IncomingMessage> {
+    type ReceiveError = JsValue;
+
+    async fn receive(&self) -> Result<IncomingMessage, Self::ReceiveError> {
+        Ok(self
+            .resubscribe()
```
dani-garcia: This doesn't feel entirely correct, the docs for `resubscribe` say:

    /// Re-subscribes to the channel starting from the current tail element.
    ///
    /// This [`Receiver`] handle will receive a clone of all values sent
    /// **after** it has resubscribed. This will not include elements that are
    /// in the queue of the current receiver. Consider the following example.


Which mean each call to `receive` will never get any buffered values, and only receive any values after `receive` has been called.

Removing the resubscribe causes errors as receive requires &mut, so we might need to do something like:

```
impl CommunicationBackend for JsCommunicationBackend {
    ...

    fn subscribe(&self) -> impl CommunicationBackendReceiver<ReceiveError = Self::ReceiveError> {
        Mutex::new(self.receive_rx.resubscribe())
    }
}

impl CommunicationBackendReceiver for Mutex<tokio::sync::broadcast::Receiver<IncomingMessage>> {
    type ReceiveError = JsValue;

    async fn receive(&self) -> Result<IncomingMessage, Self::ReceiveError> {
        Ok(self
            .lock()
            .await
            .recv()
            .await
            .map_err(|e| ChannelError(e.to_string()))?)
    }
}
```

coroiu: Ouf good point, I didn't realize this changed by moving the buffering from the platform to Rust, I'll address this asap 👍

crates/bitwarden-ipc/src/ipc_client.rs
```diff
@@ -72,26 +140,75 @@ where
...
+        timeout: Option<Duration>,
+    ) -> Result<
+        TypedIncomingMessage<Payload>,
+        TypedReceiveError<
+            <Payload as TryFrom<Vec<u8>>>::Error,
+            Crypto::ReceiveError,
+            <Com::Receiver as CommunicationBackendReceiver>::ReceiveError,
+        >,
+    > {
+        self.client.receive_typed(&self.receiver, timeout).await
```
dani-garcia, line 208: I wonder if we should inline the `receive_typed` method from `IpcClient` here, it's only used here and it's simple enough:
```
    pub async fn receive(
        &self,
        timeout: Option<Duration>,
    ) -> Result<
        TypedIncomingMessage<Payload>,
        TypedReceiveError<
            <Payload as TryFrom<Vec<u8>>>::Error,
            Crypto::ReceiveError,
            <Com::Receiver as CommunicationBackendReceiver>::ReceiveError,
        >,
    > {
        let topic = Some(Payload::name());
        let received = self.client.receive(&self.receiver, &topic, timeout).await?;
        received.try_into().map_err(TypedReceiveError::Typing)
    }
```

crates/bitwarden-ipc/src/traits/communication_backend.rs
```diff
@@ -65,19 +111,33 @@ pub mod tests {
...
+                return Err(TestCommunicationBackendReceiveError::NoQueuedMessages);
             }
+
+            Ok(self
+                .0
+                .write()
+                .await
+                .recv()
+                .await
+                .expect("Failed to receive incoming message"))
```
dani-garcia: We should grab the rwlock only once to avoid issues with other requests getting in between. Also use `.is_empty()` instead of `.len() == 0`

```suggestion
            let mut guard = self.0.write().await;

            if guard.is_empty() {
                return Err(TestCommunicationBackendReceiveError::NoQueuedMessages);
            }

            Ok(guard
                .recv()
                .await
                .expect("Failed to receive incoming message"))
        }
```

crates/bitwarden-ipc/src/wasm/communication_backend.rs
dani-garcia: This backend receiver is still resubscribing on receive, which means it won't receive any previous messages, probably needs to be updated to be similar to `TestCommunicationBackendReceiver`

coroiu: Yeah, noticed this in a follow-up PR at the same time as you wrote this comment 🥲 Will fix

crates/bitwarden-ipc/src/ipc_client.rs
coroiu: I was thinking that the req/res would build upon this and so it made sense to keep it separate, but I ended up inlining it in a follow up PR so maybe we can just merge this as is 🤷‍♂️

dani-garcia (approved):

PR #202, opened 2025-03-27
[PM-19603] Asymmetric encrypt/encapsulate interface

quexten:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-19603

## 📔 Objective

The goal of this PR is to provide a consistent, clear interface for asymmetric encryption. For asymmetric encryption we should never encrypt data, only share keys - "encapsulate". If data access is to be shared to another account, then this should be done via creating a new symmetric key, and sharing access to it.

Further, currently our asymmetric data sharing makes no proofs about authenticity of the sender, that is, the data is unsigned. This makes that clear on the publicly exposed interface, and leaves the room open to add an interface that explicitly does provide sender authenticity via signing.

This PR renames the interfaces to reflect that. Further, we drop the AsymmetricKey impl of the KeyEncryptable trait, specifically since it should *not* be able to encrypt arbitrary data, but also because a new interface that does provide sender authenticity should be the primarily used interface, which does not line up with the encryptable trait.

## ⏰ Reminders before review

- CI builds passed
  team

## 🦮 Reviewer guidelines


crates/bitwarden-crypto/src/enc_string/asymmetric.rs
```diff
@@ -246,38 +256,44 @@ XKZBokBGnjFnTnKcs7nv/O8=
...
 
     #[test]
     fn test_enc_string_rsa2048_oaep_sha1_hmac_sha256_b64() {
         let private_key = AsymmetricCryptoKey::from_pem(RSA_PRIVATE_KEY).unwrap();
-        let enc_str: &str = "6.ThnNc67nNr7GELyuhGGfsXNP2zJnNqhrIsjntEQ27r2qmn8vwdHbTbfO0cwt6YgSibDN0PjiCZ1O3Wb/IFq+vwvyRwFqF9145wBF8CQCbkhV+M0XvO99kh0daovtt120Nve/5ETI5PbPag9VdalKRQWZypJaqQHm5TAQVf4F5wtLlCLMBkzqTk+wkFe7BPMTGn07T+O3eJbTxXvyMZewQ7icJF0MZVA7VyWX9qElmZ89FCKowbf1BMr5pbcQ+0KdXcSVW3to43VkTp7k7COwsuH3M/i1AuVP5YN8ixjyRpvaeGqX/ap2nCHK2Wj5VxgCGT7XEls6ZknnAp9nB9qVjQ==|s3ntw5H/KKD/qsS0lUghTHl5Sm9j6m7YEdNHf0OeAFQ=";
+        let enc_str: &str = "6.KhZmkc7f2WYuZGm/xlKZOK4c5JSwd9JtJvmyk0R+ZCqbRnZi5XNJaqnMiJjiqeLztE97bHRGWyDPvhyIisr7jLi35vL/Znpg3QzSMEDNI7aAM2FwJbCzdUrFDa/h08edv816AL1hAOqtGmjpfRL1j+47hlAiF3/srFCeePHkj0+CmHpHN13BN1XkLKk58mETKh8ky/ZUW2s4NjZaZ/Wxh6I9sv28L+u1hekKxDOdNKBnmqsh8WRBOtmZm1ZM9WI6aPA5tXgp30vxWrc1AsZ5Ts0aVkm8UzPTWuU9d/O9ICAQkr1hX58qO6M5geP+NvaG3UGymw0zp6Hdgz239XYpKg==|AA==";
         let enc_string: AsymmetricEncString = enc_str.parse().unwrap();
 
+        let mut test_bytes = vec![0u8; 64];
+        let test_key = SymmetricCryptoKey::try_from(test_bytes.as_mut_slice()).unwrap();
```
dani-garcia: And here:
```suggestion
        let test_bytes = vec![0u8; 64];
        let test_key = SymmetricCryptoKey::try_from(test_bytes).unwrap();
```

crates/bitwarden-crypto/src/enc_string/asymmetric.rs
```diff
@@ -246,38 +256,44 @@ XKZBokBGnjFnTnKcs7nv/O8=
...
 
     #[test]
     fn test_enc_string_rsa2048_oaep_sha1_b64() {
         let private_key = AsymmetricCryptoKey::from_pem(RSA_PRIVATE_KEY).unwrap();
-        let enc_str: &str = "4.ZheRb3PCfAunyFdQYPfyrFqpuvmln9H9w5nDjt88i5A7ug1XE0LJdQHCIYJl0YOZ1gCOGkhFu/CRY2StiLmT3iRKrrVBbC1+qRMjNNyDvRcFi91LWsmRXhONVSPjywzrJJXglsztDqGkLO93dKXNhuKpcmtBLsvgkphk/aFvxbaOvJ/FHdK/iV0dMGNhc/9tbys8laTdwBlI5xIChpRcrfH+XpSFM88+Bu03uK67N9G6eU1UmET+pISJwJvMuIDMqH+qkT7OOzgL3t6I0H2LDj+CnsumnQmDsvQzDiNfTR0IgjpoE9YH2LvPXVP2wVUkiTwXD9cG/E7XeoiduHyHjw==";
+        let enc_str: &str = "4.KhZmkc7f2WYuZGm/xlKZOK4c5JSwd9JtJvmyk0R+ZCqbRnZi5XNJaqnMiJjiqeLztE97bHRGWyDPvhyIisr7jLi35vL/Znpg3QzSMEDNI7aAM2FwJbCzdUrFDa/h08edv816AL1hAOqtGmjpfRL1j+47hlAiF3/srFCeePHkj0+CmHpHN13BN1XkLKk58mETKh8ky/ZUW2s4NjZaZ/Wxh6I9sv28L+u1hekKxDOdNKBnmqsh8WRBOtmZm1ZM9WI6aPA5tXgp30vxWrc1AsZ5Ts0aVkm8UzPTWuU9d/O9ICAQkr1hX58qO6M5geP+NvaG3UGymw0zp6Hdgz239XYpKg==";
         let enc_string: AsymmetricEncString = enc_str.parse().unwrap();
 
+        let mut test_bytes = vec![0u8; 64];
+        let test_key = SymmetricCryptoKey::try_from(test_bytes.as_mut_slice()).unwrap();
```
dani-garcia: Same here:
```suggestion
        let test_bytes = vec![0u8; 64];
        let test_key = SymmetricCryptoKey::try_from(test_bytes).unwrap();
```

crates/bitwarden-crypto/src/enc_string/asymmetric.rs
```diff
@@ -246,38 +256,44 @@ XKZBokBGnjFnTnKcs7nv/O8=
...
     #[test]
     fn test_enc_string_rsa2048_oaep_sha256_b64() {
-        let private_key = AsymmetricCryptoKey::from_pem(RSA_PRIVATE_KEY).unwrap();
-        let enc_str: &str = "3.YFqzW9LL/uLjCnl0RRLtndzGJ1FV27mcwQwGjfJPOVrgCX9nJSUYCCDd0iTIyOZ/zRxG47b6L1Z3qgkEfcxjmrSBq60gijc3E2TBMAg7OCLVcjORZ+i1sOVOudmOPWro6uA8refMrg4lqbieDlbLMzjVEwxfi5WpcL876cD0vYyRwvLO3bzFrsE7x33HHHtZeOPW79RqMn5efsB5Dj9wVheC9Ix9AYDjbo+rjg9qR6guwKmS7k2MSaIQlrDR7yu8LP+ePtiSjx+gszJV5jQGfcx60dtiLQzLS/mUD+RmU7B950Bpx0H7x56lT5yXZbWK5YkoP6qd8B8D2aKbP68Ywg==";
+        let key_pair = AsymmetricCryptoKey::from_pem(RSA_PRIVATE_KEY).unwrap();
+        let enc_str: &str = "3.BfwZTwBYbU5WQ5X7Vm8yl0hYmHTRdkVACCRZYcqhcjicoaPVDEP03CIRmtnppu0aXOppoQzhw5S2OKTUaqoOGKZg7+PrmVEhjiUFfVAptInBD6XGHZ0Z3u3F+JY1E3xIFebOFiX7KLQ+7D0bJhBEnl8P7phmanKF3Cil5ayDGRpAjAsBHMwlNRKXy05YpYs3/x+V+zjlxVrBU9gYFCpacKUbxT51I8tf21ISqo6H9ZBwqDE2QUPhYJl5op7SJgySdd3YCKnsObXa8fFj2OwxGLAXJAyvF6qZyl08RO/ZYUOOOPlbC7ywXxAISw3qmrwxqpLSBqAm9BYPa/zxBnTHrA==";
         let enc_string: AsymmetricEncString = enc_str.parse().unwrap();
 
+        let mut test_key = vec![0u8; 64];
+        let test_key = SymmetricCryptoKey::try_from(test_key.as_mut_slice()).unwrap();
```
dani-garcia: We also implement TryFrom for Vec, which is slightly simpler:
```suggestion
        let test_bytes = vec![0u8; 64];
        let test_key = SymmetricCryptoKey::try_from(test_bytes).unwrap();
```

crates/bitwarden-core/src/auth/auth_request.rs
```diff
@@ -117,15 +116,20 @@ fn test_auth_request() {
         96, 31, 114, 127, 212, 187, 167, 110, 205, 116, 198, 243, 218, 72, 137, 53, 248, 43, 255,
         67, 35, 61, 245, 93,
     ];
+    let secret = secret.to_vec();
```
dani-garcia: We should change the definition of secret to be `let secret = vec![ ...  ]`, rather than defining it as a slice then converting it to vec.

crates/bitwarden-crypto/src/store/context.rs
```diff
@@ -182,40 +181,39 @@ impl<Ids: KeyIds> KeyStoreContext<'_, Ids> {
...
     /// * `encrypted_key` - The key to decrypt
-    pub fn decrypt_symmetric_key_with_asymmetric_key(
+    pub fn decapsulate_key_unsigned(
         &mut self,
-        encryption_key: Ids::Asymmetric,
+        decapsulation_key: Ids::Asymmetric,
         new_key_id: Ids::Symmetric,
-        encrypted_key: &AsymmetricEncString,
+        encapsulated_key: &AsymmetricEncString,
     ) -> Result<Ids::Symmetric> {
```
Hinton: **issue:** Comment above isn't updated.

crates/bitwarden-crypto/src/store/context.rs
```diff
@@ -182,40 +181,39 @@ impl<Ids: KeyIds> KeyStoreContext<'_, Ids> {
...
-    pub fn encrypt_symmetric_key_with_asymmetric_key(
+    /// * `encapsulation_key` - The key id used to encrypt the `encapsulated_key`. It must already
+    ///   exist in the context
+    /// * `encapsulated_key` - The key id to encrypt. It must already exist in the context
+    pub fn encapsulate_key_unsigned(
         &self,
-        encryption_key: Ids::Asymmetric,
-        key_to_encrypt: Ids::Symmetric,
+        encapsulation_key: Ids::Asymmetric,
+        encapsulated_key: Ids::Symmetric,
```
Hinton: **question:** `encapsulation_key` and `encapsulated_key` are very similar. I wonder if there are better names. https://en.wikipedia.org/wiki/Key_encapsulation_mechanism uses public key and secret key which is more distinct but the asymmetric is a keypair 🤔.

crates/bitwarden-core/src/auth/auth_request.rs
```diff
@@ -101,8 +100,8 @@ pub(crate) fn approve_auth_request(
     #[allow(deprecated)]
     let key = ctx.dangerous_get_symmetric_key(SymmetricKeyId::User)?;
 
-    Ok(AsymmetricEncString::encrypt_rsa2048_oaep_sha1(
```
Hinton, line 106: We explicitly used a fixed algorithm in some places because we wanted to avoid it being updated should the encrypt default impl change. This changes that behaviour and we should be mindful of the implications.

crates/bitwarden-crypto/src/enc_string/asymmetric.rs
Hinton, line 1: The updated naming changes the meaning of what an `AsymmetricEncString` is. It's now only encapsulated keys and is no longer arbitrary data.

We should either continue having it be generic, rename it or define a newtype with the encapsulating behaviour.

crates/bitwarden-crypto/src/store/context.rs
quexten: Proposal: The symmetric key is being shared here, so we could just call it "sharedKey", and keep encapsulation / decapsulation for the asymmetric keys.

crates/bitwarden-crypto/src/enc_string/asymmetric.rs
quexten: I agree fully, but can we defer this decision until we get to signing keys? We will very soon introduce, and then mandate all of the asymmetric sharing operations to use signcryption to verify sender authenticity. I imagine based on this, we might adjust these objects. Possibly we would name one `UnauthenticatedSharedKey`, and one `AuthenticatedSharedKey`? 

But since we don't have the sharing protocol specified yet, we may want to defer this decision?

crates/bitwarden-crypto/src/store/context.rs
quexten: Updated

crates/bitwarden-crypto/src/store/context.rs
quexten: (Changed to `shared_key` / `encapsulated_shared_key` now).

crates/bitwarden-core/src/auth/auth_request.rs
quexten: In this case I think it should be based on the public key. The public key is ephemeral, provided by the requesting device, so we know that it will understand the encrypted string format. We just need to make sure not to use new key types in the *request* before clients understand this key format for N releases.

crates/bitwarden-crypto/src/enc_string/asymmetric.rs
Hinton: Why could we not rename it `UnauthenticatedSharedKey` now? If that's a naming we're leaning towards we could just rename it now?

crates/bitwarden-crypto/src/store/context.rs
Hinton: Just an observation, but the documentation of encapsulate and decapsulate differs a bit, decapsulate description is `Decrypt a symmetric key into the context by using an already existing asymmetric key`. Encapsulate was updated to `Encapsulate and return a symmetric key from the context by using an already existing asymmetric key`

crates/bitwarden-crypto/src/enc_string/asymmetric.rs
quexten: While not used in the SDK, on the `clients` we still (temporarily) support the masterkeyhash encrypted in an `AsymmetricEncString`, until we disable that login-with-device flow. Are we OK with diverging naming compared to the clients until then? In that case I'm happy to change.

crates/bitwarden-crypto/src/store/context.rs
quexten: Updated to use "Decapsulate" for the former message

crates/bitwarden-crypto/src/enc_string/asymmetric.rs
Hinton: I would say so. `AsymmetricEncString` is not a concept in client so we're already diverging with the split into two types. Since this seems to be the direction we're going I'm fine with the further distinction. If we do need to handle the masterkeyhash we can define a newtype around SharedKey which provides a more context appropriate api.

quexten: Ok. We already have the PR open on clients disabling creation of approvals for the masterkey based login-with-device flow, so I'm also happy to rename this.

quexten: Renamed.

crates/bitwarden-crypto/src/enc_string/asymmetric.rs
```diff
@@ -246,45 +262,51 @@ XKZBokBGnjFnTnKcs7nv/O8=
...
+        let enc_string: UnauthenticatedSharedKey = enc_str.parse().unwrap();
 
+        let test_bytes = vec![0u8; 64];
+        let test_key = SymmetricCryptoKey::try_from(test_bytes).unwrap();
         assert_eq!(enc_string.enc_type(), 3);
 
-        let res: String = enc_string.decrypt_with_key(&private_key).unwrap();
-        assert_eq!(res, "EncryptMe!");
+        let res = enc_string.decapsulate_key_unsigned(&key_pair).unwrap();
+        assert_eq!(res, test_key);
```
Hinton: `test_key` is just an empty array here, would it not be more reasonable to use `derive_symmetric_key` to generate a "realistic key"?

Hinton (commented):
Question about the new tests for asymmetric. Looks good otherwise.

crates/bitwarden-crypto/src/enc_string/asymmetric.rs
quexten: Since both the result of derive_symmetric_key for a given seed, and the empty array, are equally probable sample results, and since `vec![0u8; 64];` is a valid key, I don't think it's required?

If there would be certain biases, or bugs specific to this key, that would be a much deeper founded problem, and we should stop using whatever encryption scheme has issues/biases based on the structure/contents of the key.

Hinton: The difference is that an empty vec has a higher probability of occurring due to programming errors than a random vec.

- During TDD we commonly write code that passes tests, a very reasonable impl would be to just return an empty vec in the initial pass.
- Empty vec can sometimes be used as the default case for malformed data.

quexten: Fair. I updated this to add a "generate_seeded_for_unit_tests" fn to SymmetricCryptoKey. The naming should make it clear enough that it is not meant to be used in production code, but it also has a description with a warning.

This also comes in useful for the related upcoming changes for cose, which had removed the rng from the public api available for other crates.

crates/bitwarden-crypto/src/enc_string/asymmetric.rs
```diff
@@ -148,60 +153,70 @@ impl serde::Serialize for AsymmetricEncString {
...
-    fn decrypt_with_key(&self, key: &AsymmetricCryptoKey) -> Result<Vec<u8>> {
-        use AsymmetricEncString::*;
-        match self {
-            Rsa2048_OaepSha256_B64 { data } => key.key.decrypt(Oaep::new::<sha2::Sha256>(), data),
-            Rsa2048_OaepSha1_B64 { data } => key.key.decrypt(Oaep::new::<sha1::Sha1>(), data),
+impl UnauthenticatedSharedKey {
+    /// Decapsulate a symmetric key, shared asymmetrically.
+    /// Note: The shared key does not have a sender signature and sender authenticity is not
+    /// guaranteed.
+    pub fn decapsulate_key_unsigned(
```
dani-garcia, line 188: Question: This is called `decapsulate_key_unsigned` and returns a `UnauthenticatedSharedKey`. I'm under the impression that when we say unsigned/unauthenticated we're basically meaning the same thing (no signature means no sender authenticity). 

Is there a meaninful distinction between unsigned/unauthenticated that I'm missing? Should they both be using the same terminology if not?

crates/bitwarden-crypto/src/keys/symmetric_crypto_key.rs
```diff
@@ -132,6 +141,12 @@ impl TryFrom<&mut [u8]> for SymmetricCryptoKey {
     }
 }
 
+impl From<Aes256CbcHmacKey> for SymmetricCryptoKey {
```
dani-garcia: I don't think this is used anywhere, as far as I can see

dani-garcia (commented):
Looks pretty good to me, just some naming question and some potentially unused code

crates/bitwarden-crypto/src/keys/symmetric_crypto_key.rs
```diff
@@ -53,6 +55,13 @@ impl SymmetricCryptoKey {
         SymmetricCryptoKey::Aes256CbcHmacKey(Aes256CbcHmacKey { enc_key, mac_key })
     }
 
+    /// Generate a new random [SymmetricCryptoKey] for unit tests. Note: DO NOT USE THIS
+    /// IN PRODUCTION CODE.
```
Hinton, line 63: If this is only available in bitwarden-crypto you can use `#[test]` to put it behind a compile flag.

crates/bitwarden-crypto/src/enc_string/asymmetric.rs
quexten: ~~I think naming it "UnsignedSharedKey" would not make sense. Even a signed key would not include a signature after decapsulation but only decapsulate if the signature verifies correctly.~~

So the other option is renaming the decapsulate function to something like "decapsulate_unauthenticated_key". That could work, if it helps reduce confusion?

Note: This may require a follow-up clients PR, since there the "unsigned" terminology is used.

quexten: Apologies, I think I had my mental model mixed up when I wrote this yesterday, and assumed UnauthenticatedSharedKey was *after* decapsulation. But since it is the EncString, UnauthenticatedSharedKey, or UnsignedSharedKey both work too.

quexten: Renamed to "UnsignedSharedKey".

crates/bitwarden-crypto/src/keys/symmetric_crypto_key.rs
quexten: Seems to be left over from rewriting and then scrapping something... Cleaned up now.

crates/bitwarden-crypto/src/keys/symmetric_crypto_key.rs
quexten: The idea for this function was to have something that teams can use to make their uses of generate determistic for tests, outside of the cryptoCrate. For instance  

```
#[test]
fn test_move_user_cipher_to_org_manually() {
    let org = uuid::Uuid::new_v4();
    let key = SymmetricCryptoKey::generate(rand::thread_rng());
    let org_key = SymmetricCryptoKey::generate(rand::thread_rng());
    let key_store = create_test_crypto_with_user_and_org_key(key, org, org_key);

```

quexten: This was intended to be used outside of the crate too, in case teams need a deterministically generated key.

However, looking at this more widely, I'm not sure how this would actually help since the encryption is still non-deterministic (based on IV/nonce)...

Marked it `#[cfg(test)]` for now and we can revisit later when we find cases where it would be useful?

Hinton: Sounds good.

dani-garcia (approved):

PR #203, opened 2025-03-27
[PM-20614] Cose Content Type

quexten:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-20614

## 📔 Objective

Encrypt type 7 (CoseEncrypt0 with currently Xchacha20poly1305) requires specifying a content format. Since the encrypted content is just a byte array, this alone is not enough to understand what format the data contained has. In some cases, multiple formats may be OK.

These cases include:
- Private keys
  - PKCS8 DER / CoseKey
- Symmetric keys
  - Bitwarden Custom Key Byte Serialization ("just key bytes concatenated (and padding if it is a CoseKey)"), CoseKey
- Utf8 for most cipher fields
- Future:
  - Blob Cipher: Json/Cbor?

To distinguish between these, we need to introduce a content format that is passed through until the encryption happens. For AES256-CBC-HMAC, it gets ignored, for CoseEncrypt0/Xchacha20poly1305, it gets written as a protected header onto the encstring/encrypt0 object. We do not add validation on the decryption path in this PR.

For key wrapping, when wrapping cosekeys with cosekeys, the Bitwarden Custom Key Byte Serialization with padding is not used, but the raw CoseKey without padding is used; but can be distinguished from other keys on unwrapping by inspecting the content format.

Further to ensure the content format is passed along, the encryptable trait is split into 3 traits by this PR. 
`CompositeEncryptable` is a trait that represents encryption of composite objects. Composite objects here are objects that have sub-encyrptables under the same key. For instance, a cipher is a composite encryptable, since all fields are individually encrypted. It also has some sub-composite encryptables (Uri). Composite encryptables only defer encryption to one of the other traits, and do not themselves encrypt directly, and thus also don't have a content type.

For the Utf8 content type specifically, for XChaCha20Poly1305 encryption, a padding scheme is added, in order to hide the direct relationship of ciphertext length to plaintext length. The content type written onto the encrypt0 object is a custom type to indicate this.

The since bytes `Vec<u8>`. alone are not enough to carry a content type, a new generic struct `Bytes<ConstContentType>` is introduced and returned in various places such as:
- DER private key
- DER public key
- Symmetric crypto key

The `ConstContentType` is sealed and not implementable by other crates, and type aliases are provided for more convenient / less cognitive overhead syntax.

For these, the EncryptableWithContentType trait(s) are implemented.

## ⏰ Reminders before review

- CI builds passed
  team

## 🦮 Reviewer guidelines


crates/bitwarden-crypto/src/cose.rs
```diff
@@ -15,22 +21,60 @@ use crate::{
...
+/// instance, an `EncString` in most cases contains UTF-8 encoded text. In some cases it may contain
+/// a Pkcs8 private key, or a COSE key. Specifically, for COSE keys, this allows distinguishing
+/// between the old symmetric key format, represented as `ContentFormat::OctetStream`, and the new
+/// COSE key format, represented as `ContentFormat::CoseKey`.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Serialize, Deserialize)]
+#[cfg_attr(feature = "wasm", derive(Tsify), tsify(into_wasm_abi, from_wasm_abi))]
+pub enum ContentFormat {
+    Utf8,
+    Pkcs8,
+    CoseKey,
```
quexten: Opinion question: Are we fine with OctetStream for old bitwarden keys? Should we make a custom content format - "BitwardenLegacyKey"?

crates/bitwarden-crypto/src/cose.rs
```diff
@@ -15,22 +21,60 @@ use crate::{
...
+            protected_header.content_format(CoapContentFormat::OctetStream)
+        }
+    };
+    let mut protected_header = protected_header.build();
     // This should be adjusted to use the builder pattern once implemented in coset.
     // The related coset upstream issue is:
     // https://github.com/google/coset/issues/105
     protected_header.alg = Some(coset::Algorithm::PrivateUse(XCHACHA20_POLY1305));
 
+    let encoded_plaintext = if content_format == ContentFormat::Utf8 {
```
quexten: Question: Would it be better to move this up into the match on the content_format? This would protect us against accidentally only changing one in the future and having the content format mismatched from whether the contents are actually padded?

dani-garcia: I feel like we should probably do two things here:
- Extract the condition of when we should pad to a function like `fn should_pad_content_format(ContentFormat) -> bool` and use it for both encryption/decryption, that way we'll make sure that it's always matching
- Extract the conversion of `ContentFormat<->Header` to separate functions, to simplify a bit the encrypt/decrypt functions

```rust
impl From<ContentFormat> for coset::Header {
    fn from(format: ContentFormat) -> Self {
        let header = coset::HeaderBuilder::new();
        let header = match format {
            ContentFormat::Utf8 => header.content_type(CONTENT_TYPE_PADDED_UTF8.to_string()),
            ContentFormat::Pkcs8 => header.content_format(CoapContentFormat::Pkcs8),
            ContentFormat::CoseKey => header.content_format(CoapContentFormat::CoseKey),
            ContentFormat::OctetStream => header.content_format(CoapContentFormat::OctetStream),
        };
        let mut header = header.build();
        // This should be adjusted to use the builder pattern once implemented in coset.
        // The related coset upstream issue is:
        // https://github.com/google/coset/issues/105
        header.alg = Some(coset::Algorithm::PrivateUse(XCHACHA20_POLY1305));
        header
    }
}

impl TryFrom<&coset::Header> for ContentFormat {
    type Error = CryptoError;

    fn try_from(header: &coset::Header) -> Result<Self, Self::Error> {
        let Some(ref alg) = header.alg else {
            return Err(CryptoError::EncString(
                EncStringParseError::CoseMissingAlgorithm,
            ));
        };
        if *alg != coset::Algorithm::PrivateUse(XCHACHA20_POLY1305) {
            return Err(CryptoError::WrongKeyType);
        }

        match header.content_type.as_ref() {
            Some(ContentType::Text(format)) if format == CONTENT_TYPE_PADDED_UTF8 => {
                Ok(ContentFormat::Utf8)
            }
            Some(ContentType::Assigned(CoapContentFormat::Pkcs8)) => Ok(ContentFormat::Pkcs8),
            Some(ContentType::Assigned(CoapContentFormat::CoseKey)) => Ok(ContentFormat::CoseKey),
            Some(ContentType::Assigned(CoapContentFormat::OctetStream)) => {
                Ok(ContentFormat::OctetStream)
            }
            _ => Err(CryptoError::EncString(
                EncStringParseError::CoseMissingContentType,
            )),
        }
    }
}

fn should_pad_content(format: &ContentFormat) -> bool {
    matches!(format, ContentFormat::Utf8)
}

/// Encrypts a plaintext message using XChaCha20Poly1305 and returns a COSE Encrypt0 message
pub(crate) fn encrypt_xchacha20_poly1305(
    plaintext: &[u8],
    key: &crate::XChaCha20Poly1305Key,
    content_format: ContentFormat,
) -> Result<Vec<u8>, CryptoError> {
    let mut plaintext = plaintext.to_vec();

    let protected_header: coset::Header = content_format.into();

    if should_pad_content(&content_format) {
        // Pad the data to a block size in order to hide plaintext length
        crate::keys::utils::pad_bytes(&mut plaintext, XCHACHA20_TEXT_PAD_BLOCK_SIZE);
    }

    let mut nonce = [0u8; xchacha20::NONCE_SIZE];
    let cose_encrypt0 = ... same as before ...
}

/// Decrypts a COSE Encrypt0 message, using a XChaCha20Poly1305 key
pub(crate) fn decrypt_xchacha20_poly1305(
    cose_encrypt0_message: &[u8],
    key: &crate::XChaCha20Poly1305Key,
) -> Result<(Vec<u8>, ContentFormat), CryptoError> {
    let msg = coset::CoseEncrypt0::from_slice(cose_encrypt0_message)
        .map_err(|err| CryptoError::EncString(EncStringParseError::InvalidCoseEncoding(err)))?;

    let content_format = ContentFormat::try_from(&msg.protected.header)?;

    let decrypted_message = ... same as before ...

    if should_pad_content(&content_format) {
        // Unpad the data to get the original plaintext
        let data = crate::keys::utils::unpad_bytes(&decrypted_message)
            .map_err(|_| CryptoError::InvalidPadding)?;
        return Ok((data.to_vec(), content_format));
    }
    Ok((decrypted_message.to_vec(), ContentFormat::OctetStream))
}

```

crates/bitwarden-crypto/src/cose.rs
dani-garcia: It seems reasonable to me to create our own custom content format. Attachments and file sends would probably also use the same content type, and there maybe be some others as well, like the FIDO2 private key (or is that just a base64 string?)

crates/bitwarden-crypto/src/traits/encryptable.rs
```diff
@@ -1,49 +1,116 @@
-use crate::{store::KeyStoreContext, CryptoError, EncString, KeyId, KeyIds};
+use crate::{store::KeyStoreContext, ContentFormat, CryptoError, EncString, KeyId, KeyIds};
 
-/// An encryption operation that takes the input value and encrypts it into the output value.
+/// An encryption operation that takes the input value and encrypts the fields on it recursively.
 /// Implementations should generally consist of calling [Encryptable::encrypt] for all the fields of
 /// the type.
-pub trait Encryptable<Ids: KeyIds, Key: KeyId, Output> {
+pub trait CompositeEncryptable<Ids: KeyIds, Key: KeyId, Output> {
```
dani-garcia, line 26: A couple of things about the traits:

It seems weird that we have `CompositeEncryptable`, `PrimitiveEncryptableWithContentType`, and then the other one is just called `Encryptable`, but it's also a primitive. I can think of some other ways to name them:  
- `CompositeEncryptable`,`PrimitiveEncryptable`, `PrimitiveEncryptableWithContentType`. All their names with prefixes, so there's never any confusion.
- `Encryptable`,`PrimitiveEncryptable`, `PrimitiveEncryptableWithContentType`. Composite is the only one that should be implemented outside the crypto crate, so it gets the simpler name. I would probably go with this one as it would probably make a smaller diff for this PR.

---

This can be left for a followup PR, but now that we have the split, I think it would be a good idea to ensure that the Primitive traits can't be implemented outside this crate. We can use sealed traits for that:
```rust
// This trait is only accessible as part of a pub(crate) module, so no one else would be able to implement it
pub(crate) mod private {
    pub trait Sealed {}
}
pub trait CompositeEncryptable<Ids: KeyIds, Key: KeyId, Output>: private::Sealed {
```

---

Should we also have a similar split for `Decryptable`? Presumably to validate the content types there.

---

Now that we have multiple encryptable traits, I wonder if it would be a good idea to publicly export the `traits` module and encourage users to do `use bitwarden_crypto::traits::*;` to avoid having to import them all.

crates/bitwarden-crypto/src/cose.rs
quexten: Yes, so for FIDO2, the private key seems to be a PKCS8 DER, B64 encoded (but without the typical PEM header/footer) before encrypting. So we cannot give it the pkcs8 content type, as that refers to PKCS8 DER without B64, but string/our padded string seems reasonable. (Or, I guess we could skip the b64 encoding step when encrypting with a COSE key, and mark it as PKCS8? I'm not sure whether that is worth the effort to ensure correctness with old encryption in all places...)


I'm not sure about files (attachments/send attachments); they seem like just byte buffers, so octetstream seems fine.

Hinton:
Please fill out the PR template.

quexten:
> Please fill out the PR template.

@Hinton Sorry, not sure what happened here, maybe GitHub ate the edit of the description at the time. Fixed now.

crates/bitwarden-crypto/src/traits/encryptable.rs
```diff
@@ -1,49 +1,116 @@
-use crate::{store::KeyStoreContext, CryptoError, EncString, KeyId, KeyIds};
+use crate::{store::KeyStoreContext, ContentFormat, CryptoError, EncString, KeyId, KeyIds};
```
Hinton, line 20: I want a more clear explanation about the difference between the traits. We should be mindful that we'd eventually want to migrate to blob based encryption and a more flexible data model for ciphers which will at least partially migrate us away from nested encrypt.

quexten: I'll try, though it would be helpful if you pointed out specific questions to where the PR's description falls short so I know which areas to focus on.

**CompositeEncryptable**:
This trait is used for structs that are not directly encrypted, but have child encryptables (such as the fields for a cipher). Subsequently, there is no associated content type.

Blob encryption is separate, and the split makes this very explicit. If we wanted to implement blob encryption, we would implement `PrimitiveEncryptableWithContentType`, additionally. The impl could look something like:
```
fn encrypt(...) {
  let cipher_blob: String = cipher.to_json();
  cipher_blob.encrypt(&ctx, &key) // String has content type UTF8
}
```

Depending on whether you want the cipherblob or the composite encryption (which we deprecate at that point), we would then call `.encrypt`, or `.encrypt_composite`.

**PrimitiveEncryptableWithContentType**:
This is implemented for any types that have a unique content type associated. For strings this is Utf8. The caller cannot pass along their own ContentFormat.

**Encryptable**:
This is implemented for any types that do not have a unique content type associated. The caller MUST pass along the ContentFormat.

Hinton: Apologies for the confusion, I want documentation in this file as top level documentation of the module

quexten: Ohh, okay, yeah I agree. I'll expand it.

crates/bitwarden-crypto/src/cose.rs
quexten: Nice, changed to this split for the most part. I did not move the alg validation into the `TryFrom<&coset::Header>` implementation, because it does not really have much to do with the Content Format.

crates/bitwarden-crypto/src/traits/encryptable.rs
quexten: > Should we also have a similar split for Decryptable? Presumably to validate the content types there.

We can; For now I'm ok not validating them except for CoseKey. CoseKey is the only option where we actually currently distinguish based on content type for now. So is it fine keeping that as a separate ticket?

quexten: > CompositeEncryptable,PrimitiveEncryptable, PrimitiveEncryptableWithContentType

I prefer this approach, and went with it!

crates/bitwarden-crypto/src/traits/encryptable.rs
quexten: Added

crates/bitwarden-crypto/src/cose.rs
MGibson1: If we're going with a `BitwardenLegacyKey`, is there a reason not to use this enum to precisely define content of anything that encrypted?

That definitely has value to me from a type safety perspective, but I don't know if a format flag is the right place to do that.

crates/bitwarden-crypto/src/keys/utils.rs
```diff
@@ -16,6 +16,27 @@ pub(super) fn stretch_key(key: &Pin<Box<GenericArray<u8, U32>>>) -> Result<Aes25
...
+/// For example, padded to size 4, the value 0,0 becomes 0,0,2,2.
+pub(crate) fn pad_bytes(bytes: &mut Vec<u8>, min_length: usize) {
+    // at least 1 byte of padding is required
+    let pad_bytes = min_length.saturating_sub(bytes.len()).max(1);
+    let padded_length = max(min_length, bytes.len() + 1);
+    bytes.resize(padded_length, pad_bytes as u8);
+}
+
+/// Unpads bytes that is padded using the PKCS7-like padding defined by [pad_bytes].
+/// The last N bytes of the padded bytes all have the value N.
```
MGibson1:
```suggestion
/// The last N bytes of the padded bytes all have the value N. Minimum of 1 padding byte.
```

crates/bitwarden-crypto/src/keys/utils.rs
```diff
@@ -16,6 +16,27 @@ pub(super) fn stretch_key(key: &Pin<Box<GenericArray<u8, U32>>>) -> Result<Aes25
     })
 }
 
+/// Pads bytes to a minimum length using PKCS7-like padding.
+/// The last N bytes of the padded bytes all have the value N.
```
MGibson1:
```suggestion
/// The last N bytes of the padded bytes all have the value N. Minimum of 1 padding byte.
```

crates/bitwarden-crypto/src/keys/utils.rs
```diff
@@ -16,6 +16,27 @@ pub(super) fn stretch_key(key: &Pin<Box<GenericArray<u8, U32>>>) -> Result<Aes25
...
+    // at least 1 byte of padding is required
+    let pad_bytes = min_length.saturating_sub(bytes.len()).max(1);
+    let padded_length = max(min_length, bytes.len() + 1);
+    bytes.resize(padded_length, pad_bytes as u8);
+}
+
+/// Unpads bytes that is padded using the PKCS7-like padding defined by [pad_bytes].
+/// The last N bytes of the padded bytes all have the value N.
+/// For example, padded to size 4, the value 0,0 becomes 0,0,2,2.
+pub(crate) fn unpad_bytes(padded_bytes: &[u8]) -> Result<&[u8], CryptoError> {
```
MGibson1: It seems odd these mutate, are you trying to take advantage of some `Vec` amortized resizing costs?

crates/bitwarden-crypto/src/keys/utils.rs
```diff
@@ -16,6 +16,27 @@ pub(super) fn stretch_key(key: &Pin<Box<GenericArray<u8, U32>>>) -> Result<Aes25
...
+    let pad_len = *padded_bytes.last().ok_or(CryptoError::InvalidPadding)? as usize;
+    if pad_len >= padded_bytes.len() {
+        return Err(CryptoError::InvalidPadding);
+    }
+    Ok(padded_bytes[..padded_bytes.len() - pad_len].as_ref())
```
MGibson1: **nit**: If clippy/fmt allow it parenthesis here improve readability
```suggestion
    Ok(padded_bytes[..(padded_bytes.len() - pad_len)].as_ref())
```

crates/bitwarden-crypto/src/store/context.rs
```diff
@@ -131,25 +133,54 @@ impl<Ids: KeyIds> KeyStoreContext<'_, Ids> {
...
+        let wrapping_key = self.get_symmetric_key(wrapping_key)?;
+
+        let key = match (wrapped_key, wrapping_key) {
+            (EncString::Aes256Cbc_B64 { iv, data }, SymmetricCryptoKey::Aes256CbcKey(key)) => {
+                SymmetricCryptoKey::try_from(crate::aes::decrypt_aes256(
+                    iv,
+                    data.clone(),
+                    &key.enc_key,
+                )?)?
+            }
```
MGibson1: Is this not an invalid key?

crates/bitwarden-crypto/src/traits/encryptable.rs
```diff
@@ -1,83 +1,169 @@
-use crate::{store::KeyStoreContext, CryptoError, EncString, KeyId, KeyIds};
+//! This module defines traits for encrypting data. There are three categories here.
+//!
+//! Some (legacy) encryptables are made up of many small individually encrypted items. For instance,
+//! a cipher is currently made up of many small `EncString`s and some further json objects that
+//! themselves contain `EncString`s. The use of this is generally discouraged for new designs.
+//! Still, this is generally the only trait that should be implemented outside of the crypto crate.
```
MGibson1, line 6: Is the intention that this is true in the long term? I expect that the ideal is this trait goes away and direct an evolutionary serialization (in the sense of EDD) to `&[u8]` is more desired.

crates/bitwarden-crypto/src/traits/encryptable.rs
MGibson1: Regarding trait sealing:

How important is that, really? For primitives like `Option`, `String`, `&str`, these can't be expanded outside of the crate since those crates won't own a type. It seems worthwhile allowing people to define fixed content type impls if they want to. In fact, that seems like the safer direction to go for a given type.

MGibson1 (commented):
Mostly piling onto previously existing conversations

crates/bitwarden-crypto/src/store/context.rs
quexten: Is this asking whether Aes256-CBC is still supported? Or is this asking whether this code is correctly implementing it?

The former is dropping partially, but still supported for masterkey-wrapped-userkeys (but nothing else). For the latter, I'm not sure what "invalid key" means here.

quexten: I do think we can drop  `EncString::Aes256Cbc_B64 { .. }` entirely, but that should be separate work.

crates/bitwarden-crypto/src/traits/encryptable.rs
quexten: No, we say "a cipher is **currently** made up". The implication is that there is an intent to change this. The concrete intent to change this is blob ciphers.

crates/bitwarden-crypto/src/keys/utils.rs
quexten: In this case, I think the only goal in the xchacha pr was to make the code readable.

crates/bitwarden-crypto/src/cose.rs
quexten: I'm OK with using a custom content format here.

Some things to keep in mind: Custom content formats - in contrast to standardized ones - are defined using strings. The standardized ones sadly don't offer a privateuse category:
> 0-255 | Expert Review
-- | --
256-9999 | IETF Review or IESG Approval
10000-64999 | First Come First Served
65000-65535 | Experimental use (no operational use)

(This is also the case with encstrings, that contain padded utf8 text, which are much more common right now).

The second: So far I've introduced content formats only where they are required to distinguish. (PKCS8 / CoseKey, Utf8/Octetstream CoseKey/OctetStream[or BitwardenLegacyKey].). OctetStream should be unique here, but BitwardenLegacyKey makes this much clearer.

Both of these are not a big deal, and don't hinder using it, and I'm happy to change to it. Given the context, should I change it?

crates/bitwarden-crypto/src/traits/encryptable.rs
Hinton: Per this definition SM secrets should be composite but they use Primitive? Or does this only matter to json structures and if we store values not in a json structure they are primitive?

crates/bitwarden-crypto/src/cose.rs
Hinton: I think the hassle of being to specific is the crypto system always has to be updated with new types. But maybe that is something we desire? I think what I'm asking is should consumers be allowed to define a custom content format?

The fact this is public seems to imply the caller decides the content? In reality we'd like the type system to decide this for us, since you SHOULD not be able to encrypt a key with the wrong content type.

crates/bitwarden-core/src/mobile/crypto.rs
```diff
@@ -334,7 +334,7 @@ pub(super) fn derive_pin_key(
 
     Ok(DerivePinKeyResponse {
         pin_protected_user_key,
-        encrypted_pin: pin.encrypt_with_key(user_key)?,
+        encrypted_pin: pin.encrypt_with_key(user_key, ContentFormat::Utf8)?,
```
Hinton: I am slightly concerned with `ContentFormat` leaking outside the crypto crate. Having to rely on the callers to set the content seems potentially fragile.

I wonder if we could enforce compile type safety here, a loose thought I have is what if we define

```rust
struct EncryptedString {}

impl Encryptable for EncryptedString {
  fn content_type() {
    ContentFormat::Utf8
  }
}

let encrypted: EncryptedString = EncryptedString::encrypt(pin, user_key);

let decrypted: String = encrypted.decrypt(user_key)
```

Hinton: I think @dani-garcia explored something tangentially but we had trouble getting a decent non clunky interface for it.

crates/bitwarden-crypto/src/cose.rs
```diff
@@ -15,13 +21,42 @@ use crate::{
...
+
+/// The content format describes the format of the contained bytes. Message encryption always
+/// happens on the byte level, and this allows determining what format the contained data has. For
+/// instance, an `EncString` in most cases contains UTF-8 encoded text. In some cases it may contain
+/// a Pkcs8 private key, or a COSE key. Specifically, for COSE keys, this allows distinguishing
+/// between the old symmetric key format, represented as `ContentFormat::OctetStream`, and the new
+/// COSE key format, represented as `ContentFormat::CoseKey`.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Serialize, Deserialize)]
+#[cfg_attr(feature = "wasm", derive(Tsify), tsify(into_wasm_abi, from_wasm_abi))]
+pub enum ContentFormat {
```
Hinton: **issue:** While `ContentFormat` is consumed by cose keys the fact we expect these as arguments to our public encrypt/decrypt interfaces indicates they don't belong to cose.

**suggestion:** Move it up out of cose. Maybe a top level concern?

crates/bitwarden-crypto/src/store/context.rs
```diff
@@ -131,25 +133,54 @@ impl<Ids: KeyIds> KeyStoreContext<'_, Ids> {
...
     pub fn unwrap_symmetric_key(
         &mut self,
-        encryption_key: Ids::Symmetric,
+        wrapping_key: Ids::Symmetric,
         new_key_id: Ids::Symmetric,
-        encrypted_key: &EncString,
+        wrapped_key: &EncString,
     ) -> Result<Ids::Symmetric> {
-        let mut new_key_material =
-            self.decrypt_data_with_symmetric_key(encryption_key, encrypted_key)?;
```
Hinton, line 154: It's unfortunate the unwrap diverges from decrypt since it's essentially the same operation just different content types?

Hinton:
I don't have anything blocking here. I think we could explore some alternative approaches to couple content and content type, but we'd need to spend some time to see if we can find a more intuitive interface.

crates/bitwarden-crypto/src/cose.rs
quexten: > I think what I'm asking is should consumers be allowed to define a custom content format?

No. This should probably not be done both from the COSE side (CoAP format numbers require registration, and unregistered IANA format labels that we currently use, are discouraged - vendor specific require registration), but also from our side I don't think we want consumers defining custom content formats.

Assuming we offer a way to encrypt entire structs as cbor encoded with cose, then 
they can always make a struct: 
```
{
  my_data: Bytebuf, // their custom encoded thing.
}
```
, and hand this to the crypto layer.

quexten: > The fact this is public seems to imply the caller decides the content? In reality we'd like the type system to decide this for us, since you SHOULD not be able to encrypt a key with the wrong content type.

I would like to lock this down more, but this requires further refactors that seemed out of scope. I'm happy to make a follow-up, taking away more ways that the format gets leaked outside of the crypto crate.

(or is this blocking?)

crates/bitwarden-crypto/src/cose.rs
quexten: Moved to a separate file

quexten:
> I don't have anything blocking here. I think we could explore some alternative approaches to couple content and content type, but we'd need to spend some time to see if we can find a more intuitive interface.

I agree on looking for a more intuitive interface as follow-up work. I believe `DataEnvelope` from https://github.com/bitwarden/sdk-internal/pull/308 to be such an interface, but it is not backwards compatible with existing crypto.

We may improve the interface of current crypto too. But I think this is follow-up.

crates/bitwarden-core/src/key_management/crypto.rs
```diff
@@ -604,9 +604,11 @@ pub fn make_user_signing_keys_for_enrollment(
         verifying_key: STANDARD.encode(signature_keypair.to_verifying_key().to_cose()),
         // This needs to be changed to use the correct COSE content format before rolling out to
```
dani-garcia: Is this comment relevant anymore?

crates/bitwarden-core/src/key_management/crypto.rs
```diff
@@ -604,9 +604,11 @@ pub fn make_user_signing_keys_for_enrollment(
...
         // users: https://bitwarden.atlassian.net/browse/PM-22189
-        signing_key: signature_keypair
-            .to_cose()
-            .encrypt(&mut ctx, SymmetricKeyId::User)?,
+        signing_key: signature_keypair.to_cose().encrypt(
```
dani-garcia: Probably a future enhancement, but I wonder if we should have the `to_cose()` function return some `CoseBytes` or `Bytes<CoseKey>` wrapper that implements `PrimitiveEncryptableWithContentType`, to avoid the posibility of using the wrong content type here.

crates/bitwarden-crypto/src/content_format.rs
```diff
@@ -0,0 +1,28 @@
...
+    Utf8,
+    /// Pkcs8 private key DER
+    Pkcs8,
+    /// COSE serialized CoseKey
+    CoseKey,
+    /// Bitwarden Legacy Key
+    /// There are three permissible byte values here:
+    /// - `[u8; 32]` - AES-CBC (no hmac) key. This is to be removed and banned.
+    /// - `[u8; 64]` - AES-CBC with HMAC key. This is the v1 userkey key type
+    /// - `[u8; >64]` - COSE key. Padded to be larger than 64 bytes.
```
dani-garcia, line 31: >     /// - `[u8; >64]` - COSE key. Padded to be larger than 64 bytes.

Shouldn't the COSE key use the `ContentFormat::CoseKey` variant? Or is there a distinction between the two that I'm missing?

crates/bitwarden-crypto/src/keys/utils.rs
dani-garcia: I think I was the one that initially suggested the mutability, and it was due to the fact that almost every use of these functions would take the key by reference, return a reference, and then we'd immediately call `.to_vec()`, which seems like a waste for something that in a lot of cases would not require a new allocation. We can reevaluate that if it's not the case anymore and we use the references directly now.

If we don't want the functions to take `&mut Vec<u8>`, another alternative is to take the key by value and return it again:
```rust
pub(crate) fn pad_bytes(mut bytes: Vec<u8>, min_length: usize) -> Vec<u8> {
    // at least 1 byte of padding is required
    let pad_bytes = min_length.saturating_sub(bytes.len()).max(1);
  
    let padded_length = max(min_length, bytes.len() + 1);
    bytes.resize(padded_length, pad_bytes as u8);
    bytes
}

pub(crate) fn unpad_bytes(mut padded_bytes: Vec<u8>) -> Result<Vec<u8>, CryptoError> {
    let pad_len = *padded_bytes.last().ok_or(CryptoError::InvalidPadding)? as usize;
    if pad_len >= padded_bytes.len() {
        return Err(CryptoError::InvalidPadding);
    }
    padded_bytes.truncate(padded_bytes.len() - pad_len);
    Ok(padded_bytes)
}
```

crates/bitwarden-crypto/src/store/context.rs
```diff
@@ -186,10 +219,25 @@ impl<Ids: KeyIds> KeyStoreContext<'_, Ids> {
         // or `Aes256CbcKey`, or by specifying the content format to be CoseKey, in case the
         // wrapped key is a `XChaCha20Poly1305Key`.
         match (wrapping_key_instance, key_to_wrap_instance) {
-            (Aes256CbcHmacKey(_), Aes256CbcHmacKey(_) | Aes256CbcKey(_)) => self
+            (
+                Aes256CbcHmacKey(_),
+                Aes256CbcHmacKey(_) | Aes256CbcKey(_) | XChaCha20Poly1305Key(_),
+            ) => self.encrypt_data_with_symmetric_key(
+                wrapping_key,
+                key_to_wrap_instance.to_encoded().as_slice(),
```
dani-garcia: For the AES key variants we're using `to_encoded` while for the chacha variants we use `to_encoded_raw`. For AES keys both functions return the same, so should we switch them to all use `to_encoded_raw`? It was a bit confusing to me that they were using different functions.

crates/bitwarden-crypto/src/store/context.rs
dani-garcia: Yeah, I'm wondering if it's cleaner to change `decrypt_data_with_symmetric_key` to return `Result<(Vec<u8>, Option<ContentFormat>)>` and then update this function to be like this:

```rust
    pub fn unwrap_symmetric_key(
        &mut self,
        wrapping_key: Ids::Symmetric,
        new_key_id: Ids::Symmetric,
        wrapped_key: &EncString,
    ) -> Result<Ids::Symmetric> {
        let (mut new_key_material, content_type) =
            self.decrypt_data_with_symmetric_key(wrapping_key, wrapped_key)?;

        let wrapping_key = self.get_symmetric_key(wrapping_key)?;
        let new_key = match (wrapping_key, wrapped_key, content_type) {
            (
                SymmetricCryptoKey::XChaCha20Poly1305Key(_),
                EncString::Cose_Encrypt0_B64 { .. },
                Some(ContentFormat::CoseKey),
            ) => SymmetricCryptoKey::try_from_cose(new_key_material.as_mut_slice())?,
            _ => SymmetricCryptoKey::try_from(new_key_material.as_mut_slice())?,
        };

        #[allow(deprecated)]
        self.set_symmetric_key(new_key_id, new_key)?;

        // Returning the new key identifier for convenience
        Ok(new_key_id)
    }
```

crates/bitwarden-core/src/key_management/crypto.rs
quexten: Removed

crates/bitwarden-core/src/key_management/crypto.rs
quexten: This is handled now since this PR is switched to use `SerializedBytes<T>`.

In this case, we use a marker trait for "Cose objects" as the generic parameter. And mark currently two content types (CoseKey, CoseSign1) with this marker trait.

crates/bitwarden-crypto/src/store/context.rs
quexten: Yeah, this was a bit confusing, so I've changed the return types here too to better (more safely) represent the types (I hope).

To clarify:

When wrapped with AES, or encrypted with RSA, then we have the following encodings:
Aes256-cbc: |enc_key| (32) // no content format
Aes256-cbc-hmac: |enc_key|mac_key| (64) // no content format
Xchacha/cose: |cose|padding| (>64) // no content format
The type is determined by the length.

When wrapped with Xchacha/cose, we have the following encodings:
Aes256-cbc: |enc_key| (32) // Bitwarden legacy key content format
Aes256-cbc-hmac: |enc_key|mac_key| (64) // Bitwarden legacy key content format
Xchacha/cose: |cose| (???) // CoseKey content format
The type is determined by the content format.

quexten:
> If we don't want the functions to take &mut Vec<u8>, another alternative is to take the key by value and return it again:

The decision about this feels like it's not that important here, and we can later change it. (In a different PR).

crates/bitwarden-crypto/src/content_format.rs
quexten: The distinction here is the padding. In the BitwardenLegacyKey format, 32/64 bytes are unpadded AES keys, and >64 bytes is padded COSE keys. We need this in case we don't have a content format to hint at what key it is.

When we do have a content format to indicate what key type it is, then CoseKey is not padded, but marked as CoseKey in the content format.

quexten:
@Hinton / @dani-garcia Given the discussion around not leaking out content format, I've redesigned this PR to go with the generic Byte<T> approach suggested in our slack discussion. For String, this is automatically inferred for KeyEncryptable and Encryptable. For some cases, you have to explicitly convert your untyped vecs to the typed bytes. However, with FFI improvements, I think we can later reduce these instances.

But that does mean this PR grew in size significantly.

Note: This also drops some of the zeroizing we were doing. I believe this is covered by the global zeroizing allocator however.

crates/bitwarden-core/src/auth/auth_request.rs
```diff
@@ -52,7 +52,7 @@ pub(crate) fn auth_request_decrypt_user_key(
     private_key: String,
     user_key: UnsignedSharedKey,
 ) -> Result<SymmetricCryptoKey, EncryptionSettingsError> {
-    let key = AsymmetricCryptoKey::from_der(&STANDARD.decode(private_key)?)?;
+    let key = AsymmetricCryptoKey::from_der(&STANDARD.decode(private_key)?.into())?;
```
Hinton, line 55: **question:** Do you think we could implement a generic `FromStr` for `Bytes`? Or a specialized one for `Pkcs8PrivateKeyBytes`, now that we have a specific type for it would be nice to get rid of the manual b64 decoding.

```suggestion
    let key = AsymmetricCryptoKey::from_der(private_key.parse()?)?;
```

And for other b64 conversions throughout the PR.

crates/bitwarden-core/src/client/encryption_settings.rs
```diff
@@ -49,19 +52,18 @@ impl EncryptionSettings {
...
-            AsymmetricCryptoKey, CoseSerializable, CryptoError, KeyDecryptable, SigningKey,
-        };
+        use bitwarden_crypto::{AsymmetricCryptoKey, CoseSerializable, KeyDecryptable, SigningKey};
         use log::warn;
 
         use crate::key_management::{AsymmetricKeyId, SigningKeyId, SymmetricKeyId};
 
         let private_key = {
-            let dec: Vec<u8> = private_key.decrypt_with_key(&user_key)?;
+            use bitwarden_crypto::Bytes;
```
Hinton: **suggestion:** Lift this up to line 55.

crates/bitwarden-core/src/client/encryption_settings.rs
```diff
@@ -49,19 +52,18 @@ impl EncryptionSettings {
...
 
+            let dec: Vec<u8> = private_key.decrypt_with_key(&user_key)?;
             // FIXME: [PM-11690] - Temporarily ignore invalid private keys until we have a recovery
             // process in place.
-            AsymmetricCryptoKey::from_der(&dec)
+            AsymmetricCryptoKey::from_der(&Bytes::from(dec))
```
Hinton: **suggestion:** Use `Pkcs8PrivateKeyBytes` to be more explicit.

```suggestion
            AsymmetricCryptoKey::from_der(&Pkcs8PrivateKeyBytes::from(dec))
```

crates/bitwarden-core/src/client/encryption_settings.rs
```diff
@@ -72,25 +74,28 @@ impl EncryptionSettings {
...
-            if let Some(private_key) = private_key {
-                ctx.set_asymmetric_key(AsymmetricKeyId::UserPrivateKey, private_key)?;
-            }
 
             if let Some(signing_key) = signing_key {
+                use bitwarden_crypto::Bytes;
+
+                let dec: Vec<u8> = signing_key
+                    .decrypt_with_key(&user_key)
+                    .map_err(|_| EncryptionSettingsError::InvalidSigningKey)?;
```
Hinton: **thought (non blocking, future improvement):** It'd be nice if we could eventually decrypt immediately to the Byte representation. Maybe have `decrypt_with_key` return `From<Vec<u8>>` similar to how iterators `collect` return `FromIterator`.

crates/bitwarden-core/src/client/encryption_settings.rs
```diff
@@ -72,25 +74,28 @@ impl EncryptionSettings {
...
+                let signing_key = SigningKey::from_cose(&Bytes::from(dec))
+                    .map_err(|_| EncryptionSettingsError::InvalidSigningKey)?;
                 ctx.set_signing_key(SigningKeyId::UserSigningKey, signing_key)?;
             }
+
+            if let Some(private_key) = private_key {
+                ctx.set_asymmetric_key(AsymmetricKeyId::UserPrivateKey, private_key)?;
+            }
```
Hinton: **question:** Since this moved below signing keys does that mean the order of operation is important? Do we need to document this with comments?

crates/bitwarden-core/src/key_management/crypto.rs
```diff
@@ -536,6 +536,7 @@ pub(super) fn verify_asymmetric_keys(
             .decrypt_with_key(user_key)
             .map_err(VerifyError::DecryptFailed)?;
 
+        let decrypted_private_key: Pkcs8PrivateKeyBytes = Bytes::from(decrypted_private_key);
```
Hinton:
```suggestion
        let decrypted_private_key = Pkcs8PrivateKeyBytes::from(decrypted_private_key);
```

crates/bitwarden-core/src/platform/generate_fingerprint.rs
```diff
@@ -42,7 +42,7 @@ pub enum FingerprintError {
 
 pub(crate) fn generate_fingerprint(input: &FingerprintRequest) -> Result<String, FingerprintError> {
     let key = STANDARD.decode(&input.public_key)?;
-
+    let key = SpkiPublicKeyBytes::from(key);
     Ok(fingerprint(&input.fingerprint_material, &key)?)
```
Hinton:
```suggestion
    Ok(fingerprint(&input.fingerprint_material, &key.into())?)
```

crates/bitwarden-crypto/README.md
```diff
@@ -13,7 +13,7 @@ secure.
 ## Example:
 
 ```rust
-use bitwarden_crypto::{SymmetricCryptoKey, KeyEncryptable, KeyDecryptable, CryptoError};
+use bitwarden_crypto::{SymmetricCryptoKey, KeyEncryptable, KeyDecryptable, CryptoError, ContentFormat};
```
Hinton:
```suggestion
use bitwarden_crypto::{SymmetricCryptoKey, KeyEncryptable, KeyDecryptable, CryptoError};
```

crates/bitwarden-crypto/src/content_format.rs
```diff
@@ -0,0 +1,220 @@
...
+
+/// The content format describes the format of the contained bytes. Message encryption always
+/// happens on the byte level, and this allows determining what format the contained data has. For
+/// instance, an `EncString` in most cases contains UTF-8 encoded text. In some cases it may contain
+/// a Pkcs8 private key, or a COSE key. Specifically, for COSE keys, this allows distinguishing
+/// between the old symmetric key format, represented as `ContentFormat::OctetStream`, and the new
+/// COSE key format, represented as `ContentFormat::CoseKey`.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Serialize, Deserialize)]
+#[cfg_attr(feature = "wasm", derive(Tsify), tsify(into_wasm_abi, from_wasm_abi))]
+pub enum ContentFormat {
```
Hinton: **issue:** `ContentFormat` is internal to this crate and shouldn't need to be exposed.

I think we can tweak the sealing slightly and get rid of the compiler warning. Let me know if you would be open to this. @dani-garcia curious what you're thoughts are as well.

```rust
mod private {
    use crate::ConstContentFormat;

    /// This trait is used to seal the `ConstContentFormat` trait, preventing external
    /// implementations.
    pub trait SealedConstContentFormat: ConstContentFormat {}
}

pub(crate) use private::SealedConstContentFormat;

/// This trait is used to instantiate different typed byte vectors with a specific content format,
/// using `SerializedBytes<C>`. This allows for compile-time guarantees about the content format
/// of the serialized bytes. The exception here is the escape hatch using e.g. `from(Vec<u8>)`,
/// which can still be mis-used, but has to be misused explicitly.
pub trait ConstContentFormat {
    /// Returns the content format as a `ContentFormat` enum.
    fn content_format() -> ContentFormat;
}

/// A serialized byte array with a specific content format. This is used to represent data that has
/// a specific format, such as UTF-8 encoded text, raw bytes, or COSE keys. The content
/// format is used to determine how the bytes should be interpreted when encrypting or decrypting
/// the data.
#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
pub struct Bytes<C: SealedConstContentFormat> {
    inner: Vec<u8>,
    _marker: std::marker::PhantomData<C>,
}
```

dani-garcia: You'll still get a warning that `ContentFormat` is more private than `ConstContentFormat::content_format` right?

Hinton: Sight yes. I forgot to change the visibility of one part, I guess we have to use allow

Hinton: I think we just need to do `allow` here in that case, we're a bit less worried about leaking visibility.

crates/bitwarden-vault/src/cipher/cipher.rs
```diff
@@ -425,7 +432,7 @@ impl CipherView {
         ctx: &mut KeyStoreContext<KeyIds>,
         key: SymmetricKeyId,
     ) -> Result<(), CryptoError> {
-        let old_ciphers_key = Cipher::decrypt_cipher_key(ctx, key, &self.key)?;
+        let old_ciphers_key = Cipher::decrypt_cipher_key(ctx, key, &self.key.take())?;
```
Hinton: **question:** Why did we change this to take?

crates/bitwarden-vault/src/cipher/cipher.rs
```diff
@@ -463,8 +470,9 @@ impl CipherView {
         if let Some(attachments) = &mut self.attachments {
             for attachment in attachments {
                 if let Some(attachment_key) = &mut attachment.key {
-                    let dec_attachment_key: Vec<u8> = attachment_key.decrypt(ctx, old_key)?;
-                    *attachment_key = dec_attachment_key.encrypt(ctx, new_key)?;
+                    let tmp_attachment_key_id = SymmetricKeyId::Local("attachment_key");
+                    ctx.unwrap_symmetric_key(old_key, tmp_attachment_key_id, attachment_key)?;
```
Hinton, line 476: **question:** Should we have a way to delete keys after we're done with them? CC @dani-garcia

crates/bitwarden-crypto/src/keys/key_encryptable.rs
```diff
@@ -24,6 +24,11 @@ pub trait KeyEncryptable<Key: CryptoKey, Output> {
     fn encrypt_with_key(self, key: &Key) -> Result<Output>;
 }
 
+#[allow(missing_docs)]
```
Hinton: **issue:** Not public so this allow should not be needed.

crates/bitwarden-crypto/src/keys/pin_key.rs
```diff
@@ -31,15 +32,15 @@ impl PinKey {
 
 impl CryptoKey for PinKey {}
 
-impl KeyEncryptable<PinKey, EncString> for &[u8] {
-    fn encrypt_with_key(self, key: &PinKey) -> Result<EncString> {
+impl KeyEncryptableWithContentType<PinKey, EncString> for &[u8] {
+    fn encrypt_with_key(self, key: &PinKey, content_format: ContentFormat) -> Result<EncString> {
```
Hinton: **thought:** Should the `ContentFormat` really be up to the caller and not `PinKey`? It feels like `PinKey` know's what it is and therefore knows how to structure it.

I.e. calling `pin_key.encrypt_with_key(..., ContentFormat::SPKIPublicKeyDer)` doesn't make any sense.

crates/bitwarden-crypto/src/keys/signed_public_key.rs
```diff
@@ -79,14 +82,18 @@ pub struct SignedPublicKey(pub(crate) SignedObject);
...
 
 impl TryFrom<Vec<u8>> for SignedPublicKey {
     type Error = EncodingError;
     fn try_from(bytes: Vec<u8>) -> Result<Self, EncodingError> {
-        Ok(SignedPublicKey(SignedObject::from_cose(&bytes)?))
+        Ok(SignedPublicKey(SignedObject::from_cose(&Bytes::<
+            CoseSign1ContentFormat,
+        >::from(
+            bytes
+        ))?))
```
Hinton:
```suggestion
        Ok(SignedPublicKey(SignedObject::from_cose(&CoseSign1ContentFormat::from(
            bytes
        ))?))
```

crates/bitwarden-crypto/src/enc_string/symmetric.rs
```diff
@@ -316,13 +322,13 @@ impl KeyDecryptable<SymmetricCryptoKey, Vec<u8>> for EncString {
 
 impl KeyEncryptable<SymmetricCryptoKey, EncString> for String {
     fn encrypt_with_key(self, key: &SymmetricCryptoKey) -> Result<EncString> {
-        self.as_bytes().encrypt_with_key(key)
+        Into::<Utf8Bytes>::into(self).encrypt_with_key(key)
```
Hinton:
```suggestion
        Utf8Bytes::from(self).encrypt_with_key(key)
```

crates/bitwarden-crypto/src/enc_string/symmetric.rs
```diff
@@ -316,13 +322,13 @@ impl KeyDecryptable<SymmetricCryptoKey, Vec<u8>> for EncString {
...
     }
 }
 
 impl KeyEncryptable<SymmetricCryptoKey, EncString> for &str {
     fn encrypt_with_key(self, key: &SymmetricCryptoKey) -> Result<EncString> {
-        self.as_bytes().encrypt_with_key(key)
+        Into::<Utf8Bytes>::into(self).encrypt_with_key(key)
```
Hinton:
```suggestion
        Utf8Bytes::from(self).encrypt_with_key(key)
```

crates/bitwarden-crypto/src/signing/signed_object.rs
```diff
@@ -196,22 +199,25 @@ mod tests {
 
     #[test]
     fn test_roundtrip_cose() {
-        let signed_object = SignedObject::from_cose(SIGNED_OBJECT).unwrap();
+        let signed_object =
+            SignedObject::from_cose(&Into::<CoseSign1Bytes>::into(SIGNED_OBJECT)).unwrap();
```
Hinton:
```suggestion
            SignedObject::from_cose(&<CoseSign1Bytes>::from(SIGNED_OBJECT)).unwrap();
```

crates/bitwarden-crypto/src/signing/signed_object.rs
```diff
@@ -196,22 +199,25 @@ mod tests {
...
     fn test_verify_and_unwrap_testvector() {
         let test_message = TestMessage {
             field1: "Test message".to_string(),
         };
-        let signed_object = SignedObject::from_cose(SIGNED_OBJECT).unwrap();
-        let verifying_key = VerifyingKey::from_cose(VERIFYING_KEY).unwrap();
+        let signed_object =
+            SignedObject::from_cose(&Into::<CoseSign1Bytes>::into(SIGNED_OBJECT)).unwrap();
+        let verifying_key =
+            VerifyingKey::from_cose(&Into::<CoseKeyBytes>::into(VERIFYING_KEY)).unwrap();
```
Hinton:
```suggestion
        let signed_object =
            SignedObject::from_cose(&<CoseSign1Bytes>::from(SIGNED_OBJECT)).unwrap();
        let verifying_key = VerifyingKey::from_cose(&<CoseKeyBytes>::from(VERIFYING_KEY)).unwrap();
```

crates/bitwarden-crypto/src/signing/verifying_key.rs
```diff
@@ -127,7 +129,8 @@ mod tests {
 
     #[test]
     fn test_cose_roundtrip_encode_verifying() {
-        let verifying_key = VerifyingKey::from_cose(VERIFYING_KEY).unwrap();
+        let verifying_key =
+            VerifyingKey::from_cose(&CoseKeyBytes::from(VERIFYING_KEY.to_vec())).unwrap();
```
Hinton: **issue:** No need for `to_vec` in this file. Bytes implement From u8

crates/bitwarden-crypto/src/fingerprint.rs
```diff
@@ -86,10 +87,10 @@ mod tests {
             197, 3, 219, 56, 77, 109, 47, 72, 251, 131, 36, 240, 96, 169, 31, 82, 93, 166, 242, 3,
             33, 213, 2, 3, 1, 0, 1,
         ];
-
+        let key: SpkiPublicKeyBytes = Bytes::from(key.to_vec());
```
Hinton:
```suggestion
        let key = SpkiPublicKeyBytes::from(key);
```

crates/bitwarden-crypto/src/keys/pin_key.rs
quexten: Ah, yeah, that's true. The only legit instances are encrypting a pin (string) and encrypting a (user)symmetrickey.

crates/bitwarden-vault/src/cipher/cipher.rs
quexten: Thought: It would be nice to drop only the keys created in the current function scope, but to always drop these. But that would mean pretty much creating a key-context stack with scopes..

crates/bitwarden-crypto/src/signing/verifying_key.rs
quexten: Removed to_vec from the three usages.

crates/bitwarden-vault/src/cipher/cipher.rs
quexten: Unapplied this change.

crates/bitwarden-crypto/src/keys/key_encryptable.rs
quexten: Removed

crates/bitwarden-crypto/src/keys/pin_key.rs
quexten: Updated, now the `KeyEncryptable<PinKey, ..>` is only implemented for SymmetricCryptoKey and String.

crates/bitwarden-crypto/src/content_format.rs
quexten: Ok, this is now made `pub(crate)` and the warning is ignored/allowed.

crates/bitwarden-core/src/platform/generate_fingerprint.rs
quexten: Applied

crates/bitwarden-core/src/client/encryption_settings.rs
quexten: I've ignored the decrypt path for now because it is not necessary for implementing the format changes, but I agree with the suggestion for future improvement.

crates/bitwarden-core/src/client/encryption_settings.rs
quexten: This has been replaced by Pkcs8PrivateKeyBytes, which has been moved up.

crates/bitwarden-core/src/client/encryption_settings.rs
quexten: This PR is actually the wrong place to do changes to init, and that will be done in a follow-up PR. The eventual state for this changeset is:

- If user is on v1 crypto (has AES userkey) then set userkey, and optionally private key (it may be corrupt)
- If a user is on v2 crypto (has XCHACHA/COSE key), then:
  - If Signing key is missing or undecryptable, fail init
  - If Private key is missing or undecryptable, fail init
  - If Security state is missing or signature verification fails, fail init
  
So the current logic needs to be split into two branches, but that will be done in the security state PR.

crates/bitwarden-core/src/auth/auth_request.rs
quexten: IMO `String` is just not the right type to use here and we should make these string instances `Base64String`, or even `Base64String<PrivateKey/SymmetricKey/whatever>`. FromStr would allow you to easily accidentally just put in some other type of string, but that change feels out of scope here.

For now, I have added a `Base64String` type in `bitwarden-core` and only replace the usages in the current PR with it, but I suggest we explore expanding this, adding generics, and exposing it across the FFI boundary in the long run.

crates/bitwarden-vault/src/cipher/cipher.rs
quexten: Ah, seems Dani already had created plans for this here: https://bitwarden.atlassian.net/browse/PM-18105

dani-garcia: Yeah my initial idea was that Local keys would implement Drop and be autoremoved when outside the current scope, but that turned out to be very hard to do with the current API, as the key would need to hold a mutable ref to the context.

My idea with that ticket is to have some kind of scoping like:
```rust
ctx.scope("reencrypt_attachment_keys", |ctx| {
    let tmp_attachment_key_id = ctx.unwrap_symmetric_key(old_key, attachment_key)?;
    *attachment_key = ctx.wrap_symmetric_key(new_key, tmp_attachment_key_id)?;
});
```

The idea being, scopes are named so we get nicer error reports, and local keys get dropped when near the end of their scope. This would also require https://bitwarden.atlassian.net/browse/PM-18102 to make the API around local keys nicer.

crates/bitwarden-core/src/client/encryption_settings.rs
Hinton: These are only used with the internal flag, if you build with any other flag it will cause a warning.

crates/bitwarden-core/src/client/encryption_settings.rs
Hinton: https://bitwarden.atlassian.net/browse/PM-23069

crates/bitwarden-crypto/src/keys/pin_key.rs
Hinton: Ugh, that makes me really unhappy pin and export keys are the same. We'd never really want to use the pin for strings or export for keys I guess.

Outside the scope of this PR though.

quexten: I'll file splitting export-key into the tech debt epic.

 https://bitwarden.atlassian.net/browse/PM-23070

crates/bitwarden-core/src/auth/auth_request.rs
Hinton: That's fair, I'd want the B64 implementation to reflect https://github.com/bitwarden/credential-exchange/blob/main/credential-exchange-format/src/b64url.rs though.

But is B64 even the accurate interface here? Or should the function interface actually expect a `Pkcs8PrivateKeyBytes` or `AsymmetricCryptoKey` instead, and if it has to cross the FFI interfaces should `Pkcs8PrivateKeyBytes` know how to deserialize itself? Similar to how `EncString` handles it.

crates/bitwarden-crypto/src/keys/pin_key.rs
Hinton: Thanks!

crates/bitwarden-core/src/auth/auth_client.rs
```diff
@@ -89,7 +90,7 @@ impl AuthClient {
     pub fn make_register_tde_keys(
         &self,
         email: String,
-        org_public_key: String,
+        org_public_key: Base64String,
```
Hinton: I vote to not add `Base64String` to this PR I know this was done to address one of my comments but it's not exactly what I objected to as I noted in https://github.com/bitwarden/sdk-internal/pull/203#discussion_r2168709310.

The issue here is this will be a breaking change for mobile and I'd want us to decide should the expected types from mobile and wasm be `B64`, or are do we actually want to request a specific key type? And is that defined by making `B64` generic and having a instruction of what it decodes into? Or by expecting the direct type from FFI?

Some of the problems we run into is parsing errors and how explicit parsing errors are conveyed to consumers and developers.

crates/bitwarden-core/src/auth/auth_request.rs
quexten: As for the `FromStr`, the issue is that Bytes implements another `From<String>`, which is meant for the case that you want to get `Utf8Bytes` from a string. Also implementing `FromStr` seems dangerous because they are easily confused and might conflict.

Do we want to leave it as-is for now, with the b64 changes unapplied, and revisit this when we get to improving the FFI interface?

crates/bitwarden-core/src/client/encryption_settings.rs
quexten: Should be fixed now.

crates/bitwarden-core/src/auth/auth_client.rs
quexten: Just to copy this from the other comment, I unapplied the b64 changes.

crates/bitwarden-core/src/client/encryption_settings.rs
```diff
@@ -49,19 +52,16 @@ impl EncryptionSettings {
...
 
         use crate::key_management::{AsymmetricKeyId, SigningKeyId, SymmetricKeyId};
 
         let private_key = {
             let dec: Vec<u8> = private_key.decrypt_with_key(&user_key)?;
-
             // FIXME: [PM-11690] - Temporarily ignore invalid private keys until we have a recovery
             // process in place.
-            AsymmetricCryptoKey::from_der(&dec)
+            AsymmetricCryptoKey::from_der(&Pkcs8PrivateKeyBytes::from(dec))
```
Hinton: **nit (optional):** We can use `into` here.
```suggestion
            AsymmetricCryptoKey::from_der(&dec.into())
```

crates/bitwarden-core/src/client/encryption_settings.rs
```diff
@@ -74,8 +74,10 @@ impl EncryptionSettings {
         };
         let signing_key = signing_key
             .map(|key| {
+                use bitwarden_crypto::CryptoError;
+
                 let dec: Vec<u8> = key.decrypt_with_key(&user_key)?;
-                SigningKey::from_cose(dec.as_slice()).map_err(Into::<CryptoError>::into)
+                SigningKey::from_cose(&CoseKeyBytes::from(dec)).map_err(Into::<CryptoError>::into)
```
Hinton: **nit (optional):** We can use `into` here.
```suggestion
                SigningKey::from_cose(&dec.into()).map_err(Into::<CryptoError>::into)
```

crates/bitwarden-crypto/src/content_format.rs
```diff
@@ -0,0 +1,222 @@
...
+        ContentFormat::CoseSign1
+    }
+}
+impl CoseContentFormat for CoseSign1ContentFormat {}
+/// CoseSign1Bytes is a type alias for Bytes with `CoseSign1ContentFormat`. This is used for
+/// serialized COSE Sign1 messages.
+pub type CoseSign1Bytes = Bytes<CoseSign1ContentFormat>;
+
+/// A marker trait for COSE content formats.
+pub trait CoseContentFormat {}
```
Hinton: **question:** This seems to be used before it's defined, move higher?

crates/bitwarden-crypto/src/keys/key_encryptable.rs
```diff
@@ -24,6 +24,10 @@ pub trait KeyEncryptable<Key: CryptoKey, Output> {
     fn encrypt_with_key(self, key: &Key) -> Result<Output>;
 }
 
+pub(crate) trait KeyEncryptableWithContentType<Key: CryptoKey, Output> {
```
Hinton, line 38: **suggestion:** Can we document this trait? And how it differs from `KeyEncryptable`?

crates/bitwarden-crypto/src/keys/symmetric_crypto_key.rs
```diff
@@ -145,13 +145,17 @@ impl SymmetricCryptoKey {
...
-        let mut encoded_key = self.to_encoded_raw();
-        match self {
-            Self::Aes256CbcKey(_) | Self::Aes256CbcHmacKey(_) => encoded_key,
-            Self::XChaCha20Poly1305Key(_) => {
+    pub fn to_encoded(&self) -> BitwardenLegacyKeyBytes {
+        let encoded_key = self.to_encoded_raw();
+        match encoded_key {
+            EncodedSymmetricKey::LegacyNonCoseKey(_) => {
+                let encoded_key: Vec<u8> = encoded_key.into();
+                Bytes::from(encoded_key)
```
Hinton:
```suggestion
                BitwardenLegacyKeyBytes::from(encoded_key)
```

crates/bitwarden-crypto/src/keys/symmetric_crypto_key.rs
```diff
@@ -145,13 +145,17 @@ impl SymmetricCryptoKey {
...
+            }
+            EncodedSymmetricKey::CoseKey(_) => {
+                let mut encoded_key: Vec<u8> = encoded_key.into();
                 pad_key(&mut encoded_key, Self::AES256_CBC_HMAC_KEY_LEN + 1);
-                encoded_key
+                Bytes::from(encoded_key)
```
Hinton:
```suggestion
                BitwardenLegacyKeyBytes::from(encoded_key)
```

crates/bitwarden-crypto/src/keys/symmetric_crypto_key.rs
```diff
@@ -367,11 +388,32 @@ fn pad_key(key_bytes: &mut Vec<u8>, min_length: usize) {
 /// size of the byte array the previous key types [SymmetricCryptoKey::Aes256CbcHmacKey] and
 /// [SymmetricCryptoKey::Aes256CbcKey] are 64 and 32 bytes long respectively.
 fn unpad_key(key_bytes: &[u8]) -> Result<&[u8], CryptoError> {
-    let pad_len = *key_bytes.last().ok_or(CryptoError::InvalidKey)? as usize;
-    if pad_len >= key_bytes.len() {
-        return Err(CryptoError::InvalidKey);
+    crate::keys::utils::unpad_bytes(key_bytes).map_err(|_| CryptoError::InvalidKey)
+}
+
+/// An enum to represent the different encodings of symmetric crypto keys.
```
Hinton: **suggestion:** Can we improve the documentation slightly? Calling it an enum is unecessary.

Maybe `Encoded representation of [SymmetricCryptoKey]`

crates/bitwarden-crypto/src/keys/symmetric_crypto_key.rs
```diff
@@ -367,11 +388,32 @@ fn pad_key(key_bytes: &mut Vec<u8>, min_length: usize) {
...
+pub enum EncodedSymmetricKey {
+    /// An Aes256-CBC-HMAC key, or a Aes256-CBC key
+    LegacyNonCoseKey(BitwardenLegacyKeyBytes),
```
Hinton: **suggestion:** It's generally nice to keep the enum variant naming similar to the content. `BitwardenLegacyKey` maybe?

crates/bitwarden-crypto/src/keys/device_key.rs
```diff
@@ -65,6 +65,7 @@ impl DeviceKey {
         protected_user_key: UnsignedSharedKey,
     ) -> Result<SymmetricCryptoKey> {
         let device_private_key: Vec<u8> = protected_device_private_key.decrypt_with_key(&self.0)?;
+        let device_private_key: Pkcs8PrivateKeyBytes = Bytes::from(device_private_key);
```
Hinton:
```suggestion
        let device_private_key = Pkcs8PrivateKeyBytes::from(device_private_key);
```

crates/bitwarden-crypto/src/traits/encryptable.rs
```diff
@@ -1,83 +1,171 @@
...
+//!
+//! Encrypting data directly, a content type must be provided, since an encrypted byte array alone
+//! is not enough to tell the decryption code how to interpret the decrypted bytes. For this, there
```
Hinton, line 9: I think this has improved slightly since we try to infer content type in most cases?

crates/bitwarden-crypto/src/traits/encryptable.rs
```diff
@@ -1,83 +1,171 @@
...
+//! are two traits, `PrimitiveEncryptable` and `PrimitiveEncryptableWithContentType`. The former
+//! assumes that the caller provides the content format when encrypting. The latter is a convenience
```
Hinton: This seems to imply `PrimitiveEncryptableWithContentType`. provides the pre-defined. content format?

Hinton (commented):
Some minor things I think we're good otherwise.

crates/bitwarden-crypto/src/content_format.rs
quexten: Moved.

crates/bitwarden-crypto/src/keys/key_encryptable.rs
quexten: Added docs to both this and `KeyEncryptable`, since `KeyEncryptable` was missing them.

crates/bitwarden-crypto/src/traits/encryptable.rs
quexten: These are outdated from a re-name. I'll fix the section.

crates/bitwarden-crypto/src/traits/encryptable.rs
quexten: Updated the docs to correctly reflect the situation.

crates/bitwarden-core/src/client/encryption_settings.rs
```diff
@@ -74,8 +74,10 @@ impl EncryptionSettings {
         };
         let signing_key = signing_key
             .map(|key| {
+                use bitwarden_crypto::CryptoError;
+
```
Hinton, line 78: **nit:** Move back to line 55. I personally prefer to decrease the number of `use` in the file when possible and this function already have a few imports higher up.

Hinton (approved):

gbubemismith (approved):
Vault changes look good

crates/bitwarden-crypto/src/keys/utils.rs
quexten: Let's leave this up for later refactor if we do see this as a problem.

dani-garcia (approved):

PR #212, opened 2025-04-15
Exposed temporary function to decrypt Key

gbubemismith:
## 📔 Objective
`decrypt_fido2_private_key` is exposed temporarily because the typescript clients still use the internal Fido2 authentication logic. As a result,  we need the `keyValue` on the Fido2CredentialsView to remain in the decrypted state. This workaround will be removed once the migration to the SDK-based  Fido2 Authentication is done.

## ⏰ Reminders before review

- CI builds passed
  team

## 🦮 Reviewer guidelines

### DIFF
--- a/crates/bitwarden-vault/src/cipher/cipher.rs
+++ b/crates/bitwarden-vault/src/cipher/cipher.rs
@@ -6,6 +6,7 @@ use bitwarden_core::{
 use bitwarden_crypto::{
     CryptoError, Decryptable, EncString, Encryptable, IdentifyKey, KeyStoreContext,
 };
+use bitwarden_error::bitwarden_error;
 use chrono::{DateTime, Utc};
 use schemars::JsonSchema;
 use serde::{Deserialize, Serialize};
@@ -30,6 +31,7 @@ use crate::{
     VaultParseError,
 };
 
+#[bitwarden_error(flat)]
 #[derive(Debug, Error)]
 pub enum CipherError {
     #[error(transparent)]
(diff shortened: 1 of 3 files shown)


crates/bitwarden-vault/src/cipher/cipher.rs
```diff
@@ -584,6 +584,31 @@ impl CipherView {
...
+    /// # Arguments
+    /// * `enc_key` - The encrypted key value to decrypt
+    /// * `ctx` - The key store context to use for decryption
+    ///
+    /// # Returns
+    /// * `Ok(String)` - The decrypted key value
+    /// * `Err(CryptoError)` - An error occurred during decryption
+    pub fn decrypt_key(
+        &self,
+        enc_key: EncString,
```
coroiu: suggestion: could we make this a bit more specific/limited maybe? E.g. `decrypt_fido2_private_key` so it doesn't just take a generic enc string?

coroiu: It seems you changed the name but the function itself still takes an arbitrary `EncString` and decrypts it

crates/bitwarden-vault/src/mobile/cipher_client.rs
Hinton, line 1: **question:** Is this used in mobile?

crates/bitwarden-vault/src/cipher/cipher.rs
```diff
@@ -584,6 +584,31 @@ impl CipherView {
...
+    /// This method is a temporary solution to allow typescript client access to decrypted key
+    /// values, particularly for FIDO2 credentials.
+    ///
+    /// # Arguments
+    /// * `enc_key` - The encrypted key value to decrypt
+    /// * `ctx` - The key store context to use for decryption
+    ///
+    /// # Returns
+    /// * `Ok(String)` - The decrypted key value
+    /// * `Err(CryptoError)` - An error occurred during decryption
```
Hinton: **suggestion:** Do these comments provide value? The argument values are pretty self explanatory especially with the types, and `Result<String, CryptoError>` also explains itself.

While we do encourage documentation we should keep it DRY.

Hinton (commented):
A question and a suggestion to trim down the comments slightly.

coroiu (approved):
Much better 👍

crates/bitwarden-vault/src/mobile/cipher_client.rs
gbubemismith: Nah it isn't used

Hinton: Then it should not be placed in `mobile`.

PR #213, opened 2025-04-15
[PM-19479] Client-managed Repository traits

dani-garcia:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-19479

## 📔 Objective

Implement a generic trait for accessing the client application's data storage directly. Because we want the store access to be typed, but `bitwarden_core` isn't aware of the models, `bitwarden_core` only implements a generic way to set and retrieve generic `impl Repository<T>` instances, somewhat like dependency injection. Then it's up to each team/feature crates to define which models and which stores are available.

This feature is created in a new `bitwarden-state`, which will be expanded by a separate PR with the addition of SDK-managed state (Sqlite+IndexedDB). At the moment this crate contains:
- A `Repository` trait which will be implemented by the clients (and the SDK in the future), and a `RepositoryItem` marker trait which will be used to mark which models are meant to be used with the repositories.
- A `StateRegistry` which stores all the client-managed Repositories, and in the future will also handle the SDK-managed repositories.
- A new `StateClient` subclient under platform that will be used by the applications to register their Repositories. Both the WASM and UniFFI crates also need some conversion code to implement the `Repository` traits. I've tried to simplify it as much as possible, and hide it behind a macro when that wasn't possible.

Some limitations on the current design:
- The current integration with web clients requires the State Provider definition to be `UserKeyDefinition<Record<string, T>>` to match the key-value pattern in `Repository`. This usually matches with what is being used for vault (encrypted ciphers/folders, etc), but it might fall short on other domains, like profile data, or user keys.
- There's no great way of ensuring that all the client-managed Repositories have been registered, other than keeping a list. I've tried to use the `inventory` crate to keep a global list of all the existing implementations and then validating that they've all been registered, but that doesn't work for WASM. We might be able to use the `inventory` crate and just run it in tests though.

For some documentation on how to use it, you can check the README: https://github.com/bitwarden/sdk-internal/blob/ps/state-traits/crates/bitwarden-state/README.md

The continuation of this is in https://github.com/bitwarden/sdk-internal/pull/301, which contains SDK managed repository support. The client implementation of both these PRs is in https://github.com/bitwarden/clients/pull/14839

## ⏰ Reminders before review

- CI builds passed
  team

## 🦮 Reviewer guidelines


nick-livefront (approved):
Vault ✅

crates/bitwarden-state/src/repository.rs
```diff
@@ -0,0 +1,59 @@
...
+
+/// This code is not meant to be used directly, users of this crate should use the
+/// [crate::register_repository_item] macro to register their types.
+#[doc(hidden)]
+pub mod ___internal {
+
+    // This trait is just to try to discourage users from implementing `RepositoryItem` directly.
+    pub trait Internal {}
+}
+pub(crate) use ___internal::Internal;
```
justindbaur, line 59: Nifty!

justindbaur (approved):
Looks amazing!

PR #214, opened 2025-04-15
Change LocalData timestamp fields from u32 to DateTime<Utc> to support JavaScript timestamps

gbubemismith:
## 📔 Objective
This PR updates the LocalData and LocalDataView structs to use DateTime<Utc> instead of u32 for timestamp fields (lastUsedDate and lastLaunched). The timestamp sent from the TS client can exceed the u32 max value, which throws an error.

## ⏰ Reminders before review

- CI builds passed
  team

## 🦮 Reviewer guidelines

### DIFF
--- a/crates/bitwarden-vault/src/cipher/local_data.rs
+++ b/crates/bitwarden-vault/src/cipher/local_data.rs
@@ -1,5 +1,6 @@
 use bitwarden_core::key_management::{KeyIds, SymmetricKeyId};
 use bitwarden_crypto::{CryptoError, Decryptable, Encryptable, KeyStoreContext};
+use chrono::{DateTime, Utc};
 use schemars::JsonSchema;
 use serde::{Deserialize, Serialize};
 #[cfg(feature = "wasm")]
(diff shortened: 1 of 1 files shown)


crates/bitwarden-vault/src/cipher/local_data.rs
```diff
@@ -10,17 +10,17 @@ use tsify_next::Tsify;
 #[cfg_attr(feature = "uniffi", derive(uniffi::Record))]
 #[cfg_attr(feature = "wasm", derive(Tsify), tsify(into_wasm_abi, from_wasm_abi))]
 pub struct LocalData {
-    last_used_date: Option<u32>,
-    last_launched: Option<u32>,
+    last_used_date: Option<u64>,
+    last_launched: Option<u64>,
```
Hinton: Hmm, u64 is BigInt in js world according to https://rustwasm.github.io/wasm-bindgen/reference/types/numbers.html. I wonder if we should just use Chrono date instead. @fedemkr do we use last_used_date in mobile?

fedemkr: The one inside "Cipher -> Local data" is not being used as far as I see. Only in test mocking to comply with the initializer.
I've checked in Android as well but @SaintPatrck could you confirm just in case I've missed some spot?

Hinton: @gbubemismith I think we should switch to using `DateTime<UTC>` from Chrono in that case. It's consistent with other usages.

gbubemismith: yeah that makes sense

Hinton (approved):

dani-garcia (approved):

PR #221, opened 2025-04-21
[PM-20493] Add key wrapping / wrapped key facade for encstring & expose keywrap to purecrypto

quexten:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-20493

## 📔 Objective

Refactors how wrapped keys are represented in the SDK. We now add a facade around them, that only exposes the unwrap method. Additionally, this exposes key-wrapping to wasm via PureCrypto, in order to act as a backwards compatibility layer for all clients code that has not been migrated to use the SDK entirely yet.

Note: This adds a new module - "safe" to the bitwarden-crypto crate. Any of the methods or structs exposed here should be used first and foremost before anything else within the crypto crate.

## 🦖 System evolution - "Safe module"

The safe module is meant to serve as a repository for developers to draw from first and foremost. We may want to re-consider moving all other unsafer methods (RSA encrypt, direct encrypt of data, pin-key, masterkey) into a module - hazmat - later on and moving safe to the top level, once the safe crate contains sufficient good abstractions.

The abstractions contained here should be simple to understand, easy to reason about, easy to use in combination and very hard to mis-use.

For instance, master-password unlock, and pin unlock could be represented as:

```
// envelope contains the KDF settings in the encstring serialization
let envelope: PasswordProtectedKeyEnvelope = PasswordProtectedKeyEnvelope::seal(&password, &user_key, &kdf_settings);
let user_key = envelope.unseal(&password);
```

which completely eliminates the need for developers to care about synchronizing kdf settings and salts, or even caring about things like a masterkey existing.

## ⏰ Reminders before review

- CI builds passed
  team

## 🦮 Reviewer guidelines


crates/bitwarden-vault/src/cipher/cipher.rs
```diff
@@ -456,7 +460,8 @@ impl CipherView {
         ctx: &mut KeyStoreContext<KeyIds>,
         key: SymmetricKeyId,
     ) -> Result<(), CryptoError> {
-        let old_ciphers_key = Cipher::decrypt_cipher_key(ctx, key, &self.key)?;
+        let old_ciphers_key =
+            Cipher::decrypt_cipher_key(ctx, key, &self.key.clone().map(|k| k.into()))?;
```
quexten: Is there a representation that allows us to do this better, especially without the clone? (Could use some advice from someone with more Rust experience here)

dani-garcia: In this specific case, because we're setting the key immediately after and we take the cipher view by mutable reference, you can change `clone()` by `take()`, this will take the option by value leaving a `None` behind, which avoids the clone:
```suggestion
            Cipher::decrypt_cipher_key(ctx, key, &self.key.take().map(Into::into))?;
```
This won't work for the general case where we don't want to mutate the Option though.

If this is the pattern we want to go with, I feel like what we want to do is change the `key` field in Cipher from an `EncString` to a `WrappedSymmetricKey`, which would save us from doing the conversion in the first place.

crates/bitwarden-crypto/src/safe/key_wrap.rs
```diff
@@ -0,0 +1,71 @@
...
+    fn as_ref(&self) -> &EncString {
+        &self.0
+    }
+}
+
+impl From<WrappedSymmetricKey> for EncString {
+    fn from(wrapped: WrappedSymmetricKey) -> Self {
+        wrapped.0
+    }
+}
```
dani-garcia: I feel like having these two conversions kind of defeats the point of the facade, as they allow you to extract the inner EncString accidentally quite easily.

It might be better to have some explicit functions for access to the inner value (maybe even crate-internal if possible):
```rust
    pub fn into_inner(self) -> EncString {
        self.0
    }

    pub fn as_inner(&self) -> &EncString {
        &self.0
    }
```

crates/bitwarden-crypto/src/store/context.rs
```diff
@@ -167,9 +164,10 @@ impl<Ids: KeyIds> KeyStoreContext<'_, Ids> {
         &self,
```
dani-garcia, line 164: We should probably consider renaming the function `wrap_symmetric_key` or something like that as well

crates/bitwarden-crypto/src/safe/key_wrap.rs
quexten: Agreed. I mostly want to make it so that the user has to explicitly think about and choose to convert them, I don't think banning conversion outside of the crate is possible (yet?).

crates/bitwarden-vault/src/cipher/cipher.rs
quexten: > If this is the pattern we want to go with, I feel like what we want to do is change the key field in Cipher from an EncString to a WrappedSymmetricKey, which would save us from doing the conversion in the first place.

Do we need to consider the uniffi bindings / wasm bindings here? I see they are exposed, so would this be a breaking change?

dani-garcia: I think for Uniffi and WASM `EncString` is basically a type alias to string, so assuming we do the same thing with `WrappedSymmetricKey` (implement Serialize/Deserialize and do the uniffi type conversion), it should basically work.

quexten: Should be converted for uniffi and wasm now, and has been replaced in e.g. cipher and attachment.

crates/bitwarden-crypto/src/uniffi_support.rs
```diff
@@ -43,3 +59,11 @@ impl UniffiCustomTypeConverter for UnsignedSharedKey {
         obj.to_string()
     }
 }
+
+// Uniffi doesn't emit unused types, this is a dummy record to ensure that the custom type
+// converters are emitted
```
quexten, line 64: Can be removed when https://github.com/bitwarden/sdk-internal/pull/221 is merged.

crates/bitwarden-crypto/src/safe/key_wrap.rs
```diff
@@ -0,0 +1,143 @@
...
+#[wasm_bindgen::prelude::wasm_bindgen(typescript_custom_section)]
+const TS_CUSTOM_TYPES: &'static str = r#"
+export type WrappedSymmetricKey = string;
+"#;
+
+/// A wrapped symmetric key is an an [EncString], created where a wrapping key is used to encrypt a
+/// key_to_wrap.
+///
+/// Wrapped keys such as cipher keys, or attachment keys, are used to create a layer of indirection,
+/// so that keys can be shared mor granularly, and so that data can be rotated more easily.
```
dani-garcia, line 20:
```suggestion
/// so that keys can be shared more granularly, and so that data can be rotated more easily.
```

crates/bitwarden-crypto/src/safe/key_wrap.rs
```diff
@@ -0,0 +1,143 @@
...
+    }
+}
+
+impl std::fmt::Debug for WrappedSymmetricKey {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("WrappedSymmetricKey")
+            .field("inner_enc_string", &self.as_inner())
+            .finish()
+    }
+}
```
dani-garcia, line 54: In this case we can probably just tag `WrappedSymmetricKey` with `#[derive(Debug)]` directly instead of implementing it by hand.

crates/bitwarden-crypto/src/store/context.rs
```diff
@@ -127,29 +127,26 @@ impl<Ids: KeyIds> KeyStoreContext<'_, Ids> {
...
     /// # Arguments
     ///
-    /// * `encryption_key` - The key id used to decrypt the `encrypted_key`. It must already exist
-    ///   in the context
+    /// * `wrapping_key` - The key id used to unwrap the `key_to_unwrap`. It must already exist in
+    ///   the context
     /// * `new_key_id` - The key id where the decrypted key will be stored. If it already exists, it
     ///   will be overwritten
-    /// * `encrypted_key` - The key to decrypt
+    /// * `key_to_unwrap` - The key to decrypt
```
dani-garcia, line 138: There's still a reference to decryption instead of unwrapping (`The key to decrypt` -> `The key to unwrap`). To avoid the repetition with the comment, we could also rename the parameter to more closely resemble it's type too:
```suggestion
    /// * `wrapped_key` - The key to unwrap
```

crates/bitwarden-crypto/src/keys/master_key.rs
```diff
@@ -66,17 +66,17 @@ impl MasterKey {
...
 
     /// Generate a new random user key and encrypt it with the master key.
-    pub fn make_user_key(&self) -> Result<(UserKey, EncString)> {
+    pub fn make_user_key(&self) -> Result<(UserKey, WrappedSymmetricKey)> {
         make_user_key(rand::thread_rng(), self)
     }
 
     /// Encrypt the users user key
-    pub fn encrypt_user_key(&self, user_key: &SymmetricCryptoKey) -> Result<EncString> {
+    pub fn encrypt_user_key(&self, user_key: &SymmetricCryptoKey) -> Result<WrappedSymmetricKey> {
```
dani-garcia, line 74: Question: Will we want to rename these to also use the wrap/unwrap naming?

crates/bitwarden-core/src/auth/pin.rs
```diff
@@ -70,12 +70,12 @@ mod tests {
 
         let master_key = MasterKey::derive(password, email, &kdf).unwrap();
 
-        let user_key = "2.Q/2PhzcC7GdeiMHhWguYAQ==|GpqzVdr0go0ug5cZh1n+uixeBC3oC90CIe0hd/HWA/pTRDZ8ane4fmsEIcuc8eMKUt55Y2q/fbNzsYu41YTZzzsJUSeqVjT8/iTQtgnNdpo=|dwI+uyvZ1h/iZ03VQ+/wrGEFYVewBUUl/syYgjsNMbE=";
+        let user_key: EncString = "2.Q/2PhzcC7GdeiMHhWguYAQ==|GpqzVdr0go0ug5cZh1n+uixeBC3oC90CIe0hd/HWA/pTRDZ8ane4fmsEIcuc8eMKUt55Y2q/fbNzsYu41YTZzzsJUSeqVjT8/iTQtgnNdpo=|dwI+uyvZ1h/iZ03VQ+/wrGEFYVewBUUl/syYgjsNMbE=".parse().unwrap();
```
dani-garcia, line 73: We implement `FromStr` for `WrappedSymmetricKey` as well, so if you remove the `.into()` below you can skip the type here and rust will infer it correctly as a `WrappedSymmetricKey`:
```suggestion
        let user_key = "2.Q/2PhzcC7GdeiMHhWguYAQ==|GpqzVdr0go0ug5cZh1n+uixeBC3oC90CIe0hd/HWA/pTRDZ8ane4fmsEIcuc8eMKUt55Y2q/fbNzsYu41YTZzzsJUSeqVjT8/iTQtgnNdpo=|dwI+uyvZ1h/iZ03VQ+/wrGEFYVewBUUl/syYgjsNMbE=".parse().unwrap();
```

This also probably applies to the other test cases that were updated in this PR

crates/bitwarden-uniffi/src/auth/mod.rs
```diff
@@ -110,10 +110,13 @@ impl AuthClient {
         password: String,
         encrypted_user_key: String,
     ) -> Result<String> {
+        let encrypted_user_key = encrypted_user_key
+            .parse::<EncString>()
```
dani-garcia, line 114: We can also parse into `WrappedSymmetricKey` here and remove the `.into()`
```suggestion
            .parse::<WrappedSymmetricKey>()
```

crates/bitwarden-wasm-internal/src/pure_crypto.rs
```diff
@@ -107,6 +108,27 @@ impl PureCrypto {
...
+        let wrapping_key: SymmetricCryptoKey = wrapping_key.try_into()?;
+        let wrapped_key: EncString = key_to_be_wrapped.wrap_with(&wrapping_key)?.into_inner();
+        Ok(wrapped_key.to_string())
+    }
+
+    pub fn unwrap_symmetric_key(
+        wrapped_key: String,
+        wrapping_key: Vec<u8>,
+    ) -> Result<Vec<u8>, CryptoError> {
+        let wrapped_key: WrappedSymmetricKey = EncString::from_str(&wrapped_key)?.into();
```
dani-garcia, line 127:
```suggestion
        let wrapped_key = WrappedSymmetricKey::from_str(&wrapped_key)?;
```

crates/bitwarden-vault/src/cipher/cipher.rs
```diff
@@ -453,7 +454,7 @@ impl CipherView {
         ctx: &mut KeyStoreContext<KeyIds>,
         key: SymmetricKeyId,
     ) -> Result<(), CryptoError> {
-        let old_ciphers_key = Cipher::decrypt_cipher_key(ctx, key, &self.key)?;
+        let old_ciphers_key = Cipher::decrypt_cipher_key(ctx, key, &self.key.take())?;
```
dani-garcia, line 457: Now that the keys are wrapped, the take here can be removed.

crates/bitwarden-vault/src/cipher/cipher.rs
```diff
@@ -489,8 +490,9 @@ impl CipherView {
         if let Some(attachments) = &mut self.attachments {
             for attachment in attachments {
                 if let Some(attachment_key) = &mut attachment.key {
-                    let dec_attachment_key: Vec<u8> = attachment_key.decrypt(ctx, old_key)?;
-                    *attachment_key = dec_attachment_key.encrypt(ctx, new_key)?;
+                    let tmp_attachment_key_id = SymmetricKeyId::Local("attachment_key");
+                    ctx.unwrap_symmetric_key(old_key, tmp_attachment_key_id, attachment_key)?;
+                    *attachment_key = ctx.wrap_symmetric_key(new_key, tmp_attachment_key_id)?;
```
dani-garcia, line 495: I'm wondering if we should offer a function in the context to do the unwrap/wrap for you, as we're already using it a couple of times here.

crates/bitwarden-vault/src/cipher/cipher.rs
```diff
@@ -489,8 +490,9 @@ impl CipherView {
         if let Some(attachments) = &mut self.attachments {
             for attachment in attachments {
                 if let Some(attachment_key) = &mut attachment.key {
-                    let dec_attachment_key: Vec<u8> = attachment_key.decrypt(ctx, old_key)?;
-                    *attachment_key = dec_attachment_key.encrypt(ctx, new_key)?;
+                    let tmp_attachment_key_id = SymmetricKeyId::Local("attachment_key");
```
dani-garcia, line 493: Usually we declare these ids as `const`, we should probably do the same for consistency.

dani-garcia (commented):
Looks pretty good to me, just some minor details

PR #225, opened 2025-04-25
Remove duplicated Cipher and Folder client

Hinton:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-13374

## 📔 Objective

Changes `uuid` to be represented as unknown which causes most `as` casts to fail since TS safeguards against casting "string" to unknown. Adds a new type called `OrganizationId` which is manually set to be equal to `Tagged<Uuid, "OrganizationId">` on the TS side. This prevents mixing up the type with other Uuid's.

In order to create an OrganizationId in the application you now need to pass through unknown otherwise you get the following error message: `Conversion of type 'string' to type 'Tag<"OrganizationId", never>' may be a mistake because neither type sufficiently overlaps with the other. If this was intentional, convert the expression to 'unknown' first.`

To provide some ergonomics we should add the following function in typescript:

```
export function uuid<T extends Uuid>(uuid: string): T {
  if (Utils.isGuid(uuid)) {
    return uuid as T;
  }

  throw new Error(`Invalid UUID: ${uuid}`);
}
```

## ⏰ Reminders before review

- CI builds passed
  team

## 🦮 Reviewer guidelines


crates/bitwarden-vault/src/cipher/cipher_client.rs
```diff
@@ -1,15 +1,27 @@
...
 };
 
+/// NewType representing an Organization ID. This ensures type safety.
+#[cfg_attr(
+    feature = "wasm",
+    derive(serde::Serialize, serde::Deserialize, tsify_next::Tsify),
+    tsify(into_wasm_abi, from_wasm_abi)
+)]
+#[repr(transparent)]
+pub struct OrganizationId(#[tsify(type = r#"Tagged<Uuid, "OrganizationId">"#)] pub uuid::Uuid);
```
coroiu: question: would it make sense to put org id somewhere else? I'm just thinking about what the most logical place to put it would be. If we put it in `bitwarden-vault` then it's gonna get a lot of crates depending on it, possibly hindering them from depending on others. Seems like id's are more foundational?

coroiu: suggestion: instead of making the internal field public, how about implementing `TryFrom` to convert between types?

Hinton: It probably would, I put it here since this is currently the only place consuming it right now, but we might have to resort to putting it in bitwarden_core?

Hinton: I think so, I wanted to hold off until I had implemented a macro to avoid writing throw away code but that's done now.

coroiu: That would probably be better yeah

crates/bitwarden-uuid-macro/src/lib.rs
```diff
@@ -0,0 +1,74 @@
...
+            #[tsify(type = #tsify_type)]
+            uuid::Uuid
+        );
+
+        #[cfg(not(feature = "wasm"))]
+        #[derive(serde::Serialize, serde::Deserialize)]
+        #[repr(transparent)]
+        #vis struct #ident(
+            uuid::Uuid
+        );
```
coroiu: suggestion: Why not merge these together with `cfg_attr` ?

```ts
#[cfg_attr(feature = "wasm", derive(::tsify_next::Tsify))]
```

crates/bitwarden-uuid-macro/src/lib.rs
```diff
@@ -0,0 +1,74 @@
...
+    let input = parse_macro_input!(input as IdTypeInput);
+    let ident = input.ident;
+    let vis = input.vis;
+    let name_str = ident.to_string();
+
+    let tsify_type = format!("Tagged<Uuid, \"{}\">", name_str);
+
+    let expanded = quote! {
+        #[cfg(feature = "wasm")]
+        #[derive(serde::Serialize, serde::Deserialize, tsify_next::Tsify)]
```
coroiu: suggestion: should we use `::serde::Serialize` ?
```suggestion
        #[derive(::serde::Serialize, ::serde::Deserialize, ::tsify_next::Tsify)]
```

crates/bitwarden-uuid-macro/src/lib.rs
```diff
@@ -0,0 +1,74 @@
...
+        #[cfg(feature = "uniffi")]
+        uniffi::custom_newtype!(#ident, uuid::Uuid);
+
+        impl #ident {
+            pub fn new(value: uuid::Uuid) -> Self {
+                Self(value)
+            }
+        }
+
+        impl std::str::FromStr for #ident {
```
coroiu: suggestion: not sure if someone would rename `std` but 🤷  
```suggestion
        impl ::std::str::FromStr for #ident {
```

crates/bitwarden-uuid/Cargo.toml
```diff
@@ -0,0 +1,31 @@
...
+repository.workspace = true
+license-file.workspace = true
+keywords.workspace = true
+
+[features]
+wasm = [
+    "bitwarden-uuid-macro/wasm",
+    #"dep:js-sys",
+    #"dep:tsify-next",
+    #"dep:wasm-bindgen",
```
coroiu: suggestion: clean up

crates/bitwarden-vault/Cargo.toml
```diff
@@ -18,7 +18,7 @@ keywords.workspace = true
 uniffi = [
     "bitwarden-core/uniffi",
     "bitwarden-crypto/uniffi",
-    "dep:uniffi"
+    "dep:uniffi",
```
coroiu: ah, the classical VS Code Lint-breaker

crates/bitwarden-uuid-macro/src/lib.rs
```diff
@@ -0,0 +1,74 @@
+use proc_macro::TokenStream;
+use quote::quote;
+use syn::{
+    parse::{Parse, ParseStream},
+    parse_macro_input, Ident, Visibility,
+};
+
+#[proc_macro]
+pub fn uuid(input: TokenStream) -> TokenStream {
```
coroiu, line 11: question: do you think it would make sense to add tests?

crates/bitwarden-uuid-macro/src/lib.rs
Hinton: Because of the inner `#[tsify(type = #tsify_type)]`

coroiu: Oh is it not possible to use `cfg_attr` with that?

crates/bitwarden-uuid-macro/src/lib.rs
Hinton: Probably but I'd have to look into how macros are supposed to be tested 😓

coroiu: Just check out `bitwarden-error/tests` 👍

crates/bitwarden-uuid-macro/src/lib.rs
dani-garcia: Those inner field attributes are parsed by the proc macros themselves, so unless tsify has support for cfg_attr itself, it's usually not allowed.

Hinton: Huh, seems to work in this scenario.

crates/bitwarden-uuid-macro/src/lib.rs
```diff
@@ -0,0 +1,69 @@
...
+    let vis = input.vis;
+    let name_str = ident.to_string();
+
+    let tsify_type = format!("Tagged<Uuid, \"{}\">", name_str);
+    let doc_string = format!(" NewType wrapper for `{}`", name_str);
+
+    let expanded = quote! {
+        #[doc = #doc_string]
+        #[cfg_attr(feature = "wasm", derive(::tsify_next::Tsify), tsify(into_wasm_abi, from_wasm_abi))]
+        #[derive(::serde::Serialize, ::serde::Deserialize)]
```
dani-garcia: Do we know if just having `#[derive(::serde::Serialize, ::serde::Deserialize)]` and `#[repr(transparent)]` makes serde serialize it as a string, instead of a struct with one string field? I know serde has it's own `#[serde(transparent)]`, but not sure if it's needed here. May be a good idea to add a serialization/deserialization test just in case.

It's also a good idea to derive Clone and Copy for this, potentially also Eq,Hash if we ever want to use them as keys in a hashmap

crates/bitwarden-uuid-macro/src/lib.rs
coroiu: Lol well, sometimes I get lucky xD

crates/bitwarden-uuid-macro/src/lib.rs
Hinton: Seems towork.

dani-garcia (approved):

PR #226, opened 2025-04-25
Unify WASM crypto client with mobile

dani-garcia:
## 📔 Objective

With the removal of the lifetimes of the clients, we can remove the wasm-specific crypto client and instead use the one in bitwarden-core.

## ⏰ Reminders before review

- CI builds passed
  team

## 🦮 Reviewer guidelines


crates/bitwarden-core/src/mobile/crypto_client.rs
```diff
@@ -39,6 +43,24 @@ impl CryptoClient {
...
+    /// Crypto initialization not required.
+    pub fn verify_asymmetric_keys(
+        &self,
+        request: VerifyAsymmetricKeysRequest,
+    ) -> Result<VerifyAsymmetricKeysResponse, CryptoError> {
+        verify_asymmetric_keys(request)
+    }
+}
+
+impl CryptoClient {
```
dani-garcia, line 63: I've split the `impl CryptoClient` into two blocks, one with wasm_bindgen for the functions we're currently exporting to wasm, and one without it for the ones that we're not using.

We could also export all of them if we wanted, but to keep the PR simple I only include the ones actually in use.

crates/bitwarden-core/Cargo.toml
```diff
@@ -49,6 +50,7 @@ tsify-next = { workspace = true, optional = true }
 uniffi = { workspace = true, optional = true, features = ["tokio"] }
 uuid = { workspace = true }
 wasm-bindgen = { workspace = true, optional = true }
+wasm-bindgen-futures = { workspace = true, optional = true }
```
dani-garcia, line 54: This dep is used internally for the wasm-bindgen macro as cryptoclient has some async functions.

crates/bitwarden-core/src/mobile/crypto.rs
Hinton, line 1: **suggestion:** Could we lift this out of mobile?

crates/bitwarden-core/src/mobile/crypto.rs
```diff
@@ -23,6 +25,7 @@ use crate::{
 
 /// Catch all error for mobile crypto operations.
 #[allow(missing_docs)]
+#[bitwarden_error(flat)]
```
Hinton, line 28: The `MobileCryptoError` doesn't quite make sense if this is used from wasm.

Hinton (dismissed):

crates/bitwarden-core/src/mobile/crypto.rs
dani-garcia: Sounds reasonable yeah, do you think we should just lift it into the root? Or maybe into the key-management folder?

Hinton: I think we can lift it into KM, @quexten would you be alright with that, I think everything currently falls under KM domain.

quexten: Seems reasonable!

My question here (non-blocking, but needs clarification): Does key-management (the folder) really refer to the KM team, or the key-management capabilities (key store, key context) of the sdk?

The KM domain is not directly the same (Biometrics, Crypto Primitives).

dani-garcia: So the way I'm thinking about this is like this:

The crypto client that I'm moving to the `key_management` folder is mostly a thin compatibility layer on top of the `bitwarden-crypto` types to make the functionality more usable for FFI (By exposing a client, instead of relying on types like `MasterKey`, etc, which aren't usable in FFI). In that sense, once KM officially owns `bitwarden-crypto`, this folder should come with, as it's fairly tied to the API exposed by `bitwarden-crypto`. In the future, if we find a way to expose the normal `bitwarden-crypto` API through FFI this module could be removed.

The other thing in the `key_management` folder is the definition of `KeyIds` used by the Key Store from `bitwarden-crypto`. I think of this and the Key Store as part of KM's field to expose safe to use crypto, so in my mind it would be part of your domain. The reason I defined these outside the crypto crate was to ensure the system was generic and easily expandable, but I'm not sure if that's worth it. If not, we can always move it into `bitwarden-crypto` directly and simplify ownership.

crates/bitwarden-core/src/mobile/crypto.rs
dani-garcia: I changed it to CryptoClientError now

Hinton (dismissed):

addisonbeck (approved):

PR #256, opened 2025-04-29
[PM-20957] Update UniFFI to 0.29.1

dani-garcia:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-20957

## 📔 Objective

This replaces the renovate PR as the changes are fairly big.

Update UniFFI to version 0.29.1. This requires bumping MSRV to 1.82, as that's the requirement from UniFFI.

The migration consists of a few parts:
- Replaced all the uses of `UniffiCustomTypeConverter` by the simpler `custom_type!` macro. Note that for types defined in other crates, we still use the same macro, but with the `remote,` value added.
- Reworked all the uses of `ffi_converter_forward!`:
    - If they were forwarding bitwarden defined types, they aren't needed anymore.
    - If they were forwarding types from third party crates, they were replaced by `use_remote_type!`
- Replaced a usage of `uniffi::deps::log` by the `log` crate directly.
- While replacing the implementation of `LinkedId`, I've noticed that we had two implementations of `TryFrom<U32>`, the one used by uniffi based on deserialize, and a hardcoded map. I've combined them in one using the deserialize strategy.

I've tested this on the Android and iOS native apps, and they both compile and decrypt the vault without issues.

## ⏰ Reminders before review

- CI builds passed
  team

## 🦮 Reviewer guidelines


crates/bitwarden-vault/src/cipher/linked_id.rs
```diff
@@ -91,37 +81,9 @@ pub enum IdentityLinkedIdType {
...
-            412 => Ok(LinkedIdType::Identity(IdentityLinkedIdType::Ssn)),
-            413 => Ok(LinkedIdType::Identity(IdentityLinkedIdType::Username)),
-            414 => Ok(LinkedIdType::Identity(IdentityLinkedIdType::PassportNumber)),
-            415 => Ok(LinkedIdType::Identity(IdentityLinkedIdType::LicenseNumber)),
-            416 => Ok(LinkedIdType::Identity(IdentityLinkedIdType::FirstName)),
-            417 => Ok(LinkedIdType::Identity(IdentityLinkedIdType::LastName)),
-            418 => Ok(LinkedIdType::Identity(IdentityLinkedIdType::FullName)),
-            _ => Err(MissingFieldError("LinkedIdType")),
-        }
+    fn try_from(val: u32) -> Result<Self, Self::Error> {
```
dani-garcia, line 84: The uniffi conversion was done through the deserialize impl, while this tryfrom was implemented using a hardcoded map, while I was at it I unified both of them to use the deserialize impl.

Hinton (dismissed):

crates/bitwarden-core/src/uniffi_support.rs
```diff
@@ -2,47 +2,22 @@
...
-
-    fn from_custom(obj: Self) -> Self::Builtin {
-        obj.to_string()
-    }
-}
+uniffi::custom_type!(DateTime, std::time::SystemTime, {
+    remote,
+    try_lift: |val| Ok(Self::from(val)),
+    lower: |obj| obj.into(),
+});
```
coroiu: suggestion: trivial conversion using `From`/`TryFrom` traits is built-in into the macro from my understanding. I was able to simplify it like this:

```suggestion
uniffi::custom_type!(DateTime, std::time::SystemTime, { remote });
```

However I got blocked on the `use_remote_type` stuff so I'm not 100% sure if it actually worked

crates/bitwarden-vault/src/cipher/linked_id.rs
```diff
@@ -16,22 +16,12 @@ pub enum LinkedIdType {
...
-
-    fn from_custom(obj: Self) -> Self::Builtin {
-        obj.into()
-    }
-}
+uniffi::custom_type!(LinkedIdType, u32, {
+    try_lift: |val| {
+        LinkedIdType::try_from(val).map_err(|e| e.into())
+    },
+    lower: |obj| obj.into(),
```
coroiu: suggestion: here too, I think we could skip `lower`

```suggestion
```

It doesn't cause any compilation errors and from what I can see in the source this should be supported

https://github.com/mozilla/uniffi-rs/blob/4e0a9e96a546014af4cecba1ca382c4074cb7fb2/uniffi_macros/src/custom.rs#L184-L197

coroiu (commented):
Looks good, just some small simplification comments

crates/bitwarden-core/src/uniffi_support.rs
dani-garcia: Nice, checked the generated code and tested the ios and android apps, seems to be working the same

crates/bitwarden-vault/src/cipher/linked_id.rs
dani-garcia: Done, also skipped the try_lift, as it's doing `Ok(try_into()?)`, which is basically the same thing I was doing mapping the error.

coroiu: Oh yeah true, nice!

coroiu (commented):
Let's goooooo

coroiu (approved):
Second time's the charm

PR #258, opened 2025-05-01
[BEEEP] Add ThreadBoundRunner

coroiu:
## 📔 Objective

Introduce ThreadBoundRunner, a utility that owns a non-Send state pinned to a single thread. It exposes a Send API (`run_in_thread`) for submitting async tasks that operate on this state. Tasks are dispatched in order and run on the owning thread, enabling safe cross-thread interaction with thread-bound resources.

Or in other words, it's a utility that allows us to easily depend on JavaScript objects across the WASM divide in the rest of our core infrastructure without having to worry about the single-threaded nature of WASM/JavaScript. 

## ⏰ Reminders before review

- CI builds passed
  team

## 🦮 Reviewer guidelines


crates/bitwarden-threading/src/thread_bound_dispatcher.rs
```diff
@@ -0,0 +1,182 @@
...
+                function: function_wrapper,
+            })
+            .await
+            .map_err(|e| CallError::ChannelSend(e.to_string()))?;
+
+        return_channel_rx
+            .await
+            .map_err(|e| CallError::ChannelReceive(e.to_string()))
+    }
+}
```
dani-garcia: By changing this function definition slightly (moving the Future to be a generic type and removing the `Pin<Box>>`, I can get the calls to this function to be a bit simpler:
```suggestion
    pub async fn call<F, Fut, Output>(&self, function: F) -> Result<Output, CallError>
    where
        F: FnOnce(Arc<ThreadState>) -> Fut + Send + 'static,
        Fut: Future<Output = Output>,
        Output: Send + Sync + 'static,
    {
        let (return_channel_tx, return_channel_rx) = tokio::sync::oneshot::channel();

        let request = CallRequest {
            function: Box::new(|state| {
                Box::pin(async move {
                    let result = function(state);
                    return_channel_tx.send(result.await).unwrap_or_else(|_| {
                        log::warn!(
                            "ThreadBoundDispatcher failed to send result back to the caller"
                        );
                    });
                })
            }),
        };

        self.call_channel_tx
            .send(request)
            .await
            .map_err(|e| CallError::ChannelSend(e.to_string()))?;

        return_channel_rx
            .await
            .map_err(|e| CallError::ChannelReceive(e.to_string()))
    }
```

```rust
        async fn get(&self, id: String) -> Cipher {
            self.0
                 // Easier than `move |state| { Box::pin(async move {`
                .call(|state| async move {
                    let js_value_cipher = state.get(id).await;
                    let cipher: Cipher = serde_wasm_bindgen::from_value(js_value_cipher)
                        .expect("Failed to convert JsValue to Cipher");
                    cipher
                })
                .await
                .unwrap()
        }
```

crates/bitwarden-threading/src/thread_bound_dispatcher.rs
```diff
@@ -0,0 +1,182 @@
...
+impl<ThreadState> ThreadBoundDispatcher<ThreadState>
+where
+    ThreadState: 'static,
+{
+    pub fn new(state: ThreadState) -> Self {
+        let (call_channel_tx, mut call_channel_rx) =
+            tokio::sync::mpsc::channel::<CallRequest<ThreadState>>(1);
+
+        spawn_local(async move {
+            let state = Arc::new(state);
```
dani-garcia: The `Arc` never leaves this locally spawned task so we don't need the thread-safety and can safely use an `Rc` and avoid the overhead of the atomic operations.

I feel like there might even be a way to just pass `&state` but probably not worth it adding all the lifetime bounds.

crates/bitwarden-threading/src/thread_bound_dispatcher.rs
```diff
@@ -0,0 +1,182 @@
...
+                    log::warn!("ThreadBoundDispatcher failed to send result back to the caller");
+                });
+            })
+        });
+        self.call_channel_tx
+            .send(CallRequest {
+                function: function_wrapper,
+            })
+            .await
+            .map_err(|e| CallError::ChannelSend(e.to_string()))?;
```
dani-garcia: I wonder if handling these errors is necessary, the only way this send can fail is if the receiver thread is dropped, but the only way for that to happen is if `Self` is dropped, which obviously can't happen.

Same thing for `return_channel_rx`, that can only fail if `return_channel_tx` gets dropped before calling `return_channel_tx.send`, which the only way I can imagine happening is if the provided function panics maybe?.

If we can show that these errors can't happen, we can change the function to return `-> Output` and the users have one less error type to worry about.

crates/bitwarden-threading/src/thread_bound_dispatcher.rs
coroiu: Ou, that's as close to an `AsyncFn` closure we're gonna get without actually using one, nice 👍

crates/bitwarden-threading/src/thread_bound_dispatcher.rs
coroiu: > I feel like there might even be a way to just pass &state

Haha oh yeah... that's what I started out with but the issue is (was?) that the reference crosses over multiple closure bounds which causes it to escape and make `rustc` sad. Your suggestion above might make it easier though. I'll try to give it a shot but it's probably not worth a massive increase in complexity

crates/bitwarden-threading/src/thread_bound_dispatcher.rs
coroiu: Yeah the runner thread might panic causing the receiver to drop. I've been thinking about spawning secondary threads to actually run the command, mostly as away to allow for multiple commands to run at the same time, but that should also protect the main thread from any panic. That might require `Sync` though

crates/bitwarden-threading/src/thread_bound_dispatcher.rs
coroiu: Oh yeah, confirmed, still works great!

crates/bitwarden-threading/src/thread_bound_dispatcher.rs
coroiu: I tried borrow again but I don't think it's possible, at least, it's beyond me to get Rust to understand all the lifetime bounds

dani-garcia: Yeah figured that was the case for the references, we should still be able to swap `Arc` for `Rc` though.

crates/bitwarden-threading/src/thread_bound_dispatcher.rs
dani-garcia: We should also be able to use catch_unwind to catch the panic in the thread, and then send it to the calling thread:

```rust
    pub async fn run_in_thread<F, Fut, Output>(&self, function: F) -> Output
    where
        F: FnOnce(Arc<ThreadState>) -> Fut + Send + 'static,
        Fut: Future<Output = Output> + UnwindSafe,
        Output: Send + Sync + 'static,
    {
        let (return_channel_tx, return_channel_rx) = tokio::sync::oneshot::channel();
        let request = CallRequest {
            function: Box::new(|state| {
                Box::pin(async move {
                    use futures::future::FutureExt;
                    let result = function(state).catch_unwind().await;
                    return_channel_tx.send(result).unwrap_or_else(|_| {
                        log::warn!(
                            "ThreadBoundDispatcher failed to send result back to the caller"
                        );
                    });
                })
            }),
        };

        self.call_channel_tx
            .send(request)
            .await
            .expect("channel must not be closed");
        let result = return_channel_rx
            .await
            .expect("a response must be available");

        match result {
            Ok(value) => value,
            Err(e) => std::panic::resume_unwind(e),
        }
    }
```

coroiu: I played around with this but it introduces some limitations on the `ThreadState`. E.g. I added a pure rust example in `tests/standard-tokio.rs` which uses `ciphers: std::collections::HashMap<String, Cipher>` to store ciphers, but because we don't have a mutable reference to `ThreadState` we need to use some other mechanism like `Mutex` but apparently Mutex does not implement `UnwindSafe` which causes the whole closure to not implement it which then fails compilation

dani-garcia: Welp, it was worth a try hah. We should be avoiding panics as a general rule so it's probably not a big deal if the thread doesn't handle them anyway.

coroiu: Yeah, we might be able to get something working using e.g. https://stackoverflow.com/questions/59426545/rust-paniccatch-unwind-no-use-in-webassembly but I'll let that be a later problem

crates/bitwarden-threading/src/thread_bound_dispatcher.rs
coroiu: Done! 👍

crates/bitwarden-threading/README.md
```diff
@@ -0,0 +1,11 @@
+# bitwarden-threading
+
+Utility crate for Bitwarden SDK to handle threading and async quirks in FFI contexts.
+
+## WASM Testing
+
+To run the WASM tests, you can use the following command:
+
+```bash
+cargo test --target wasm32-unknown-unknown --all-features -- --nocapture
```
dani-garcia, line 10: Should we add this to the CI test workflow as well?

crates/bitwarden-threading/src/thread_bound_runner.rs
```diff
@@ -0,0 +1,227 @@
...
+#![allow(unused_variables)]
+
+use std::{future::Future, pin::Pin, rc::Rc};
+
+use bitwarden_error::bitwarden_error;
+use thiserror::Error;
+#[cfg(not(target_arch = "wasm32"))]
+use tokio::task::spawn_local;
+#[cfg(target_arch = "wasm32")]
+use wasm_bindgen_futures::spawn_local;
```
dani-garcia, line 11: The `wasm_bindgen_futures` crate is only in the dev-dependencies, so this will fail to build.

crates/bitwarden-threading/Cargo.toml
```diff
@@ -0,0 +1,34 @@
...
+authors.workspace = true
+edition.workspace = true
+rust-version.workspace = true
+homepage.workspace = true
+repository.workspace = true
+license-file.workspace = true
+keywords.workspace = true
+
+[features]
+wasm = ["dep:tsify-next", "dep:wasm-bindgen", "dep:js-sys"] # WASM support
```
dani-garcia: We're not using the feature or these dependencies at the moment, so we can probably remove them.

crates/bitwarden-threading/README.md
coroiu: We probably should but I'm getting too many compilation errors to fix in this PR 🥲

dani-garcia: We can probably just test this crate only, we don't have wasm specific tests anywhere else atm anyway:
```
cargo test --target wasm32-unknown-unknown -p bitwarden-threading --all-features -- --nocapture
```

coroiu: Done 👍

crates/bitwarden-threading/src/thread_bound_runner.rs
coroiu: Fixed!

crates/bitwarden-threading/Cargo.toml
coroiu: Removed!

.github/workflows/rust-test.yml
```diff
@@ -38,13 +38,45 @@ jobs:
...
+          components: rust-src
+
+      - name: Cache cargo registry
+        uses: Swatinem/rust-cache@f0deed1e0edfc6a9be95417288c0e1099b1eeec3 # v2.7.7
+
+      - name: Install wasm-bindgen-cli
+        run: cargo install wasm-bindgen-cli --version 0.2.100
+
+      - name: Test WASM
+        run: cargo test --target wasm32-unknown-unknown -p bitwarden-threading -p bitwarden-error --all-features
```
dani-garcia: Can we also add `-p bitwarden-wasm-internal` here? We don't have tests on it at the moment, but it should compile for WASM and if we add tests in the future we might not notice that they are not running in CI

crates/bitwarden-threading/tests/wasm.rs
```diff
@@ -0,0 +1,93 @@
...
+
+#[derive(Clone, Debug, Tsify, Serialize, Deserialize, PartialEq, Eq)]
+#[tsify(into_wasm_abi, from_wasm_abi)]
+struct Cipher {
+    id: String,
+    name: String,
+    password: String,
+}
+
+// #[wasm_bindgen]
```
dani-garcia: Remove this comment?

crates/bitwarden-threading/Cargo.toml
```diff
@@ -0,0 +1,37 @@
...
+license-file.workspace = true
+keywords.workspace = true
+
+[dependencies]
+bitwarden-error = { workspace = true }
+log = { workspace = true }
+serde = { workspace = true }
+serde_json = { workspace = true }
+thiserror = { workspace = true }
+tokio = { features = ["sync", "time"], workspace = true }
```
dani-garcia: I'm pretty sure we need the "rt" feature here as well, as we're using `spawn_local`. It's probably working because cargo is unifying the features from the workspace.

crates/bitwarden-threading/src/thread_bound_runner.rs
```diff
@@ -0,0 +1,228 @@
...
+        let (call_channel_tx, mut call_channel_rx) =
+            tokio::sync::mpsc::channel::<CallRequest<ThreadState>>(1);
+
+        spawn_local(async move {
+            let state = Rc::new(state);
+            while let Some(request) = call_channel_rx.recv().await {
+                let state = state.clone();
+                spawn_local(async move {
+                    (request.function)(state).await;
+                });
```
dani-garcia: This can probably be simplified a little, as the function is already a future, but it's not a big deal either way:
```suggestion
                spawn_local((request.function)(state.clone()));
```

dani-garcia (commented):
Some small comments but otherwise LGTM

.github/workflows/rust-test.yml
coroiu: Yeah fair enough 🤷

crates/bitwarden-threading/Cargo.toml
coroiu: Ouf good catch, that's a real gotcha. It's interesting that the editor uses the dev-dependencies when analyzing

dani-garcia (approved):

PR #264, opened 2025-05-06
[PM-20615] Only process incoming messages once

coroiu:
## 📔 Objective

- Create a new thread/task and delegate message processing to it
- Make sure CryptoProvider only sees each incoming message once
- Make everything multi-thread safe
  - Enables a single thread to take ownership of message processing
  - but will also allow us to use the IPC framework with `BitwardenClient`

## ⏰ Reminders before review

- CI builds passed
  team

## 🦮 Reviewer guidelines


coroiu:
Sorry for assigning this to you too @dani-garcia, but I don't think anyone else has experience with Send + Sync

coroiu:
I'm debugging the tests now but apparently they only fail when you run `cargo test` and not `cargo test -p bitwarden-ipc --lib`

Edit: Fixed

crates/bitwarden-ipc/src/ipc_client.rs
```diff
@@ -1,22 +1,27 @@
...
     Com: CommunicationBackend,
-    Ses: SessionRepository<Session = Crypto::Session>,
+    Ses: SessionRepository<Crypto::Session>,
 {
     crypto: Crypto,
     communication: Com,
     sessions: Ses,
+
+    incoming: RwLock<Option<tokio::sync::broadcast::Receiver<IncomingMessage>>>,
+    cancellation_handle: RwLock<Option<tokio::sync::watch::Sender<bool>>>,
```
dani-garcia: By the way there's a `CancellationToken` in tokio-util, which more or less does something similar: https://docs.rs/tokio-util/latest/tokio_util/sync/struct.CancellationToken.html

crates/bitwarden-ipc/src/ipc_client.rs
```diff
@@ -42,170 +47,270 @@ pub struct IpcClientTypedSubscription<Crypto, Com, Ses, Payload>
...
+
+            let mut client_incoming = client.incoming.write().await;
+            *client_incoming = Some(client_rx);
+            drop(client_incoming);
+
+            await_init_tx
+                .send(())
+                .expect("Sending init signal should not fail");
+
+            loop {
```
dani-garcia: Is await_init_tx used to wait until the subscription is done? if so we could just subscribe outside the task.
```suggestion
    let com_receiver = client.communication.subscribe().await;
    let (client_tx, client_rx) = tokio::sync::broadcast::channel(CHANNEL_BUFFER_CAPACITY);

    let mut client_incoming = client.incoming.write().await;
    *client_incoming = Some(client_rx);
    drop(client_incoming);
        let future = async move {
            loop {
```

crates/bitwarden-ipc/src/ipc_client.rs
```diff
@@ -42,170 +47,270 @@ pub struct IpcClientTypedSubscription<Crypto, Com, Ses, Payload>
...
+        let incoming = self.incoming.read().await;
+        let cancellation_handle = self.cancellation_handle.read().await;
+        incoming.is_some() && cancellation_handle.is_some()
+    }
+
+    pub async fn stop(self: &Arc<Self>) {
+        let mut incoming = self.incoming.write().await;
+        if let Some(receiver) = incoming.take() {
+            drop(receiver);
+        }
```
dani-garcia: You can also take the value and assign it to _ to immediately drop it.
```suggestion
        let _ = incoming.take();
```

crates/bitwarden-ipc/src/ipc_client.rs
```diff
@@ -241,63 +346,57 @@ mod tests {
...
+                    tokio::time::sleep(Duration::from_secs(600)).await;
+                    Err("Simulated timeout".to_string())
+                }
+            }
         }
     }
 
     #[tokio::test]
     async fn returns_send_error_when_crypto_provider_returns_error() {
+        // TODO: THIS LOCKS UP!
```
dani-garcia: Is this happening still?

crates/bitwarden-ipc/src/wasm/communication_backend.rs
```diff
@@ -68,12 +66,61 @@ impl JsCommunicationBackend {
     }
 }
 
-impl CommunicationBackend for JsCommunicationBackend {
-    type SendError = JsValue;
+impl From<&JsCommunicationBackend> for ThreadSafeJsCommunicationBackend {
```
dani-garcia: Minor thing, but it seems weird that a From impl would ever spawn a task, it feels like this would be better off as a separate function in  `JsCommunicationBackend`

crates/bitwarden-ipc/src/ipc_client.rs
coroiu: Yeah, I didn't want to include yet another crate, but I'm having issues with `tokio::time` which I'll have to replace with a cancellation handle like this so I'll probably just have to add it

crates/bitwarden-ipc/src/ipc_client.rs
coroiu: Hmm, I'm not sure I follow, which task do you mean?

crates/bitwarden-ipc/src/ipc_client.rs
coroiu: No, old comment, good catch!

crates/bitwarden-ipc/src/wasm/communication_backend.rs
coroiu: Yeah no that's fair, fixed!

crates/bitwarden-ipc/src/ipc_client.rs
dani-garcia: I mean that currently in the code we create a future that we spawn in the background, but then we use a oneshot channel to synchronize and wait for initialization. If we instead do the initialization outside the task, we can avoid having to do that synchronization. With some pseudocode, this is what we currently have:

```rust
let (await_init_tx, await_init_rx) = tokio::sync::oneshot::channel();

// We create a future that gets spawned in the background
let future = async move {
    // Do some initialization
    some_init_stuff().await;

    // Notify that we're initialized
    await_init_tx.send(());

    // Process messages
    loop { ... }
};

// Spawn the task and...
spawn_future(future);

///  wait for the initialization to complete before continuing
await_init_rx.await;
```

if we pull the initialization code out of the future, then we can avoid the explicit syncrhonization:

```rust
// Do some initialization before the future is created, now we don't need to wait as it's already done
some_init_stuff().await;

// We create a future that gets spawned in the background
let future = async move {
    // Process messages
    loop { ... }
};

// Spawn the task and...
spawn_future(future);
```

crates/bitwarden-threading/src/thread_bound_runner.rs
coroiu, line 1: Apparently I had commited this change to the wrong branch, I'm adding it here instead

crates/bitwarden-threading/src/thread_bound_runner.rs
```diff
@@ -25,22 +25,56 @@ struct CallRequest<ThreadState> {
...
 /// allows other threads to submit tasks (function pointers or closures) that operate on the
 /// thread-bound state.
 ///
 /// Tasks are queued via an internal channel and are executed sequentially on the owning thread.
 ///
 /// # Example
-/// ```ignore
-/// let runner = ThreadBoundRunner::new(my_state);
+/// ```
+/// # tokio_test::block_on(tokio::task::LocalSet::new().run_until(async {
```
dani-garcia, line 39: This crate is not a dev dependency so the build is failing

crates/bitwarden-ipc/src/traits/communication_backend.rs
```diff
@@ -1,14 +1,22 @@
...
+
 use crate::message::{IncomingMessage, OutgoingMessage};
 
 /// This trait defines the interface that will be used to send and receive messages over IPC.
 /// It is up to the platform to implement this trait and any necessary thread synchronization and
 /// broadcasting.
-pub trait CommunicationBackend {
-    type SendError;
+pub trait CommunicationBackend: Send + Sync + 'static {
+    type SendError: Debug + Send + Sync + 'static;
```
dani-garcia, line 9: Non-blocking thought: We could also have the Errors require the Error trait, which requires Debug on its own, though this might require creating a new error type for the tests, so maybe not worth it.
```suggestion
    type SendError: std::error::Error + Send + Sync + 'static;
```

crates/bitwarden-ipc/src/traits/session_repository.rs
```diff
@@ -4,20 +4,19 @@ use tokio::sync::RwLock;
...
 
-pub trait SessionRepository {
-    type Session;
-    type GetError;
-    type SaveError;
-    type RemoveError;
+pub trait SessionRepository<Session>: Send + Sync + 'static {
+    type GetError: Send + Sync + 'static;
+    type SaveError: Send + Sync + 'static;
+    type RemoveError: Send + Sync + 'static;
```
dani-garcia: Should these errors not also require Debug (or Error) to keep it consistent with the others?

crates/bitwarden-ipc/src/wasm/communication_backend.rs
```diff
@@ -69,11 +77,18 @@ impl JsCommunicationBackend {
...
     async fn send(&self, message: OutgoingMessage) -> Result<(), Self::SendError> {
-        self.sender.send(message).await
+        let result = self
+            .sender
+            .run_in_thread(|sender| async move {
+                sender.send(message).await.map_err(|e| format!("{:?}", e))
+            })
+            .await;
+
+        Ok(result.map_err(|e| e.to_string())??)
```
dani-garcia: Clippy lint here about unnecessary `?`

crates/bitwarden-ipc/src/ipc_client.rs
```diff
@@ -42,170 +47,255 @@ pub struct IpcClientTypedSubscription<Crypto, Com, Ses, Payload>
...
-        IpcClientTypedSubscription {
-            receiver: self.communication.subscribe().await,
+        Ok(IpcClientTypedSubscription {
+            receiver: self
+                .incoming
+                .read()
+                .await
+                .as_ref()
+                .ok_or(SubscribeError::NotStarted)?
+                .resubscribe(),
```
dani-garcia: Non-blocking thought: Some of the code for `IpcClientTypedSubscription` is fairly similar to the code in `IpcClientSubscription`, I wonder if we could define it like `IpcClientTypedSubscription<T>(IpcClientSubscription, PhantomData<T>)` or something like that to avoid repetition in their subscribe and receive functions.

crates/bitwarden-ipc/src/traits/communication_backend.rs
coroiu: Not using `std::error` allow us to use an arbitrary `JsValue` from a captured exception and bubble that up. Though of course, now that we are going across thread boundaries that is no longer possible because of other reasons, but still, maybe some other platform can take advantage of it?

dani-garcia: Yeah that's reasonable enough, we can always add the bound later if it turns out we only use Errors.

crates/bitwarden-ipc/src/ipc_client.rs
coroiu: Fixed, much simpler!

crates/bitwarden-ipc/src/wasm/communication_backend.rs
coroiu: Fixed!

crates/bitwarden-ipc/src/ipc_client.rs
coroiu: replaced with token 👍

crates/bitwarden-threading/src/thread_bound_runner.rs
coroiu: Fixed!

crates/bitwarden-ipc/src/traits/session_repository.rs
coroiu: Fixed!

crates/bitwarden-ipc/src/ipc_client.rs
coroiu: This change meant I could remove the `client` reference which simplified a lot of generics 👍

crates/bitwarden-threading/src/lib.rs
```diff
@@ -1,3 +1,8 @@
+#[cfg(all(target_arch = "wasm32", not(feature = "wasm")))]
+compile_error!(
+    "The `wasm` feature must be enabled to use the `bitwarden-ipc` crate in a WebAssembly environment."
```
dani-garcia, line 5: bitwarden-threading doesn't have any feature defined in cargo.toml, does it? Why is this needed?

I wonder if we should decide whether we want to use target_arch = "wasm32" or feature = "wasm" and just update the whole repo to be consistent.

crates/bitwarden-ipc/src/ipc_client.rs
```diff
@@ -1,233 +1,288 @@
...
+            incoming: RwLock::new(None),
+            cancellation_token: RwLock::new(None),
         })
     }
 
+    pub async fn start(self: &Arc<Self>) {
+        let cancellation_token = CancellationToken::new();
+        let mut self_cancellation_token = self.cancellation_token.write().await;
+
+        *self_cancellation_token = Some(cancellation_token.clone());
```
dani-garcia: We're not dropping `self_cancellation_token`, so we hold the lock a bit longer than necessary.
We can avoid this if we use `Option::replace()` and don't store the lock guard in a variable:

```suggestion
        self.cancellation_token
            .write()
            .await
            .replace(cancellation_token.clone());
```

crates/bitwarden-ipc/src/ipc_client.rs
```diff
@@ -1,233 +1,288 @@
...
+
+        let com_receiver = self.communication.subscribe().await;
+        let (client_tx, client_rx) = tokio::sync::broadcast::channel(CHANNEL_BUFFER_CAPACITY);
+
+        let mut client_incoming = self.incoming.write().await;
+        *client_incoming = Some(client_rx);
+        let _ = client_incoming;
```
dani-garcia: Here we're dropping correctly, but I feel like using replace makes it easier to parse.
```suggestion
        self.incoming.write().await.replace(client_rx);
```

crates/bitwarden-ipc/src/ipc_client.rs
```diff
@@ -1,233 +1,288 @@
...
+        tokio::spawn(future);
+
+        #[cfg(target_arch = "wasm32")]
+        wasm_bindgen_futures::spawn_local(future);
+    }
+
+    pub async fn is_running(self: &Arc<Self>) -> bool {
+        let incoming = self.incoming.read().await;
+        let cancellation_token = self.cancellation_token.read().await;
+        incoming.is_some() && cancellation_token.is_some()
```
dani-garcia: I don't think this is an issue here, but we should be careful when holding two locks together, if any other parts of the code are doing the same thing in the opposite order, we can find ourselves with a deadlock. 

If we move the is_some to the variables, we avoid holding the locks and remove the problem from happening.
```suggestion
        let has_incoming = self.incoming.read().await.is_some();
        let has_cancellation_token = self.cancellation_token.read().await.is_some();
        has_incoming && has_cancellation_token
```

crates/bitwarden-ipc/src/ipc_client.rs
```diff
@@ -1,233 +1,288 @@
...
-            Crypto::ReceiveError,
-            <Com::Receiver as CommunicationBackendReceiver>::ReceiveError,
-        >,
-    > {
+    /// Setting the cancellation_token to `None` will wait indefinitely.
+    pub async fn receive(
+        &mut self,
+        cancellation_token: Option<CancellationToken>,
+    ) -> Result<IncomingMessage, ReceiveError> {
         let receive_loop = async {
```
dani-garcia: Now that we're using select! for the cancellation here, we can probably structure it as a `loop {select! {} }` like the main loop in `start` for consistency:
```rust
    pub async fn receive(
        &mut self,
        cancellation_token: Option<CancellationToken>,
    ) -> Result<IncomingMessage, ReceiveError> {
        let cancellation_token = cancellation_token.unwrap_or_default();

        loop {
            select! {
                _ = cancellation_token.cancelled() => {
                    return Err(ReceiveError::Cancelled);
                }
                result = self.receiver.recv() => {
                    let received = result?;
                    if self.topic.is_none() || received.topic == self.topic {
                        return Ok::<IncomingMessage, ReceiveError>(received);
                    }
                },
            }
        }
    }
```

crates/bitwarden-ipc/src/ipc_client.rs
```diff
@@ -414,17 +468,27 @@ mod tests {
...
+                .expect("Serialization should not fail"),
+        );
 
         let received_message = subscription.receive(None).await.unwrap();
 
         assert_eq!(received_message, typed_message);
     }
 
     #[tokio::test]
+    // async fn skips_message_if_it_was_not_deserializable() {
```
dani-garcia: Remove comment?

dani-garcia (commented):
Some comments about the locking we're doing, but otherwise LGTM

crates/bitwarden-threading/src/lib.rs
coroiu: I added a `wasm` feature. I needed to add it because the crates that depend on `threading` use features and so I had to either convert them to `target_arch` too or convert `threading` to use `feature`.

Yeah, I think features has some advantage but I can't remember what 🙃 So far adding the `wasm` flag to a non-wasm build (which we do when we run `cargo test --all-features`) hasn't caused any issues because `wasm-bindgen` has a goal to always compile on all target and instead throw runtime errors if you try to use it on an unsupported target. But now, we have this thing:

```rs
#[cfg(not(target_arch = "wasm32"))]
use tokio::task::spawn_local;
#[cfg(all(target_arch = "wasm32", feature = "wasm"))]
use wasm_bindgen_futures::spawn_local;
```

which is actively used in unit-tests and so all of a sudden we cannot just do 

```rs
#[cfg(not(featgure = "wasm"))]
use tokio::task::spawn_local;
#[cfg(feature = "wasm")]
use wasm_bindgen_futures::spawn_local;
```

because that would cause `wasm_bindgen_futures::spawn_local` to be called during a regular unit test which would cause a runtime error.

So either we remove `wasm` from `cargo test` (`cargo test --all-features --except wasm` [???](https://internals.rust-lang.org/t/pre-rfc-cargo-test-default-features-except-and-co/20850/6)) or we use `target_arch`

crates/bitwarden-ipc/src/ipc_client.rs
coroiu: Nice, thanks for the tip! 👍  Fixed!

crates/bitwarden-ipc/src/ipc_client.rs
coroiu: Fixed!

crates/bitwarden-ipc/src/ipc_client.rs
coroiu: Thanks for catching this, locks seem to be an easy footgun, it's not super obvious (unless you're familiar with how rust works) that this change would actually drop the locks. Fixed!

crates/bitwarden-ipc/src/ipc_client.rs
coroiu: Good point, fixed!

crates/bitwarden-ipc/src/ipc_client.rs
coroiu: Removed!

dani-garcia (dismissed):

dani-garcia (approved):

PR #274, opened 2025-05-12
[PM-18102] Use opaque autogenerated local key ids

dani-garcia:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-18102

## 📔 Objective

Currently any function that adds/decrypts a key needs to provide the KeyId under which it can be stored. This is fairly cumbersome and error prone as it can lead to accidentally using the wrong key. 

This PR changes all the functions that store a key in a context to instead generate and return a unique ID, which simplifies the code and makes the code less error prone.

## ⏰ Reminders before review

- CI builds passed
  team

## 🦮 Reviewer guidelines


crates/bitwarden-core/src/key_management/crypto.rs
```diff
@@ -735,13 +733,10 @@ pub(crate) fn make_v2_keys_for_v1_user(
 
     // New user key
     let user_key = SymmetricCryptoKey::make_xchacha20_poly1305_key();
-    #[allow(deprecated)]
-    ctx.set_symmetric_key(temporary_user_key_id, user_key.clone())?;
```
dani-garcia, line 755: Note that this key was set but never used, so I've removed it here

quexten:
Noting for reviewers: I'll take over this PR for the time being (https://bitwarden.atlassian.net/browse/PM-18102?focusedCommentId=113387)

coroiu (dismissed):

gbubemismith (dismissed):
Vault changes look good

crates/bitwarden-crypto/src/store/context.rs
mzieniukbw, line 196: ❓ Are we going to cleanup the deprecated functions with an equivalent `add_local_symmetric_key` later (same for asymmetric and signing keys) and remove the local key id support from `set_symmetric_key` function ?

mzieniukbw (dismissed):

crates/bitwarden-crypto/src/store/context.rs
quexten: Yeah, I think that makes sense. Considering this is a fairly small nit I'll leave it as follow-up work.

quexten:
Had to resolve some merge conflicts.

mzieniukbw (approved):

coroiu (approved):

Hinton (approved):
Do we ever need to clear the local keys?

gbubemismith (approved):

quexten:
> Do we ever need to clear the local keys?

@Hinton 

Considering the key ids are randomly generated now and the collision chance is basically zero, I don't think we need to. Dropping the context will also drop the keys.

I will defer to @dani-garcia though, maybe he has thought about other use-cases here before.

dani-garcia:
> > Do we ever need to clear the local keys?
> 
> @Hinton
> 
> Considering the key ids are randomly generated now and the collision chance is basically zero, I don't think we need to. Dropping the context will also drop the keys.
> 
> I will defer to @dani-garcia though, maybe he has thought about other use-cases here before.

@quexten 

That's correct security-wise, the current UUIDs basically guarantee no posibility of collision so clearing them is not needed as the context teardown will take care of them all eventually.

There's still a small concern that if a lot of operations are done on that context, the keys will keep growing until the context is dropped. For example, if you were decrypting 5K ciphers individually, you might end up with 5K local keys at the end. This is why the `encrypt_list` and `decrypt_list` operations clear the local keys themselves, and why we recommend using those high level functions instead of operating on the context directly: 
https://github.com/bitwarden/sdk-internal/blob/ac7150235ee7f3e362419a9240ba233a2e45f53b/crates/bitwarden-crypto/src/store/mod.rs#L337-L338

Ideally I'd like this to be handled automatically by introducing some scope concept to the keystore context and key identifiers, but I haven't found a satisfying solution just yet. Something like: 
```rust
ctx.scope(|| {
    let key = ctx.decrypt_key(somedata);

   use_the_key(ctx, key);
   // Key gets automatically deleted from the store at the end of this scope
};
```

PR #275, opened 2025-05-14
[PM-18042] Build request response structure

coroiu:
## ⏰ Reminders before review

- CI builds passed
  team

## 🦮 Reviewer guidelines


crates/bitwarden-ipc/src/ipc_client.rs
```diff
@@ -186,6 +219,23 @@ where
...
+    pub async fn register_rpc_handler<H>(self: &Arc<Self>, handler: H)
+    where
+        H: RpcHandler + Send + Sync + 'static,
+        H::Request: RpcRequest + TryFrom<Vec<u8>> + TryInto<Vec<u8>>,
+        <H::Request as RpcRequest>::Response: TryFrom<Vec<u8>> + TryInto<Vec<u8>>,
+        <H::Request as TryFrom<Vec<u8>>>::Error: std::fmt::Display,
+        <H::Request as TryInto<Vec<u8>>>::Error: std::fmt::Display,
+        <<H::Request as RpcRequest>::Response as TryFrom<Vec<u8>>>::Error: std::fmt::Display,
+        <<H::Request as RpcRequest>::Response as TryInto<Vec<u8>>>::Error: std::fmt::Display,
+    {
```
coroiu: thought: I'm still very divided on this, maybe we should create a follow-up ticket where we simply hardcode the wire-format conversion

crates/bitwarden-ipc/src/rpc/request_message.rs
```diff
@@ -0,0 +1,32 @@
...
+    }
+}
+
+impl TryFrom<RpcRequestMessage> for Vec<u8> {
+    type Error = serde_json::Error;
+
+    fn try_from(value: RpcRequestMessage) -> Result<Self, Self::Error> {
+        serde_json::to_vec(&value)
+    }
+}
```
coroiu: thought: this is another thing I'm a bit divided on; the levels of serialization. At this point we don't know how to deserialize the request (because we don't know its type) but we still need to wrap it somehow and so we end up with a nested `Vec<u8>` field which contains the actual request data. This will then be deserialized by the rpc handler

coroiu: Technically RPC can be seen as built on top of the req/res paradigm that the IPC client provides at its core, that's why we are actually defining a serilalization format (`serde_json`) here, because it's RPC over `TypedMessages`

crates/bitwarden-ipc/src/ipc_client.rs
dani-garcia: Question: What if we change `RpcRequest` to require Serialize and DeserializeOwned?
```rust
pub trait RpcRequest: serde::ser::Serialize + serde::de::DeserializeOwned {
    type Response: serde::ser::Serialize + serde::de::DeserializeOwned;

    /// Used to identify handlers
    fn name() -> String;
}
```

This would avoid a lot of the generic type bounds on all the methods, and I feel like most messages are going to just depend on serde to serialize/deserialize the data anyway. That would also centralize the wire-format to the IPC client, and we can swap to another one as long as it supports serde, which most formats do.

crates/bitwarden-ipc/src/wasm/error.rs
```diff
@@ -0,0 +1,38 @@
+use bitwarden_error::bitwarden_error;
+use thiserror::Error;
+
+use crate::{ipc_client::RequestError, rpc::error::RpcError};
+
+#[derive(Debug, Error, PartialEq, Eq)]
+#[bitwarden_error(flat, export_as = "RequestError")]
+pub enum JsRequestError {
```
dani-garcia: Is this error duplicated because the original has the generic type? If so, I wonder if it won't be better to just remove the generic and use `Send(Box<dyn Debug>)` or `Send(String)` directly on the main type.

crates/bitwarden-ipc/src/wasm/ipc_client.rs
```diff
@@ -15,11 +15,16 @@ use crate::{
...
 #[wasm_bindgen(js_name = IpcClient)]
 pub struct JsIpcClient {
-    // TODO: Change session provider to a JS-implemented one
-    client: Arc<
+    #[wasm_bindgen(skip)]
+    /// The underlying IPC client instance. Use this to create WASM-compatible functions
+    /// that interact with the IPC client, e.g. to register RPC handlers, trigger RPC requests,
+    /// send typed messages, etc. For examples see
+    /// [wasm::ipc_register_discover_handler](crate::wasm::ipc_register_discover_handler).
+    pub client: Arc<
```
dani-garcia, line 23: This should probably be pub(crate), and then we don't need the skip

coroiu: But then other teams won't be able to make their own functions using this?

dani-garcia: You mean to let them run `register_rpc_handler` in their crates? Yeah that's fair, I didn't think of that

crates/bitwarden-ipc/src/wasm/error.rs
coroiu: Yeah you're probably right, I did it with the other types because if how they were jumping across threads and `JsValue` couldn't do that anyways, but this one was still executed on the same thread. But should probably just do it with this one too for consistency

crates/bitwarden-ipc/src/ipc_client.rs
coroiu: Lol VS Code renders your comment with the german flag 
image

I'll rewrite this for RPC and then we can follow-up with a similar rewrite for the top-level message structs

dani-garcia: That's hilarious 😆

crates/bitwarden-ipc/src/wasm/error.rs
coroiu: Fixed 👍

crates/bitwarden-ipc/Cargo.toml
```diff
@@ -20,15 +20,18 @@ wasm = [
 ] # WASM support
 
 [dependencies]
+async-trait = { workspace = true }
 bitwarden-error = { workspace = true }
 bitwarden-threading = { workspace = true }
+erased-serde = "0.4.6"
```
dani-garcia:
```suggestion
erased-serde = ">=0.4.6, <0.5"
```

crates/bitwarden-ipc/src/rpc/request.rs
```diff
@@ -0,0 +1,8 @@
+use serde::{de::DeserializeOwned, Serialize};
+
+pub trait RpcRequest: Serialize + DeserializeOwned + 'static {
+    type Response: Serialize + DeserializeOwned + 'static;
+
+    /// Used to identify handlers
+    fn name() -> String;
```
dani-garcia: For some other traits, instead of a function we use an associated constant, like on `PayloadTypeName`, should we do the same here?

crates/bitwarden-ipc/src/rpc/request_message.rs
```diff
@@ -0,0 +1,59 @@
...
+
+    pub fn request_id(&self) -> &str {
+        &self.partial.request_id
+    }
+
+    pub fn request_type(&self) -> &str {
+        &self.partial.request_type
+    }
+
+    pub fn full<T>(&self) -> Result<RpcRequestMessage<T>, RpcError>
```
dani-garcia: `full` is a bit of a strange name for a function, maybe `deserialize_full` to be more clear?

dani-garcia (commented):
Some minor comments now that we have everything together, the rest looks pretty good to me!

crates/bitwarden-ipc/src/rpc/request.rs
coroiu: Ah, yes, I only recently discovered that you could use constants like this (during a review I think) so I just missed updating this :)

dani-garcia (approved):

PR #278, opened 2025-05-19
Add-userid-to-encryption-methods

MGibson1:
This is not the owning entity of the key, but the user that is unlocked that triggered the request

## ⏰ Reminders before review

- CI builds passed
  team

## 🦮 Reviewer guidelines

### DIFF
--- a/crates/bitwarden-core/src/client/internal.rs
+++ b/crates/bitwarden-core/src/client/internal.rs
@@ -178,10 +178,7 @@ impl InternalClient {
     }
 
     pub fn get_user_id(&self) -> Option<Uuid> {
-        self.user_id
-            .read()
-            .expect("RwLock is not poisoned")
-            .clone()
+        self.user_id.read().expect("RwLock is not poisoned").clone()
     }
 
     #[cfg(feature = "internal")]
--- a/crates/bitwarden-core/src/client/internal.rs
+++ b/crates/bitwarden-core/src/client/internal.rs
@@ -1,4 +1,4 @@
-use std::sync::{Arc, RwLock};
+use std::sync::{Arc, OnceLock, RwLock};
 
 use bitwarden_crypto::KeyStore;
 #[cfg(any(feature = "internal", feature = "secrets"))]
@@ -13,6 +13,7 @@ use super::login_method::ServiceAccountLoginMethod;
 use crate::{
     auth::renew::renew_token,
     client::{encryption_settings::EncryptionSettings, login_method::LoginMethod},
+    error::UserIdAlreadySetError,
     key_management::KeyIds,
     DeviceType,
 };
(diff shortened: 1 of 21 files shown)


crates/bitwarden-core/src/client/internal.rs
```diff
@@ -172,6 +173,14 @@ impl InternalClient {
         &self.key_store
     }
 
+    pub fn set_user_id(&self, user_id: Uuid) {
+        *self.user_id.write().expect("RwLock is not poisoned") = Some(user_id);
+    }
+
```
coroiu: issue?: it seems odd to have a public function that changes the `user_id` without any additional context required. I feel like the `user_id` should only ever change when the client is initialized and possibly never again after that. Maybe there's some edge case where we'd want a client without Crypto but tied to a user? If that is the case then I think this function should instead be called something like `init_user_id()` and only allow itself to be called once. However, it's probably more likely that we'll split up the client into user and non-user scoped clients and then just require the `user_id` on creation.

Either way, I think we should remove the ability to change the `user_id`. My opinion is that it should be bundled into the `initialize_user_crypto` functions below, but I'm not married to that solution

dani-garcia: Yeah I agree, separate user-bound and non-user bounds clients would allow us to require user id during construction and enforce no changes to it, until we have those we can use a `OnceLock` to ensure the ID is only initialized once:
```rust
    pub(crate) user_id: OnceLock<Uuid>,

    pub fn init_user_id(&self, user_id: Uuid) -> Result<(), UserIdAlreadySetError> {
        self.user_id
            .set(user_id)
            .map_err(|_| UserIdAlreadySetError)
    }

    pub fn get_user_id(&self) -> Option<Uuid> {
        self.user_id.get().map(|x| *x)
    }
```

MGibson1: There definitely are user-bound, but not crypto-inited clients -- that's a locked SDK.

The thought was, that I only _need_ the `UserId` during encryption (for now) and I'm not 100% familiar with our current usages and I wanted to minimize the impact of the breaking change.

I can certainly revise. I agree it makes sense to init userId only once -- that's an easy change -- Actually making the pure, environment-bound, and user-bound client contexts feel out of scope here, though. Do you agree?

coroiu: Oh yes splitting them up is way out of scope, we're gonna be working on that with Auth if I remember correctly. I was just thinking ahead a little, I could've been a little clearer there :) I think `OnceLock` that Dani suggested is a perfectly sound solution until we eventually start splitting stuff

coroiu (approved):

crates/bitwarden-core/src/client/internal.rs
MGibson1: This is done (468bcda), but there is an unfortunate side effect that crypto initialization can now error with a userId already set.

This will have to be updated once the clients are split, but for now is in line with the expectation that UserId is sent into crypto initialization. Once I implemented it, I got to thinking that for the same reasons it makes sense to error should multiple `init_user_crypto` requests be sent into a single client, so hopefully this can just morph into a `CryptoAlreadyInitialized` error or some such.

PR #279, opened 2025-05-19
[PM-21691] - Added tree and test with implementation for CollectionView

jrmccannon:
## 🎟️ Tracking
[PM-21691](https://bitwarden.atlassian.net/browse/PM-21691)

## 📔 Objective
This splits the collections code out of the vault crate and adds a Tree implementation for representing the parent child relationship between nested collections.

This ideally will then be used as a unified representation for the nested collection structure across all clients, instead of each one having to implement their own.

## ⏰ Reminders before review

- CI builds passed
  team

## 🦮 Reviewer guidelines

[PM-21691]: https://bitwarden.atlassian.net/browse/PM-21691?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


crates/bitwarden-vault/src/tree.rs
```diff
@@ -0,0 +1,275 @@
...
+}
+
+pub struct Tree<T: TreeItem> {
+    pub nodes: Vec<TreeNode>,
+    pub items: Vec<TreeIndex<T>>,
+    path_to_node: HashMap<Vec<String>, usize>
+}
+
+impl<T: TreeItem> Tree<T> {
+    pub fn from_items(items: &mut Vec<T>) -> Self {
```
coroiu: suggestion: I would take the value here since we are modifying it in unexpected ways inside of the function (sorting). It might also mean that we can get away with not cloning the items when you put them in the tree since you'd have ownership of them

crates/bitwarden-vault/src/tree.rs
```diff
@@ -0,0 +1,275 @@
...
+    pub parent_idx: Option<usize>,
+    pub children_idx: Vec<usize>,
+    pub path: Vec<String>
+}
+
+impl TreeNode {
+    pub fn new<T: TreeItem>(id: usize, parent_idx: Option<usize>, children_idx: Vec<usize>, index: TreeIndex<T>) -> Self {
+        TreeNode {
+            id,
+            item_id: index.data.id().unwrap(),
```
coroiu: issue: unwrap is not allowed

suggestion: either remove the ability for this to be optional or use `Result:Err` somehow

crates/bitwarden-vault/src/collection.rs
```diff
@@ -85,3 +86,22 @@ impl TryFrom<CollectionDetailsResponseModel> for Collection {
...
 }
+
+impl TreeItem for CollectionView {
+    fn id(&self) -> Option<Uuid> {
+        self.id
+    }
+
+    fn short_name(&self) -> &str {
+        &self.path().last().unwrap()
+    }
```
coroiu: issue: unwrap is discouraged/not allowed, and while we allow `expect` it should be reserved for the very inlikely "This should never happen" scenarios.

crates/bitwarden-vault/src/tree.rs
```diff
@@ -0,0 +1,275 @@
+use std::collections::HashMap;
+use std::fmt::Debug;
+use uuid::Uuid;
+
+pub trait TreeItem: Clone + Debug {
+    fn id(&self) -> Option<Uuid>;
```
coroiu: suggestion: since you are using `unwrap` on this that must mean that you **_really_** don't expect this to be `None`, in which case I'd suggest removing the option completely
```suggestion
    fn id(&self) -> Uuid;
```

crates/bitwarden-vault/src/tree.rs
```diff
@@ -0,0 +1,275 @@
...
+            let children = node.children;
+            
+            assert_eq!(children.len(), 2);
+            assert_eq!(item.id(), Some(parent_id));
+            assert_eq!(item.short_name(), "parent");
+            assert_eq!(item.path(), ["parent"]);
+            assert!(parent.is_none());
+        } else {
+            panic!("Node not found");
+        }
```
coroiu: suggestion: simplify by expecting a value

```suggestion
        let node = Tree::from_items(&mut items)
            .get_item_by_id(parent_id)
            .expect("Node not found");
        
        let item = node.item;
        let parent = node.parent;
        let children = node.children;
            
        assert_eq!(children.len(), 2);
        assert_eq!(item.id(), Some(parent_id));
        assert_eq!(item.short_name(), "parent");
        assert_eq!(item.path(), ["parent"]);
        assert!(parent.is_none());
```

crates/bitwarden-vault/src/tree.rs
```diff
@@ -0,0 +1,275 @@
...
+            TestItem {
+                id: Uuid::new_v4(),
+                name: "parent/child2".to_string()
+            },
+        ];
+
+        let node_option = Tree::from_items(&mut items)
+            .get_item_by_id(child_1_id);
+
+        if let Some(node) = node_option {
```
coroiu: suggestion: same here, simplify with `expect`

crates/bitwarden-vault/src/tree.rs
```diff
@@ -0,0 +1,275 @@
...
+            assert_eq!(item.short_name(), "child1");
+            assert_eq!(item.path(), ["parent", "child1"]);
+            assert_eq!(parent.unwrap().id, parent_id);
+        } else {
+            panic!("Node not found");
+        }
+    }
+
+    #[test]
+    fn given_collection_with_two_children_where_there_parent_node_does_not_exist_children_are_returned_correctly() {
```
coroiu: typo
```suggestion
    fn given_collection_with_two_children_where_their_parent_node_does_not_exist_children_are_returned_correctly() {
```

coroiu (commented):
Just some minor comments, but very nice overall, +1 for the tests 👍

package.json
```diff
@@ -21,5 +21,8 @@
   "devDependencies": {
     "@openapitools/openapi-generator-cli": "2.20.2",
     "prettier": "3.5.3"
+  },
+  "dependencies": {
```
gbubemismith: ❓ Is this intentional?

crates/bitwarden-vault/src/collection_client.rs
```diff
@@ -24,6 +34,79 @@ impl CollectionsClient {
...
+        }
+    }
+}
+
+#[cfg_attr(feature = "wasm", wasm_bindgen)]
+pub struct CollectionViewTree {
+    tree: Tree<CollectionView>,
+}
+
+#[allow(unused)]
```
gbubemismith: ⚠️ I think we can remove this

gbubemismith (commented):
Looks good, just noticed a few things

package.json
jrmccannon: It is not. thank you.

gbubemismith (dismissed):

package-lock.json
gbubemismith, line 1: I think you forgot to remove this as well

gbubemismith (commented):
Just one last comment

gbubemismith (approved):

justindbaur (approved):

PR #283, opened 2025-05-22
Add more threading utilities

coroiu:
## 📔 Objective

Add sleep and cancellation token 

## ⏰ Reminders before review

- CI builds passed
  team

## 🦮 Reviewer guidelines


crates/bitwarden-threading/src/time.rs
```diff
@@ -0,0 +1,56 @@
...
+
+        sleep(Duration::from_millis(100)).await;
+
+        let end = Date::now();
+        let elapsed = end - start;
+
+        assert!(elapsed >= 90.0, "Elapsed time was less than expected");
+    }
+
+    // #[cfg(not(target_arch = "wasm32"))]
```
dani-garcia: This comment should probably be removed

crates/bitwarden-threading/src/cancellation_token.rs
```diff
@@ -0,0 +1,66 @@
+pub use tokio_util::sync::CancellationToken;
+
+#[cfg(target_arch = "wasm32")]
+pub mod wasm {
```
dani-garcia, line 4: If the whole file is wasm-only, we should probably just put the cfg in the lib.rs, and then we avoid one nesting level here:
```
#[cfg(target_arch = "wasm32")]
pub mod cancellation_token;
```

crates/bitwarden-threading/src/cancellation_token.rs
```diff
@@ -0,0 +1,66 @@
...
+    use wasm_bindgen::prelude::*;
+
+    use super::*;
+
+    #[wasm_bindgen]
+    extern "C" {
+        #[wasm_bindgen]
+        #[derive(Clone)]
```
dani-garcia, line 12: Question: Is this safe to be clonable when it contains a pointer to a JS value? I wonder how that works internally

crates/bitwarden-threading/src/cancellation_token.rs
```diff
@@ -0,0 +1,66 @@
...
+            let token_clone = token.clone();
+            let closure = Closure::new(move || {
+                token_clone.cancel();
+            });
+            self.add_event_listener("abort", &closure);
+            closure.forget(); // Transfer ownership to the JS runtime
+
+            token
+        }
+    }
```
dani-garcia, line 73: We should probably add a basic test to ensure this works:
```rust
        fn should_abort_wasm() {
            use super::*;

            console_error_panic_hook::set_once();

            let abort = AbortController::new();
            let signal = abort.signal();

            let token = signal.to_cancellation_token();

            assert!(!token.is_cancelled(), "Token should not be cancelled");
            assert!(!signal.aborted(), "Signal should not be aborted");

            abort.abort(JsValue::from("Test reason"));

            assert!(token.is_cancelled(), "Token should be cancelled");
            assert!(signal.aborted(), "Signal should be aborted");
        }
```

dani-garcia: Also, I wonder if we should add a comment indicating that the returned CancellationToken is only for receiving the cancellation signal. Currently you can do something like this:
```rust
let abort = AbortController::new();
let signal = abort.signal();

// This will do nothing, you have to use abort.abort();
signal.to_cancellation_token().cancel();

// Still not aborted
assert!(!signal.aborted(), "Signal is not aborted");
```

crates/bitwarden-threading/src/cancellation_token.rs
coroiu: It's not, I'm re-exporting the cancellation token as a convenience regardless of wasm.

crates/bitwarden-threading/src/cancellation_token.rs
coroiu: You mean this? Or is there a subtle difference I'm missing?

https://github.com/bitwarden/sdk-internal/blob/b7115f6f96994d50f648aaab4998080f956efe8c/crates/bitwarden-threading/tests/cancellation_token/wasm.rs#L33-L51

Added comment!

dani-garcia: Oh they're on the other file, didn't see them 🤦

crates/bitwarden-threading/src/cancellation_token.rs
coroiu: Yeah I was surprised too but it's perfectly safe. The imported type is just a wrapper around `JsValue` which itself is just a wrapper around a pointer to the actual value stored in a table

```
impl Clone for JsValue {
    #[inline]
    fn clone(&self) -> JsValue {
        unsafe {
            let idx = __wbindgen_object_clone_ref(self.idx);
            JsValue::_new(idx)
        }
    }
}
```

crates/bitwarden-threading/src/cancellation_token.rs
coroiu: Hehe, decided to put them there so I can test the public interface. I've forgotten `pub` too many times

crates/bitwarden-threading/src/cancellation_token.rs
dani-garcia: Interesting, so Clone gives you two pointers to the same AbortController here right? I guess that makes sense following JS semantics of variable assignment.

```rust
let abort1 = AbortController::new();
let abort2 = abort1.clone();

abort1.abort(JsValue::from("Test reason"));
assert!(abort2.signal().aborted(), "Abort2 is also aborted")
```

coroiu: Yeah exactly, they act like an `Rc`. You can even check the total amount of `JsValue` pointers in existence using `externref_heap_live_count`

dani-garcia (approved):

PR #294, opened 2025-05-26
Remove vault folder from wasm-internal

Hinton:
## 📔 Objective

Removes the vault folder from wasm-internal and relies on the upstream implementations instead. Note this adds `wasm_bindgen` to everything that should be exposed which is a bit clunky.

## ⏰ Reminders before review

- CI builds passed
  team

## 🦮 Reviewer guidelines


crates/bitwarden-vault/src/totp_client.rs
```diff
@@ -1,10 +1,42 @@
...
+    /// - `key` - Can be:
+    ///     - A base32 encoded string
+    ///     - OTP Auth URI
+    ///     - Steam URI
+    /// - `time_ms` - Optional timestamp in milliseconds
+    #[wasm_bindgen(js_name = "generate_totp")]
+    pub fn generate_totp_wasm(
+        &self,
+        key: String,
+        time_ms: Option<f64>,
```
Hinton, line 28: DateTimes 😢

crates/bitwarden-vault/src/totp_client.rs
```diff
@@ -30,3 +61,12 @@ impl VaultClient {
         generate_totp_cipher_view(&mut key_store.context(), view, time)
     }
 }
+
+#[cfg_attr(feature = "wasm", wasm_bindgen)]
+impl VaultClient {
+    pub fn totp(&self) -> TotpClient {
```
dani-garcia: This would be easier to find if it was placed in the `vault_client` file with the other `VaultClient` methods, I think.

crates/bitwarden-vault/src/totp_client.rs
dani-garcia: If we're having a separate WASM function to deal with the parameters, we could use `js_sys::Date` or `js_sys::Number` here and then convert, that'll probably be a breaking change for the current SDK API though, so maybe later.

dani-garcia (dismissed):

crates/bitwarden-vault/src/totp_client.rs
Hinton: It would, but it follows the existing patterns in the crate.

Hinton: Done in https://github.com/bitwarden/sdk-internal/pull/295

gbubemismith (dismissed):
Thanks for this change

gbubemismith (dismissed):

dani-garcia (approved):

gbubemismith (approved):

PR #296, opened 2025-05-28
Update OpenAPI version and templates

dani-garcia:
## 📔 Objective

Update the version of openapi generator that we use, and use the latest templates. Because of the autogenerated code, this PR is fairly big so I've split it into easier to parse commits. I recommend reviewing them individually:
- [Update versions](https://github.com/bitwarden/sdk-internal/commit/5b8c947b09923c889c9374fea31fb396b8b540ee): Very straightforward, just bumps all the packages. This doesn't change the generated code in any way

- [Update template files](https://github.com/bitwarden/sdk-internal/commit/82d44298b252ff17a36972ad6821c9f15a8f56ed): This updates the template files with the latest in the openAPI repo. Note that this is the templates as they are upstream, without our changes. The next commit will apply only our changes, so it's easy to review and rebase in the future. I basically deleted the contents of the `support/openapi-templates` and pasted the contents from here: https://github.com/OpenAPITools/openapi-generator/tree/05e672d85672a2f7a97fe0d3601aadc26d43e572/modules/openapi-generator/src/main/resources/rust Best way to validate it's correct would be to do the same thing and check that the files match.

- [Make changes to templates](https://github.com/bitwarden/sdk-internal/commit/272d541e4a3ed8a2f1b52e8b414726b81fba1900). The changes applied are as follows:
  -  Use the `uuid` crate type instead of strings.
  - Removed the double optional pattern, which was applied when types were `nullable & !required`. We don't differentiate between nullable and !required, so the double optional only adds complexity.
  - Updated `Cargo.toml` so it's more in line with ours.
  - Ignored some clippy lints
  - The new templates already contain support for int enums, so that's one less patch we have to do

- [🤖 Generated code](https://github.com/bitwarden/sdk-internal/commit/fff91275fb1db5fc380d935e8dd17c9f5ab8235a): This is the autogenerated bindings built using `./support/build-api.sh` on the same server version as last time https://github.com/bitwarden/server/commit/29dc69a77b7b1a781dc93cbbcd4bf5475184ddf1 . I haven't used a newer version to reduce the number of changes on this PR. The changes are entirely because of template updates, so rather than review it I'd recommend trying to build the bindings on top and seeing if they match:
  - Models are functionally identical, their Display impl is slightly different but has the same output.
  - API functions are restructured internally, and they now check for a JSON content type before trying to deserialize the response.
  - I've kept the Cargo.toml changes as they were before, because our version is updated with renovate updates.

## ⏰ Reminders before review

- CI builds passed
  team

## 🦮 Reviewer guidelines


coroiu (dismissed):
Confirmed that re-generating the code yields in the same result on my machine as in this PR 👍

coroiu (approved):

PR #301, opened 2025-06-09
[PM-12612] SDK-Managed Repository support

dani-garcia:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-12612

## 📔 Objective

A continuation of https://github.com/bitwarden/sdk-internal/pull/213, this PR implements some simple SDK managed data store based on SQLite (on non-wasm) and IndexedDB (on wasm).

Both databases are wrapped in a `Database` trait and conditionally compiled based on platform. Then from each database we can get multiple `Repository` implementations that can be used to read and write data persistently. Each repository is mapped to a separate table in SQLite and Object Store in IndexedDb.

Some limitations of the current system:
- The database currently needs to be initialized as a separate step to both the SdkClient and the client-managed repositories, I feel like ideally we can do that as part of Client initialization, but that would require us to make initialization async. I've left that for the future.
- Currently we don't have any indexes beyond the main key, but both sqlite and indexeddb support adding more. 
- The current structure doesn't have any migration capabilities, which are required for indexeddb to create the object stores, this PR temporarily adds version numbers to the `register_repository_item` calls, just to get something working, but note that this is removed in https://github.com/bitwarden/sdk-internal/pull/410 and replaced with a better migration system.

## ⏰ Reminders before review

- CI builds passed
  team

## 🦮 Reviewer guidelines


crates/bitwarden-state/src/sdk_managed/sqlite.rs
```diff
@@ -0,0 +1,114 @@
...
+
+        // Set WAL mode for better concurrency
+        db.execute("PRAGMA journal_mode = WAL;", [])?;
+
+        let transaction = db.transaction()?;
+
+        for reg in registrations {
+            transaction.execute(
+                "CREATE TABLE IF NOT EXISTS ?1 (key TEXT PRIMARY KEY, value TEXT NOT NULL);",
+                [reg.name],
```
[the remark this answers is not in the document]
coroiu: What, really? xD

Edit: apparently, yes

crates/bitwarden-threading/src/thread_bound_runner.rs
```diff
@@ -80,11 +80,19 @@ pub struct CallError(String);
...
 /// on the same thread, such as GUI toolkits, WebAssembly contexts, or other thread-bound
 /// environments.
-#[derive(Clone)]
 pub struct ThreadBoundRunner<ThreadState> {
     call_channel_tx: tokio::sync::mpsc::Sender<CallRequest<ThreadState>>,
 }
 
+// This is not implemented using derive to remove the implicit bound on `ThreadState: Clone`
+impl<ThreadState> Clone for ThreadBoundRunner<ThreadState> {
+    fn clone(&self) -> Self {
```
coroiu, line 94: Nice!

suggestion: Now that we have a custom implementation, could you add some more explicit code docs to it?

Maybe something like this?
```
/// Makes a clone of the runner handle.
///
/// This creates another handle to the same underlying runner object.
/// The underlying state is not duplicated; all clones refer to the same
/// instance.
```

coroiu (commented):
Just a small comment

crates/bitwarden-threading/src/thread_bound_runner.rs
dani-garcia: Yeah good point, I've added a doc comment like that. Also added a comment to mention that sdk and client managed state arent mutually exclusive to the readme in a763767e6546e3bcf29f20ffddc60dd8b1d900f2 and 6d9a0955cd3e61dc7a287c8f581c1fd4eb6f6ddb.

dani-garcia:
Build errors already in main, fixed in https://github.com/bitwarden/sdk-internal/pull/423

coroiu (approved):
Let's goo
image

PR #306, opened 2025-06-12
`CipherListView` updates

nick-livefront:
## 🎟️ Tracking

Tangential to [PM-22134](https://bitwarden.atlassian.net/browse/PM-22134)

## 📔 Objective

Two updates to ensure parity when migrating from `CipherView` to `CipherListView` on the clients repo
- Added `local_data` so the type propagates to the clients repo and can be set there.
- Added  `CopiableCipherFields` enum + property.
  - In the browser the copy menu / actions are dynamic based on the populated fields on the cipher.  The `CipherListView` doesn't contain these details. This array is filled with the available fields so the extension can still check their availability. When a value is needed the full cipher is decrypted. 
- Added `has_old_attachments` to identify attachments following old encryption methods. This populates UI signaling to the user that the attachment needs to be re-encrypted. 

## 🦮 Reviewer guidelines

[PM-22134]: https://bitwarden.atlassian.net/browse/PM-22134?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/crates/bitwarden-vault/src/cipher/card.rs
+++ b/crates/bitwarden-vault/src/cipher/card.rs
@@ -6,7 +6,7 @@ use serde::{Deserialize, Serialize};
 use tsify_next::Tsify;
 
 use super::cipher::CipherKind;
-use crate::VaultParseError;
+use crate::{cipher::cipher::CopyableCipherFields, Cipher, VaultParseError};
 
 #[derive(Serialize, Deserialize, Debug, Clone)]
 #[serde(rename_all = "camelCase", deny_unknown_fields)]
@@ -143,6 +143,20 @@ impl CipherKind for Card {
 
         Ok(build_subtitle_card(brand, number))
     }
+
+    fn get_copyable_fields(&self, _: &Cipher) -> Vec<CopyableCipherFields> {
+        [
+            self.number
+                .as_ref()
+                .map(|_| CopyableCipherFields::CardNumber),
+            self.code
+                .as_ref()
+                .map(|_| CopyableCipherFields::CardSecurityCode),
+        ]
+        .into_iter()
+        .flatten()
+        .collect()
+    }
 }
 
 /// Builds the subtitle for a card cipher
@@ -178,7 +192,52 @@ fn build_subtitle_card(brand: Option<String>, number: Option<String>) -> String 
 
 #[cfg(test)]
 mod tests {
+    use bitwarden_core::key_management::create_test_crypto_with_user_key;
+    use bitwarden_crypto::SymmetricCryptoKey;
+
+    use crate::{CipherRepromptType, CipherType};
+
     use super::*;
+
+    fn encrypt_test_string(string: &str) -> EncString {
+        let key = SymmetricCryptoKey::try_from("hvBMMb1t79YssFZkpetYsM3deyVuQv4r88Uj9gvYe0+G8EwxvW3v1iywVmSl61iwzd17JW5C/ivzxSP2C9h7Tw==".to_string()).unwrap();
+        let key_store = create_test_crypto_with_user_key(key);
+        let key = SymmetricKeyId::User;
+        let mut ctx = key_store.context();
+
+        string.to_string().encrypt(&mut ctx, key).unwrap()
+    }
+
+    fn create_cipher_for_card(card: Card) -> Cipher {
+        Cipher {
+            id: Some("090c19ea-a61a-4df6-8963-262b97bc6266".parse().unwrap()),
+            organization_id: None,
+            folder_id: None,
+            collection_ids: vec![],
+            r#type: CipherType::Login,
+            key: None,
+            name: encrypt_test_string("My test cipher"),
+            notes: None,
+            login: None,
+            identity: None,
+            card: Some(card),
+            secure_note: None,
+            ssh_key: None,
+            favorite: false,
+            reprompt: CipherRepromptType::None,
+            organization_use_totp: false,
+            edit: true,
+            permissions: None,
+            view_password: true,
+            local_data: None,
+            attachments: None,
+            fields: None,
+            password_history: None,
+            creation_date: "2024-01-01T00:00:00.000Z".parse().unwrap(),
+            deleted_date: None,
+            revision_date: "2024-01-01T00:00:00.000Z".parse().unwrap(),
+        }
+    }
 
     #[test]
     fn test_build_subtitle_card_visa() {
@@ -232,5 +291,41 @@ mod tests {
 
         let subtitle = build_subtitle_card(brand, number);
         assert_eq!(subtitle, "*4444");
+    }
+    #[test]
+    fn test_get_copyable_fields_code() {
+        let card = Card {
+            cardholder_name: None,
+            exp_month: None,
+            exp_year: None,
+            code: Some(encrypt_test_string("123")),
+            brand: None,
+            number: None,
+        };
+
+        let cipher = create_cipher_for_card(card.clone());
+        let copyable_fields = card.get_copyable_fields(&cipher);
+
+        assert_eq!(
+            copyable_fields,
+            vec![CopyableCipherFields::CardSecurityCode]
+        );
+    }
+
+    #[test]
+    fn test_get_copyable_fields_number() {
+        let card = Card {
+            cardholder_name: None,
+            exp_month: None,
+            exp_year: None,
+            code: None,
+            brand: None,
+            number: Some(encrypt_test_string("4242424242424242")),
+        };
+
+        let cipher = create_cipher_for_card(card.clone());
+        let copyable_fields = card.get_copyable_fields(&cipher);
+
+        assert_eq!(copyable_fields, vec![CopyableCipherFields::CardNumber]);
     }
 }
--- a/crates/bitwarden-vault/src/cipher/card.rs
+++ b/crates/bitwarden-vault/src/cipher/card.rs
@@ -195,9 +195,8 @@ mod tests {
     use bitwarden_core::key_management::create_test_crypto_with_user_key;
     use bitwarden_crypto::SymmetricCryptoKey;
 
-    use crate::{CipherRepromptType, CipherType};
-
     use super::*;
+    use crate::{CipherRepromptType, CipherType};
 
     fn encrypt_test_string(string: &str) -> EncString {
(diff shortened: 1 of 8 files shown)


crates/bitwarden-vault/src/cipher/cipher.rs
```diff
@@ -209,6 +227,14 @@ pub struct CipherListView {
...
     pub deleted_date: Option<DateTime<Utc>>,
     pub revision_date: DateTime<Utc>,
+
+    /// Fields on the cipher that are populated. This can be used in the
+    /// UI to determine the visibility of copy actions without needing
+    /// the full cipher details.
+    pub copiable_fields: Vec<CopiableCipherFields>,
+
+    // Only added to the list view for typing purposes, it won't be retrieved from the server
+    pub local_data: Option<LocalDataView>,
```
Hinton: **question:** Is this be able to access `last_used_date` and `last_launched`? I think we should wire it up similar to how it's done on `CipherView`. Currently `local_data` would be empty since it's always empty on `Cipher` but if it starts being populated it will have the same behaviour for `CipherListView` and `CipherView`.

crates/bitwarden-vault/src/cipher/cipher.rs
```diff
@@ -176,6 +176,24 @@ pub enum CipherListViewType {
     SshKey,
 }
 
+/// Available fields on a cipher and can be copied from a the list view in the UI.
+#[derive(Serialize, Deserialize, Debug, PartialEq)]
+#[cfg_attr(feature = "wasm", derive(Tsify), tsify(into_wasm_abi, from_wasm_abi))]
+#[cfg_attr(feature = "uniffi", derive(uniffi::Enum))]
+pub enum CopiableCipherFields {
```
Hinton: **nit**: I think `Copyable` is more commonly used in modern english. If we keep `Copiable` we should add it to the wordlist, `.vscode/settings.json` since it's currently not detected as a valid word.

crates/bitwarden-vault/src/cipher/cipher.rs
```diff
@@ -409,6 +435,102 @@ impl Cipher {
...
+    ) {
+        let any_populated = encs.iter().any(|enc| enc.is_some());
+        if any_populated {
+            fields.push(field_name);
+        }
+    }
+
+    /// Returns a list of copiable field names for this cipher,
+    /// based on the cipher type and populated properties.
+    fn get_copiable_fields(&self) -> Vec<CopiableCipherFields> {
```
Hinton: **issue:** This file actually contains some anti patterns related to subtitle, and this feature is a bit rough to test right now.

**suggestion:** This PR motivated me to revisit subtitle and I created https://github.com/bitwarden/sdk-internal/pull/307. We should lift up the implementations into the cipher type structs in a similar way.

**suggestion:** I think you can get rid of the mutability, especially when extracted to smaller functions you can have something like:

```rust
[
    card.number.map(|_| CopiableCipherFields::CardNumber),
    card.code.map(|_| CopiableCipherFields::CardSecurityCode),
]
.into_iter()
.filter_map(|opt| opt)
.collect()
```

crates/bitwarden-vault/src/cipher/cipher.rs
nick-livefront: Good catch, yes! Those details are needed for sorting in the lists. I updated the logic to match `CipherView`. [eb5a944](https://github.com/bitwarden/sdk-internal/pull/306/commits/eb5a9448041efcd33f5b0bfd993bde42228e18d3)

crates/bitwarden-vault/src/cipher/cipher.rs
nick-livefront: Agreed: [b2694b6](https://github.com/bitwarden/sdk-internal/pull/306/commits/b2694b68e8790c22fd7ceadc61706fab2adf3396)

crates/bitwarden-vault/src/cipher/cipher.rs
nick-livefront: Oh nice! This is way better, I really like the delegation into the sub types. [fe0e02c](https://github.com/bitwarden/sdk-internal/pull/306/commits/fe0e02caead9ceab92151b0e21be9de66a7c2295) I will pull your changes in and refactor to use `getKind` after your PR merges.

On the rust side of things I would consider myself still learning so if there are other pragmatic fixes that I can make here, feel free to point them out! I'm happy to learn on the fly here.

nick-livefront:
@Hinton I rebased off of `main` to utilize your changes and I added another attribute I missed: `has_old_attachments`. 

This is good for another review!

crates/bitwarden-vault/src/cipher/card.rs
```diff
@@ -233,4 +291,40 @@ mod tests {
...
+        let card = Card {
+            cardholder_name: None,
+            exp_month: None,
+            exp_year: None,
+            code: Some(encrypt_test_string("123")),
+            brand: None,
+            number: None,
+        };
+
+        let cipher = create_cipher_for_card(card.clone());
```
Hinton: **suggestion:** I think we can just skip creating a Cipher and call this directly on `card`? Avoids needing all the boilerplate code for creating ciphers.

crates/bitwarden-vault/src/cipher/card.rs
```diff
@@ -233,4 +291,40 @@ mod tests {
         let subtitle = build_subtitle_card(brand, number);
         assert_eq!(subtitle, "*4444");
     }
+    #[test]
+    fn test_get_copyable_fields_code() {
+        let card = Card {
+            cardholder_name: None,
+            exp_month: None,
+            exp_year: None,
+            code: Some(encrypt_test_string("123")),
```
Hinton: **suggestion:** We never decrypt these so we can just use any valid enc string? Saves some more boilerplate in tests.

crates/bitwarden-vault/src/cipher/cipher.rs
```diff
@@ -219,6 +240,16 @@ pub struct CipherListView {
     pub creation_date: DateTime<Utc>,
     pub deleted_date: Option<DateTime<Utc>>,
     pub revision_date: DateTime<Utc>,
+
+    /// Fields on the cipher that are populated. This can be used in the
+    /// UI to determine the visibility of copy actions without needing
+    /// the full cipher details.
```
Hinton: **suggestion (nit):** We can be a bit more succinct without losing context.

```suggestion
    /// Hints for the presentation layer for which fields can be copied.
```

crates/bitwarden-vault/src/cipher/cipher.rs
```diff
@@ -219,6 +240,16 @@ pub struct CipherListView {
...
+    pub copyable_fields: Vec<CopyableCipherFields>,
+
+    /// Indicates if the cipher has old attachments that need to be re-uploaded
+    pub has_old_attachments: bool,
```
Hinton: **suggestion (nit):** Move this up to `attachments`

crates/bitwarden-vault/src/cipher/cipher.rs
```diff
@@ -380,6 +411,38 @@ impl Cipher {
...
+                .ssh_key
+                .as_ref()
+                .map(|ssh_key| ssh_key.get_copyable_fields(self))
+                .unwrap_or_default(),
+            CipherType::SecureNote => self
+                .secure_note
+                .as_ref()
+                .map(|secure_note| secure_note.get_copyable_fields(self))
+                .unwrap_or_default(),
+        }
```
Hinton: **suggestion:** You should be able to use `get_kind` here and simplify it. `get_subtitle` should be a good example:

```suggestion
        self.get_kind()
            .map(|sub| sub.get_copyable_fields())
            .unwrap_or_default()
```

nick-livefront: 🤦 [6d424ea](https://github.com/bitwarden/sdk-internal/pull/306/commits/6d424ea49732fdf4d0caba8e454c1d164bd9bce2)

crates/bitwarden-vault/src/cipher/card.rs
nick-livefront: For all cipher types except `SecureNotes` this is true. The `SecureNotes` type references the `notes` property on the cipher. But you're right, this is just added boilerplate for all other cipher types. I changed the parameter to an `Option` and then removed the boilerplate from the tests that do not utilize it. [a8035b8](https://github.com/bitwarden/sdk-internal/pull/306/commits/a8035b8ee7aa35041953fa174c3d7fbbee898834)

crates/bitwarden-vault/src/cipher/card.rs
nick-livefront: Agreed [8cea293](https://github.com/bitwarden/sdk-internal/pull/306/commits/8cea29354f9dbe3e3499bc22fc600a8771181804)

Is there an easy way to generate encrypted strings? I was using `println` to retrieve an actual encrypted string, maybe there is a better way?

crates/bitwarden-vault/src/cipher/cipher.rs
nick-livefront: [ac7f600](https://github.com/bitwarden/sdk-internal/pull/306/commits/ac7f6004f8e67fada62fa14f2986a0505d9b4ec1)

crates/bitwarden-vault/src/cipher/cipher.rs
nick-livefront: I took this as, move it up next to the attachments property. Let me know if I misread that! [12bf038](https://github.com/bitwarden/sdk-internal/pull/306/commits/12bf038885635767bd11eb97c3275576e38eada7)

crates/bitwarden-vault/src/cipher/card.rs
Hinton: Ah, notes 😠.

crates/bitwarden-vault/src/cipher/card.rs
Hinton: You could copy them from the web apps or cli.

Hinton (approved):
Thanks for iterating on this. Looks good to me.

gbubemismith (approved):
This definitely makes things easier on the client 🚀

addisonbeck (approved):

PR #313, opened 2025-06-17
[PM-22845] Add account key rotation

quexten:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-22845

## 📔 Objective

The decrypted signing key is never available to Typescript, and we need key-rotation to re-encrypt the signing key. This PR implements a part of key-rotation in the crypto crate - specifically re-encrypting the private and signing key.

Long-term, the key-rotation could should be ported at higher levels to the crypto crate.

## ⏰ Reminders before review

- CI builds passed
  team

## 🦮 Reviewer guidelines


crates/bitwarden-crypto/src/key_rotation/mod.rs
```diff
@@ -0,0 +1,49 @@
...
+    /// The user's public key, signed by the signing key
+    signed_public_key: String,
+    // The user's public key, without signature
+    public_key: String,
+    // The user's private key, encrypted with the user key
+    private_key: EncString,
+}
+
+/// Re-encrypts the user's keys with the provided symmetric key for a v2 user.
+pub fn get_v2_rotated_account_keys<Ids: crate::KeyIds>(
```
Hinton: **question:** Key rotation feels very application bound and not like a low level cryptographic primitive? I think this is further enhanced by bitwarden-core just acting as a wrapper around this function.

I presume this was done to avoid making `get_asymetric_key` and `get_signing_key public?

quexten: Yes;

As far as I understand we do not want any further new uses that expose keys out of `bitwarden-crypto`, so this had to go into the crypto crate to achieve that. Is there another pattern we can use here? (Maybe we do allow (non-deprecated) getting wrapped(encrypted) keys out of a context?)

Hinton: I think it's important to consider the public API of crates. In this case we have a `get_v2_rotated_account_keys` which anyone can call for any reason. It provides a way into the "locked" down areas of the Context.

To start with we can move rotate into context that gets rid of the `pub(crate)`. Prefixing it with `dangerous` similar to other operations we don't want to maintain long term also seems reasonable? Ideally we'd find a way to get rid of the response struct but we'd essentially just be duplicating it ... and keeping it where it is seems easier.

Having `dangerous` wrappers to extract keys until we can lift in the whole key rotation also seems fine.

Hinton:
How do we envision the final key rotation to behave? Context generates new keys for you, it stores them in context and exposes encrypted and signed variants?

Currently we'd need `ctx.rotate(private_key_id, signing_key_id)`, having to pass in your private key id and signing key id and possibly your current user id seems clunky and potentially error prone. Ideally we'd have a way to bind keys together?

quexten:
> How do we envision the final key rotation to behave? Context generates new keys for you, it stores them in context and exposes encrypted and signed variants?

> Currently we'd need ctx.rotate(private_key_id, signing_key_id), having to pass in your private key id and signing key id and possibly your current user id seems clunky and potentially error prone. Ideally we'd have a way to bind keys together?

Does final refer to within the context of the current changeset, or in the long run?
For the latter, I think we need proper state management, since currently keys are provided by the host (`clients`, or `ios/android`). So changing keys in context currently needs to be propagated to the host, and then fed back into new sdk instances.

In the long run, with state living in the sdk, calling rotate would:
- In core, start the rotation
- In crypto, perform the rotation, update the context, and return the encrypted versions
- In core, send the keys to API and persist to state
- The client would stay logged in

In the short term, what core does right now, is only done in `clients`. Worse, `clients` needs at least the decrypted userkey returned. Additionally, we don't really support SDK instances just updating their own keys, and an SDK instance is always dependent on the key observables from state. Finally, a key rotation forces a logout currently, so the SDK will be wiped directly afterwards anyways.

So, currently the rotation is intended to be used with a fresh, locally generated SDK instance for the current set of keys. Based on this, the sdk returns a rotated set of keys. The SDK only ever touches the local context, and does not modify key-context outside of local scope. Then, the client takes the result, sends it to the server and logs out: https://github.com/bitwarden/clients/blob/91538819767e46f4b6045d8ff115ddda1aff1843/apps/web/src/app/key-management/key-rotation/user-key-rotation.service.ts#L309

crates/bitwarden-crypto/src/key_rotation/mod.rs
```diff
@@ -0,0 +1,39 @@
...
+    let signed_public_key =
+        ctx.make_signed_public_key(current_user_private_key_id, current_user_signing_key_id)?;
+
+    Ok(RotatedUserKeys {
+        verifying_key: signing_key.to_verifying_key().to_cose(),
+        signing_key: signing_key.to_cose().encrypt_with_key(new_user_key)?,
+        signed_public_key: signed_public_key.into(),
+        public_key: private_key.to_public_key().to_der()?,
+        private_key: private_key.to_der()?.encrypt_with_key(new_user_key)?,
+    })
```
Hinton: **question:** Any reason to not inline this in context? It would mean we can change back the visability of `get_signing_key and `get_asymmetric_key`.

quexten: Hmmm.... To me the key-context should really only implement CRUD for keys, and not more complex operations. As these operations keep growing, the key-context would grow immensely. If the visibility is the only problem, then we can change the function to take key refs instead of id + context, which is OK within the crypto crate (similar to what we do in the sign/sign_detached methods.

In the long-term, I do think we need to figure out the boundaries of what exactly the key-context should handle. (cc @dani-garcia since you designed it initially).

quexten: The functions are now private again on the context, and it passes the key references to the function.

crates/bitwarden-crypto/src/key_rotation/mod.rs
quexten: It is prefixed with dangerous now.

crates/bitwarden-crypto/src/key_rotation/mod.rs
dani-garcia: Yeah I think I agree with your view @quexten, the key context should be kept to CRUD otherwise it's going to get very bloated (it's already quite large as it is).

If we can do key rotation safely without adding it as part of the key context, then I think that's the best option. If we needed to add key rotation for some reason, we'd probably have to think about restructuring the key context a bit organizationally at least.

quexten: So, do I keep it as it is now (with the proxy function on the context that gets the keyids and calls the actual implementation with key references), or do I revert the change / change it to be completely separate from the context?

(The latter would make the "dangerous" get functions visible, but only within the crate, which in my opinion is fine).

Hinton: We could make the functions public to super and move key rotation into the same directory. I think that would limit the visibility while still keeping context CRUD?

I'm fine with moving it back if we don't think `pub(super)` will work.

quexten: Ok, it's moved to the store module now, and the `get` functions are made `pub(super)`.

crates/bitwarden-core/src/key_management/crypto.rs
```diff
@@ -613,6 +616,56 @@ pub fn make_user_signing_keys_for_enrollment(
...
+            private_key: rotated.private_key,
+        }
+    }
+}
+
+/// Gets a set of new wrapped account keys for a user, given a new user key.
+///
+/// In the current implementation, it just re-encrypts any existing keys. This function expects a
+/// user to be a v2 user; that is, they have a signing key, a cose user-key, and a private key
+pub fn get_v2_rotated_account_keys(
```
Hinton: **issue:** Should this be public? We expose it through crypto crypto client which generally means we can mark it as pub(crate)?

crates/bitwarden-crypto/src/keys/asymmetric_crypto_key.rs
```diff
@@ -17,14 +17,14 @@ pub enum PublicKeyEncryptionAlgorithm {
...
-#[derive(Clone)]
+#[derive(Clone, PartialEq, Debug)]
 pub(crate) enum RawPublicKey {
     RsaOaepSha1(RsaPublicKey),
 }
 
 /// Public key of a key pair used in a public key encryption scheme. It is used for
 /// encrypting data.
-#[derive(Clone)]
+#[derive(Clone, PartialEq, Debug)]
```
Hinton: **question:** Are we concerned about implementing `Debug` here? We would potentially print the primes, or is this fine because they are "only" public keys?

Hinton (commented):
One small blocker and a question. I think we're good otherwise but I'd appreciate @dani-garcia thoughts on the store.

crates/bitwarden-crypto/src/keys/asymmetric_crypto_key.rs
quexten: The private-key's primes are not in the pubkey, just the modulus (not prime), and exponent, so no direct concern there.

Still, that is a good point. Usually this would be fine, but our current "RotateableKeySets" (TDE, Passkey-PRF login) rely on the assumption that public keys are secret (HPKE in `mode_psk` would allow to weaken this assumption eventually, for a future redesign of this feature...). Public keys (the modulus) leaking would allow the server / MITM to inject/overwrite userkeys.

So in this case, I'll remove them to make sure.

crates/bitwarden-core/src/key_management/crypto.rs
quexten: I agree, missed this one, and it is fixed.

crates/bitwarden-core/src/auth/auth_request.rs
```diff
@@ -91,7 +91,9 @@ pub(crate) fn approve_auth_request(
     client: &Client,
     public_key: String,
 ) -> Result<UnsignedSharedKey, ApproveAuthRequestError> {
-    let public_key = AsymmetricPublicCryptoKey::from_der(&STANDARD.decode(public_key)?)?;
+    let public_key = AsymmetricPublicCryptoKey::from_der(&SpkiPublicKeyBytes::from(
```
dani-garcia, line 94: I've noticed that we do `Bytes::from(STANDARD.decode())` very often, I wonder if it would be simpler to have a `Bytes::from_b64()` function for that. Not as part of this PR though, just a possible future improvement.

crates/bitwarden-core/src/key_management/crypto.rs
```diff
@@ -613,6 +616,56 @@ pub fn make_user_signing_keys_for_enrollment(
...
+    let key_store = client.internal.get_key_store();
+    let ctx = key_store.context();
+
+    dangerous_get_v2_rotated_account_keys(
+        &SymmetricCryptoKey::try_from(user_key)?,
+        AsymmetricKeyId::UserPrivateKey,
+        SigningKeyId::UserSigningKey,
+        &ctx,
+    )
+    .map(|rotated| rotated.into())
```
dani-garcia: Small non-blocking stylistic nit:
```suggestion
    .map(Into::into)
```

crates/bitwarden-core/src/key_management/crypto.rs
```diff
@@ -613,6 +616,56 @@ pub fn make_user_signing_keys_for_enrollment(
     })
 }
 
+/// A rotated set of account keys for a user
+#[derive(Serialize, Deserialize, Debug, JsonSchema)]
+#[serde(rename_all = "camelCase", deny_unknown_fields)]
+#[cfg_attr(feature = "uniffi", derive(uniffi::Record))]
+#[cfg_attr(feature = "wasm", derive(Tsify), tsify(into_wasm_abi, from_wasm_abi))]
+pub struct RotateUserKeysResponse {
```
dani-garcia, line 624: Question, Is this duplicate struct only needed because we don't have FFI support for the `Bytes` types yet?

Hinton (dismissed):

dani-garcia (dismissed):
LGTM, just some non blocking comments

crates/bitwarden-core/src/key_management/crypto.rs
quexten: Yes. I felt that FFI for bytes is outside of scope for this PR. I think another path to drop it is if we move more of key rotation to the sdk.

dani-garcia (approved):

PR #316, opened 2025-06-18
Refactor: Use serde instead of TryFrom<Vec<u8>> traits in IPC

coroiu:
## 📔 Objective

Refactoring the IPC to use serde instead. Putting this in a separate PR to make reviewing easier

## ⏰ Reminders before review

- CI builds passed
  team

## 🦮 Reviewer guidelines


crates/bitwarden-ipc/src/serde_utils.rs
```diff
@@ -0,0 +1,18 @@
...
+use serde::{Deserialize, Serialize};
+
+pub(crate) type SerializeError = serde_json::Error;
+pub(crate) type DeserializeError = serde_json::Error;
+
+pub(crate) fn to_vec<T: Serialize>(value: &T) -> Result<Vec<u8>, serde_json::Error> {
+    serde_json::to_vec(value)
+}
+
+pub(crate) fn from_slice<T: for<'de> Deserialize<'de>>(
```
dani-garcia: You should be able to use `DeserializeOwned` which is the same thing and slightly easier to understand: https://docs.rs/serde/latest/serde/de/trait.DeserializeOwned.html

coroiu: You are too quick haha

dani-garcia: 😎

crates/bitwarden-ipc/src/ipc_client.rs
```diff
@@ -292,19 +292,15 @@ where
...
-        <Request as TryInto<Vec<u8>>>::Error: std::fmt::Display,
-        <Request::Response as TryFrom<Vec<u8>>>::Error: std::fmt::Display,
+        Request: RpcRequest,
     {
         let request_id = uuid::Uuid::new_v4().to_string();
-        let mut response_subscription: IpcClientTypedSubscription<RpcResponseMessage> =
-            self.subscribe_typed().await?;
+        let mut response_subscription: IpcClientTypedSubscription<
+            IncomingRpcResponseMessage<Request::Response>,
+        > = self.subscribe_typed().await?;
```
dani-garcia: Non blocking nit, but if you move the type to the function, you can let the compiler infer some of the types:
```suggestion
        let mut response_subscription = self
            .subscribe_typed::<IncomingRpcResponseMessage<_>>()
            .await?;
```

crates/bitwarden-ipc/src/ipc_client.rs
```diff
@@ -315,7 +311,9 @@ where
         }
         .try_into()
         .map_err(
```
dani-garcia: I think this map_err can be simplified a bit, the error type is always the same:
```
        .map_err(|e: serde_utils::DeserializeError| {
            RequestError::<_>::RpcError(RpcError::RequestSerializationError(e.to_string()))
```

crates/bitwarden-ipc/src/ipc_client.rs
```diff
@@ -389,7 +375,7 @@ where
                 }
                 .try_into()
                 .map_err(
```
dani-garcia: Same for this error:
```
                .map_err(|e: serde_utils::SerializeError| HandleError::Serialize(e.to_string()))?;
```

crates/bitwarden-ipc/src/ipc_client.rs
```diff
@@ -367,19 +356,16 @@ where
...
 
-                let response = handlers
-                    .handle(&request.request_type, request.request)
-                    .await;
+                let response = handlers.handle(&request).await;
 
-                let response_message = RpcResponseMessage {
-                    request_id: request.request_id,
-                    request_type: request.request_type,
+                let response_message = OutgoingRpcResponseMessage {
```
dani-garcia, line 362: We're only ever using `OutgoingRpcResponseMessage` with `request_id` and `request_type` as `&str` and we always immediately serialize it. I wonder if we could just put a lifetime on it to avoid cloning the values.

crates/bitwarden-ipc/src/ipc_client.rs
```diff
@@ -454,9 +439,11 @@ where
...
     ) -> Result<TypedIncomingMessage<Payload>, TypedReceiveError> {
         let received = self.0.receive(cancellation_token).await?;
-        received
-            .try_into()
-            .map_err(|e: TryFromError| TypedReceiveError::Typing(e.to_string()))
+        received.try_into().map_err(
+            |e: <TypedIncomingMessage<Payload> as TryFrom<IncomingMessage>>::Error| {
+                TypedReceiveError::Typing(e.to_string())
+            },
+        )
```
dani-garcia:
```suggestion
        received
            .try_into()
            .map_err(|e: serde_utils::DeserializeError| TypedReceiveError::Typing(e.to_string()))
```

dani-garcia (commented):
Looks pretty good to me, erased serde seems to have improved the generic bounds of this a lot. I only have some small nits here and there.

I wonder why rust doesn't seem able to correctly guess the error types in all the .try_into().map_err() functions, that's annoying.

crates/bitwarden-ipc/src/ipc_client.rs
coroiu: Nice, fixed!

crates/bitwarden-ipc/src/ipc_client.rs
coroiu: I thought I'd try to keep these errors a little bit more dynamic since changing how serialization happens might change the error type. But on the other hand, it's probably just YAGNI. I already have the `DeserializeError` alias, probably never gonna need more.

Fixed!

crates/bitwarden-ipc/src/ipc_client.rs
coroiu: Changed to `&'a str` 👍

dani-garcia (approved):

PR #318, opened 2025-06-19
Lift in user_id into login method

Hinton:
## 📔 Objective

Refactors `user_id` to be part of the Login Method. This seems like a more logical place to put it since in a fully SDK based flow you can always get the id as part of the access token. And it decreases us from three states to two.

## ⏰ Reminders before review

- CI builds passed
  team

## 🦮 Reviewer guidelines

### DIFF
--- a/crates/bitwarden-core/src/auth/auth_client.rs
+++ b/crates/bitwarden-core/src/auth/auth_client.rs
@@ -2,6 +2,8 @@
 use bitwarden_crypto::{
     CryptoError, DeviceKey, EncString, Kdf, TrustDeviceResponse, UnsignedSharedKey,
 };
+#[cfg(feature = "internal")]
+use uuid::Uuid;
 
 #[cfg(feature = "secrets")]
 use crate::auth::login::{login_access_token, AccessTokenLoginRequest, AccessTokenLoginResponse};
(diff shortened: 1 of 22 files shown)


coroiu (commented):
failing builds

dani-garcia (commented):
I feel like moving the user_uuid into login_method defeats a bit the initial point of the change, which was to have an immutable user id associated to an SDK client instance. Because of compatibility we couldn't just include it as part of client settings, so we changed it to be set-once only.

By moving it into the LoginMethod we're back to a situation where a client could be modified to be for a different user. Maybe what we really need to do is to also allow setting login_method once.

Hinton:
I think the more we can couple these things and make them immutable the better. I would be for changing LoginMethod to be once lock, there might be edge cases where you want to move between login methods i.e. logging in using username but transitioning to TDE?

dani-garcia:
I think making as many fields immutable or initialized during construction makes sense, but there may be a lot of edge cases since a lot of things can change. 

For example, on mobile a lot of calls to `init_crypto` also set the login_method, so we'd need to be careful as mobile calls `init_crypto` multiple times if the user inputs the wrong password (see https://github.com/bitwarden/sdk-internal/pull/321, which is trying to solve some similar issues on mobile).

We also store email and KDF settings in LoginMethod, which potentially would need changing during key rotation or email change, though we're not handling those right now.

MGibson1:
@Hinton can you update or close this PR?

PR #321, opened 2025-06-19
Ignore the error when setting the user id twice to the same value

dani-garcia:
## 📔 Objective

The mobile clients will call `init_crypto` multiple times on the same client if the user types the wrong password/pin.

The simplest fix here is to make it so that `init_user_id` only errors when there is an attempted UUID change, and trying to set the same UUID is just ignored. The UUID will only be set the very first call either way.

Ideally we'd do as mentioned in  #278, and just provide this ID during initialization once we have a pure/user-bound client split, but that's a bigger API break.

## ⏰ Reminders before review

- CI builds passed
  team

## 🦮 Reviewer guidelines


coroiu (approved):
Clever fix 👍

PR #322, opened 2025-06-20
[PM-22861] Account security state

quexten:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-22861

## 📔 Objective

This adds the initial version of account security state to the sdk, on init of the sdk state, and for enrollment to v2 crypto.

Account security state is an attestation to a specific version and featureset that is enabled. The version is used to prevent cryptographic downgrades of the account state.

This will prevent downgrades in the following ways:
- Arbitrary old states cannot be created by the server, because the state is signed for a user by the user's signing key, the server can only record entire states
- A signed in session will not accept a sync with an older state
- Some sign in mechanisms will not accept a sync with an older state (login-with-device 2.0, PRF 2.0, both still to be specified)

A security version then is used to lock down specific features. If a feature is to be disabled, then a migrator is needed. After migrating the account not to use the retired feature, the version is incremented and any subsequent syncs with that format included are rejected.

This PR stores the account security state on the client for v2 users. However, besides validating that the state is correctly signed, it is not yet used.

## ⏰ Reminders before review

- CI builds passed
  team

## 🦮 Reviewer guidelines


Hinton:
Security state seems like it would fit better in the bitwarden-core crate than crypto, similar to EncryptionSettings.

quexten:
> Security state seems like it would fit better in the bitwarden-core crate than crypto, similar to EncryptionSettings.

The reason it lives in `crypto` is because the namespace definition lives in `crypto`, and we (so far) commented which specific struct a namespace is for. We can not do that if the struct lives somewhere else. We want to make sure the namespace is only ever used with one struct, to avoid any possibility for cross-protocol attacks where multiple structs are valid for the serialized payload that is signed.

That does make me wonder if namespace should really be a generic parameter, similar to how key context is done with the actual key ids being defined in core.

Hinton:
This seems to only be enforced by convention though, there's nothing in the type system providing compiler level safety?

> That does make me wonder if namespace should really be a generic parameter, similar to how key context is done with the actual key ids being defined in core.

I think so, or leaving the code signing concepts in crypto but lifting out the application logic. Or splitting up the crypto crate into multiple more focused areas where some provide low level utilities but other mixing domain logic.

quexten:
> This seems to only be enforced by convention though, there's nothing in the type system providing compiler level safety?

Yeah, that is true, currently this relies on people reading the documentation.

I think we can change this in the long term to:
- Remove namespace as a separate parameter
- Make signable objects have a trait that forces them to provide a namespace
  - This is similar to what we do for content formats

This does not provide complete safety, but at least makes accidental misuse less likely.
I'm open to better suggestions. We may be able to use procmacros to provide a more safe solution that forces a namespace variant to always have one associated struct, but I'm not too familiar here.

quexten:
> I think so, or leaving the code signing concepts in crypto but lifting out the application logic. Or splitting up the crypto crate into multiple more focused areas where some provide low level utilities but other mixing domain logic.

I think resorting the crypto crate is something we wanted to do anyways after "user crypto v2" is done; I recall we had a DM about this with you and Dani, for how this would look. I'd like to defer resorting the crate for after the bigger changes here are done however. Are we OK with putting this in `bitwarden-crypto` for now, and then looking at this again after the resort is done? (Either moving it into a different place within `crypto` or moving it out to `core`?

Hinton:
I think I'd prefer to just put it in core right now and not have a link in the documentation. It's weird that we have crypto errors that are only thrown by other crates. We can always lift it in later if we think it fits better.

quexten:
Moving this to draft until the preceeding PR is merged.

crates/bitwarden-core/src/key_management/crypto.rs
```diff
@@ -574,45 +583,97 @@ pub(super) fn verify_asymmetric_keys(
...
 }
 
-/// Makes a new set of signing keys for a user, which should only be done during
-/// once. This also signs the public key with the signing key
-/// and returns the signed public key.
-pub fn make_user_signing_keys_for_enrollment(
+/// Initializes the user's cryptographic state for v2 users.
+/// If the client already contains a v1 user, then this user's private-key will be
+/// re-used.
+pub(crate) fn make_keys_for_user_crypto_v2(
```
quexten: This moves more to the SDK than had been previously in the SDK. Since this API is not used yet a breaking change seemed OK.

crates/bitwarden-core/src/client/encryption_settings.rs
```diff
@@ -50,49 +54,85 @@ impl EncryptionSettings {
...
+
+            // Everything MUST decrypt.
+            let signing_key: Vec<u8> = signing_key.decrypt_with_key(&user_key)?;
+            let signing_key = SigningKey::from_cose(&CoseKeyBytes::from(signing_key))
+                .map_err(|_| EncryptionSettingsError::InvalidSigningKey)?;
+            let private_key: Vec<u8> = private_key.decrypt_with_key(&user_key)?;
+            let private_key =
+                AsymmetricCryptoKey::from_der(&Pkcs8PrivateKeyBytes::from(private_key))
+                    .map_err(|_| EncryptionSettingsError::InvalidPrivateKey)?;
+            let _security_state: SecurityState = security_state
```
quexten: This is currently only decrypted to validate that the state is correct, and present. But it is not used yet.

crates/bitwarden-core/src/client/encryption_settings.rs
```diff
@@ -50,49 +56,85 @@ impl EncryptionSettings {
...
+                .verify_and_unwrap(&signing_key.to_verifying_key())
+                .map_err(|_| EncryptionSettingsError::InvalidSecurityState)?;
+
+            #[allow(deprecated)]
```
quexten: Making the entire block as allow deprecated since each of the context functions requires this. This will be removed when we remove external access to these set functions.

Question for reviewers: Would it make sense to split these into a non-deprecated `pub(crate)` and to a deprecated `pub fn dangerous_...` similar to the get functions?

crates/bitwarden-crypto/src/lib.rs
```diff
@@ -36,6 +36,9 @@ pub use store::{
 };
 mod cose;
 pub use cose::CoseSerializable;
+/// The `SecurityState` module provides functionality to cryptographically attest to which features
+/// are allowed to be used for a user.
+pub mod security_state;
```
dani-garcia: Small nit: On this file we seem to be doing private module and then `pub use` for the types that we want public.

We also don't need a comment for the module as the module already has a mod level comment using `//!`

crates/bitwarden-crypto/src/security_state.rs
```diff
@@ -0,0 +1,200 @@
...
+    }
+
+    /// Returns the version of the security state
+    pub fn version(&self) -> u64 {
+        self.version
+    }
+}
+
+/// Signs the `SecurityState` with the provided signing key ID from the context.
+pub fn sign<Ids: KeyIds>(
```
dani-garcia: This takes `SecurityState` as parameter, should it not be part of the `impl SecurityState` block?

crates/bitwarden-crypto/src/store/context.rs
```diff
@@ -308,6 +309,15 @@ impl<Ids: KeyIds> KeyStoreContext<'_, Ids> {
         Ok(key_id)
     }
 
+    /// Makes a new asymmetric encryption key using the current default algorithm, and stores it in
+    /// the context
+    pub fn make_asymmetric_key(&mut self, key_id: Ids::Asymmetric) -> Result<Ids::Asymmetric> {
```
dani-garcia, line 314: We should have a followup to rename `generate_symmetric_key` to match the asymmetric and signing variants, so they all match

crates/bitwarden-core/src/client/encryption_settings.rs
```diff
@@ -50,49 +56,85 @@ impl EncryptionSettings {
         user_key: SymmetricCryptoKey,
         private_key: EncString,
         signing_key: Option<EncString>,
+        security_state: Option<SignedSecurityState>,
```
dani-garcia: Suggestion: If `signing_key` and `security_state` will both always be defined when using a chacha key, I wonder if it would simplify the logic if instead of having two optional values, we combined both parameters into a struct and made that optional instead.

crates/bitwarden-core/src/client/encryption_settings.rs
```diff
@@ -50,49 +56,85 @@ impl EncryptionSettings {
...
+
+            // Both of these are required for v2 users
+            let signing_key = signing_key.ok_or(EncryptionSettingsError::Crypto(
+                CryptoError::SecurityDowngrade("Signing key is required for v2 users".to_string()),
+            ))?;
+            let security_state = security_state.ok_or(EncryptionSettingsError::Crypto(
+                CryptoError::SecurityDowngrade(
+                    "Security state is required for v2 users".to_string(),
+                ),
+            ))?;
```
dani-garcia: We don't need to specify `EncryptionSettingsError::Crypto`  when `CryptoError` implements `Into<EncryptionSettingsError>`:
```suggestion
            let signing_key = signing_key.ok_or(CryptoError::SecurityDowngrade(
                "Signing key is required for v2 users".to_string(),
            ))?;
            let security_state = security_state.ok_or(CryptoError::SecurityDowngrade(
                "Security state is required for v2 users".to_string(),
            ))?;
```

crates/bitwarden-core/src/client/encryption_settings.rs
```diff
@@ -50,49 +56,85 @@ impl EncryptionSettings {
...
-            })
-            .transpose()?;
-
-        // FIXME: [PM-18098] When this is part of crypto we won't need to use deprecated methods
-        #[allow(deprecated)]
-        {
-            let mut ctx = store.context_mut();
-            ctx.set_symmetric_key(SymmetricKeyId::User, user_key)?;
-            if let Some(private_key) = private_key {
+        use crate::key_management::{AsymmetricKeyId, SymmetricKeyId};
```
dani-garcia: This method is getting pretty large, it's probably a good idea to split it into two internal methods, for v1 and v2 users

crates/bitwarden-core/src/key_management/crypto.rs
```diff
@@ -574,45 +583,97 @@ pub(super) fn verify_asymmetric_keys(
...
+    let signing_key = SigningKey::make(SignatureAlgorithm::Ed25519);
+    #[allow(deprecated)]
+    ctx.set_signing_key(temporary_signing_key_id, signing_key.clone())?;
+
+    // Sign existing public key
+    let signed_public_key = ctx.make_signed_public_key(private_key_id, temporary_signing_key_id)?;
+    #[allow(deprecated)]
+    let public_key = ctx
+        .dangerous_get_asymmetric_key(private_key_id)?
+        .to_public_key();
```
dani-garcia: We already get the private key in line 633, right?
```suggestion
    let public_key = private_key.to_public_key();
```

dani-garcia (commented):
Some initial review comments

crates/bitwarden-core/src/client/encryption_settings.rs
quexten: Nice, applied!

crates/bitwarden-crypto/src/security_state.rs
quexten: Yeah, I had assumed the struct needed to be generic to `Ids` if the impl was generic, but that is not the case in Rust. Moved.

Also this was the reason for the `security_state` mod being public, since i'd rather have callers use `security_state::sign` than just `sign`. But since we impl on the struct now, we can just `security_state.sign`.

crates/bitwarden-crypto/src/lib.rs
quexten: Fixed (https://github.com/bitwarden/sdk-internal/pull/322#discussion_r2185291175).

crates/bitwarden-crypto/src/store/context.rs
quexten: Agreed, I had seen this but kept it up for later follow-up to now grow this PR.

crates/bitwarden-core/src/client/encryption_settings.rs
quexten: Split into two functions now.

crates/bitwarden-core/src/client/encryption_settings.rs
quexten: I've moved the entire set of keys into enum variants now. The type of symmetric key is also enforced by the type system.

quexten: Note this required an `allow(large_enum_variant)`. I feel that this is OK since we only have one instance of this enum in the sdk.

crates/bitwarden-core/Cargo.toml
```diff
@@ -45,6 +45,7 @@ rand = ">=0.8.5, <0.9"
 reqwest = { workspace = true }
 schemars = { workspace = true }
 serde = { workspace = true }
+serde_bytes = ">=0.11.17, <0.12"
```
quexten: Do we need to move this to the workspace? Both crypto and core use it at this point.

crates/bitwarden-core/src/client/encryption_settings.rs
```diff
@@ -72,27 +129,62 @@ impl EncryptionSettings {
...
-            }
+        Ok(())
+    }
+
+    #[cfg(feature = "internal")]
+    fn init_v2(
+        user_key: XChaCha20Poly1305Key,
+        private_key: EncString,
+        signing_key: Option<EncString>,
+        security_state: Option<SignedSecurityState>,
```
dani-garcia: These two are currently optional, but we always pass Some() into them, we should remove the Option if it's not needed, which should allow us to remove the downgrade error handling and just let the type system deal with it.

crates/bitwarden-core/src/client/encryption_settings.rs
```diff
@@ -72,27 +129,62 @@ impl EncryptionSettings {
...
-            })
-            .transpose()?;
 
         // FIXME: [PM-18098] When this is part of crypto we won't need to use deprecated methods
         #[allow(deprecated)]
         {
             let mut ctx = store.context_mut();
             ctx.set_symmetric_key(SymmetricKeyId::User, user_key)?;
             if let Some(private_key) = private_key {
+                use crate::key_management::AsymmetricKeyId;
```
dani-garcia: We're already importing this above, I don't think it's needed here

crates/bitwarden-core/src/client/encryption_settings.rs
dani-garcia: I'd probably add a code comment on top of that line mentioning this, to avoid someone accidentally deleting it thinking it's not in use.

crates/bitwarden-core/src/client/encryption_settings.rs
dani-garcia: Looks good to me, yeah that warning can be safely ignored for this case, I think.

crates/bitwarden-core/Cargo.toml
dani-garcia: Yeah I think that sounds like a good idea here.

crates/bitwarden-core/src/client/internal.rs
```diff
@@ -230,8 +238,35 @@ impl InternalClient {
...
         private_key: EncString,
         signing_key: Option<EncString>,
+        security_state: Option<SignedSecurityState>,
     ) -> Result<(), EncryptionSettingsError> {
-        EncryptionSettings::new_decrypted_key(user_key, private_key, signing_key, &self.key_store)?;
+        match user_key {
+            SymmetricCryptoKey::Aes256CbcHmacKey(ref user_key) => {
+                EncryptionSettings::new_decrypted_key(
+                    AccountEncryptionKeys::V1 {
+                        user_key: user_key.clone(),
```
dani-garcia, line 270: Question: I think this clone is because we can't destructure  a `SymmetricCryptoKey` as it's deriving ZeroizeOnDrop which implements Drop?

Probably better off for a future change, but knowing that all the variants implement ZeroizeOnDrop themselves, maybe we can remove it from `SymmetricCryptoKey` itself and replace it with something like this, which would allow you to destructure here and avoid the clones:
```rust
/// All the variants implement ZeroizeOnDrop+Drop themselves,
/// so we don't need to implement Drop again here
impl ZeroizeOnDrop for SymmetricCryptoKey {}
const _: () = {
    fn assert_zeroize_on_drop<T: zeroize::ZeroizeOnDrop>() {}
    fn assert_all() {
        assert_zeroize_on_drop::<Aes256CbcKey>();
        assert_zeroize_on_drop::<Aes256CbcHmacKey>();
        assert_zeroize_on_drop::<XChaCha20Poly1305Key>();
    }
};
```

dani-garcia (commented):
Nice, yeah I think the AccountEncryptionKeys makes it much more clear what options are expected here, a couple of more comments I have, but other than that things are looking pretty good.

crates/bitwarden-core/src/client/encryption_settings.rs
quexten: Whoops, this was non-intentional. I forgot to make them non-optional after splitting.

quexten: Resolved

crates/bitwarden-core/src/client/encryption_settings.rs
quexten: Added a comment

crates/bitwarden-core/src/client/internal.rs
quexten: Yeah, that makes sense. I'll take a note to look at this later on.

quexten:
Moving back to draft to clean up the interface a bit more.

crates/bitwarden-core/src/key_management/mod.rs
```diff
@@ -18,6 +18,9 @@ mod crypto_client;
 #[cfg(feature = "internal")]
 pub use crypto_client::CryptoClient;
 
+mod security_state;
+pub use security_state::{SecurityState, SignedSecurityState};
```
dani-garcia: These can probably be internal-only
```suggestion
#[cfg(feature = "internal")]
mod security_state;
#[cfg(feature = "internal")]
pub use security_state::{SecurityState, SignedSecurityState};
```

crates/bitwarden-crypto/src/error.rs
```diff
@@ -69,6 +69,9 @@ pub enum CryptoError {
 
     #[error("Encoding error, {0}")]
     EncodingError(#[from] EncodingError),
+
+    #[error("Uninitialized error")]
+    CryptoStateError(CryptoStateError),
```
dani-garcia: The error message doesn't match the error type here, and we should add `#[from]` so the error has the appropiate source.

crates/bitwarden-crypto/src/error.rs
```diff
@@ -77,6 +80,16 @@ pub enum UnsupportedOperation {
     EncryptionNotImplementedForKey,
 }
 
+/// Signifies that the state is invalid from a cryptographic perspective, such as a required
+/// security value missing, or being invalid
+#[derive(Debug, Error)]
+pub enum CryptoStateError {
+    /// The security state is not present, but required for this user. V2 users must always
+    /// have a security state, V1 users cannot have a security state.
+    #[error("Security state is not set to state")]
```
dani-garcia: This error message is a bit confusing, would it be better to have `Security state is required but missing`?

crates/bitwarden-crypto/src/store/key_rotation.rs
```diff
@@ -19,27 +22,25 @@ pub struct RotatedUserKeys {
...
-            .encrypt_with_key(new_user_key)?,
-        signed_public_key: signed_public_key.into(),
+        signing_key: current_signing_key.to_cose().encrypt_with_key(&user_key)?,
+        signed_public_key,
         public_key: current_private_key.to_public_key().to_der()?,
-        private_key: current_private_key
-            .to_der()?
-            .encrypt_with_key(new_user_key)?,
+        private_key: current_private_key.to_der()?.encrypt_with_key(&user_key)?,
     })
```
dani-garcia: Small suggestion, we're calling to_public_key() twice, so that can be reused. Also if we move user_key at the bottom, we can avoid the clone.
```rust
let public_key = current_private_key.to_public_key();
let signed_public_key =
    SignedPublicKeyMessage::from_public_key(&public_key)?.sign(current_signing_key)?;

Ok(RotatedUserKeys {
    verifying_key: current_signing_key.to_verifying_key().to_cose(),
    signing_key: current_signing_key.to_cose().encrypt_with_key(&user_key)?,
    signed_public_key,
    public_key: public_key.to_der()?,
    private_key: current_private_key.to_der()?.encrypt_with_key(&user_key)?,
    user_key,
})
```

crates/bitwarden-core/src/key_management/security_state.rs
```diff
@@ -0,0 +1,199 @@
...
+    }
+}
+
+impl FromStr for SignedSecurityState {
+    type Err = EncodingError;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        let bytes = STANDARD
+            .decode(s)
+            .map_err(|_| EncodingError::InvalidCborSerialization)?;
```
dani-garcia: This error is mentioning cbor serialization, but this is using base64 and then cose

crates/bitwarden-core/src/key_management/crypto.rs
```diff
@@ -652,18 +674,43 @@ impl From<RotatedUserKeys> for RotateUserKeysResponse {
...
+        .read()
+        .expect("RwLock is not poisoned")
+        .to_owned();
+    let security_state =
+        signed_security_state
+            .as_ref()
+            .to_owned()
+            .ok_or(CryptoError::CryptoStateError(
+                CryptoStateError::MissingSecurityState,
+            ))?;
```
dani-garcia: These two can be combined into one, I think:
```rust
    let security_state = client
        .internal
        .security_state
        .read()
        .expect("RwLock is not poisoned")
        .as_ref()
        .to_owned()
        .ok_or(CryptoError::CryptoStateError(
            CryptoStateError::MissingSecurityState,
        ))?;
```

crates/bitwarden-core/src/client/internal.rs
```diff
@@ -80,6 +86,8 @@ pub struct InternalClient {
     pub(crate) external_client: reqwest::Client,
 
     pub(super) key_store: KeyStore<KeyIds>,
+    #[cfg(feature = "internal")]
+    pub(crate) security_state: RwLock<Arc<Option<SecurityState>>>,
```
dani-garcia: We don't need the Arc here, we never hold onto the value for long, as we always make a clone of it. `RwLock<Option<SecurityState>>` would do just fine.

crates/bitwarden-core/src/key_management/crypto.rs
```diff
@@ -574,76 +584,88 @@ pub(super) fn verify_asymmetric_keys(
...
+) -> Result<UserCryptoV2KeysResponse, CryptoError> {
     let key_store = client.internal.get_key_store();
     let mut ctx = key_store.context();
 
-    // Make new keypair and sign the public key with it
-    let signature_keypair = SigningKey::make(SignatureAlgorithm::Ed25519);
-    let temporary_signature_keypair_id = SigningKeyId::Local("temporary_key_for_rotation");
+    let temporary_user_key_id = SymmetricKeyId::Local("temporary_user_key");
+    let temporary_signing_key_id = SigningKeyId::Local("temporary_signing_key");
+    let temporary_private_key_id = AsymmetricKeyId::Local("temporary_private_key");
```
dani-garcia: This could be simplified a bit with #274, we might want to try to prioritize it.

crates/bitwarden-crypto/src/error.rs
quexten: Apologies, this was left over during a rename.

crates/bitwarden-core/src/key_management/security_state.rs
quexten: Yeah, this function only does base64 deserialization, for cose it doesn't even deserialize but just creates the typed bytes. I'll add a new variant for base64 deserialization issues.

crates/bitwarden-core/src/key_management/crypto.rs
quexten: Nice, much more readable!

quexten: @dani-garcia Had to drop the `as_ref()` too from the example to fix the lifetime / borrowing issues.

crates/bitwarden-core/src/key_management/crypto.rs
quexten: I'll try to pull this in as one of the next tech debt items to work on. The developer ergonomics will improve by a lot.

crates/bitwarden-crypto/src/util.rs
```diff
@@ -53,9 +53,16 @@ pub fn pbkdf2(password: &[u8], salt: &[u8], rounds: u32) -> [u8; PBKDF_SHA256_HM
...
+pub struct FromStrVisitor<T>(std::marker::PhantomData<T>);
 impl<T> FromStrVisitor<T> {
-    pub(crate) fn new() -> Self {
+    /// Create a new `FromStrVisitor` for the given type.
+    pub fn new() -> Self {
+        Self::default()
+    }
+}
+impl<T> Default for FromStrVisitor<T> {
+    fn default() -> Self {
```
Hinton, line 65: non-blocking: This doesn't quite fit `bitwarden-crypto` and should probably be lifted out to another crate.

crates/bitwarden-crypto/src/store/key_rotation.rs
```diff
@@ -19,27 +22,26 @@ pub struct RotatedUserKeys {
 
 /// Re-encrypts the user's keys with the provided symmetric key for a v2 user.
```
Hinton: **question:** Is this comment still accurate now that we don't have a `new_user_key`?

crates/bitwarden-crypto/src/error.rs
```diff
@@ -69,6 +69,9 @@ pub enum CryptoError {
 
     #[error("Encoding error, {0}")]
     EncodingError(#[from] EncodingError),
+
+    #[error("Crypto state error, {0}")]
+    CryptoStateError(#[from] CryptoStateError),
```
Hinton: **question:** Do we envision every crypto operation to be throwing these errors? If not we might want to flip and have `CryptoStateError` include `CryptoError`.

We're trying to make errors more specific and have smaller enums where possible.

crates/bitwarden-core/src/key_management/security_state.rs
```diff
@@ -0,0 +1,199 @@
...
+
+impl schemars::JsonSchema for SignedSecurityState {
+    fn schema_name() -> String {
+        "SecurityState".to_string()
+    }
+
+    fn json_schema(generator: &mut schemars::r#gen::SchemaGenerator) -> schemars::schema::Schema {
+        generator.subschema_for::<String>()
+    }
+}
```
Hinton: **question:** Will this be exposed to SM SDK? If not I don't think we need to derive schema here.

`Schemars` is used exclusively to produce language bindings for secrets manager sdk, https://github.com/bitwarden/sdk-sm/tree/main?tab=readme-ov-file#schemas.

crates/bitwarden-core/src/client/internal.rs
```diff
@@ -217,11 +225,15 @@ impl InternalClient {
         user_key: EncString,
         private_key: EncString,
         signing_key: Option<EncString>,
+        security_state: Option<SignedSecurityState>,
```
Hinton: **suggestion:** We are starting to have quite a few arguments to these methods which are similar in type, this could easily be mixed up.

Since all 3 functions share what is essentially the crypto state it might be useful to create a struct and pass it in everywhere.

```rust
struct KeyState {
    private_key: EncString,
    signing_key: Option<EncString>,
    security_state: Option<SignedSecurityState>,
}

pub(crate) fn initialize_user_crypto_master_key(
    &self,
    master_key: MasterKey,
    user_key: EncString,
    key_state: KeyState,
```

You could also expose this struct in `InitUserCryptoRequest` and avoid having to map it.

crates/bitwarden-core/src/key_management/crypto.rs
```diff
@@ -574,76 +584,88 @@ pub(super) fn verify_asymmetric_keys(
...
+    security_state: SignedSecurityState,
+    /// The security state's version
+    security_version: u64,
 }
 
-/// Makes a new set of signing keys for a user, which should only be done during
-/// once. This also signs the public key with the signing key
-/// and returns the signed public key.
-pub fn make_user_signing_keys_for_enrollment(
+/// Initializes the user's cryptographic state for v2 users.
```
Hinton: **question**: What does `Initializes` mean here? I don't see this setting security state in the client.

crates/bitwarden-core/src/client/encryption_settings.rs
```diff
@@ -89,10 +143,44 @@ impl EncryptionSettings {
...
+    }
+
+    #[cfg(feature = "internal")]
+    fn init_v2(
+        user_key: XChaCha20Poly1305Key,
+        private_key: EncString,
+        signing_key: EncString,
+        security_state: SignedSecurityState,
+        store: &KeyStore<KeyIds>,
+        sdk_security_state: &RwLock<Option<SecurityState>>,
```
Hinton, line 158: out-of-scope: Hmm, having to pass in the SDK's state and write to it seems quite unfortunate. Outside the scope of this PR though.

crates/bitwarden-core/src/key_management/crypto.rs
```diff
@@ -870,10 +918,70 @@ mod tests {
...
+
+        let master_key = MasterKey::derive(
+            "asdfasdfasdf",
+            "test@bitwarden.com",
+            &Kdf::PBKDF2 {
+                iterations: NonZeroU32::new(100_000).unwrap(),
+            },
+        )
+        .unwrap();
+        let enrollment_response = make_keys_for_user_crypto_v2(&client).unwrap();
```
Hinton: **question:** Do we need a test for when a client does not contain a user per the `make_keys_for_user_crypto_v2` documentation.

> If the client already contains a v1 user, then this user's private-key will be re-used.

crates/bitwarden-core/src/key_management/security_state.rs
```diff
@@ -0,0 +1,199 @@
...
+///
+/// It contains a version, which can only ever increment. Based on the version, old formats and
+/// features are blocked. This prevents a server from downgrading a user's account features, because
+/// only the user can create this signed object.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct SecurityState {
+    /// The entity ID is a permanent, unchangeable, unique identifier for the object this security
+    /// state applies to. For users, this is the user ID, which never changes.
+    entity_id: ByteBuf,
```
Hinton: Are we avoiding using `Uuid` here in case that assumption ever changes?

crates/bitwarden-core/src/key_management/security_state.rs
```diff
@@ -0,0 +1,199 @@
...
+impl SignedSecurityState {
+    /// Verifies the signature of the `SignedSecurityState` using the provided `VerifyingKey`.
+    pub fn verify_and_unwrap(
+        self,
+        verifying_key: &VerifyingKey,
+    ) -> Result<SecurityState, CryptoError> {
+        self.0
+            .verify_and_unwrap(verifying_key, &SigningNamespace::SecurityState)
+    }
+}
```
Hinton, line 98: **issue:** The struct is implemented before it's defined. For readability we can consider changing the order.

crates/bitwarden-core/src/key_management/security_state.rs
quexten: I'm OK with changing it to UUID, but I also didn't check whether UUID encodes efficiently here using `serde`. Bytebuf specifically is chosen here because it does.

crates/bitwarden-crypto/src/store/key_rotation.rs
quexten: No, this was not accurate. Nice catch!

crates/bitwarden-crypto/src/util.rs
quexten: I'll keep that refactor out of this PR. Do we have a crate where this *should* live? (`crypto` currently only depends on `error` but no other sdk crates).

crates/bitwarden-crypto/src/error.rs
quexten: I think `CryptoStateError` should be used when the state the SDK is in is invalid. Usually this is prevented by the type system, but sometimes it is not (specifically for the security state it is not).

For any actions where the input is wrong - an encstring is not decryptable because it has been tampered with, etc. - a `CryptoError` is used. I agree with making the enums smaller, but I don't think `CryptoStateError` should include `CryptoError`.

crates/bitwarden-core/src/key_management/security_state.rs
quexten: Dropped this. Not planned to be used in SM-SDK for now, but I don't know if that's the case long-term.

crates/bitwarden-core/src/key_management/crypto.rs
quexten: Updated to:
> /// Creates the user's cryptographic state for v2 users. This includes ensuring signature keypair is present,
/// a signed public key is present, a security state is present and signed, and the user key is a Cose key.
/// If the client already contains a v1 user, then this user's private-key will be
/// re-used.

crates/bitwarden-core/src/client/encryption_settings.rs
quexten: Discussed in slack, but yes I agree, and we'll change this in tech-debt cleanup.

crates/bitwarden-core/src/key_management/crypto.rs
quexten: Added a test, that matches that this throws a "missing security state" error. Do we want a specific error detecting a v1 user?

crates/bitwarden-crypto/src/error.rs
Hinton: Just to confirm, the expectation is that any decrypt or encrypt operation (i.e. where `CryptoError` is currently primarily used) can throw `CryptoStateError` and we plan on doing so?

crates/bitwarden-core/src/auth/login/password.rs
```diff
@@ -34,6 +34,8 @@ pub(crate) async fn login_password(
     let response = request_identity_tokens(client, input, &password_hash).await?;
 
     if let IdentityTokenResponse::Authenticated(r) = &response {
+        use crate::client::internal::UserKeyState;
+
```
Hinton: **suggestion:** Lift this up to line 26.

crates/bitwarden-core/src/auth/tde.rs
```diff
@@ -41,10 +44,13 @@ pub(super) fn make_register_tde_keys(
...
     client.internal.initialize_user_crypto_decrypted_key(
         user_key.0,
-        key_pair.private.clone(),
-        // Note: Signing keys are not supported on registration yet. This needs to be changed as
-        // soon as registration is supported.
-        None,
+        UserKeyState {
+            private_key: key_pair.private.clone(),
+            // Note: Signing keys are not supported on registration yet. This needs to be changed as
+            // soon as registration is supported.
```
Hinton: **suggestion:** Link to a jira ticket?

```suggestion
// TODO (PM- ): Add support for signing keys during registration.
```

crates/bitwarden-core/src/client/internal.rs
```diff
@@ -14,17 +14,31 @@ use uuid::Uuid;
...
+use crate::key_management::SecurityState;
 use crate::{
     auth::renew::renew_token, client::login_method::LoginMethod, error::UserIdAlreadySetError,
     key_management::KeyIds, DeviceType,
 };
 #[cfg(feature = "internal")]
+use crate::{
+    client::encryption_settings::AccountEncryptionKeys, key_management::SignedSecurityState,
+};
+#[cfg(feature = "internal")]
```
Hinton: **issue:** Please combine the imports with the same cfg flags.

crates/bitwarden-core/src/key_management/security_state.rs
quexten: Switched to UUID after verifying it serializes efficiently.

crates/bitwarden-core/src/key_management/crypto.rs
```diff
@@ -574,76 +567,89 @@ pub(super) fn verify_asymmetric_keys(
...
+    let temporary_user_key_id = SymmetricKeyId::Local("temporary_user_key");
+    let temporary_signing_key_id = SigningKeyId::Local("temporary_signing_key");
+    let temporary_private_key_id = AsymmetricKeyId::Local("temporary_private_key");
+
+    // If the user already has a private key, use it. Otherwise, create a temporary one.
+    let private_key_id = if ctx.has_asymmetric_key(AsymmetricKeyId::UserPrivateKey) {
+        AsymmetricKeyId::UserPrivateKey
+    } else {
+        ctx.make_asymmetric_key(temporary_private_key_id)?
+    };
```
Hinton: **question:** Are we concerned that users with invalid private keys will get a new private key in this process?

crates/bitwarden-core/src/key_management/crypto.rs
Hinton: Not sure I follow. `make_keys_for_user_crypto_v2` explicitly tells you it can be run without a user. But it's never tested.

crates/bitwarden-crypto/src/util.rs
Hinton: No, not currently.

crates/bitwarden-core/src/key_management/crypto.rs
quexten: Hmm, maybe this suggests this function should be split up into:
`make_v2_keys_for_new_user` (this one is not needed for now, only later when registration is implemented) and `make_v2_keys_for_v1_user`. The former always generates a new private key, the latter always forces a private key to be present. That way, we can handle `v1` users that have a corrupt private key through an explicit process (private key regen.. whenever that is turned on).

crates/bitwarden-core/src/client/internal.rs
quexten: Done

crates/bitwarden-core/src/key_management/crypto.rs
quexten: Renamed the function now and added assertions + tests validating the behavior. This also introduces a new `CryptoStateError` variant.

crates/bitwarden-core/src/key_management/crypto.rs
quexten: `make_keys_for_user_crypto_v2` now became `make_v2_keys_for_v1_user`, and tests are added. Whenever we get to registration, we should handle the uninitialized flow differently to ensure each state is valid.

crates/bitwarden-core/src/auth/tde.rs
quexten: Linked!

crates/bitwarden-core/src/auth/login/password.rs
quexten: Done

crates/bitwarden-crypto/src/error.rs
quexten: Hmm, thinking about this, my initial line of thinking was not accurate. Really the `CryptoStateError` can only ever occur outside of the `crypto` crate currently, and forseeably. I think your initial suggestion was right. I'll move this error to the `core` crate and rename it to `StatefulCryptoError`, which signifies that it is not necessarily only about the state being wrong.

This error type is meant to include some of the regular crypto error variants (via including crypto error), and then some extra ones about the state being wrong (no private key, wrong user crypto version), and these extra ones don't occur in the crypto crate itself.

Hinton (approved):

dani-garcia (approved):

PR #328, opened 2025-06-24
[PM-22812] Attachments get corrupted when downgrading from cipherkeys

gbubemismith:
## 🎟️ Tracking
https://bitwarden.atlassian.net/browse/PM-22812

## 📔 Objective
Editing a cipher with attachments causes attachment corruption when switching between cipher-key encryption and user-key encryption in either direction. The SDK's AttachmentView was not providing the decrypted attachment key needed for re-encryption scenarios. When the cipher gets re-encrypted between different encryption contexts (cipher-key ↔ user-key) during save, the client needs the decrypted attachment key to properly re-encrypt it under the new encryption context. Without this, null gets posted for the attachment key, breaking decryption.

This is a temporary solution during the migration from TypeScript to the SDK. The decrypted_key field should be removed once all encryption logic is handled within the SDK.

Cleanup tracked in: https://bitwarden.atlassian.net/browse/PM-23005

## ⏰ Reminders before review

- CI builds passed
  team

## 🦮 Reviewer guidelines

### DIFF
--- a/crates/bitwarden-vault/src/cipher/attachment.rs
+++ b/crates/bitwarden-vault/src/cipher/attachment.rs
@@ -172,6 +172,7 @@ impl Decryptable<KeyIds, SymmetricKeyId, AttachmentView> for Attachment {
         ctx: &mut KeyStoreContext<KeyIds>,
         key: SymmetricKeyId,
     ) -> Result<AttachmentView, CryptoError> {
+        #[cfg(feature = "wasm")]
         let decrypted_key = if let Some(attachment_key) = &self.key {
             let content_key_id = ctx.unwrap_symmetric_key(key, ATTACHMENT_KEY, attachment_key)?;
 
--- a/crates/bitwarden-vault/src/cipher/attachment.rs
+++ b/crates/bitwarden-vault/src/cipher/attachment.rs
@@ -43,7 +43,7 @@ pub struct AttachmentView {
     /// process. It will be removed once the encryption/decryption logic is
     /// fully migrated to the SDK.
     ///
-    /// **Ticket**: https://bitwarden.atlassian.net/browse/PM-23005
+    /// **Ticket**: <https://bitwarden.atlassian.net/browse/PM-23005>
     ///
     /// Do not rely on this field for long-term use.
     #[cfg(feature = "wasm")]
--- a/crates/bitwarden-vault/src/cipher/attachment.rs
+++ b/crates/bitwarden-vault/src/cipher/attachment.rs
@@ -36,6 +36,18 @@ pub struct AttachmentView {
     pub size_name: Option<String>,
     pub file_name: Option<String>,
     pub key: Option<EncString>,
+    /// The decrypted attachmentkey in base64 format.
+    ///
+    /// **TEMPORARY FIELD**: This field is a temporary workaround to provide
+    /// decrypted attachment keys to the TypeScript client during the migration
+    /// process. It will be removed once the encryption/decryption logic is
+    /// fully migrated to the SDK.
+    ///
+    /// **Ticket**: https://bitwarden.atlassian.net/browse/PM-23005
+    ///
+    /// Do not rely on this field for long-term use.
+    #[cfg(feature = "wasm")]
+    pub decrypted_key: Option<String>,
 }
 
 #[allow(missing_docs)]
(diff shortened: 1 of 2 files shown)


crates/bitwarden-vault/src/cipher/attachment.rs
```diff
@@ -36,6 +36,18 @@ pub struct AttachmentView {
...
+    ///
+    /// **TEMPORARY FIELD**: This field is a temporary workaround to provide
+    /// decrypted attachment keys to the TypeScript client during the migration
+    /// process. It will be removed once the encryption/decryption logic is
+    /// fully migrated to the SDK.
+    ///
+    /// **Ticket**: <https://bitwarden.atlassian.net/browse/PM-23005>
+    ///
+    /// Do not rely on this field for long-term use.
+    #[cfg(feature = "wasm")]
```
quexten, line 49: Question: As far as I'm aware, we don't use the SDK to make API calls on mobile yet. I think this means that the same bug *has* to be present on mobile currently. We may need to expose this to mobile too?

quexten: Ah, probably nevermind, we also expose the encrypted key, which in this case is what we want to update with the server. I think this also means we can just replace the whole `clients` code for re-encrypting the key, and just use the encrypted key from the attachment view from the sdk directly?

gbubemismith: You are probably thinking about this [PR](https://github.com/bitwarden/server/pull/5998) which addresses the current issue we have in production

quexten: No, this is the correct PR. I guess I'm wondering: Why do we need to expose the key material on WASM but don't for uniffi / mobile. Mobile also supports editing ciphers that have cipherkeys and attachments, and would downgrade them right?

Is the difference that we don't the sdk for *encryption* of attachment views / cipher views yet, wheras on mobile we do?

gbubemismith: Oh, I don't expect mobile to have the same issues cause they use the encryption and decryption from the SDK, the SDK handles the keys. On the TS clients, we won't need the decrypted key field anymore once we migrate the cipher encryption to use the SDK

>  the difference that we don't the sdk for encryption of attachment views / cipher views yet, wheras on mobile we do?

Yes

quexten: Nice, thanks.

shane-melton (dismissed):

quexten (dismissed):

crates/bitwarden-vault/src/cipher/attachment.rs
```diff
@@ -160,13 +172,26 @@ impl Decryptable<KeyIds, SymmetricKeyId, AttachmentView> for Attachment {
         ctx: &mut KeyStoreContext<KeyIds>,
         key: SymmetricKeyId,
     ) -> Result<AttachmentView, CryptoError> {
+        let decrypted_key = if let Some(attachment_key) = &self.key {
```
dani-garcia, line 176: This entire if can also be under a `#[cfg(feature = "wasm")]` flag to avoid decrypting the key when not needed in mobile.

crates/bitwarden-vault/src/cipher/cipher.rs
```diff
@@ -343,7 +343,7 @@ impl Decryptable<KeyIds, SymmetricKeyId, CipherView> for Cipher {
             permissions: self.permissions,
             view_password: self.view_password,
             local_data: self.local_data.decrypt(ctx, ciphers_key).ok().flatten(),
-            attachments: self.attachments.decrypt(ctx, ciphers_key).ok().flatten(),
+            attachments: self.attachments.decrypt(ctx, ciphers_key)?,
```
dani-garcia: Is this change needed here?

dani-garcia (commented):
LGTM just two small changes

crates/bitwarden-vault/src/cipher/cipher.rs
gbubemismith: Oh, left over from a test I was doing. Nice catch

dani-garcia (approved):

shane-melton (approved):

ilya-sarantsev:
@gbubemismith hi! I see this one is fixed, and now I have `[error: cannot decrypt]` in place of the attachment that I couldn't download. I still can't download the file. Also, as I understand it, I have an unknown number of attachments that are similarly corrupted. Any idea how to proceed to fix this?

PR #329, opened 2025-06-25
[PM-22593] improve initialization process for database and repositories

coroiu:
## ⏰ Reminders before review

- CI builds passed
  team

## 🦮 Reviewer guidelines

### DIFF
--- a/crates/bitwarden-state/src/sdk_managed/configuration.rs
+++ b/crates/bitwarden-state/src/sdk_managed/configuration.rs
@@ -1,11 +1,16 @@
+use std::path::PathBuf;
+
 #[derive(Debug)]
 /// Configuration for the database used by the SDK.
 pub enum DatabaseConfiguration {
     /// SQLite configuration, used on native platforms
     Sqlite {
+        /// The name of the SQLite database. Different users should have different database
+        /// names to avoid conflicts.
+        db_name: String,
         /// The file path to the SQLite database. Databases for different users should be stored in
         /// different files.
-        file_path: String,
+        folder_path: PathBuf,
     },
 
     /// IndexedDB configuration, used on WebAssembly platforms
--- a/crates/bitwarden-state/src/sdk_managed/configuration.rs
+++ b/crates/bitwarden-state/src/sdk_managed/configuration.rs
@@ -1,0 +1,17 @@
+#[derive(Debug)]
+/// Configuration for the database used by the SDK.
+pub enum DatabaseConfiguration {
+    /// SQLite configuration, used on native platforms
+    Sqlite {
+        /// The file path to the SQLite database. Databases for different users should be stored in
+        /// different files.
+        file_path: String,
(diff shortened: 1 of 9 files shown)


crates/bitwarden-state/src/sdk_managed/noop/mod.rs
dani-garcia, line 1: This file doesn't seem used?

crates/bitwarden-state/src/sdk_managed/configuration.rs
```diff
@@ -0,0 +1,17 @@
+#[derive(Debug)]
+/// Configuration for the database used by the SDK.
+pub enum DatabaseConfiguration {
+    /// SQLite configuration, used on native platforms
+    Sqlite {
+        /// The file path to the SQLite database. Databases for different users should be stored in
+        /// different files.
+        file_path: String,
```
dani-garcia: We could use `PathBuf` here, and just call `PathBuf::from(String)` from the Uniffi struct.

Non-blocking thought: SQLite tends to create a couple of files (the main db and the wal/shm files). It might be confusing if we receive a file path and also create a couple of extra files, should we take a directory instead? And then we can handle creating the files inside ourselves.

dani-garcia (commented):
Just a couple of small comments, otherwise lgtm

crates/bitwarden-state/src/sdk_managed/configuration.rs
coroiu: Hmm yeah, in that case I can add `db_name` as an additional option so that we know what to name the files

crates/bitwarden-state/src/sdk_managed/configuration.rs
```diff
@@ -0,0 +1,22 @@
...
+#[derive(Debug)]
+/// Configuration for the database used by the SDK.
+pub enum DatabaseConfiguration {
+    /// SQLite configuration, used on native platforms
+    Sqlite {
+        /// The name of the SQLite database. Different users should have different database
+        /// names to avoid conflicts.
+        db_name: String,
+        /// The file path to the SQLite database. Databases for different users should be stored in
+        /// different files.
```
dani-garcia: Thought: I think now that the `db_name` is separate and we say should be different per-user, this doesn't need to be as restrictive right? 

We could have the same folder for all users as long as db_name is different.

dani-garcia (approved):
LGTM, one small non blocking thought about the sqlite separation

crates/bitwarden-state/src/sdk_managed/configuration.rs
coroiu: Argh forgot to update the comment, you are right, sorry for making you comment on such tiny things 🥲 
I'll fix before marking this PR ready for review!

dani-garcia (approved):

PR #331, opened 2025-06-25
[PM-22136] Expose encrypt_fido2_credentials to support cipher encryption in TS clients

shane-melton:
## 🎟️ Tracking

[PM-22136](https://bitwarden.atlassian.net/browse/PM-22136)

## 📔 Objective

The TS clients use fully decrypted FIDO2 credentials which need to be re-encrypted separately before using the SDK to encrypt the rest of the Cipher. Otherwise, the FIDO2 credential encryption key will be lost.

Once the decrypted FIDO2 Credentials are removed from the LoginView in TS, this method can be removed.

Related Clients PR: https://github.com/bitwarden/clients/pull/15337

## ⏰ Reminders before review

- CI builds passed
  team

## 🦮 Reviewer guidelines

[PM-22136]: https://bitwarden.atlassian.net/browse/PM-22136?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/crates/bitwarden-vault/src/cipher/cipher_client.rs
+++ b/crates/bitwarden-vault/src/cipher/cipher_client.rs
@@ -6,7 +6,7 @@ use wasm_bindgen::prelude::*;
 use super::EncryptionContext;
 use crate::{
     Cipher, CipherError, CipherListView, CipherView, DecryptError, EncryptError,
-    Fido2CredentialFullView, Fido2CredentialView,
+    Fido2CredentialFullView,
 };
 
 #[allow(missing_docs)]
--- a/crates/bitwarden-vault/src/cipher/cipher_client.rs
+++ b/crates/bitwarden-vault/src/cipher/cipher_client.rs
@@ -1,10 +1,10 @@
 use bitwarden_core::{Client, OrganizationId};
-use bitwarden_crypto::IdentifyKey;
+use bitwarden_crypto::{Encryptable, IdentifyKey};
 #[cfg(feature = "wasm")]
 use wasm_bindgen::prelude::*;
 
 use super::EncryptionContext;
-use crate::{Cipher, CipherError, CipherListView, CipherView, DecryptError, EncryptError};
+use crate::{Cipher, CipherError, CipherListView, CipherView, DecryptError, EncryptError, Fido2CredentialFullView, Fido2CredentialView};
 
 #[allow(missing_docs)]
 #[cfg_attr(feature = "wasm", wasm_bindgen)]
@@ -65,6 +65,24 @@ impl CiphersClient {
         let key_store = self.client.internal.get_key_store();
         let credentials = cipher_view.decrypt_fido2_credentials(&mut key_store.context())?;
         Ok(credentials)
+    }
+    
+    /// Temporary method used to re-encrypt FIDO2 credentials for a cipher view. 
+    /// Necessary while the FIDO2 credentials remain in the LoginView in an encrypted form.
+    /// Used by the TS clients to encrypt FIDO2 credentials separately before encrypting the remaining cipher view.
+    /// TODO: Remove once FIDO2 credentials have been removed from the LoginView.
+    pub fn encrypt_fido2_credentials(
+        &self,
+        cipher_view: CipherView,
+        fido2_credentials: Fido2CredentialFullView,
+    ) -> Result<crate::Fido2Credential, EncryptError> {
+        let key_store = self.client.internal.get_key_store();
+        let key = cipher_view.key_identifier();
+        let cipher_key = Cipher::decrypt_cipher_key(&mut key_store.context(), key, &cipher_view.key)?;
+        
+        let fido2_credential = fido2_credentials.encrypt(&mut key_store.context(), cipher_key)?;
+        
+        Ok(fido2_credential)
     }
 
     #[allow(missing_docs)]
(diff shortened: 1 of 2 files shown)


crates/bitwarden-vault/src/cipher/cipher_client.rs
```diff
@@ -67,6 +70,26 @@ impl CiphersClient {
         Ok(credentials)
     }
 
+    /// Temporary method used to re-encrypt FIDO2 credentials for a cipher view.
+    /// Necessary while the FIDO2 credentials remain in the LoginView in an encrypted form.
+    /// Used by the TS clients to encrypt FIDO2 credentials separately before encrypting the
+    /// remaining cipher view. TODO: Remove once FIDO2 credentials have been removed from the
+    /// LoginView.
```
Hinton: Please add a jira ticket for the this work, would it just be to revert this commit?

crates/bitwarden-vault/src/cipher/cipher_client.rs
```diff
@@ -67,6 +70,26 @@ impl CiphersClient {
...
+    pub fn encrypt_fido2_credentials(
```
Hinton: **issue:** Please limit this to wasm using `cfg(feature="wasm")`

Hinton (commented):
Can we please provide some more information about **why** this is needed? The mobile clients uses this implementation so if we have issues with fido 2 credentials not being encrypted correctly that would apply to mobile as well and is something we need to address.

> Necessary while the FIDO2 credentials remain in the LoginView in an encrypted form.

As in you want to expose them decrypted? That shouldn't be necessary ever. Or you want to remove it once the SDK can load from state?

It's there in encrypted form so that we can generate a cipher key if needed as we have to re-encrypt the whole object.

gbubemismith:
Changes look good. Just waiting for the request changes to come in before approving

shane-melton:
@Hinton 
>Can we please provide some more information about why this is needed? The mobile clients uses this implementation so if we have issues with fido 2 credentials not being encrypted correctly that would apply to mobile as well and is something we need to address.

The TS clients do not currently use the SDK for [creating fido 2 credentials](https://github.com/bitwarden/clients/blob/5eca3a591667766ff8dd42b67642ce960391004e/libs/common/src/platform/services/fido2/fido2-authenticator.service.ts#L184-L190). So, whenever we [encrypt ciphers with new credentials](https://github.com/bitwarden/clients/blob/1f8a56a5c7c59e14dcd4fb9ac89eef1252b4effe/libs/common/src/vault/services/default-cipher-encryption.service.ts#L33-L42) with the SDK we do not have an encrypted copy to pass along to the SDK. In order to use the SDK for all encryption, we need to expose this method in the meantime to explicitly encrypt the fido2 credentials.

Mobile uses the SDK for [Fido2 credential creation](https://github.com/bitwarden/android/blob/917aaac3a60daf9a409e4744d3c0e7b8d23459ab/app/src/main/kotlin/com/x8bit/bitwarden/data/credentials/manager/BitwardenCredentialManagerImpl.kt#L300-L301) so they don't need to worry about encrypting the credentials manually. 

Once the TS clients use the SDK for Fido2 credential management ([PM-8313](https://bitwarden.atlassian.net/browse/PM-8313)) we'll be able to remove this _and_ the decrypted Fido2Credentials from the LoginView in the TS clients.

> Or you want to remove it once the SDK can load from state?

Yep, we can remove the [credentials from the LoginView](https://github.com/bitwarden/sdk-internal/blob/d7b9687e63b42eb12e5a8b127ac305209aba1f83/crates/bitwarden-vault/src/cipher/login.rs#L311-L312) once the SDK can load from state. Then we won't need to always pass the encrypted fido 2 credentials back and forth (as far as I understand it)


[PM-8313]: https://bitwarden.atlassian.net/browse/PM-8313?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

crates/bitwarden-vault/src/cipher/cipher_client.rs
shane-melton: Added JIRA ticket number [d8e596b](https://github.com/bitwarden/sdk-internal/pull/331/commits/d8e596b01893d060c53648c07c03074a17c160ed)

And yes, we could just revert this PR.

crates/bitwarden-vault/src/cipher/cipher_client.rs
shane-melton: Done! [d8e596b](https://github.com/bitwarden/sdk-internal/pull/331/commits/d8e596b01893d060c53648c07c03074a17c160ed)

Hinton (changes requested):
There is a `set_new_fido2_credentials` on `CipherView` for this purpose which is currently used internally in the sdk for the fido authenticator. This can probably be exposed with some simple glue code.

https://github.com/bitwarden/sdk-internal/blob/f2bc7084ffa2a0a427b2b215bd0d39944f9cb705/crates/bitwarden-vault/src/cipher/cipher.rs#L563-L575


```rust
cipher_view.set_new_fido2_credentials(creds);
```

shane-melton:
@Hinton Thanks for pointing that out! Updated in [efaf7fd](https://github.com/bitwarden/sdk-internal/pull/331/commits/efaf7fdf4797ed49f1fc45ac63bc243700d51235)

gbubemismith (approved):

Hinton (approved):
LGTM

PR #332, opened 2025-06-25
[PM-22515] Update cipher decrypt list to return successful decryptions and failure metadata

gbubemismith:
## 🎟️ Tracking
https://bitwarden.atlassian.net/browse/PM-22515

## 📔 Objective
Currently, calling ClientCiphers.decryptList(ciphers:) is an all-or-nothing operation. You’ll either get back a list decrypted ciphers if everything was successful or a single error if any cipher is unable to be decrypted. We’d like to get back anything that was successfully decrypted along with a list of any ciphers which failed to decrypt.

## ⏰ Reminders before review

- CI builds passed
  team

## 🦮 Reviewer guidelines

### DIFF
--- a/crates/bitwarden-crypto/src/store/mod.rs
+++ b/crates/bitwarden-crypto/src/store/mod.rs
@@ -284,18 +284,16 @@ impl<Ids: KeyIds> KeyStore<Ids> {
             .map(|chunk| {
                 let mut ctx = self.context();
 
-                let mut result = Vec::with_capacity(chunk.len());
-
-                for item in chunk {
-                    let key = item.key_identifier();
-                    result.push(
-                        item.decrypt(&mut ctx, key)
-                            .map_err(|_| ErrorView::from(item)),
-                    );
-                    ctx.clear_local();
-                }
-
-                result
+                chunk
+                    .iter()
+                    .map(|item| {
+                        let result = item
+                            .decrypt(&mut ctx, item.key_identifier())
+                            .map_err(|_| ErrorView::from(item));
+                        ctx.clear_local();
+                        result
+                    })
+                    .collect::<Vec<_>>()
             })
             .flatten()
             .partition_map(|result| match result {
--- a/crates/bitwarden-crypto/src/store/mod.rs
+++ b/crates/bitwarden-crypto/src/store/mod.rs
@@ -259,6 +259,54 @@ impl<Ids: KeyIds> KeyStore<Ids> {
         res
     }
 
+    /// Decrypt a list of items using this key store, returning a tuple of successful and failed items.
+    ///
+    /// # Arguments
+    /// * `data` - The list of items to decrypt.
+    ///
+    /// # Returns
+    /// A tuple containing two vectors: the first vector contains the successfully decrypted items,
+    /// and the second vector contains the items that failed to decrypt along with their error views.
+    pub fn decrypt_list_with_failures<
+        'a,
+        Key: KeyId,
+        Data: Decryptable<Ids, Key, Output> + IdentifyKey<Key> + Send + Sync + 'a,
+        Output: Send + Sync,
+        ErrorView: From<&'a Data> + Send + Sync,
+    >(
+        &self,
+        data: &'a [Data],
+    ) -> (Vec<Output>, Vec<ErrorView>) {
+        let results: Vec<_> = data
+            .par_chunks(batch_chunk_size(data.len()))
+            .map(|chunk| {
+                let mut ctx = self.context();
+                let mut successes = Vec::new();
+                let mut failures = Vec::new();
+
+                for item in chunk {
+                    let key = item.key_identifier();
+                    match item.decrypt(&mut ctx, key) {
+                        Ok(decrypted) => successes.push(decrypted),
+                        Err(_) => failures.push(ErrorView::from(item)),
+                    }
+                    ctx.clear_local();
+                }
+
+                (successes, failures)
+            })
+            .collect();
+
+        let (mut successes, mut failures) = (Vec::new(), Vec::new());
+
+        for (successes_chunk, failures_chunk) in results {
+            successes.extend(successes_chunk);
+            failures.extend(failures_chunk);
+        }
+
+        (successes, failures)
+    }
+
     /// Encrypt a list of items using this key store. The keys returned by
     /// `data[i].key_identifier()` must already be present in the store, otherwise this will
     /// return an error. This method will try to parallelize the encryption of the items, for
--- a/crates/bitwarden-crypto/src/store/mod.rs
+++ b/crates/bitwarden-crypto/src/store/mod.rs
@@ -24,7 +24,7 @@
 
 use std::sync::{Arc, RwLock};
 
-use rayon::prelude::*;
+use rayon::{iter::Either, prelude::*};
 
 use crate::{Decryptable, Encryptable, IdentifyKey, KeyId, KeyIds};
 
@@ -279,34 +279,31 @@ impl<Ids: KeyIds> KeyStore<Ids> {
         &self,
         data: &'a [Data],
     ) -> (Vec<Output>, Vec<ErrorView>) {
-        let results: Vec<_> = data
+        let results: (Vec<_>, Vec<_>) = data
             .par_chunks(batch_chunk_size(data.len()))
             .map(|chunk| {
                 let mut ctx = self.context();
-                let mut successes = Vec::new();
-                let mut failures = Vec::new();
+
+                let mut result = Vec::with_capacity(data.len());
 
                 for item in chunk {
                     let key = item.key_identifier();
-                    match item.decrypt(&mut ctx, key) {
-                        Ok(decrypted) => successes.push(decrypted),
-                        Err(_) => failures.push(ErrorView::from(item)),
-                    }
+                    result.push(
+                        item.decrypt(&mut ctx, key)
+                            .map_err(|_| ErrorView::from(item)),
+                    );
                     ctx.clear_local();
                 }
 
-                (successes, failures)
+                result
             })
-            .collect();
+            .flatten()
+            .partition_map(|result| match result {
+                Ok(output) => Either::Left(output),
+                Err(err) => Either::Right(err),
+            });
 
-        let (mut successes, mut failures) = (Vec::new(), Vec::new());
-
-        for (successes_chunk, failures_chunk) in results {
-            successes.extend(successes_chunk);
-            failures.extend(failures_chunk);
-        }
-
-        (successes, failures)
+        results
     }
 
     /// Encrypt a list of items using this key store. The keys returned by
--- a/crates/bitwarden-crypto/src/store/mod.rs
+++ b/crates/bitwarden-crypto/src/store/mod.rs
@@ -284,7 +284,7 @@ impl<Ids: KeyIds> KeyStore<Ids> {
             .map(|chunk| {
                 let mut ctx = self.context();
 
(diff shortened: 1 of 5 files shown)


crates/bitwarden-vault/src/cipher/cipher_client.rs
```diff
@@ -57,6 +60,17 @@ impl CiphersClient {
         Ok(cipher_views)
     }
 
+    #[allow(missing_docs)]
```
dani-garcia: Please add a comment here, it can be the same one that is on the uniffi crate.

crates/bitwarden-vault/src/cipher/cipher_client.rs
```diff
@@ -203,6 +173,66 @@ mod tests {
...
+        let result = client
+            .vault()
+            .ciphers()
+            .decrypt_list_with_failures(vec![valid_cipher]);
+
+        assert_eq!(result.successes.len(), 1);
+        assert!(result.failures.is_empty());
+        assert_eq!(result.successes[0].name, "234234");
+    }
+
```
dani-garcia, line 236: It could be nice to also have a test with a failure too

crates/bitwarden-crypto/src/store/mod.rs
```diff
@@ -259,6 +259,56 @@ impl<Ids: KeyIds> KeyStore<Ids> {
...
+
+        let (mut successes, mut failures) = (Vec::new(), Vec::new());
+
+        for (successes_chunk, failures_chunk) in results {
+            successes.extend(successes_chunk);
+            failures.extend(failures_chunk);
+        }
+
+        (successes, failures)
+    }
```
dani-garcia: I think we can optimize this code and the allocations a bit if we keep it inside the rayon context:
```rust
    pub fn decrypt_list_with_failures<
        'a,
        Key: KeyId,
        Data: Decryptable<Ids, Key, Output> + IdentifyKey<Key> + Send + Sync + 'a,
        Output: Send + Sync,
        ErrorView: From<&'a Data> + Send + Sync,
    >(
        &self,
        data: &'a [Data],
    ) -> (Vec<Output>, Vec<ErrorView>) {
        use rayon::iter::Either;

        let results: (Vec<_>, Vec<_>) = data
            .par_chunks(batch_chunk_size(data.len()))
            .map(|chunk| {
                let mut ctx = self.context();

                let mut result = Vec::with_capacity(data.len());

                for item in chunk {
                    let key = item.key_identifier();
                    result.push(
                        item.decrypt(&mut ctx, key)
                            .map_err(|_| ErrorView::from(item)),
                    );
                    ctx.clear_local();
                }

                result
            })
            .flatten()
            .partition_map(|result| match result {
                Ok(output) => Either::Left(output),
                Err(err) => Either::Right(err),
            });

        results
    }
```

gbubemismith: Thanks 🙌🏼. This looks nice. Fixed 8e8dde18a7434b7fe2f378e0d9002a01af503742

crates/bitwarden-vault/src/cipher/cipher_client.rs
gbubemismith: Added c27d05f404c7be955f0b890c996fc60855927ab4

crates/bitwarden-vault/src/cipher/cipher_client.rs
gbubemismith: Added 17809c4cf10ca4028d577564c0a7942c2775b8c0

dani-garcia (dismissed):

crates/bitwarden-crypto/src/store/mod.rs
```diff
@@ -259,6 +259,53 @@ impl<Ids: KeyIds> KeyStore<Ids> {
...
+                for item in chunk {
+                    let key = item.key_identifier();
+                    result.push(
+                        item.decrypt(&mut ctx, key)
+                            .map_err(|_| ErrorView::from(item)),
+                    );
+                    ctx.clear_local();
+                }
+
+                result
```
quexten: Seems we can make this a little bit more concise (unless I'm missing the reason to use the mut vec).

```suggestion
                let mut ctx = self.context();

                chunk
                    .iter()
                    .map(|item| {
                        let result = item
                            .decrypt(&mut ctx, item.key_identifier())
                            .map_err(|_| ErrorView::from(item));
                        ctx.clear_local();
                        result
                    })
                    .collect::<Vec<_>>()
```

quexten (changes requested):
Mostly seems good, I do think we can make the iteration a little bit more concise and drop the mut Vec.

crates/bitwarden-crypto/src/store/mod.rs
quexten: @dani-garcia since you approved above / suggested the pre-allocated Vec: Is the mut vec simply an optimization so that we pre-allocate once? If we do want to keep the pre-allocated mut vec, could we add a comment stating we pre-allocate for performance optimization?

gbubemismith: Looks like the mut vec is a memory allocation optimisation. However, I prefer this closure approach. I'll wait to get @dani-garcia's feedback. I don't think there would be a huge difference in performance.

dani-garcia: Yeah, it's a micro optimization really, but was easy to do so figured why not. In this example, doing chunk.iter().map().collect() will also apply the same optimization internally, so we can do it as well (The FromIterator of Vec is smart enough to use the SizeHint trait to preallocate the correct amount)

gbubemismith: 57d83ecd4709a7111a519ca03b45de79ae3823f3

dani-garcia (dismissed):

crates/bitwarden-crypto/src/store/mod.rs
quexten: Nice, thanks for the small cleanup!

quexten (dismissed):

crates/bitwarden-crypto/src/store/mod.rs
```diff
@@ -259,6 +259,51 @@ impl<Ids: KeyIds> KeyStore<Ids> {
...
+                    .map(|item| {
+                        let result = item
+                            .decrypt(&mut ctx, item.key_identifier())
+                            .map_err(|_| ErrorView::from(item));
+                        ctx.clear_local();
+                        result
+                    })
+                    .collect::<Vec<_>>()
+            })
+            .flatten()
```
Hinton: **nit**: You can use `flat_map` above which is the same as `map().flatten()`

crates/bitwarden-vault/src/cipher/cipher.rs
```diff
@@ -249,6 +249,64 @@ pub struct CipherListView {
...
+
+#[allow(missing_docs)]
+#[derive(Serialize, Deserialize, Debug, PartialEq)]
+#[serde(rename_all = "camelCase", deny_unknown_fields)]
+#[cfg_attr(feature = "uniffi", derive(uniffi::Record))]
+#[cfg_attr(feature = "wasm", derive(Tsify), tsify(into_wasm_abi, from_wasm_abi))]
+pub struct DecryptCipherListResult {
+    pub successes: Vec<CipherListView>,
+    pub failures: Vec<CipherDecryptionErrorView>,
+}
```
Hinton: **question (non blocking):** What is the user experience we want? I notice the error view is just returning data that doesn't require us to decrypt.

Could we make this even simpler and return `Cipher` for the failed entries? We would need to handle the mapping in the TS side, but if we only display `decryption failed` that's not particularly worse?

```suggestion
pub struct DecryptCipherListResult {
    pub successes: Vec<CipherListView>,
    pub failures: Vec<Cipher>,
}
```

What are the plans for decrypting individual items, do we need a graceful handling of those items?

crates/bitwarden-vault/src/cipher/cipher.rs
```diff
@@ -249,6 +249,64 @@ pub struct CipherListView {
...
+    pub collection_ids: Vec<Uuid>,
+    pub r#type: CipherType,
+    pub favorite: bool,
+    pub organization_use_totp: bool,
+    pub has_fido2: bool,
+    pub has_totp: bool,
+    pub deleted_date: Option<DateTime<Utc>>,
+}
+
+#[allow(missing_docs)]
```
Hinton: **issue:** Please do not add `#[allow(missing_docs)]`. We currently have a decent amount of `allow` statements in the code but that is primarily code we haven't updated yet. At a minimum always document the struct, and just skip the fields.

Hinton (commented):
I'd be curious if we would be alright with simplify the decrypt with failures and just return back the failed entries. We'd get rid of around 60 lines of code.

crates/bitwarden-vault/src/cipher/cipher.rs
gbubemismith: Great point! My original thinking was to return a smaller payload. I agree with returning the `Cipher`, as it adds some flexibility to how the clients plan to handle things.

>What are the plans for decrypting individual items, do we need a graceful handling of those items?

I think what we have for individual items is enough; we should never get to decrypting a cipher that has failed from the decrypt list, we show a modal instead with a message to contact CS

crates/bitwarden-crypto/src/store/mod.rs
gbubemismith: Fixed 659741c30d8d2dd07152e693a4b73ee7ef595950

crates/bitwarden-vault/src/cipher/cipher.rs
gbubemismith: Updated 925d199969fd3d0c8f0b9f18ed4204f7d6aa0e31

crates/bitwarden-crypto/src/store/mod.rs
```diff
@@ -259,6 +259,48 @@ impl<Ids: KeyIds> KeyStore<Ids> {
...
+    /// and the second vector contains the original items that failed to decrypt.
+    pub fn decrypt_list_with_failures<
+        'a,
+        Key: KeyId,
+        Data: Decryptable<Ids, Key, Output> + IdentifyKey<Key> + Send + Sync + 'a,
+        Output: Send + Sync,
+    >(
+        &self,
+        data: &'a [Data],
+    ) -> (Vec<Output>, Vec<&'a Data>) {
```
Hinton, line 279: @dani-garcia do you think it would be nicer to always take ownership of the failed items? I.e. require `Data` to implement clone and clone the failures on line 298.

I have a hard time imagining scenarios where we don't eventually clone things.

```suggestion
    pub fn decrypt_list_with_failures<
        Key: KeyId,
        Data: Decryptable<Ids, Key, Output> + IdentifyKey<Key> + Send + Sync + Clone,
        Output: Send + Sync,
    >(
        &self,
        data: &[Data],
    ) -> (Vec<Output>, Vec<Data>) {
```

Hinton (approved):
Thanks for iterating, I think we got rid of much complexity.

crates/bitwarden-crypto/src/store/mod.rs
dani-garcia: There may be a few cases where not doing cloning here could be an improvement, for example if we ever do some filtering of the decrypted cipher list on the SDK side, we might be able to only do cloning after filtering.

I agree that it's probably a very small improvement though.

dani-garcia (approved):

PR #333, opened 2025-06-26
Ensure we can always build bitwarden-sm

Hinton:
## 📔 Objective

We need to ensure `bitwarden-sm` is always buildable to avoid breaking sdk-sm.

## ⏰ Reminders before review

- CI builds passed
  team

## 🦮 Reviewer guidelines

### DIFF
--- a/crates/bitwarden-core/src/client/internal.rs
+++ b/crates/bitwarden-core/src/client/internal.rs
@@ -10,9 +10,10 @@ use bitwarden_state::registry::StateRegistry;
 use chrono::Utc;
 use uuid::Uuid;
 
-use super::encryption_settings::EncryptionSettings;
+#[cfg(any(feature = "internal", feature = "secrets"))]
+use crate::client::encryption_settings::EncryptionSettings;
 #[cfg(feature = "secrets")]
-use super::login_method::ServiceAccountLoginMethod;
+use crate::client::login_method::ServiceAccountLoginMethod;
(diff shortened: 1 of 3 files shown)


quexten (approved):
LGTM for crypto changes

crates/bitwarden-core/src/client/internal.rs
```diff
@@ -10,9 +10,10 @@ use bitwarden_state::registry::StateRegistry;
 use chrono::Utc;
 use uuid::Uuid;
 
-use super::encryption_settings::EncryptionSettings;
+#[cfg(any(feature = "internal", feature = "secrets"))]
```
dani-garcia, line 13: Non blocking thought: I've noticed that we have a couple of cfgs like this in the crate, even though I don't think it makes much sense to use the `bitwarden-core` crate without at least one of the two features enabled.

I'm wondering if we should just remove them all and just add this to the lib.rs file:
```rust
#[cfg(not(any(feature = "internal", feature = "secrets")))]
compile_error!("Either the 'internal' or 'secrets' feature must be enabled.");
```

dani-garcia (approved):

crates/bitwarden-core/src/client/internal.rs
Hinton: Yea that'd be fine for me but I think it would break running cargo hack so something we can look into later.

PR #335, opened 2025-06-27
[PM-24127] Implement password-protected key envelope

quexten:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-24127
## 📔 Objective

The current masterkey logic is complex to understand for using teams (auth), and also prone to error. When any setting changes / gets out of sync, such as the email, or kdf, then decryption fails. The masterkey is further too widely scoped, used both in an authentication protocol, and in unlock decryption.

This PR introduces a PasswordProtectedKeyEnvelope. The goal is to protect a symmetric key with a password securely. Internally, this uses a KDF, and the KDF settings (argon2 parameters, and random salt) are stored on the serialized object.

That means that the only thing needed to unlock this structure is the correct password, everything else is stored on the object, making this process much less error prone. At the same time the interface is easier to use.

An example is provided to show usage.

A follow-up PR will add an unlock method / enrollment for PIN based on this new cryptographic API.

Note: Only argon2 is supported here. The PasswordProtectedKeyEnvelope's settings are completely decoupled from the account settings, and we don't need to provide backwards compatibility to non-recommended legacy cryptographic algorithms (pbkdf2).

## ⏰ Reminders before review

- CI builds passed
  team

## 🦮 Reviewer guidelines


crates/bitwarden-crypto/examples/protect_key_with_password.rs
```diff
@@ -0,0 +1,117 @@
...
+    // For this, the `PasswordProtectedKeyEnvelope` is used.
+
+    // Alice has a vault protected with a symmetric key. She wants this protected with a PIN.
+    let vault_key = ctx
+        .generate_symmetric_key(ExampleSymmetricKey::VaultKey)
+        .expect("Generating vault key should work");
+
+    // Seal the key with the PIN
+    // The KDF settings are chosen for you, and do not need to be separately tracked or synced
+    // Next, story this protected key envelope on disk.
```
Thomas-Avery:
```suggestion
    // Next, store this protected key envelope on disk.
```

crates/bitwarden-crypto/examples/protect_key_with_password.rs
```diff
@@ -0,0 +1,117 @@
...
+    let pin = "1234";
+    let envelope =
+        PasswordProtectedKeyEnvelope::seal(vault_key, pin, &ctx).expect("Sealing should work");
```
Thomas-Avery, line 33: 💭  
It is a little odd here in the editor to be passing in a `pin` for the `password` parameter. I wonder if a more umbrella term like `SecretProtectedKeyEnvelope::seal(key_to_seal, secret, &ctx)` would be a better fit?

We already have thought of a few use cases, and I'm sure the types of authentication secrets used will expand in the future.

I could see an argument for some consumers being more confused with `secret` terminology.  I have no problems moving forward with `PasswordProtectedKeyEnvelope` just think it's worth talking about/considering.

crates/bitwarden-crypto/examples/protect_key_with_password.rs
```diff
@@ -0,0 +1,117 @@
...
+    let mut disk = MockDisk::new();
+
+    // Alice wants to protect a key with a password.
+    // For example to:
+    // - Protect her vault with a pin
+    // - Protect her exported vault with a password
+    // - Protect a send with a URL fragment secret
+    // For this, the `PasswordProtectedKeyEnvelope` is used.
+
+    // Alice has a vault protected with a symmetric key. She wants this protected with a PIN.
```
Thomas-Avery:
```suggestion
    // Alice has a vault protected with a symmetric key. She wants the symmetric key protected with a PIN.
```

crates/bitwarden-crypto/examples/protect_key_with_password.rs
```diff
@@ -0,0 +1,117 @@
...
+        .expect("The password should be valid");
+    disk.save(
+        "vault_key_envelope",
+        (&envelope).try_into().expect("Saving envelope should work"),
+    );
+
+    // Alice wants to change the protected key. This requires creating a new envelope
+    ctx.generate_symmetric_key(ExampleSymmetricKey::VaultKey)
+        .expect("Generating vault key should work");
+    let envelope = PasswordProtectedKeyEnvelope::seal(ExampleSymmetricKey::VaultKey, "0000", &ctx)
```
Thomas-Avery, line 60:
```suggestion
   let new_key = ctx.generate_symmetric_key(ExampleSymmetricKey::VaultKey)
        .expect("Generating vault key should work");
    let envelope = PasswordProtectedKeyEnvelope::seal(new_key, "0000", &ctx)
        .expect("Sealing should work");
```

Don't we need to pass in the newly generated key?

crates/bitwarden-crypto/examples/protect_key_with_password.rs
```diff
@@ -0,0 +1,117 @@
...
+//! [PasswordProtectedKeyEnvelope].
+
+use bitwarden_crypto::{
+    key_ids,
+    safe::{PasswordProtectedKeyEnvelope, PasswordProtectedKeyEnvelopeError},
+    KeyStore, KeyStoreContext,
+};
+
+fn main() {
+    let key_story = KeyStore::<ExampleIds>::default();
```
Thomas-Avery:
```suggestion
    let key_store = KeyStore::<ExampleIds>::default();
```

crates/bitwarden-crypto/src/safe/README.md
```diff
@@ -0,0 +1,12 @@
+# Bitwarden-crypto safe module
+
+The safe module contains a high-level set of tools used in building protocols and features involving
+cryptography. Whenever possible, a feature should be build with features from this module, before
+opting to build with any other, more lower-level primitives in the `bitwarden-crypto` crate.
```
```suggestion
The safe module provides high-level cryptographic tools for building secure protocols and features. When developing new features, use this module first before considering lower-level primitives from other parts of `bitwarden-crypto`.
```

crates/bitwarden-crypto/src/safe/README.md
```diff
@@ -0,0 +1,12 @@
...
+
+## Password-protected key envelope
+
+The password protected key envelope should be used, when the goal is to protect a symmetric key with
+a password, for example for locking a vault with a PIN/Password, for protecting exports with a
+password, etc. Internally, a KDF is used to protect against brute-forcing, but this is not exposed
+to the consumer. The consumer only provides a password and key.
```
```suggestion
Use the password protected key envelope to protect a symmetric key with a password. Examples include:

- locking a vault with a PIN/Password
- protecting exports with a password

Internally, the module uses a KDF to protect against brute-forcing, but it does not expose this to the consumer. The consumer only provides a password and key.
```

crates/bitwarden-crypto/src/safe/password_protected_key_envelope.rs
```diff
@@ -0,0 +1,572 @@
...
+};
+
+/// A password-protected key envelope can seal a symmetric key, and protect it with a password. It
+/// does so by using a Key Derivation Function (KDF), to increase the difficulty of brute-forcing
+/// the password.
+///
+/// The KDF parameters such as iterations and salt are stored in the key-envelope and do not have to
+/// be provided.
+///
+/// Internally, Argon2 as the KDF, and XChaCha20-Poly1305 are used to encrypt the message.
```
Thomas-Avery:
```suggestion
/// The KDF parameters such as iterations and salt are stored in the envelope and do not have to
/// be provided.
///
/// Internally, Argon2 is used as the KDF and XChaCha20-Poly1305 is used to encrypt the key.
```

crates/bitwarden-crypto/examples/protect_key_with_password.rs
Thomas-Avery: Looking back at this after reviewing the implementation. This is actually the `id` of the key inside the `key_store` that we are passing around. Makes more sense, feel free to disregard.

crates/bitwarden-crypto/src/safe/password_protected_key_envelope.rs
```diff
@@ -0,0 +1,572 @@
...
+
+        let kdf_settings: Argon2RawSettings =
+            (&recipient.unprotected).try_into().map_err(|_| {
+                PasswordProtectedKeyEnvelopeError::ParsingError(
+                    "Invalid or missing KDF parameters".to_string(),
+                )
+            })?;
+        let envelope_key = derive_key(&kdf_settings, password)
+            .map_err(|_| PasswordProtectedKeyEnvelopeError::KdfError)?;
+        let nonce: [u8; 24] = self
```
Thomas-Avery:
```suggestion
        let nonce: [u8; crate::xchacha20::NONCE_SIZE] = self
```

Should we be using the same constant as we did in `seal_ref_with_settings`?

crates/bitwarden-crypto/src/safe/password_protected_key_envelope.rs
```diff
@@ -0,0 +1,572 @@
...
+    cose_encrypt: coset::CoseEncrypt,
+}
+
+impl<Ids: KeyIds> PasswordProtectedKeyEnvelope<Ids> {
+    /// Seals a symmetric key with a password, using the current default KDF parameters and a random
+    /// salt.
+    ///
+    /// This should never fail, except for memory allocation error, when running the KDF.
+    pub fn seal(
+        key_to_seal: Ids::Symmetric,
```
Thomas-Avery, line 60:
```suggestion
        keyslot_to_seal: Ids::Symmetric,
```
Something to make it clearer, this is the ID of the key. Maybe this is fine how it is and we just rely on the type system.

crates/bitwarden-crypto/src/safe/password_protected_key_envelope.rs
```diff
@@ -0,0 +1,572 @@
...
+                .map_err(|_| {
+                    PasswordProtectedKeyEnvelopeError::ParsingError(
+                        "Invalid Argon2 salt".to_string(),
+                    )
+                })?,
+        })
+    }
+}
+
+fn make_salt() -> [u8; 32] {
```
Thomas-Avery: We have `32` in a few places as the fixed salt size. Consider making a constant instead of "magic numbers".

Thomas-Avery (changes requested):
I really like this new interface, great work. Some minor things to take a look at from me.

crates/bitwarden-crypto/examples/protect_key_with_password.rs
quexten: I wonder what happened here, this typo seems to be a pattern...

crates/bitwarden-crypto/examples/protect_key_with_password.rs
quexten: We have a tech debt ticket inherited from Dani's initial work on this here: https://bitwarden.atlassian.net/browse/PM-18102, which would update the api to not require specifying a static keyslot that could have collisions, but always getting a new ID by the API.

That would make this API more intuitive / less error prone to use.

crates/bitwarden-crypto/src/safe/password_protected_key_envelope.rs
quexten: Hmm, `keyslot_to_seal` suggests that you are sealing the keyslot, but you are sealing the key referenced by/"stored in" the keyslot. That said, I'm OK with this naming too.

I believe the forms we currently use are `key_id*`, `key*`, inconsistently. Sometimes even within a single function's set of arguments: https://github.com/bitwarden/sdk-internal/blob/481349272195b330c52610f64b4b7f3559b0e25b/crates/bitwarden-crypto/src/store/context.rs#L153

So we should probably figure out a naming scheme to move towards.

To me, `key` without any suffix, feels natural because just like when passing a reference to a struct `&MyStruct`, you are omitting the fact that it is a reference usually, letting the consumer infer that by the type system, similarly, here, the fact that it is a key referenced in a keyslot, should be apparent from the type, and the fact that most APIs (and eventually all APIs) in `bitwarden-crypto` that are exposed only ever work with keyslots.

However, I can see `keyslot` being useful too, I'd only like to move away from `key_id`, since we now use that in another context (cose key ids).

crates/bitwarden-crypto/examples/protect_key_with_password.rs
quexten: A pin is also a password, just typically with lower entropy, but granted that may not be clear to callers. On the other hand, `Secret` seems to capture more than just PIN/Password. A symmetric key is also a secret? Symmetric key cryptography is also known as **secret**-key cryptography (and asymmetric as public-key cryptography).

Let's move forward with `PasswordProtectedKeyEnvelope` for now, and re-evaluate in case we find that we are adding more variants.

crates/bitwarden-crypto/src/safe/password_protected_key_envelope.rs
quexten: Replaced all the magic numbers and moved the constants up in the file.

crates/bitwarden-crypto/examples/protect_key_with_password.rs
Thomas-Avery: Sounds good, thank you for the discussion.

crates/bitwarden-crypto/src/safe/password_protected_key_envelope.rs
Thomas-Avery: I thought about this some more. My preference here is to keep with `key_to_seal` or use your suggestion `key`.

Can we get argument comment blocks on these public methods? Like https://github.com/bitwarden/sdk-internal/blob/481349272195b330c52610f64b4b7f3559b0e25b/crates/bitwarden-crypto/src/store/context.rs#L143-L149 I think that will make things clear for consumers.

quexten: `key` alone should generally not be used, sorry my phrasing above was a bit confusing because it ended up suggesting that. I meant `key` + purpose, or `key_id` + purpose, so something like `wrapping_key`, or `key_to_be_wrapped`, but *never* key alone.


On the point of function arguments, in rust function arguments are generally not documented, and they (rust team) have been explicit about removing an examples around it: https://bitwarden.slack.com/archives/C054ZQSBS49/p1754994308725229

The referenced code was added by me, while I still did not know about this. The contributing docs are still lacking on this end currently.

justindbaur (dismissed):

Thomas-Avery (dismissed):
LGTM

.vscode/settings.json
```diff
@@ -15,6 +15,7 @@
     "encryptable",
     "Hkdf",
     "Hmac",
+    "keyslot",
```
Hinton: **issue:** Keyslot was removed.

crates/bitwarden-crypto/src/store/context.rs
```diff
@@ -309,6 +309,18 @@ impl<Ids: KeyIds> KeyStoreContext<'_, Ids> {
...
+    pub(crate) fn make_cose_symmetric_key(
+        &mut self,
+        key_id: Ids::Symmetric,
+    ) -> Result<Ids::Symmetric> {
+        let key = SymmetricCryptoKey::make_xchacha20_poly1305_key();
+        #[allow(deprecated)]
+        self.set_symmetric_key(key_id, key)?;
+        Ok(key_id)
+    }
+
```
Hinton, line 323: **question:** This is a pretty basic helper only used in a single test file. Do you anticipate this being used in more places?

crates/bitwarden-crypto/src/safe/README.md
Hinton, line 1: **question:** Are we planning on implementing more _safe_ interfaces, and what decides if something is _safe_ or not? Is everything else in the care _unsafe_? Such as decrypting `EncString`s?

I suppose the goal is to ask people to use the higher level abstractions here rather than rolling their own using primitives?

crates/bitwarden-crypto/examples/protect_key_with_password.rs
```diff
@@ -0,0 +1,118 @@
...
+
+    // Alice wants to change her password; also her KDF settings are below the minimums.
+    // Re-sealing will update the password, and KDF settings.
+    let envelope = envelope
+        .reseal(pin, "0000")
+        .expect("The password should be valid");
+    disk.save(
+        "vault_key_envelope",
+        (&envelope).try_into().expect("Saving envelope should work"),
+    );
```
Hinton: **question:** Do you have a strategy for how to migrate infrequently changed keys? You could imagine a user setting up pin once and keeping it for years. In which time we may have upgraded the minimum kdf settings.

crates/bitwarden-crypto/src/cose.rs
```diff
@@ -22,10 +23,16 @@ use crate::{
...
+pub(crate) const ARGON2_SALT: i64 = -71001;
+pub(crate) const ARGON2_ITERATIONS: i64 = -71002;
+pub(crate) const ARGON2_MEMORY: i64 = -71003;
+pub(crate) const ARGON2_PARALLELISM: i64 = -71004;
+
 // Note: These are in the "unregistered" tree: https://datatracker.ietf.org/doc/html/rfc6838#section-3.4
 // These are only used within Bitwarden, and not meant for exchange with other systems.
 const CONTENT_TYPE_PADDED_UTF8: &str = "application/x.bitwarden.utf8-padded";
-const CONTENT_TYPE_BITWARDEN_LEGACY_KEY: &str = "application/x.bitwarden.legacy-key";
+pub(crate) const CONTENT_TYPE_BITWARDEN_LEGACY_KEY: &str = "application/x.bitwarden.legacy-key";
```
Hinton: **question:** Does this need to be public and is there a reason we can't use `impl TryFrom<&coset::Header> for ContentFormat {` where this is consumed?

crates/bitwarden-crypto/src/cose.rs
```diff
@@ -221,6 +228,52 @@ pub trait CoseSerializable<T: CoseContentFormat + ConstContentFormat> {
...
+            (Label::Int(label_value), ciborium::Value::Integer(int_value))
+                if *label_value == target_label =>
+            {
+                Some(*int_value)
+            }
+            _ => None,
+        })
+        .ok_or(CoseExtractError::MissingValue(value_name.to_string()))?
+        .into())
+}
```
Hinton: **nit**: You can use `ok_or_else` here and get rid of the `?` and wrapping `Ok`.

```suggestion
pub(crate) fn extract_integer(
    header: &Header,
    target_label: i64,
    value_name: &str,
) -> Result<i128, CoseExtractError> {
    header
        .rest
        .iter()
        .find_map(|(label, value)| match (label, value) {
            (Label::Int(label_value), ciborium::Value::Integer(int_value))
                if *label_value == target_label =>
            {
                Some(*int_value)
            }
            _ => None,
        })
        .map(Into::into)
        .ok_or_else(|| CoseExtractError::MissingValue(value_name.to_string()))
}
```

crates/bitwarden-crypto/src/cose.rs
```diff
@@ -221,6 +228,52 @@ pub trait CoseSerializable<T: CoseContentFormat + ConstContentFormat> {
...
+            _ => None,
+        })
+        .ok_or(CoseExtractError::MissingValue(value_name.to_string()))
+}
+
+#[derive(Debug, Error)]
+pub(crate) enum CoseExtractError {
+    #[error("Missing value {0}")]
+    MissingValue(String),
+}
```
Hinton, line 275: **question:** Do you anticipate more error variants the above functions? If not you extract the error from the enum.

```rust
#[derive(Debug, Error)]
#[error("Missing value {0}")]
pub struct MissingValueError(String);
```

crates/bitwarden-crypto/src/safe/password_protected_key_envelope.rs
```diff
@@ -0,0 +1,590 @@
...
+    /// Seals a key reference with a password. This function is not public since callers are
+    /// expected to only work with key store references.
+    fn seal_ref(
+        key_to_seal: &SymmetricCryptoKey,
+        password: &str,
+    ) -> Result<Self, PasswordProtectedKeyEnvelopeError> {
+        Self::seal_ref_with_settings(
+            key_to_seal,
+            password,
+            &Argon2RawSettings::default_for_platform(),
```
Hinton: **question:** Would we ever want to apply a more restrictive default if we get say a numeric low digit pin? Like bumping up the iteration count to slightly slow down brute force attacks.

crates/bitwarden-crypto/src/safe/password_protected_key_envelope.rs
```diff
@@ -0,0 +1,590 @@
...
+        // There must be exactly one recipient in the COSE Encrypt object, which contains the KDF
+        // parameters.
+        if self.cose_encrypt.recipients.len() != 1 {
+            return Err(PasswordProtectedKeyEnvelopeError::ParsingError(
+                "Invalid number of recipients".to_string(),
+            ));
+        }
+
+        let recipient = self.cose_encrypt.recipients.first().ok_or(
+            PasswordProtectedKeyEnvelopeError::ParsingError("Missing recipient".to_string()),
```
Hinton: **issue (non blocking):** This error can never happen, you just checked recipients length above.

I'm not sure if it's a significant improvement but you could maybe write it as

```rust
let recipient = self
    .cose_encrypt
    .recipients
    .first()
    .filter(|_| self.cose_encrypt.recipients.len() == 1)
    .ok_or_else(|| {
        PasswordProtectedKeyEnvelopeError::ParsingError(
            "Invalid number of recipients".to_string(),
        )
    })?;
```

crates/bitwarden-crypto/src/safe/password_protected_key_envelope.rs
```diff
@@ -0,0 +1,590 @@
...
+    fn try_from(value: &Vec<u8>) -> Result<Self, Self::Error> {
+        let cose_encrypt = coset::CoseEncrypt::from_slice(value)?;
+        Ok(PasswordProtectedKeyEnvelope {
+            _phantom: PhantomData,
+            cose_encrypt,
+        })
+    }
+}
+
+/// Raw argon2 settings differ from the KDF struct defined for existing master-password unlock.
```
Hinton: **nit:** I think you could link to the existing KDF.

```suggestion
/// Raw argon2 settings differ from the [Kdf::Argon2id] struct defined for existing master-password unlock.
```

Having two ways to represent argon kdf settings is annoying though. I suppose KiB would be more true to the spec and library as we currently multiply it ourselves.

crates/bitwarden-crypto/src/safe/password_protected_key_envelope.rs
```diff
@@ -0,0 +1,590 @@
...
+            // clients except desktop.
+            Self {
+                iterations: 3,
+                memory: 64 * 1024, // 64 MiB
+                parallelism: 4,
+                salt: make_salt(),
+            }
+        }
+    }
+}
```
Hinton: **question:** Does the use of platform specific KDF settings imply these should be device bound and not be synced? I could see a desire to share PasswordProtected envelopes.

**suggestion:** Highlight this to consumers to prevent any potential misuse. Or ideally name them local or device bound.

crates/bitwarden-crypto/src/safe/password_protected_key_envelope.rs
```diff
@@ -0,0 +1,590 @@
...
+        header.alg = Some(coset::Algorithm::PrivateUse(ALG_ARGON2ID13));
+        header
+    }
+}
+
+impl TryInto<Params> for &Argon2RawSettings {
+    type Error = PasswordProtectedKeyEnvelopeError;
+
+    fn try_into(self) -> Result<Params, PasswordProtectedKeyEnvelopeError> {
+        Params::new(self.memory, self.iterations, self.parallelism, Some(32))
```
Hinton: **question:** What is this magic number? Can we name it?

Hinton (commented):
Looks great, a couple of things I observed.

The open questions I have, do we want to bake in a way to upgrade kdf's in the future? Clear up up any potential confusion of what _safe_ means, and a magic number.

crates/bitwarden-crypto/src/store/context.rs
quexten: We currently have `generate_symmetric_key`, which always generates aes256-cbc-hmac keys. I'm still undecided whether the caller should get a choice about what key type to create, or whether this should be decided by the crypto crate. (Most likely we have to give the caller a choice, since we want to be mindful about where we upgrade permanent vault formats.)

I would like to use this function in more places to get test coverage for both the old and new key type. At the same time it's flagged for test only, so that noone else uses it just yet, thinking it's fine to consume.

crates/bitwarden-crypto/src/safe/README.md
quexten: Yeah, the goal here is higher-level abstractions with the general goals of:

- Safe to use
- Easy mental model (high level)
- Hard to misuse
Though, I don't think we have exact criteria defined yet.

Each of these should have an easy to understand usage story associated with it. "I want to protect a key, and have a password. That is done with a PasswordProtectedKeyEnvelope". 

The next safe interface I thought about proposing - it is still draft / RFC - is DataEnvelope: https://github.com/bitwarden/sdk-internal/pull/336, with the usage story: "I have a struct of data and want to protect it. That is done with a DataEnvelope, and I receive back a sealed envelope and an envelope key".

crates/bitwarden-crypto/examples/protect_key_with_password.rs
quexten: For the pin-specific case this can be done by either:
- Using re-seal(pin) on unlock with the pin
- Using the `userkey_encrypted_pin`

I'm not sure whether the FFI interface should force this migration by returning an optional update version of the envelope, or whether this should be a manual process called by each client. 


For master-password, we would only be able to use re-seal(master-password) on login/unlock with MP, since we don't store `userkey_encrypted_master_password`.

crates/bitwarden-crypto/src/safe/password_protected_key_envelope.rs
quexten: Renamed.

As for the question, its' fine to sync these, but syncing these has a few implications:

- If the target device cannot handle the settings, this may cause issues (such as we are currently still seeing with high argon2 memory + iOS autofill) and differing unsealing times

- Differing unsealing times may be OK, if it is either short on all devices, or slow on some but is also an infrequent action (login)

These are not different from the currently shared master-key-encrypted-userkey, which has the same constraints.

crates/bitwarden-crypto/src/safe/password_protected_key_envelope.rs
quexten: I think that makes sense. The argon2 RFC [[1]](https://datatracker.ietf.org/doc/rfc9106/) has a procedure specified on how to pick argon2 settings. They do also specify a method where you first pick a fixed memory setting that is uniformly safe, then auto-adjust the iteration count based on acceptable unlock time. We could implement this fairly easily.

However, only applying it on weak passwords would leak that the password is a weak password by looking at the envelope's KDF settings. I'm not sure if that's something we are concerned about.

crates/bitwarden-crypto/src/cose.rs
quexten: Updated the consumer to:
```
SymmetricCryptoKey::try_from(
            match ContentFormat::try_from(&self.cose_encrypt.protected.header).map_err(|_| {
                PasswordProtectedKeyEnvelopeError::ParsingError(
                    "Invalid content format".to_string(),
                )
            })? {
                ContentFormat::BitwardenLegacyKey => EncodedSymmetricKey::BitwardenLegacyKey(
                    BitwardenLegacyKeyBytes::from(key_bytes),
                ),
                ContentFormat::CoseKey => {
                    EncodedSymmetricKey::CoseKey(CoseKeyBytes::from(key_bytes))
                }
                _ => {
                    return Err(PasswordProtectedKeyEnvelopeError::ParsingError(
                        "Unknown or unsupported content format".to_string(),
                    ));
                }
            },
        )
```

And made the content type non-public again.

crates/bitwarden-crypto/src/safe/password_protected_key_envelope.rs
quexten: Cleaned this up

crates/bitwarden-crypto/src/cose.rs
quexten: I think there are more variants that we get to as we re-use these functions in other places that parse cose (decrypt, verify signatures) and want better errors.

- Duplicate values (should this throw?)
- Value was found, but has wrong type (currently mapped to "MissingValue")

quexten:
Noting here: I lifted in a fair bit of non-pin-specific changes from the follow-up PR to this PR since the last set of reviews.

crates/bitwarden-crypto/src/safe/password_protected_key_envelope.rs
Hinton: The difference with user key is that the user picks the values. This interface does not allow you to customize it by design and therefore if someone creates a value outside of iOS and it's synced it would be the same as not having a lower limit for iOS in the first place.

quexten: I think for the synced vs local use-case, we should just have two separate functions. Maybe the synced case would be `seal_with_custom_kdf`, where the caller provides the KDF setting. 

Since we don't use this for the synced use-case currently, we don't implement that functionality yet.

crates/bitwarden-crypto/examples/protect_key_with_password.rs
Hinton: Since this is a new interface now would be the optimal time to decide how it should be handled as the scope to introduce this later will be significantly larger. **But it's also completely valid to postpone it.**

Ideally there would be a pattern that forces the consumer to handle it, since it's often easy to miss. Throwing an error on initialization and forcing them through an upgrade path before being allowed to consume maybe. But this could also easily be missed as we probably don't have a good way to test 

Going Vec<u8> to a wrapper, `enum { Envelope(PaasswordProtectedKeyenvelope), UpgradeRequired }` maybe?

crates/bitwarden-crypto/src/cose.rs
Hinton: That's fair.

crates/bitwarden-crypto/src/safe/password_protected_key_envelope.rs
Hinton: Hmm true, you could let the caller decide if it's "high probability of low entropy" like pins since I suspect many users use pin when they have complex MP. But I'm good marking this as out of scope, mostly wanted you to think of this scenario.

crates/bitwarden-crypto/src/safe/README.md
Hinton: I think it's worth adding a reference to the `safe` module from the root readme in crypto? It might be hard for people to find it otherwise.

```
## High level abstractions

The crate provides high level abstractions under the `safe` namespace. These should be heavily prioritized over the lower level primitives when developing new functionality.
```

crates/bitwarden-crypto/src/safe/password_protected_key_envelope.rs
Hinton: What about including a `Purpose` enum when you create the envelope? That would make it clear for consumers you shouldn't sync this key yet. But I don't have a strong preference and this should be seen as **non blocking**.

```rust
enum Purpose {
  Local,
  // Synced TBD
}
```

Hinton (approved):
My primary concerns are resolved. I'd still like to see some elegant way to force upgrades as adding it later will be significantly more complicated. And a note to the root readme to look in `safe` before implementing anything.

But I'll leave it up to the team to decide if it happens now or when the need arises.

crates/bitwarden-crypto/src/safe/password_protected_key_envelope.rs
quexten: Yeah, that looks very nice. I'll not add it in this PR, but will look into adding this as follow-up.

crates/bitwarden-crypto/examples/protect_key_with_password.rs
Thomas-Avery: 💭 
I would think we want whatever owns the storage layer for the envelopes to own being responsible for migrating these. At the moment, that would be the clients but could be the SDK in the future. `enum { Envelope(PaasswordProtectedKeyenvelope), UpgradeRequired }`  reminds me of the ASP.NET Core Identity's `PasswordHasher` and makes sense to me since it is flexible to storage ownership changes.

For TS clients, KM could just own a storage service layer `unsealFromStorage(pin)` and that would handle saving a new `reseal` if `UpgradeRequired` on use.

@quexten would this be a big increase in scope?

quexten: I think we should look into this whenever we also want to take on the work to manage PINs with SDK state management, especially since it is otherwise tricky to make forced upgrades work nicely with crypto initialization of the SDK. And the scope increase to do that here would be significant.

Thomas-Avery (approved):
LGTM

justindbaur (approved):

PR #336, opened 2025-06-27
[PM-26459] Implement data envelope

quexten:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-26459
https://bitwarden.atlassian.net/wiki/spaces/EN/pages/2052653095/Draft+Tech+Breakdown+-+DataEnvelope

## 📔 Objective

> [!IMPORTANT]  
> The confluence document has a more detailed breakdown. This is a shortened version.

Consumers want to protect complex structs such as:
- Vault exports
- Ciphers (vault items)
- Organization reports

Currently, they serialize these themselves (organization reports are serialized as json), and encrypt fields individually. This is slow to handle during encrypt/decrypt, and complex to maintain. The server now has to know the same representation as clients. At the same time, this can be abused by the server omitting certain fields, or swapping encrypted fields encrypted under the same key.

EncStrings are not the correct abstraction layer. A new abstraction / API that is safe and easier to use (an impossible to misuse) is needed.

### Security Definition

**Attacker model:** The relevant attacker model here is a fully compromised server. That’s what E2E encryption / zero knowledge protect against.

**Security Goals:** 

**SG1:** The structure should not be modifiable (malleable)
- Reason: This can be abused by the attacker by tampering with the encrypted item in (to the developer) unexpected ways

**SG2:** The attacker should not be able to infer anything about the contents of the structure aside from length
- Reason: We have a “Zero knowledge” product and customers expect their data to not be readable by anyone but them.

**SG3:** The structure should only be encryptable in the specific section of code that the developer intends
 - I.e it should not be possible to swap an encrypted vault item into an encrypted “user settings” “slot”, since the behavior here is undefined.
 - Reason: Same as SG1

This PR implements a "DataEnvelope". The DataEnvelope solves the problem of encrypting entire structs. The caller just provides a struct that is Serializable/Deserializable, and gets back an encrypted blob. The caller does not decide serialization.

Further, we force the creation of a "content encryption key" by the interface. This ensures that teams do not create a coupling to keys. When implementing key-rotation we do not want to re-upload data, but only re-upload re-encrypted keys, and the content-encryption-key enforces this key being present.

An example is provided to show usage.

## ⏰ Reminders before review

- CI builds passed
  team

## 🦮 Reviewer guidelines


crates/bitwarden-crypto/src/safe/data_envelope_namespace.rs
```diff
@@ -0,0 +1,55 @@
...
+
+/// Data envelopes are domain-separated within bitwarden, to prevent cross protocol attacks.
+///
+/// A new struct shall use a new data envelope namespace. Generally, this means
+/// that a data envelope namespace has exactly one associated valid message struct. Internal
+/// versioning within a namespace is permitted and up to the domain owner to ensure is done
+/// correctly.
+///
+/// If there is a new version of a message added, it should (generally) use a new namespace, since
+/// this prevents downgrades to the old type of message, and makes optional fields unnecessary.
```
nikwithak: What are you considering as a "new version" of a message, as opposed to "internal versioning" in this case? e.g. with cipher versioning, we may end up with multiple variations of the data field, and a growing list of types of data (wrapped as an enum, [for example](https://github.com/bitwarden/sdk-internal/pull/455/files#diff-fff269bd815a538fcc6ff54b2738abc2154a1ddae96ac37083e7de821f2de74fR13-R19)). Would that mean we need a new namespace in this file every time we introduce a change that isn't backwards compatible?

quexten: Internal versioning seems fine for vault items. The main thing we want to achieve here is to limit the scope of where a struct is valid, since currently all items / data encrypted under a key is validly decryptable in all other places (i.e a cipher key could be decrypted and interpreted as a username).

I agree, one namespace per cipher version is overkill. Since vault item versions can be captured as a single serializable struct/enum, only a single namespace is needed.

I would suggest just using an externally (or internally, depending on vault's preference) tagged enum here and passing that to the envelope. Each cipher version would be a variant of the enum, and lives within the same namespace, and you don't need to do anything else aside from making a new enum variant (which you presumably wanted to do anyways?).

https://serde.rs/enum-representations.html

quexten: I've updated the docs on dataenvelope and the example with a versioned vault item.

crates/bitwarden-crypto/src/ensure.rs
Hinton, line 1: I don't know if these macros provide much value, at best they save 2 lines of code but they also interup the control flow. I know we have precedence of this with `require!` ...

crates/bitwarden-crypto/src/error.rs
```diff
@@ -68,6 +71,9 @@ pub enum CryptoError {
     #[error("Signature error, {0}")]
     Signature(#[from] SignatureError),
 
+    #[error("DataEnvelope error, {0}")]
+    DataEnvelopeError(#[from] DataEnvelopeError),
```
Hinton: **question:** Could we inverse the relationship between `CryptoError` and `DataEvenlopeError`? I'm hesitant to continue expanding this error with more non generic scenarios.

crates/bitwarden-crypto/src/keys/symmetric_crypto_key.rs
```diff
@@ -71,6 +71,38 @@ impl PartialEq for Aes256CbcHmacKey {
 pub struct XChaCha20Poly1305Key {
     pub(crate) key_id: [u8; KEY_ID_SIZE],
     pub(crate) enc_key: Pin<Box<GenericArray<u8, U32>>>,
+    /// Controls which key operations are allowed with this key. Note: Only checking decrypt is
+    /// implemented right now. Further, disabling decrypt will also disable unwrap. The only
+    /// use-case so far is `DataEnvelope`.
```
Hinton: **question:** Do we want to track implementation status in jira?

crates/bitwarden-crypto/src/keys/symmetric_crypto_key.rs
```diff
@@ -71,6 +71,38 @@ impl PartialEq for Aes256CbcHmacKey {
...
+    #[zeroize(skip)]
+    pub(crate) key_operations: Vec<KeyOperation>,
```
Hinton: **question:** `key_operations` is somewhat ambiguous. Is this operations the key supports, or something else? (I.e. would calling it `supported_operations` be more clear, although this matches the cbor spec ...)

crates/bitwarden-crypto/src/safe/data_envelope.rs
```diff
@@ -0,0 +1,665 @@
...
+    envelope_data: CoseEncrypt0Bytes,
+}
+
+impl Clone for DataEnvelope {
+    fn clone(&self) -> Self {
+        DataEnvelope {
+            envelope_data: self.envelope_data.clone(),
+        }
+    }
+}
```
Hinton: **question:** I think deriving Clone does the same?

crates/bitwarden-crypto/src/safe/data_envelope.rs
```diff
@@ -0,0 +1,665 @@
...
+            .finish()
+    }
+}
+
+impl FromStr for DataEnvelope {
+    type Err = DataEnvelopeError;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        let data = B64::try_from(s).map_err(|_| {
+            DataEnvelopeError::ParsingError("Invalid DataEnvelope Base64 encoding".to_string())
```
Hinton: **question:** Should we just propagate the b64 error rather than converting it to a string?

crates/bitwarden-crypto/src/safe/data_envelope_namespace.rs
```diff
@@ -0,0 +1,52 @@
...
+/// Data envelopes are domain-separated within bitwarden, to prevent cross protocol attacks.
+///
+/// A new struct shall use a new data envelope namespace. Generally, this means
+/// that a data envelope namespace has exactly one associated valid message struct. Internal
+/// versioning within a namespace is permitted and up to the domain owner to ensure is done
+/// correctly.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum DataEnvelopeNamespace {
+    /// The namespace for vault items
+    VaultItem = 1,
```
Hinton, line 12: **question:** This is generally called `Cipher` in other places. Could this cause confusion?

Hinton (commented):
Overall looks great, some general thoughts.

crates/bitwarden-crypto/src/ensure.rs
quexten: Imo, in this case it's not so much about saving 2 lines, versus readability. The ensure is used in a cluster of checks. Thus, matches or an equality check seem (slightly) harder to parse than something that states "ensure_{matches/equal}":

```
ensure_matches!(msg.protected.header.alg, Some(coset::Algorithm::PrivateUse(XCHACHA20_POLY1305)) => DataEnvelopeError::DecryptionError);
ensure_equal!(msg.protected.header.key_id, cek.key_id => DataEnvelopeError::WrongKey);
ensure_equal!(envelope_namespace, *namespace => DataEnvelopeError::InvalidNamespace);
ensure_equal!(content_format, CONTENT_TYPE_PADDED_CBOR => DataEnvelopeError::UnsupportedContentFormat);
```

vs 
```
if !matches(msg.protected.header.alg, Some(coset::Algorithm::PrivateUse(XCHACHA20_POLY1305))) {
  return Err(DataEnvelopeError::DecryptionError);
}
if !msg.protected.header.key_id == cek.key_id {
  return Err(DataEnvelopeError::WrongKey);
}
if !envelope_namespace == *namespace {
  return Err(DataEnvelopeError::InvalidNamespace);
}
if !content_format == CONTENT_TYPE_PADDED_CBOR {
  return Err(DataEnvelopeError::UnsupportedContentFormat);
}

File: crates/bitwarden-crypto/src/keys/symmetric_crypto_key.rs (hunk 3 above)
Comment (quexten) [reply crates/bitwarden-crypto/src/keys/symmetric_crypto_key.rs]: Added a ticket to the comment  https://bitwarden.atlassian.net/browse/PM-27513.

File: crates/bitwarden-crypto/src/error.rs (hunk 2 above)
Comment (quexten) [reply crates/bitwarden-crypto/src/error.rs]: Removed it from the CryptoError.

File: crates/bitwarden-crypto/src/safe/data_envelope.rs (hunk 5 above)
Comment (quexten) [reply crates/bitwarden-crypto/src/safe/data_envelope.rs]: This may be left over from a previous iteration from the PR where the internals where more complex, not sure. Removed.

File: crates/bitwarden-crypto/src/safe/data_envelope.rs (hunk 6 above)
Comment (quexten) [reply crates/bitwarden-crypto/src/safe/data_envelope.rs]: Makes sense, changed!

File: crates/bitwarden-crypto/src/safe/data_envelope_namespace.rs (hunk 7 above)
Comment (quexten) [reply crates/bitwarden-crypto/src/safe/data_envelope_namespace.rs:12]: The vault team is free to rename this entry; However I'll argue that despite precedent of using the term for a long time, Cipher is extremely confusing, especially in the context of an end-to-end encrypted app. Thus I opted for simpler naming here.

I've added a comment clarifying that this is the same as "ciphers"

File: crates/bitwarden-crypto/examples/seal_struct.rs [hunk 8]
```diff
@@ -0,0 +1,112 @@
...
+    let mut disk = MockDisk::new();
+
+    let my_item: MyItem = MyItemV1 {
+        a: 42,
+        b: "Hello, World!".to_string(),
+    }
+    .into();
+
+    // Seal the item into an encrypted blob, and store the content-encryption-key in the context.
+    let (sealed_item, cek) = DataEnvelope::seal(my_item, &mut ctx).expect("Sealing should work");
```
Thomas-Avery: I would rewrite this comment to make it clear the `seal` stores the cek into context and we get the ID back. Maybe even rename `cek`, but that is obvious by the type.

crates/bitwarden-crypto/src/safe/README.md
```diff
@@ -14,3 +16,15 @@ include:
...
+
+Use the data envelope to protect a struct (document) of data. Examples include:
+
+- protecting a vault item
+- protecting metadata (name, etc.) of a collection
+- protecting a vault report
+
+The serialization of the data and the creation of a content encryption key is handled internally.
+Calling the API with a decrypted struct, the content encryption key and the encrypted data are
+returned.
```
```suggestion
Calling the API with a decrypted struct, the content encryption key ID and the encrypted data are
returned.
```

⛏️  Make it clearer we return the ID form the ctx store.

crates/bitwarden-crypto/src/safe/data_envelope.rs
```diff
@@ -0,0 +1,656 @@
...
+/// The struct serialization format is stable. You may not mutate the struct so that old structs
+/// no longer deserialize correctly. A change such as adding an optional field to your struct is
+/// permitted, anything breaking old formats is not.
+///
+/// Ideally, when creating a new struct, create a test vector (a sealed DataEnvelope for a test
+/// value), and create a unit test ensuring that it permanently deserializes correctly.
+///
+/// To make breaking changes, introduce a new version. This should use the
+/// `generate_versioned_sealable!` macro to auto-generate the versioning code. Please see the
+/// examples directory.
```
Thomas-Avery, line 44: ❓ I think I'm just confused here.

Trying to make sure I understand what a breaking change is in this context. 

Would adding a new optional field be a "breaking change" and require a new version? 

Is it that deserialization backwards compatibility is required per version? 

So I could have

```
struct MyItemV1 {
    a: u32,
    b: String,
}
```

And edit it to ?

```
struct MyItemV1 {
    a: u32,
    b: String,
    optional: Option<String>,
}
```

crates/bitwarden-crypto/src/safe/data_envelope.rs
```diff
@@ -0,0 +1,656 @@
...
+    }
+}
+
+#[cfg(feature = "wasm")]
+impl FromWasmAbi for DataEnvelope {
+    type Abi = <String as FromWasmAbi>::Abi;
+
+    unsafe fn from_abi(abi: Self::Abi) -> Self {
+        let s = unsafe { String::from_abi(abi) };
+        Self::from_str(&s).expect("Invalid DataEnvelope")
```
Thomas-Avery: ⚠️  panicking here instead of using `unwrap_throw`

crates/bitwarden-crypto/src/safe/data_envelope.rs
```diff
@@ -0,0 +1,656 @@
...
+            .to_vec()
+            .map(CoseEncrypt0Bytes::from)
+            .map_err(|_| DataEnvelopeError::EncodingError)?;
+
+        // Disable key operations other than decrypt on the CEK
+        cek.disable_key_operation(coset::iana::KeyOperation::Encrypt)
+            .disable_key_operation(coset::iana::KeyOperation::WrapKey)
+            .disable_key_operation(coset::iana::KeyOperation::UnwrapKey);
+
+        Ok((DataEnvelope { envelope_data }, cek.clone()))
```
Thomas-Avery:
```suggestion
        Ok((DataEnvelope { envelope_data }, cek))
```

I don't think the clone is needed?

Thomas-Avery (changes requested):
Looking good, have a few minor things and some questions/confusion around versioning.

crates/bitwarden-crypto/src/safe/data_envelope.rs
quexten: A breaking change is anything that serde can no longer deserialize. Usually this means changing the type of a field, adding a new field. If you are adding a new field, then as you point out above you can edit it to add the Option<T>. The consumer now has to match this after deserializing, so they may want to opt for a V2 struct anyways, since the above just progressively gets them to a looser and looser type format.

I don't want to tie this to rust or serde specifically, but in serde you can also specify a default value like so: https://serde.rs/attr-default.html.

quexten:
Claude, re: Critical issues:

1. The CEK keyid collision is just incoherent nonsense. Even if the key id collides, the key is different so nothing can be swapped between envelopes and SG1, SG3 are not broken.

2. This is not correct, we don't pass in any ephemeral AAD yes, but cose creates AAD from the protected headers. 

3. No, it does not. The mac is validated before the padding. Only the correct padded message decrypts.

4. This is not a security issue but a usability issue, we just don't want developers to mis-use the key for unintended operations.

As the review is not useful, I'm muting it.

crates/bitwarden-crypto/src/keys/symmetric_crypto_key.rs
quexten: Updated to `supported_operations`

crates/bitwarden-crypto/src/safe/data_envelope.rs
Thomas-Avery: Okay I would recommend making it clearer in these comments that a breaking change is any changes to the struct that would make old data no longer deserialize.

Something like:

> You may not mutate the struct so that old structs no longer deserialize correctly

To 

Struct modifications must maintain backward compatibility with existing serialized data. Changes that break deserialization are considered breaking changes and require a new version and struct.

quexten: Applied the suggested phrasing.

Thomas-Avery (approved):

crates/bitwarden-crypto/src/ensure.rs
Hinton: I'm just saying that in the first case to understand what's actually going in I need to jump into the macros. The syntax is somewhat confusing with the last argument being an expression.

Unwrapped everything I need to know to follow the execution flow is there.

crates/bitwarden-crypto/src/safe/data_envelope_namespace.rs
Hinton: CC @bitwarden/team-vault-dev

Hinton (approved):

crates/bitwarden-crypto/src/ensure.rs
quexten: That makes sense. These were intended to make readability nicer, if they don't I'll undo them. 

Looking for additional feedback @mzieniukbw and @Thomas-Avery. Do you have thoughts on readability here?

quexten: Actually, I'll make this an item for the next KM dev sync. I'll make a follow-up PR to undo the macros assuming the outcome is that they don't provide readability benefit.

quexten: Update: undoing this in a follow-up PR.

PR #337, opened 2025-06-30
[PM-23189] Add ClientManagedTokens trait

Hinton:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-23189

## 📔 Objective

Adds support for client managed access tokens alongside SDK managed access tokens.

It does so by splitting `Tokens` into `SdkManaged` and `ClientManaged`. `ClientManaged` relies on the consumer implementing a trait with `get_access_token` function.

In WASM this is exposed as a `TokenProvider` interface which can be implemented in TS clients using:

```ts
class JsTokenProvider implements TokenProvider {
  constructor(private apiService: ApiService) {}

  async get_access_token(): Promise<string> {
    return await this.apiService.getActiveBearerToken();
  }
}
```

## ⏰ Reminders before review

- CI builds passed
  team

## 🦮 Reviewer guidelines

### DIFF
--- a/crates/bitwarden-vault/src/folder_client.rs
+++ b/crates/bitwarden-vault/src/folder_client.rs
@@ -1,6 +1,5 @@
 use bitwarden_api_api::{apis::folders_api, models::FolderRequestModel};
 use bitwarden_core::{require, ApiError, Client, MissingFieldError};
-use bitwarden_crypto::Decryptable;
 use bitwarden_error::bitwarden_error;
 use bitwarden_state::repository::RepositoryError;
 use chrono::Utc;
--- a/crates/bitwarden-vault/src/folder_client.rs
+++ b/crates/bitwarden-vault/src/folder_client.rs
@@ -71,7 +71,8 @@ impl FoldersClient {
 
     /// Create a new folder and save it to the server.
     pub async fn create(&self, request: FolderAddEditRequest) -> Result<Folder, CreateFolderError> {
-        // TODO: We should probably not use a Folder model here, but rather create FolderRequestModel directly?
+        // TODO: We should probably not use a Folder model here, but rather create
+        // FolderRequestModel directly?
         let folder = self.encrypt(FolderView {
             id: None,
             name: request.name,
--- a/crates/bitwarden-vault/src/folder_client.rs
+++ b/crates/bitwarden-vault/src/folder_client.rs
@@ -1,51 +1,16 @@
-use bitwarden_api_api::{apis::folders_api, models::FolderRequestModel};
-use bitwarden_core::{require, ApiError, Client, MissingFieldError};
-use bitwarden_error::bitwarden_error;
-use bitwarden_state::repository::RepositoryError;
-use chrono::Utc;
-use serde::{Deserialize, Serialize};
-use thiserror::Error;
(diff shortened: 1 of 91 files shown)


crates/bitwarden-vault/src/folder_client.rs
```diff
@@ -35,4 +70,56 @@ impl FoldersClient {
         let views = key_store.decrypt_list(&folders)?;
         Ok(views)
     }
+
+    /// Create a new folder and save it to the server.
+    pub async fn create(
+        &self,
+        request: FolderAddEditRequest,
```
Hinton: We could probably implement `Encryptable` directly to `FolderAddEditRequest` which would encrypt into a `FolderRequestModel`?

crates/bitwarden-vault/src/folder_client.rs
```diff
@@ -35,4 +70,56 @@ impl FoldersClient {
...
+        .await
+        .map_err(ApiError::from)?;
+
+        let folder: Folder = resp.try_into()?;
+
+        self.client
+            .platform()
+            .state()
+            .get_client_managed::<Folder>()
+            .ok_or(MissingFieldError("Folder not found in state"))?
```
Hinton: **TODO**: Return a proper error. Might be nice if `get_client_managed` returned a `Result` instead and we didn't have to map it to error. Feels like a common use case? CC: @dani-garcia

crates/bitwarden-wasm-internal/src/platform/mod.rs
```diff
@@ -30,11 +31,17 @@ impl StateClient {
...
 impl StateClient {
     pub fn register_cipher_repository(&self, store: CipherRepository) {
         let store = store.into_channel_impl();
         self.0.platform().state().register_client_managed(store)
     }
+
+    pub fn register_folder_repository(&self, store: FolderRepository) {
+        let store = store.into_channel_impl();
+        self.0.platform().state().register_client_managed(store)
+    }
```
Hinton: @dani-garcia I wonder if the `create_wasm_repository` macro could wire this up for us?

crates/bitwarden-vault/src/folder_client.rs
dani-garcia: That sounds resonable yeah, I can imagine that we'll always consider it an error, so returning an option is going to be more cumbersome.

My hopes was that we could get into a situation were we can verify at compile time or startup that all the repositories have been registered and then this function would just panic, but haven't found a good way yet.

crates/bitwarden-wasm-internal/src/platform/mod.rs
dani-garcia: Yeah I think so, generating functions can be a bit opaque, so on #301 I've moved to a single `initialize_state` function that just receives all the repository impls to see how much of it could be macro generated.

My current thought is that we can just have some struct like:
```rust
struct RepositoryImpls {
 folder: FolderRepository, 
 cipher: CipherRepository,
}
```

And that could be easier to generate with a macro, maybe?

Hinton: Maybe.  Something to think about.

dani-garcia (approved):
LGTM

PR #339, opened 2025-06-30
[PM-22271] Expose KDF calculation in PureCrypto

quexten:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-22271
https://github.com/bitwarden/clients/pull/15401

## 📔 Objective

We want to drop our three different argon2 implementations on `clients` (desktop-native rustcrypto, patched argon2-browser WASM, native C++ module on cli). Each of these have impactful bugs (https://bitwarden.atlassian.net/browse/PM-17450, https://bitwarden.atlassian.net/browse/PM-22217), and tech debt (dependency updates, and the patching scripts for node modules) associated.

In this case, there is a small amount of tech debt introduced in the SDK because we introduce a deprecated function on PureCrypto. This small amount of tech debt introduced is far outweight by the dependencies that can be dropped on ts. (The clients PR: image)

Specifically, since we want to fix immediate bugs, we do not yet do the work of porting all higher-level usages (masterpassword-unlock, masterkeyhash auth, pin unlock/enrollment, send password "key"(hash) deriving) to the SDK. This is follow-up feature work, and larger in scope, and involves other teams (auth,tools). Once this follow-up feature-work is complete, the tech-debt function introduced in this ticket will be removed, and a tech-debt ticket for this is linked. 

The tech debt function is clearly marked as `dangerous_` and `deprecated` with a note.

## ⏰ Reminders before review

- CI builds passed
  team

## 🦮 Reviewer guidelines

### DIFF
--- a/crates/bitwarden-wasm-internal/src/pure_crypto.rs
+++ b/crates/bitwarden-wasm-internal/src/pure_crypto.rs
@@ -2,7 +2,7 @@ use std::str::FromStr;
 
 use bitwarden_core::key_management::{KeyIds, SymmetricKeyId};
 #[allow(deprecated)]
-use bitwarden_crypto::dangerous_derive_kdf_output;
+use bitwarden_crypto::dangerous_derive_kdf_material;
 use bitwarden_crypto::{
     AsymmetricCryptoKey, AsymmetricPublicCryptoKey, BitwardenLegacyKeyBytes, CoseKeyBytes,
     CoseSerializable, CoseSign1Bytes, CryptoError, Decryptable, EncString, Kdf, KeyDecryptable,
@@ -313,14 +313,14 @@ impl PureCrypto {
             .map(|pk| pk.to_vec())
     }
 
-    /// Derive the KDF output for a [bitwarden_crypto::Kdf] configuration.
-    pub fn derive_kdf_output(
+    /// Derive output of the KDF for a [bitwarden_crypto::Kdf] configuration.
+    pub fn derive_kdf_material(
         password: &[u8],
         salt: &[u8],
         kdf: Kdf,
     ) -> Result<Vec<u8>, CryptoError> {
         #[allow(deprecated)]
-        dangerous_derive_kdf_output(password, salt, &kdf)
+        dangerous_derive_kdf_material(password, salt, &kdf)
     }
 }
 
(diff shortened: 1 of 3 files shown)


crates/bitwarden-wasm-internal/src/pure_crypto.rs
```diff
@@ -310,6 +312,16 @@ impl PureCrypto {
             .map(|public_key| public_key.to_der())?
             .map(|pk| pk.to_vec())
     }
+
+    /// Derive the KDF output for a [bitwarden_crypto::Kdf] configuration.
+    pub fn derive_kdf_output(
```
dani-garcia: Should this not be called `derive_kdf_key` to match with the `dangerous_*` function it's calling and with `KdfDerivedKeyMaterial::derive_kdf_key`?

quexten: Hmm, I guess the thought was that purecrypto already indicates that it's not a function that is generally available and should only be used in low level crypto services, which is what the `dangerous_` prefix already indicates, but I'll still add it for consistency.

dani-garcia: Sorry, I don't think I was clear enough. My question wasn't about the prefix but about the ending, the rest of the functions are `derive_kdf_key` while this one was `derive_kdf_output`, so I was wondering why this one wasn't called `derive_kdf_key` as well.  I agree that the prefix is not really needed here.

quexten: Ah, whoops. I renamed both to have the suffix `kdf_output`. I'm trying to call anything that is not a SymmetricKey, PrivateKey/PublicKey, SigningKey/VerifyingKey a key because that's just confusing. Since the KDF output needs to be stretched to be used as a key, it's not a key.

dani-garcia: Ah yeah, wanting to separate the actual keys from the KDF output makes sense to me.

Question, is what we call KDF output here the same as `KdfDerivedKeyMaterial`? That also mentions that it should not be used as a key without stretching first, but it has a different naming so I'm not sure if they refer to the exact same concept.

https://github.com/bitwarden/sdk-internal/blob/e398fba3bd6230362c8731b3f1571b0c86999cf0/crates/bitwarden-crypto/src/keys/kdf.rs#L20-L28

quexten: I think they are the same thing, and could really be named the same. The (non-meaningful) minor difference here is that "KdfOutput" has type `Vec<u8>`, where KdfKeyMaterial specifically restricts how it can be used through the typesystem (even though it also just contains `GenericArray<u8,..>`. 

Would it be easier to name the function "dangerous_derive_kdf_material"?

dani-garcia: Yeah I wasn't sure if there was a meaningful difference that I was missing. As that's not the case it would be good to be consistent.

quexten: Renamed.

dani-garcia (approved):

Hinton (approved):

PR #343, opened 2025-07-04
Update bindings to use the `reqwest-trait` template

dani-garcia:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-25542

## 📔 Objective

Note: The PR is absolutely massive because the generated code is very verbose (something that I plan to optimize in the future) and github is pretty bad about filtering out whitespace changes 😞. To facilitate review I include some recommendations at the bottom of this block.

This template is trait based which should allow easier mocking. It also follows a similar pattern to our clients which I think would make it easier to use:

```rust
let config = client.internal.get_api_configurations().await;

// Old API
let res = bitwarden_api_api::apis::projects_api::projects_create(
    &config.api,
    input.organization_id,
    project,
).await?;

// New API
let res = config.api_client
    .projects_api()
    .create(input.organization_id, project)
    .await?;
```

The provided mocks are also typesafe and easy to use:
```rust
        let api_client = ApiClient::new_mocked(move |mock| {
            mock.folders_api
                .expect_post()
                .returning(move |model| {
                    Ok(FolderResponseModel {
                        id: Some(folder_id),
                        name: Some(model.unwrap().name),
                        revision_date: Some("2025-01-01T00:00:00Z".to_string()),
                        object: Some("folder".to_string()),
                    })
                })
                .once();
        });
```
### PR structure

The PR is split in four commits:
- `Update OpenAPI and templates to latest version`: Updates the openapi generator and the templates to the latest version. These files should match with the files at https://github.com/OpenAPITools/openapi-generator/tree/ee40887d47f6d7a16318772f49c63b8eb0553488/modules/openapi-generator/src/main/resources/rust I've also gone ahead and deleted some unused files to make it clearer which files are being used.
- `Add our changes to the templates`
  - Updates to templates 
    - `README.mustache`: Added the server commit hash to the readme
    - `model.mustache` and `model_mod.mustache`: Remove double option pattern, as we don't differentiate between not required and nullable, and they just add noise.
    - `api.mustache`: Made async-traits ?Send to support WASM
    - `api.mustache`: Used our custom `x-action-name` instead of `operationID`, so that function names are just `functionName` instead of `ControllerName_FunctionName`
    - `api.mustache`: Filtered deprecated operations out, no need to generate code for them when they shouldn't be used
    - `api.mustache`: Use `uuid` instead of `&str` for request parameters
    - `api.mustache`: Replaced all the function lifetimes by a single one. Previously it would create one per parameter, which is more flexible but made for some terrible function signatures.
    - `api_mod.mustache`: Replaced the root Api trait and implementations by a single ApiClient enum with both real and mock impls. This is functionally the same but allows the compiler to better optimize out the unused functions (from 3.7MB to 2.2MB WASM blob size)
-  `🤖 Generate bindings`: Generate the bindings using the template from the previous step. this is based on the server bindings from https://github.com/bitwarden/server/commit/d384c0cfe60ec02226479d3cde200e4d785a7e50
-  `Update code to new api`. Updated our code to build with the new template, most changes are very simple API adjustments, the biggest changes are in the `client/internal.rs` file and consist in modifications to how we initialize and update the `ApiConfigurations` struct to match the newer API.

### PR review recommendation

As this PR edits the generated bindings, the total changes are huge and difficult to review. As such, I've written a suggested review guide for the teams.

#### For vault/auth
Filter by team owned, the vault/auth related changes are fairly small:
- small update in the /sync code to use the new API
- updated the folder API code to use the new API and new mocks
- updated the send access code to use the new configuration

If filtering by team owned doesn't work, the non-platform changes are contained entirely in the fourth commit: https://github.com/bitwarden/sdk-internal/pull/343/commits/320ea80580773795b9a5fdeb3b0f480d0bcbcfba

#### For platform
Due to the big autogenerated changes, I recommend reviewing commit by commit. The first, third and fourth commits should be reviewed as normal. The second commit, marked as 🤖, should instead be reviewed by regenerating the bindings and making sure that nothing is changed:
```sh
cd server
git checkout c93c3464732c93c9be593a3a55b032c029c4bd6f
pwsh ./dev/generate_openapi_files.ps1

cd ../sdk-internal
git pull
git checkout ps/migrate-to-openapi-rust-trait
./support/build-api.sh

git status
# This should report a clean work tree
```

## ⏰ Reminders before review

- CI builds passed
  team

## 🦮 Reviewer guidelines


Hinton:
Should we open upstream issues and see if any of the changes can upstreamed?

It would be interesting to see how the mocked usage would be as a comparison to using wiremock.

dani-garcia:
I'm planning to upstream the doc link issue at least, which clearly seems like a mistake. Probably also the removing Send bound, which seems safe. The rest could be harder:
- Using str for uuids seems intentional for the templates, as it's happening in all variants, even though they use the uuid type in other places.
- The lifetimes change is mostly stylistic, as it simplifies the generated code a bit, but there may be cases where it's useful to have separately named lifetimes, not sure.
- The APIclient initialization is fairly subjective on whether it should eagerly initialize all the subclients or to initialize them on each call. To me it makes more sense to do it on each call than to initialize the 60 of them at the same time, but for smaller APIs maybe that doesn't matter

The mocks look something like this: https://github.com/bitwarden/sdk-internal/commit/77e23333fe926a244be33b8ac51158f4f5bb922d

support/build-api.sh
```diff
@@ -30,8 +30,23 @@ npx openapi-generator-cli generate \
     -o crates/bitwarden-api-identity \
     --package-name bitwarden-api-identity \
     -t ./support/openapi-template \
-    --additional-properties=packageVersion=$VERSION,packageDescription=\"Api bindings for the Bitwarden Identity API.\"
+    --additional-properties=library=reqwest-trait,mockall,topLevelApiClient,packageVersion=$VERSION,packageDescription=\"Api bindings for the Bitwarden Identity API.\"
 
 rustup toolchain install nightly
+
+# Rustfmt has what looks like a bug, where it requires multiple passes to format the code.
+# For example with code like this:
```
Hinton: Yea some of the rules we use are unstable.

crates/bitwarden-api-identity/src/apis/configuration.rs
```diff
@@ -36,7 +36,7 @@ impl Configuration {
 impl Default for Configuration {
     fn default() -> Self {
         Configuration {
-            base_path: "http://localhost".to_owned(),
+            base_path: "https://identity.bitwarden.com".to_owned(),
```
Hinton, line 39: Could this be potentially dangerous? We should never use the default config as we don't know if it's EU or US.

Hinton (dismissed):

dani-garcia:
Moving to draft for now as the compiler can't seem to optimize the unused trait impls away, leading to 2x file sizes for the final wasm bundle. I'll look into modifying the generated code to not be as resource heavy.

Edit: I have some improvements to the generated code that reduce the size of the binary considerably, but they are deeper changes into the templates that will benefit from upstreaming, so for now I've commented out the use of traits in this PR, and I will revisit that in a separate PR when I'm back from vacation.

crates/bitwarden-api-identity/src/apis/configuration.rs
dani-garcia: We never use this default, we use the value from `ClientSettings`:
https://github.com/bitwarden/sdk-internal/blob/c3d62e346690a08d704f5d1afd4074b94e77bfab/crates/bitwarden-core/src/client/client_settings.rs#L38-L45

dani-garcia:
Moving to draft while the swagger generation improvements on server get merged

crates/bitwarden-core/src/client/internal.rs
```diff
@@ -50,13 +50,52 @@ impl From<&InitUserCryptoRequest> for UserKeyState {
...
 
 #[allow(missing_docs)]
-#[derive(Debug, Clone)]
 pub struct ApiConfigurations {
-    pub identity: bitwarden_api_identity::apis::configuration::Configuration,
-    pub api: bitwarden_api_api::apis::configuration::Configuration,
+    pub identity_client: bitwarden_api_identity::apis::ApiClient,
+    pub api_client: bitwarden_api_api::apis::ApiClient,
+    pub identity_config: bitwarden_api_identity::apis::configuration::Configuration,
+    pub api_config: bitwarden_api_api::apis::configuration::Configuration,
```
Hinton, line 57: **question:** Can these be private or `pub(crate)`?

dani-garcia: `api_config` can be private, but `identity_config` needs to be pub as its used by auth because they can't use bindings for `connect/token`. I left both public for symmetry, but I can make `api_config` private for now.

Hinton: Urgh, let's not do it for now.

Hinton (approved):
I've reviewed all code except the auto generated which i've confirmed matches the output from manually running it.

JaredSnider-Bitwarden (approved):
Auth file change LGTM!

Jingo88 (approved):

nikwithak (approved):

addisonbeck (approved):

PR #347, opened 2025-07-09
[PM-23638] Fix URI checksums not being generated when cipherkey is not present

quexten:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-23638

## 📔 Objective

While we only enforce decrypting correct checksums when a cipherkey is present for that cipher, we must always upgrade to url checksums for newly encrypted ciphers, regardless of whether the cipherkey is present or not.

This is the case on Typescript (https://github.com/bitwarden/clients/blob/682f1f83d9c211e886d6f05fc557eddb25c4c5c2/libs/common/src/vault/services/cipher.service.ts#L1660). This allows us to enforce the presence of icon URLs after users went through the key-rotation for v2 encryption based on the account security version *instead of* being based on cipherkeys.

## ⏰ Reminders before review

- CI builds passed
  team

## 🦮 Reviewer guidelines

### DIFF
--- a/crates/bitwarden-vault/src/cipher/cipher.rs
+++ b/crates/bitwarden-vault/src/cipher/cipher.rs
@@ -296,11 +296,7 @@ impl CompositeEncryptable<KeyIds, SymmetricKeyId, Cipher> for CipherView {
         let ciphers_key = Cipher::decrypt_cipher_key(ctx, key, &self.key)?;
 
         let mut cipher_view = self.clone();
-
-        // For compatibility reasons, we only create checksums for ciphers that have a key
-        if cipher_view.key.is_some() {
-            cipher_view.generate_checksums();
-        }
+        cipher_view.generate_checksums();
 
         Ok(Cipher {
             id: cipher_view.id,


dani-garcia (approved):
LGTM not sure why I did that initially, the checksums are pretty safe to create.

Now that the logic doesn't consider the cipher key, we should think about lifting the checksum creation into the LoginUriView encryptable implementation directly. This will make sure it's always updated correctly.
```rust
impl CompositeEncryptable<LoginUri> for LoginUriView {
    fn encrypt_composite(&self, ctx, key) -> Result<LoginUri> {
        let checksum = self.uri.as_ref().map(|u| {
            use sha2::Digest;
            let uri_hash = sha2::Sha256::new().chain_update(u.as_bytes()).finalize();
            STANDARD.encode(uri_hash.as_slice())
        });

        Ok(LoginUri {
            uri: self.uri.encrypt(ctx, key)?,
            r#match: self.r#match,
            uri_checksum: checksum.encrypt(ctx, key)?,
        })
    }
}
```

nick-livefront (approved):

PR #348, opened 2025-07-11
[Shared Unlock] [BEEEP/IS/PM-34080] Add noise IPC crypto provider

quexten:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-34080

## 📔 Objective

Adds a compile-time flagged crypto provider for IPC, implemented via noise (snow).

A handshake provides break-in recovery, since a new key is established interactively, encrypted and authenticated under the previous tunnel. It provides forward secrecy, in the sense that there are no long-term keys that can be compromised. Any given keypair can only decrypt keys in the current handshake period, but not the previous or next handshake's keys.

Please note the handshake is changeable, here we support two handshakes, one that is FIPS compatible, and one that is not. We can also use the same mechanism to switch to post-quantum handshakes later, provided snow gains support for these.

## ⏰ Reminders before review

- CI builds passed
  team

## 🦮 Reviewer guidelines


crates/bitwarden-ipc/src/crypto_provider/noise/crypto_provider.rs
```diff
@@ -0,0 +1,359 @@
...
+    async fn send(
+        &self,
+        communication: &Com,
+        sessions: &Ses,
+        message: OutgoingMessage,
+    ) -> Result<(), Self::SendError> {
+        // Send operations *MUST* be seralized, otherwise nonce re-use may happen since
+        // concurrent sends may aquire the same copy of the transport state before nonce
+        // updating.
+        let _send_guard = SEND_GUARD.lock().await;
```
quexten: This one was pretty interesting. Basically, when two concurrent sends happen then two clients may receive the same copy of the state with nonce N and hold it in the function stack. Then, both update the state by one, and save it to the state to N+1. Both encrypt with N.

With both gcm and chacha this results in catastrophic cryptographic failure, and reveals the xor of the two plaintexts of the messages.

The send guard here should prevent this.

quexten: @coroiu Do you think there is a better pattern here? I'm not exactly sure if this is the way the sessions are meant to be interacted with or if sessions in general didn't mean to account for concurrent updating / left that up to the crypto provider.

crates/bitwarden-threading/src/thread_bound_runner.rs
```diff
@@ -84,6 +84,12 @@ pub struct ThreadBoundRunner<ThreadState> {
     call_channel_tx: tokio::sync::mpsc::Sender<CallRequest<ThreadState>>,
 }
 
+// Safety: ThreadBoundRunner only contains a Sender which is Send + Sync.
+// The ThreadState is never directly accessed from ThreadBoundRunner, only through
+// the channel which ensures proper thread confinement.
+unsafe impl<ThreadState> Send for ThreadBoundRunner<ThreadState> {}
+unsafe impl<ThreadState> Sync for ThreadBoundRunner<ThreadState> {}
+
```
coroiu: issue: I think these can be removed without impacting anything

crates/bitwarden-ipc/src/wasm/generic_session_repository.rs
```diff
@@ -9,12 +9,13 @@
 use std::sync::Arc;
 
 use crate::{
+    crypto_provider::noise::crypto_provider::NoiseCryptoProviderState,
     traits::{InMemorySessionRepository, SessionRepository},
     wasm::JsSessionRepository,
 };
 
 // TODO: Change session type when implementing encryption
```
coroiu: nit: remove the TODO comment :)

coroiu (commented):
One big comment, other than that everything looks good from my side

crates/bitwarden-ipc/src/crypto_provider/noise/crypto_provider.rs
coroiu: Hmm, I don't think there any better ways of achieving this. If we try to add locking/syncing mechanisms to the repository trait we'd need a way of actually implementing them using the actual storage backend. I'm not sure that's worth it when we can just do this in the specific crypto provider

coroiu (dismissed):

.github/CODEOWNERS
```diff
@@ -44,6 +44,7 @@ crates/bitwarden-send/** @bitwarden/team-tools-dev @bitwarden/team-platform-dev
 # KM Co-owned files
 crates/bitwarden-core/src/key_management/** @bitwarden/team-key-management-dev @bitwarden/team-sdk-sme
 crates/bitwarden-wasm-internal/src/pure_crypto.rs @bitwarden/team-key-management-dev @bitwarden/team-sdk-sme
+crates/bitwarden-ipc/src/crypto_provider/noise/ @bitwarden/team-key-managemnet-dev @bitwarden/team-sdk-sme
```
Thomas-Avery: Would be good to have someone from `@bitwarden/team-sdk-sme` review if they're co owning it?

crates/bitwarden-ipc/Cargo.toml
```diff
@@ -25,17 +27,21 @@ async-trait = { workspace = true }
 bitwarden-error = { workspace = true }
 bitwarden-ffi = { workspace = true }
 bitwarden-threading = { workspace = true }
+ciborium = { version = ">=0.2.2, <0.3" }
```
Thomas-Avery:
```suggestion
ciborium = { workspace = true }
```

In main it looks like we already have this in the root Cargo.toml

crates/bitwarden-ipc/src/wasm/js_session_repository.rs
```diff
@@ -41,6 +41,9 @@ extern "C" {
 /// Thread safe JavaScript implementation of the `SessionRepository` trait for IPC sessions.
 pub struct JsSessionRepository(ThreadBoundRunner<RawJsSessionRepository>);
 
+unsafe impl Send for RawJsSessionRepository {}
+unsafe impl Sync for RawJsSessionRepository {}
```
Thomas-Avery: It would be nice to have some context and document why these unsafe implementations are okay. To be honest it's hard to follow what's going on here from the scope of this review.

.github/CODEOWNERS
Thomas-Avery:
```suggestion
crates/bitwarden-ipc/src/crypto_provider/noise/ @bitwarden/team-key-management-dev @bitwarden/team-sdk-sme
```

⚠️

crates/bitwarden-ipc/src/crypto_provider/noise/transport_state.rs
```diff
@@ -0,0 +1,360 @@
...
+    }
+}
+
+impl PersistentTransportState {
+    /// Encrypts the message, mutates the state and returns the transport frame to be sent over
+    /// IPC.s
+    pub(crate) fn send(&mut self, payload: Payload) -> Result<TransportFrame, ()> {
+        // Increase nonce. WARNING: Re-used nonces lead to catastrophic
+        // crypto failure. Ensure this increases always.
+        self.send_nonce += 1;
```
Thomas-Avery: 💭   https://doc.rust-lang.org/book/ch03-02-data-types.html#integer-overflow

Release mode overflow behavior sounds scary here (and in general) based on the documentation of nonce reuse. In practice it shouldn't be possible with the re. handshake interval. But maybe it's worth adding `checked_add` here to be defensive.

crates/bitwarden-ipc/src/crypto_provider/noise/transport_state.rs
```diff
@@ -0,0 +1,360 @@
...
+    // For the responder, send_key = r2i and receive_key = i2r.
+    send_key: SymmetricKey,
+    receive_key: SymmetricKey,
+
+    // Noise transport messages include a nonce that must be unique for every message encrypted
+    // with the same key. The nonce increases monotonically with every sent/received message
+    // and is never reset to a lower value. Re-using nonces results in catastrophic
+    // cryptographic failure.
+    send_nonce: u64,
+    // For receiving, skipping nonces is alowed, but never going back.
```
Thomas-Avery:
```suggestion
    // For receiving, skipping nonces is allowed, but never going back.
```

crates/bitwarden-ipc/src/crypto_provider/noise/crypto_provider.rs
```diff
@@ -0,0 +1,366 @@
...
+    }
+}
+
+/// The raw frame that is sent via IPC.
+#[derive(Serialize, Deserialize)]
+pub(super) enum Frame {
+    // Handshake Frames
+    HandshakeStart(HandshakeStartMessage),
+    HandshakeFinish(HandshakeFinishMessage),
+    // After the handsahke is done, transport frames are used to wrap ciphertexts
```
Thomas-Avery:
```suggestion
    // After the handshake is done, transport frames are used to wrap ciphertexts
```

crates/bitwarden-ipc/src/crypto_provider/noise/crypto_provider.rs
```diff
@@ -0,0 +1,366 @@
...
+    type SendError = ();
+    type ReceiveError = ();
+
+    async fn send(
+        &self,
+        communication: &Com,
+        sessions: &Ses,
+        message: OutgoingMessage,
+    ) -> Result<(), Self::SendError> {
+        // Send operations *MUST* be seralized, otherwise nonce re-use may happen since
```
Thomas-Avery:
```suggestion
        // Send operations *MUST* be serialized, otherwise nonce re-use may happen since
```

crates/bitwarden-ipc/src/crypto_provider/noise/crypto_provider.rs
```diff
@@ -0,0 +1,366 @@
...
+        // concurrent sends may aquire the same copy of the transport state before nonce
```
Thomas-Avery:
```suggestion
        // concurrent sends may acquire the same copy of the transport state before nonce
```

crates/bitwarden-ipc/src/crypto_provider/noise/handshake.rs
```diff
@@ -0,0 +1,214 @@
+//! This module implements the Noise NN handshake for IPC.
+//! Note: NN does not provide any sort of authentication, and the keys each
+//! side uses are just trusted. This means that a MITM with active tampering is possible and
+//! accepted. Thereby it is necessary that either the threat model of the application using IPC
+//! assumes that the IPC channel is not exposed to MITM attacks, or that the transport layer
+//! prevents MITM with active tampering.
+//!
+//! Protocol flow:
+//! 1. Initiator -> Ressponder: `HandshakeStartMessage { ciphersuite, noise_frame }`
```
Thomas-Avery:
```suggestion
//! 1. Initiator -> Responder: `HandshakeStartMessage { ciphersuite, noise_frame }`
```

crates/bitwarden-ipc/src/crypto_provider/noise/transport_state.rs
```diff
@@ -0,0 +1,360 @@
...
+        f.debug_struct("SymmetricKey").finish()
+    }
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub(crate) struct PersistentTransportState {
+    // The symmetric algorithm used for transport encryption
+    transport_cipher: TransportCipher,
+
+    // Noise has two keys, the initatior to responder key (i2r) and the responder to initiator key
```
Thomas-Avery:
```suggestion
    // Noise has two keys, the initiator to responder key (i2r) and the responder to initiator key
```

crates/bitwarden-ipc/src/crypto_provider/noise/transport_state.rs
```diff
@@ -0,0 +1,360 @@
...
+    fn default() -> Self {
+        if cfg!(feature = "fips") {
+            Self::Aes256Gcm
+        } else {
+            Self::ChaCha20Poly1305
+        }
+    }
+}
+
+/// A newtype for symmetric keys used in noise. A noise key is always 256-bits.
```
Thomas-Avery, line 19:
```suggestion
/// A new type for symmetric keys used in noise. A noise key is always 256-bits.
```

crates/bitwarden-ipc/src/crypto_provider/noise/transport_state.rs
```diff
@@ -0,0 +1,360 @@
...
+#[derive(Clone, Serialize, Deserialize)]
+pub(crate) struct SymmetricKey(pub(crate) [u8; KEY_SIZE]);
```
Thomas-Avery: 💭  I wonder if it would be a good idea to name this more specific to the use case. Like `NoiseSymmetricKey`.

crates/bitwarden-ipc/src/crypto_provider/noise/transport_state.rs
```diff
@@ -0,0 +1,360 @@
...
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub(crate) struct PersistentTransportState {
+    // The symmetric algorithm used for transport encryption
+    transport_cipher: TransportCipher,
+
+    // Noise has two keys, the initatior to responder key (i2r) and the responder to initiator key
+    // (r2i). For the initiator, send_key = i2r and receive_key = r2i.
+    // For the responder, send_key = r2i and receive_key = i2r.
```
Thomas-Avery:
```suggestion
    // (r2i). 
    // For the initiator, send_key = i2r and receive_key = r2i.
    // For the responder, send_key = r2i and receive_key = i2r.
```

⛏️  Having a newline here makes it easier to read IMO.

crates/bitwarden-ipc/src/crypto_provider/noise/transport_state.rs
```diff
@@ -0,0 +1,360 @@
...
+
+impl AsRef<[u8]> for Payload {
+    fn as_ref(&self) -> &[u8] {
+        &self.0
+    }
+}
+
+impl PersistentTransportState {
+    /// Encrypts the message, mutates the state and returns the transport frame to be sent over
+    /// IPC.s
```
Thomas-Avery:
```suggestion
    /// IPC.
```

crates/bitwarden-ipc/src/crypto_provider/noise/transport_state.rs
```diff
@@ -0,0 +1,360 @@
...
+            nonce: self.send_nonce,
+        })
+    }
+
+    /// Decrypts the transport frame, mutates the state and returns the plaintext message.
+    pub(crate) fn receive(
+        &mut self,
+        transport_frame: &TransportFrame,
+    ) -> Result<Payload, ReceiveError> {
+        // Try decryption with current receive key first.
```
Thomas-Avery:
```suggestion

```
This comment makes me think we will attempt something else but we don't? Maybe just remove it.

crates/bitwarden-ipc/src/crypto_provider/noise/mod.rs
```diff
@@ -0,0 +1,73 @@
...
+//!     |--- HandshakeStart ------------------>|
+//!     |<-- HandshakeFinish ------------------|
+//!     |                                      |
+//!     |     [Both derive transport keys]     |
+//!     |                                      |
+//!     |--- TransportFrame (encrypted) ------>|
+//!     |<-- TransportFrame (encrypted) -------|
+//!     |              ...                     |
+//!     |                                      |
+//!     | [After REHANDSHAKE_INTERVAL (300s),  |
```
Thomas-Avery:
```suggestion
//!     | [After REHANDSHAKE_INTERVAL          |
```

Consider leaving out the current value as often these docs drift when constants are updated and this is missed.

crates/bitwarden-ipc/src/crypto_provider/noise/handshake.rs
```diff
@@ -0,0 +1,214 @@
...
+            state: handshake_state,
+        }
+    }
+
+    pub(crate) fn write_start_message(&mut self) -> Result<HandshakeStartMessage, WriteError> {
+        let mut buf = [0u8; super::NOISE_MAX_MESSAGE_LEN];
+        let len = self
+            .state
+            .write_message(&[], &mut buf)
+            .map_err(|_| WriteError)?;
```
Thomas-Avery, line 100: We lose a lot of helpful troubleshooting context when we discard library errors like this. I would suggest at a minimum logging as debug and if we aren't worried about content logging as an error. 

This feedback applies to every instance we are just suppressing the error to replace with our custom ones.

crates/bitwarden-ipc/src/crypto_provider/noise/crypto_provider.rs
```diff
@@ -0,0 +1,366 @@
...
+
+        let handshake_frame = Frame::HandshakeStart(message);
+        communication
+            .send(OutgoingMessage {
+                payload: handshake_frame.to_cbor(),
+                destination: destination.clone(),
+                topic: None,
+            })
+            .await
+            .map_err(|_| ())?;
```
Thomas-Avery: ❓  Silently failing? This seems odd, would expect at least a log. 

Applies to several occurrences in this file.

crates/bitwarden-ipc/src/crypto_provider/noise/crypto_provider.rs
```diff
@@ -0,0 +1,366 @@
...
+                    continue;
+                };
+                let Ok(response_frame) = Frame::from_cbor(&incoming.payload) else {
+                    continue;
+                };
+                if let Frame::HandshakeFinish(handshake_finish) = response_frame {
+                    let _ = initiator
+                        .read_response_message(&handshake_finish)
+                        .map_err(|_| ());
+                    break;
```
Thomas-Avery: ⚠️  If I'm understanding this correctly `read_response_message` can error but that won't propagate to the wider timeout error handling?

Then we immediately process assuming a success.

Thomas-Avery (changes requested):
Did the best I could to review with my limited understanding of the wider IPC implementation. Let me know if I'm miss understanding something.

crates/bitwarden-ipc/src/crypto_provider/noise/crypto_provider.rs
Thomas-Avery: Also shouldn't we be checking/filtering that the `HandshakeFinish` is coming from the `destination` provided in `perform_handshake`?

crates/bitwarden-ipc/Cargo.toml
```diff
@@ -25,17 +27,21 @@ async-trait = { workspace = true }
...
 serde_json = { workspace = true }
+snow = { version = "0.9.6", features = ["risky-raw-split"] }
 thiserror = { workspace = true }
 tokio = { features = ["sync", "time", "rt"], workspace = true }
 tracing = { workspace = true }
 tsify = { workspace = true, optional = true }
 uuid = { workspace = true }
 wasm-bindgen = { workspace = true, optional = true }
 wasm-bindgen-futures = { workspace = true, optional = true }
+web-time = "1.1.0"
```
Thomas-Avery: 💭  This is a dependency only when using WASM. Should we make it optional with the wasm feature enabling it?

Also curious if you looked at using js-sys for this instead? https://docs.rs/js-sys/latest/js_sys/struct.Date.html looks like we are already using that for JSON parsing in IPC. It comes with wasm-bindgen.

.github/CODEOWNERS
quexten: Actually, I think we should remove the co-ownership in light of https://github.com/bitwarden/sdk-internal/pull/950

crates/bitwarden-ipc/src/crypto_provider/noise/transport_state.rs
quexten: @Thomas-Avery actually newtype is correct here https://doc.rust-lang.org/rust-by-example/generics/new_types.html

crates/bitwarden-ipc/src/wasm/js_session_repository.rs
quexten: Actually, seems we can remove the impl's at this point. I'm not sure why they were needed in the first place, maybe @coroiu's changes to make this use the threadbound runner removed the need?

crates/bitwarden-ipc/src/crypto_provider/noise/transport_state.rs
quexten: It's a u64 so there is no way to actually reach u64 max value. To be clear, if we assume 100 bytes per message, this would be 1.8 zettabytes. For a 5 minute re-handshake interval, you would have to encrypt at 3.3 Petabyte per second to reach this amount.

crates/bitwarden-ipc/src/crypto_provider/noise/transport_state.rs
quexten: I've changed the visibility to pub(super) now so that it is only available in the noise module. Considering it is only available in the noise module, it is already clear that it is a noise struct, and the prefix would be unexpected (considering all other structs also do not have a prefix).

crates/bitwarden-ipc/src/crypto_provider/noise/handshake.rs
quexten: That,s fair, but TBH, I would expect IPC delivery errors to be logged in the IPC layer, not by us. We should just expect it to work.

However, crypto errors should be logged in our layer of the stack. I'll add those.

If we do want transport layer logs, we should propose these to platform's owned code.

crates/bitwarden-ipc/src/crypto_provider/noise/crypto_provider.rs
quexten: @Thomas-Avery I cleaned up the error handling a bit with more intentional logging, and a few error cases that we do pass through, which should fix this.

crates/bitwarden-ipc/src/crypto_provider/noise/crypto_provider.rs
quexten: Please see the other comments; We should not log all IPC transport failures, but the concerns of our code. I've added logs for our concerns, and made it so that we pass through underlying IPC failures through the stack where appropriate.

crates/bitwarden-ipc/Cargo.toml
quexten: Seems we can drop web-time.

crates/bitwarden-ipc/src/wasm/js_session_repository.rs
coroiu: Yes these can be removed, I thought I already did that? Maybe I only did that in our IS PR :shrug: Either way, you are both correct, `ThreadBoundRunner` implements these properly, and no it's not okay to use `unsafe` :)

crates/bitwarden-ipc/src/crypto_provider/noise/transport_state.rs
Thomas-Avery: My concern is while all that is true right now, if how this is used changes in the future it's a gotcha that isn't immediately obvious. Adding the defensive now seems to have no downside and helps with future maintainability?

quexten: Okay, I've added an expect that will panic the app in the event that this happens, along with a comment explaining the assumption that this cannot happen.

(As long as we increment nonces one-by-one it is not possible to send this many messages within the lifetime of a Bitwarden app).

crates/bitwarden-ipc/src/crypto_provider/noise/transport_state.rs
```diff
@@ -9,25 +9,16 @@ use crate::crypto_provider::noise::NOISE_MAX_MESSAGE_LEN;
...
-        } else {
-            Self::ChaCha20Poly1305
-        }
-    }
-}
-
 /// A newtype for symmetric keys used in noise. A noise key is always 256-bits.
 #[derive(Clone, Serialize, Deserialize)]
-pub(crate) struct SymmetricKey(pub(crate) [u8; KEY_SIZE]);
+pub(super) struct SymmetricKey(pub(crate) [u8; KEY_SIZE]);
```
Thomas-Avery, line 21: Does this need `ZeroizeOnDrop` like our `SymmetricCryptoKey`?

crates/bitwarden-ipc/src/crypto_provider/noise/crypto_provider.rs
Thomas-Avery: This is hard to reason about. The errors from the loop are silently ignored and we continue on and it won't hit the timeout block?

Thomas-Avery, line 226: 💭  Sends are serialized but receives are not. But they write to the same shared state, is possible for the receive nonce to get out of sync here? 

Something like: 


- Process A: sends ->  {sendNonce N+1,  receiveNonce M} (doesn't save to state yet)
- Process B: sends a message to process A at the same time
- Process A: receives -> gets current state  {sendNonce N,  receiveNonce M}
- Process A: finishes receives saves nonce from payload {sendNonce N,  receiveNonce M+1}
- Process A: finishes sending and saves state  {sendNonce N+1,  receiveNonce M}

Receive nonce in process A is now out of sync? Should be M+1 but send over writes it to M.

Thomas-Avery (changes requested):
Looking good, a few more things after looking at this again and should be good.

crates/bitwarden-ipc/src/crypto_provider/noise/crypto_provider.rs
Thomas-Avery: Section I was referring to. Not sure what is going on with GitHub here. 

```
      // Wait for the handshake response (with timeout)
        let Ok(_) = timeout(Duration::from_secs(HANDSHAKE_TIMEOUT_SECS), async {
            loop {
                let incoming = receiver.receive().await;
                let incoming = incoming.map_err(|_| NoiseCryptoProviderError::TransportReceive)?;
                // For concurrent handshakes, ignore messages
                if incoming.source.to_endpoint() != destination {
                    continue;
                }
                let Ok(response_frame) = Frame::from_cbor(&incoming.payload) else {
                    warn!("Received malformed cbor message during handshake, ignoring");
                    continue;
                };
                if let Frame::HandshakeFinish(handshake_finish) = response_frame {
                    if initiator.read_response_message(&handshake_finish).is_err() {
                        error!("Failed to read handshake response message");
                        return Err(NoiseCryptoProviderError::HandshakeProtocol);
                    }
                    break;
                }
            }
            Ok(())
        })
        .await
        else {
            info!(
                "Noise handshake with {:?} timed out after {} seconds",
                destination, HANDSHAKE_TIMEOUT_SECS
            );
            return Err(NoiseCryptoProviderError::Timeout);
        };

```

Thomas-Avery: Was reading old version the new guard fixes this and this can be ignored.

crates/bitwarden-ipc/src/crypto_provider/noise/transport_state.rs
quexten: Can't hurt to add it at any rate.

crates/bitwarden-ipc/src/crypto_provider/noise/crypto_provider.rs
quexten: Should be more readable now... Please let me know if there is anything in particular that needs to be further clarified.

Thomas-Avery (approved):
LGTM, looks like CI is failing but might be unrelated.

coroiu (approved):

PR #350, opened 2025-07-16
[PM-23085] Add encrypt_with_key method to cipher_client

shane-melton:
## 🎟️ Tracking

[PM-23085](https://bitwarden.atlassian.net/browse/PM-23085)

## 📔 Objective

Implement `encrypt_with_key` for the `cipher_client` so that the TS clients can re-encrypt a cipher with a new user key during key rotation. Once key rotation is implemented fully in the SDK ([PM-23084](https://bitwarden.atlassian.net/browse/PM-23084)) this method can either be removed or updated to accept a symmetric key identifier instead of the raw base64 encoded key.

## ⏰ Reminders before review

- CI builds passed
  team

## 🦮 Reviewer guidelines

[PM-23085]: https://bitwarden.atlassian.net/browse/PM-23085?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[PM-23084]: https://bitwarden.atlassian.net/browse/PM-23084?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


crates/bitwarden-vault/src/cipher/cipher.rs
```diff
@@ -459,6 +461,39 @@ impl CipherView {
...
+            self.key = Some(ctx.wrap_symmetric_key(new_wrapping_key, cipher_key)?);
+            return Ok(());
+        }
+
+        // The cipher does not have a key, we must reencrypt all attachment keys and FIDO2
+        // credentials individually
+        self.reencrypt_attachment_keys(ctx, old_key, new_wrapping_key)?;
+        self.reencrypt_fido2_credentials(ctx, old_key, new_wrapping_key)?;
+
+        Ok(())
```
gbubemismith: ❓ (non-blocking)  What do you think about using  here?
``` rust
if self.key.is_some() {

} else {

}

Ok(())
```
Since both paths are valid business logic, it might make the intent clearer

gbubemismith (commented):
This looks good, just one question

crates/bitwarden-vault/src/cipher/cipher.rs
```diff
@@ -459,6 +461,39 @@ impl CipherView {
...
+    }
```
Hinton: **issue:** We have a `move_to_organization` function on `Cipher` that does this already pretty much. Except it has additional checks for attachments. To avoid maintaining two implementations we should consolidate them.

**suggestion:** Extract the logic from `move_to_organization` into `reencrypt_cipher_keys` so that `move_to_organization` is just

```rust
pub fn move_to_organization(
    &mut self,
    organization_id: Uuid,
) -> Result<(), CipherError> {
    let new_key = SymmetricKeyId::Organization(organization_id);
    self.reencrypt_cipher_keys(ctx, new_key)
}
```

Hinton:
@quexten Any KM requests for how this interface should look?

crates/bitwarden-vault/src/cipher/cipher_client.rs
```diff
@@ -46,6 +46,61 @@ impl CiphersClient {
...
+        let user_id = self
+            .client
+            .internal
+            .get_user_id()
+            .ok_or(EncryptError::MissingUserId)?;
+        let key_store = self.client.internal.get_key_store();
+        let mut ctx = key_store.context();
+
+        // Set the new key in the key store context
+        const NEW_KEY_ID: SymmetricKeyId = SymmetricKeyId::Local("new_key");
```
Hinton: **suggestion:** This is `new_cipher_key` in other places.
```suggestion
        const NEW_KEY_ID: SymmetricKeyId = SymmetricKeyId::Local("new_cipher_key");
```

crates/bitwarden-vault/src/cipher/cipher_client.rs
```diff
@@ -46,6 +46,61 @@ impl CiphersClient {
...
+            && self
+                .client
+                .internal
+                .get_flags()
+                .enable_cipher_key_encryption
+        {
+            cipher_view.generate_cipher_key(&mut ctx, NEW_KEY_ID)?;
+        } else {
+            cipher_view.reencrypt_cipher_keys(&mut ctx, NEW_KEY_ID)?;
+        }
```
Hinton, line 90: It's unfortunate we still need the cipher_client to handle the feature flag ...

crates/bitwarden-vault/src/cipher/cipher.rs
shane-melton: Great point! Consolidated in [1997917](https://github.com/bitwarden/sdk-internal/pull/350/commits/199791747f6c18e6c168319fa38d3561c5d76128)

crates/bitwarden-vault/src/cipher/cipher_client.rs
shane-melton: [c5648d0](https://github.com/bitwarden/sdk-internal/pull/350/commits/c5648d04bcbed4ec8bfbcec8fd9c38133f8de8ad)

crates/bitwarden-vault/src/cipher/cipher_client.rs
shane-melton: Agreed, it would be great to be able to pass the application owned flags (`configService`) into the SDK client so that the application and SDK flags would stay synced.

crates/bitwarden-vault/src/cipher/cipher.rs
shane-melton: Updated in [1997917](https://github.com/bitwarden/sdk-internal/pull/350/commits/199791747f6c18e6c168319fa38d3561c5d76128)

crates/bitwarden-vault/src/cipher/cipher_client.rs
Hinton: The problem here is more that flags are only available in clients and not the encrypt interface.

crates/bitwarden-vault/src/cipher/cipher.rs
```diff
@@ -459,6 +461,43 @@ impl CipherView {
...
+            self.key = Some(ctx.wrap_symmetric_key(new_wrapping_key, cipher_key)?);
+        } else {
+            // The cipher does not have a key, we must reencrypt all attachment keys and FIDO2
+            // credentials individually
+            self.reencrypt_attachment_keys(ctx, old_key, new_wrapping_key)?;
+            self.reencrypt_fido2_credentials(ctx, old_key, new_wrapping_key)?;
+        }
+
+        Ok(())
+    }
```
Hinton: **nit** The diff could maybe be minimized by moving this function below `move_to_organization`.

crates/bitwarden-vault/src/cipher/cipher.rs
```diff
@@ -526,32 +565,22 @@ impl CipherView {
         Ok(())
     }
 
-    #[allow(missing_docs)]
+    /// Moves the cipher to an organization by re-encrypting the cipher keys with the organization
+    /// key and assigning the organization ID to the cipher.
+    ///
+    /// # Arguments
+    /// * `ctx` - The key store context where the cipher keys will be re-encrypted
+    /// * `organization_id` - The ID of the organization to move the cipher to
```
Hinton, line 536: **nit**: Rust doesn't have a convention of documenting arguments. When possible prefer to make the names self explainatory such that you don't need docs.

In this case both `ctx` and `organization_id` are quite self explanatory.

Hinton (dismissed):

crates/bitwarden-vault/src/cipher/cipher_client.rs
```diff
@@ -46,6 +46,61 @@ impl CiphersClient {
...
+    /// If the cipher has a CipherKey, it will be re-encrypted with the new key.
+    /// If the cipher does not have a CipherKey and CipherKeyEncryption is enabled, one will be
+    /// generated using the new key. Otherwise, the cipher's data will be encrypted with the new
+    /// key directly.
+    ///
+    /// # Arguments
+    /// * `cipher_view` - The cipher view to encrypt.
+    /// * `new_key_b64` - The base64 encoded symmetric key to use for encryption.
+    #[cfg(feature = "wasm")]
+    pub fn encrypt_with_key(
```
quexten: Suggestion: Make it clear from the naming that this is only meant for rotation. People, and LLMs, often don't read documentation carefully and may assume that this is OK to use. `encrypt_cipher_for_rotation`

quexten:
> @quexten Any KM requests for how this interface should look?

Looks mostly OK from my side, as an interim, until key-rotation is moved into the SDK at a high level. However, if this is only merged/used with full SDK encrypt/decrypt, could we think about just enabling cipherkeys always, and drop the flag dependence?

The main concern left with cipherkeys is around the TS code, and around testing. And this will be QA'd heavily either way. That would drop complexity and risk from multiple code paths.

(Thoughts on "expanding" the scope to enable cipherkeys always here @shane-melton ?)

crates/bitwarden-vault/src/cipher/cipher.rs
shane-melton: Good to know, removed in [0fe8dcc](https://github.com/bitwarden/sdk-internal/pull/350/commits/0fe8dcc58ee5e090a0fe60b1e64c1e86b17cd3d7)

crates/bitwarden-vault/src/cipher/cipher.rs
shane-melton: Done! [0fe8dcc](https://github.com/bitwarden/sdk-internal/pull/350/commits/0fe8dcc58ee5e090a0fe60b1e64c1e86b17cd3d7)

crates/bitwarden-vault/src/cipher/cipher_client.rs
shane-melton: Great point, updated in [0fd0eb5](https://github.com/bitwarden/sdk-internal/pull/350/commits/0fd0eb50e1709abbd9acfc4c49a1847f7c7aa1b1)

shane-melton:
@quexten I agree that the complexity of having cipher keys encryption in this awkward state is not ideal and I would very much like to have it always on here too. However, I don't think we're ready for that yet based on [PM-16556](https://bitwarden.atlassian.net/browse/PM-16556) still being on Hold until we get to the bottom of issues we ran into originally.

[PM-16556]: https://bitwarden.atlassian.net/browse/PM-16556?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

Hinton (approved):

gbubemismith (approved):

PR #353, opened 2025-07-22
Set opt-level z

Hinton:
## 📔 Objective

Set opt-level to `z` and enable _fat_ `lto`.

From testing this seems to result in a minimal performance decrease. Decrypting a vault of 2000 items in wasm goes from 1470ms to ~1500ms. While the `.wasm` filesize decreases from 3.5M to 1.9M.

## ⏰ Reminders before review

- CI builds passed
  team

## 🦮 Reviewer guidelines

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -104,8 +104,9 @@ opt-level = 1
 
 # Turn on LTO on release mode
 [profile.release]
-lto = "thin"
 codegen-units = 1
+lto = true
+opt-level = "z"
 
 # Stripping the binary reduces the size by ~30%, but the stacktraces won't be usable anymore.
 # This is fine as long as we don't have any unhandled panics, but let's keep it disabled for now


dani-garcia (approved):
Sounds reasonable, glad to know it doesn't affect performance too negatively. It would also be good to benchmark an argon2 operation, as I've seen those be quite sensitive to optimization level in the past.

I'm wondering if we should enable this only for WASM and let mobile take the normal opt level, as they probably won't care as much about build size, but seeing the small performance diff it might not be worth the effort.

As an aside question, I also see that a few lines under this we have a commented out `strip = true`, does enabling that or setting `strip = "debuginfo"` help reduce the size further? The stacktraces for rust panics in wasm are not great either way so it probably won't be a big loss.

Hinton:
I explored strip but I didn't see a noticeable improvement in file size.

PR #354, opened 2025-07-22
Create bitwarden-auth crate

dani-garcia:
## 📔 Objective

Created a new auth crate for the work that auth is moving to the SDK, and defined the WASM API for the send access functionality as a basic skeleton.

Some things left open:
- We already have an `AuthClient` in `bitwarden-core`, it's not used for WASM but it is for UniFFI. This PR leaves both clients in place but the goal is to migrate the older clients functionality over to the auth crate or somewhere else, and keep it out of core.
- Because we already have the `AuthClient` in core, I've made the extension trait for auth use a function name like `auth_new`, this is to avoid conflicts with the other function and should be renamed once the older `AuthClient` is deleted.
- Currently `AuthClient` is wrapping a bitwarden-core `Client`, so it can reuse the `ApiConfigurations` defined in core, this seems like it would complicate the usage of the auth crate, so I'm thinking we should extract the `ApiConfigurations` and maybe part of the token handling code to a separate bitwarden-api crate that both core and auth can use without depending on eachother?

## ⏰ Reminders before review

- CI builds passed
  team

## 🦮 Reviewer guidelines

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -7,6 +7,8 @@
 # Platform is the default owner for all files
 * @bitwarden/team-platform-dev
 
+crates/bitwarden-auth/** @bitwarden/team-auth-dev @bitwarden/team-platform-dev
+
 crates/bitwarden-vault/** @bitwarden/team-vault-dev @bitwarden/team-platform-dev
 
 # Temporarily owned by multiple teams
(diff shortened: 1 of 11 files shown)


justindbaur (approved):

JaredSnider-Bitwarden (approved):
TY!

PR #358, opened 2025-07-24
PM-23655: Map WI-FI -> Note

abergs:
## 🎟️ Tracking
https://bitwarden.atlassian.net/browse/PM-23655

## 📔 Objective

Maps wifi into a note on import

## ⏰ Reminders before review

- CI builds passed
  team

## 🦮 Reviewer guidelines


crates/bitwarden-exporters/src/cxf/secure_note.rs
Hinton, line 1: I think this file should be called wifi.

crates/bitwarden-exporters/src/cxf/secure_note.rs
```diff
@@ -0,0 +1,167 @@
...
+    if let Some(hidden) = &wifi.hidden {
+        fields.push(Field {
+            name: Some("Hidden".to_string()),
+            value: Some(if hidden.value.0 { "true" } else { "false" }.to_string()),
+            r#type: FieldType::Boolean as u8,
+            linked_id: None,
+        });
+    }
+
+    fields
```
Hinton: **suggestion:** Rather than using a mutable vector we can use iterator chaining.

```suggestion
      [
        wifi.ssid.as_ref().map(|ssid| Field {
            name: Some("SSID".to_string()),
            value: Some(ssid.value.0.clone()),
            r#type: FieldType::Text as u8,
            linked_id: None,
        }),
        wifi.passphrase.as_ref().map(|passphrase| Field {
            name: Some("Passphrase".to_string()),
            value: Some(passphrase.value.0.clone()),
            r#type: FieldType::Hidden as u8,
            linked_id: None,
        }),
        wifi.network_security_type.as_ref().map(|security| {
            let security_str = match &security.value {
                credential_exchange_format::EditableFieldWifiNetworkSecurityType::Unsecured => "Unsecured",
                credential_exchange_format::EditableFieldWifiNetworkSecurityType::WpaPersonal => "WPA Personal",
                credential_exchange_format::EditableFieldWifiNetworkSecurityType::Wpa2Personal => "WPA2 Personal",
                credential_exchange_format::EditableFieldWifiNetworkSecurityType::Wpa3Personal => "WPA3 Personal",
                credential_exchange_format::EditableFieldWifiNetworkSecurityType::Wep => "WEP",
                credential_exchange_format::EditableFieldWifiNetworkSecurityType::Other(s) => s,
            };
            Field {
                name: Some("Network Security Type".to_string()),
                value: Some(security_str.to_string()),
                r#type: FieldType::Text as u8,
                linked_id: None,
            }
        }),
        wifi.hidden.as_ref().map(|hidden| Field {
            name: Some("Hidden".to_string()),
            value: Some(if hidden.value.0 { "true" } else { "false" }.to_string()),
            r#type: FieldType::Boolean as u8,
            linked_id: None,
        }),
    ]
    .into_iter()
    .filter_map(|f| f)
    .collect()
```

crates/bitwarden-exporters/src/cxf/secure_note.rs
```diff
@@ -0,0 +1,167 @@
...
+
+        // Security type field
+        assert_eq!(fields[2].name, Some("Network Security Type".to_string()));
+        assert_eq!(fields[2].value, Some("WPA2 Personal".to_string()));
+        assert_eq!(fields[2].r#type, FieldType::Text as u8);
+
+        // Hidden field
+        assert_eq!(fields[3].name, Some("Hidden".to_string()));
+        assert_eq!(fields[3].value, Some("false".to_string()));
+        assert_eq!(fields[3].r#type, FieldType::Boolean as u8);
```
Hinton: **suggestion:** Similar to how we test it in `login`, https://github.com/bitwarden/sdk-internal/blob/1ba460c77a0704e15ef3b4c4104c6554f6061d9d/crates/bitwarden-exporters/src/cxf/login.rs#L300-L344 we can compare the vector directly i.e.

```rust
assert_eq!(
  fields
  vec![
    // ....
]);
```

crates/bitwarden-exporters/src/cxf/secure_note.rs
```diff
@@ -0,0 +1,167 @@
...
+        security_type: Option<EditableFieldWifiNetworkSecurityType>,
+        hidden: Option<bool>,
+    ) -> WifiCredential {
+        WifiCredential {
+            ssid: ssid.map(|s| EditableField {
+                id: None,
+                value: EditableFieldString(s.to_string()),
+                label: None,
+                extensions: None,
+            }),
```
Hinton: **suggestion:** Use the `from` available for the `EditableField` variants.

```suggestion
            ssid: ssid.map(|s| EditableFieldString(s.to_string())),
```

Hinton: At that point do we need the helper?

crates/bitwarden-exporters/src/cxf/secure_note.rs
```diff
@@ -0,0 +1,167 @@
...
+            }
+            credential_exchange_format::EditableFieldWifiNetworkSecurityType::Wpa2Personal => {
+                "WPA2 Personal"
+            }
+            credential_exchange_format::EditableFieldWifiNetworkSecurityType::Wpa3Personal => {
+                "WPA3 Personal"
+            }
+            credential_exchange_format::EditableFieldWifiNetworkSecurityType::Wep => "WEP",
+            credential_exchange_format::EditableFieldWifiNetworkSecurityType::Other(s) => s,
+        };
```
Hinton: **suggestion:** Extract this to a separate function that we can also test.

crates/bitwarden-exporters/src/cxf/secure_note.rs
```diff
@@ -0,0 +1,167 @@
...
+    let mut fields = Vec::new();
+
+    // SSID: Text field
+    if let Some(ssid) = &wifi.ssid {
+        fields.push(Field {
+            name: Some("SSID".to_string()),
+            value: Some(ssid.value.0.clone()),
+            r#type: FieldType::Text as u8,
+            linked_id: None,
+        });
```
Hinton: **suggestion:** It would be useful for us to have a helper for creating custom fields.

```Rust
fn create_field(name: &str, value: str, type: FieldType) {}

crate_field("SSID", ssid.value.0, FieldType::Text)
```

Hinton (changes requested):
Thanks!

I've suggested some general patterns that should make it a bit more similar to the codebase and improve the readability.

crates/bitwarden-exporters/src/cxf/secure_note.rs
abergs: Implemented in another PR, is that ok?

crates/bitwarden-exporters/src/cxf/secure_note.rs
abergs: That aligns with my other PRs where I changed my approach (to minimize conflicts) so I agree

crates/bitwarden-exporters/src/cxf/secure_note.rs
abergs: Thanks! To me this seems harder to read, but I'm sure that's just because I'm still getting used to rust. Happy to change it.

crates/bitwarden-exporters/src/cxf/secure_note.rs
Hinton: Up to you, you will have to go back and clean this up afterwards.

abergs:
Thank you for the feedback @Hinton, it's been incorporated.

crates/bitwarden-exporters/src/cxf/wifi.rs
```diff
@@ -0,0 +1,233 @@
...
+impl From<(&str, String, FieldType)> for Field {
+    fn from((name, value, field_type): (&str, String, FieldType)) -> Self {
+        Field {
+            name: Some(name.to_string()),
+            value: Some(value),
+            r#type: field_type as u8,
+            linked_id: None,
+        }
+    }
+}
```
Hinton: **suggestion:** Implementing `From` for a tuple is quite confusing and I think we'd be better of having a regular function, similar to the one in the tests.

**issue:** We're accepting a `&str`, and `String` in the same function. The `&str` is then immediately copied using `to_string`. For consistency we should change it all to be owned strings (i.e. `String`). (The reason we use `String` instead of `&str` is that in case we already have a `String` we would end up with an unnecessary copy of the string.

crates/bitwarden-exporters/src/cxf/wifi.rs
```diff
@@ -0,0 +1,233 @@
...
+    }
+
+    #[test]
+    fn test_wifi_to_fields_all_fields() {
+        let wifi = create_wifi_credential(
+            Some("MyWiFi"),
+            Some("secret123"),
+            Some(EditableFieldWifiNetworkSecurityType::Wpa2Personal),
+            Some(false),
+        );
```
Hinton: **suggestion:** Skip the helper function and just write it out. While it's _slightly_ more code (barely) it reduces the complexity of understanding what the test does.

```suggestion
        let wifi = WifiCredential {
            ssid: Some("MyWiFi".to_string().into()),
            passphrase: Some("secret123".to_string().into()),
            network_security_type: Some(EditableFieldWifiNetworkSecurityType::Wpa2Personal.into()),
            hidden: Some(false.into()),
        };
```

crates/bitwarden-exporters/src/cxf/wifi.rs
```diff
@@ -0,0 +1,233 @@
...
+
+    /// Helper function to create a Field for testing
+    fn create_field(name: &str, value: &str, field_type: FieldType) -> Field {
+        Field {
+            name: Some(name.to_string()),
+            value: Some(value.to_string()),
+            r#type: field_type as u8,
+            linked_id: None,
+        }
+    }
```
Hinton: **issue:** This function is the same as the `From` tuple.

**suggestion:** Remove it and rely on the previous implementation.

crates/bitwarden-exporters/src/cxf/wifi.rs
```diff
@@ -0,0 +1,233 @@
...
+
+        assert_eq!(
+            fields,
+            vec![
+                create_field("SSID", "MyWiFi", FieldType::Text),
+                create_field("Passphrase", "secret123", FieldType::Hidden),
+                create_field("Network Security Type", "WPA2 Personal", FieldType::Text),
+                create_field("Hidden", "false", FieldType::Boolean),
+            ]
+        );
```
Hinton: An alternative is to fully write out the fields in the test.

```suggestion
assert_eq!(
            fields,
            vec![
                Field {
                    name: Some("SSID".to_string()),
                    value: Some("MyWiFi".to_string()),
                    r#type: 0,
                    linked_id: None
                },
                Field {
                    name: Some("Passphrase".to_string()),
                    value: Some("secret123".to_string()),
                    r#type: 1,
                    linked_id: None
                },
                Field {
                    name: Some("Network Security Type".to_string()),
                    value: Some("WPA2 Personal".to_string()),
                    r#type: 0,
                    linked_id: None
                },
                Field {
                    name: Some("Hidden".to_string()),
                    value: Some("false".to_string()),
                    r#type: 2,
                    linked_id: None
                }
            ]
        );
```

If we keep the current helper we should change it to the same implementation as the code, and add unit tests to validate it produces the expected output.

Hinton (dismissed):

crates/bitwarden-exporters/src/cxf/editable_field.rs
```diff
@@ -0,0 +1,225 @@
...
+        value: Some(field.field_value()),
+        r#type: field.field_type() as u8,
+        linked_id: None,
+    }
+}
+
+/// Trait to convert EditableField types to Field values and types
+pub(super) trait EditableFieldToField {
+    fn field_value(&self) -> String;
+    fn field_type(&self) -> FieldType;
```
dani-garcia: Suggestion, `field_type` never uses `&self`, so I'd probably remove it from the trait. In fact it might be a better idea to change the definition to:
```rust
pub(super) trait EditableFieldToField {
    const FIELD_TYPE: FieldType;
    fn field_value(&self) -> String;
}
```

This way it's more clear that a field type will never change?

crates/bitwarden-exporters/src/cxf/editable_field.rs
```diff
@@ -0,0 +1,225 @@
...
+    }
+
+    fn field_type(&self) -> FieldType {
+        FieldType::Hidden
+    }
+}
+
+impl EditableFieldToField for EditableField<EditableFieldBoolean> {
+    fn field_value(&self) -> String {
+        if self.value.0 { "true" } else { "false" }.to_string()
```
dani-garcia: booleans already get stringified as `true/false`, so we can just call tostring on it directly
```suggestion
        self.value.0.to_string()
```

crates/bitwarden-exporters/src/cxf/editable_field.rs
```diff
@@ -0,0 +1,225 @@
...
+/// Convert WiFi security type enum to human-readable string
+fn security_type_to_string(security_type: &EditableFieldWifiNetworkSecurityType) -> &str {
+    match security_type {
+        EditableFieldWifiNetworkSecurityType::Unsecured => "Unsecured",
+        EditableFieldWifiNetworkSecurityType::WpaPersonal => "WPA Personal",
+        EditableFieldWifiNetworkSecurityType::Wpa2Personal => "WPA2 Personal",
+        EditableFieldWifiNetworkSecurityType::Wpa3Personal => "WPA3 Personal",
+        EditableFieldWifiNetworkSecurityType::Wep => "WEP",
+        EditableFieldWifiNetworkSecurityType::Other(s) => s,
+    }
```
dani-garcia: This is more of a personal thing, but when matching enums with big names, I like to star import the variants to avoid the repetition:
```rust
    use EditableFieldWifiNetworkSecurityType::*;
    match security_type {
        Unsecured => "Unsecured",
        WpaPersonal => "WPA Personal",
        Wpa2Personal => "WPA2 Personal",
        Wpa3Personal => "WPA3 Personal",
        Wep => "WEP",
        Other(s) => s,
    }
```

crates/bitwarden-exporters/src/cxf/editable_field.rs
```diff
@@ -0,0 +1,225 @@
...
+
+use crate::Field;
+
+/// Helper function to create a Field from any EditableField type
+pub(super) fn create_field<T>(name: String, field: &T) -> Field
+where
+    T: EditableFieldToField,
+{
+    Field {
+        name: Some(name),
```
dani-garcia: Almost every time we call this function, we're passing it a &str and calling .into() each time. It feels like it would be simpler to do something like:
```rust
pub(super) fn create_field<T>(name: impl Into<String>, field: &T) -> Field
where
    T: EditableFieldToField,
{
    Field {
        name: Some(name.into()),
```
And then functions could pass a &str or a String as needed.

crates/bitwarden-exporters/src/cxf/editable_field.rs
```diff
@@ -0,0 +1,225 @@
...
+        value: Some(field.field_value()),
+        r#type: field.field_type() as u8,
+        linked_id: None,
+    }
+}
+
+/// Trait to convert EditableField types to Field values and types
```
dani-garcia: We should probably be a bit more clear in who owns each type, as they have very similar names and they're easy to confuse.
```suggestion
/// Trait to convert CXP EditableField types to Bitwarden Field values and types
```

dani-garcia (commented):
Looks pretty good, I have some comments

crates/bitwarden-exporters/src/cxf/editable_field.rs
Hinton: It is, I initially thought we might at some point need to dynamically determine it but I can't think of a reason to do so beyond future proofing.

dani-garcia (approved):

abergs:
Thank you @dani-garcia and @Hinton for your fast and helpful review on this.

PR #361, opened 2025-07-25
[PM-23639] Add CXF Import mappings

abergs:
## 🎟️ Tracking
https://bitwarden.atlassian.net/browse/PM-23639

## 📔 Objective

This PR adds the CXF sample from the repo so that we can use it for reference in unit tests.

**It also servers as the feature branch for adding CXF Import mappings for multiple CXF data types.**

## ⏰ Reminders before review

- CI builds passed
  team

## 🦮 Reviewer guidelines


abergs:
How do I tell Sonar to relax?

crates/bitwarden-exporters/src/cxf/import.rs
```diff
@@ -177,10 +181,38 @@ struct GroupedCredentials {
...
+    fn load_sample_cxf() -> Result<Vec<ImportingCipher>, CxfError> {
+        use std::fs;
+
+        // Read the actual CXF example file
+        let cxf_data = fs::read_to_string("resources/cxf_example.json")
+            .expect("Should be able to read cxf_example.json");
+
+        // Workaround for library bug: the example file has "integrityHash" but the library expects
+        // "integrationHash"
+        let fixed_cxf_data = cxf_data.replace("\"integrityHash\":", "\"integrationHash\":");
```
Hinton: This should be fixed now, you can update the credential exchange lib to use the git commit.

crates/bitwarden-exporters/src/cxf/import.rs
```diff
@@ -15,9 +15,13 @@ use crate::{
...
 pub(crate) fn parse_cxf(payload: String) -> Result<Vec<ImportingCipher>, CxfError> {
-    let account: CxfAccount = serde_json::from_str(&payload)?;
+    let header: Header = serde_json::from_str(&payload)?;
 
-    let items: Vec<ImportingCipher> = account.items.into_iter().flat_map(parse_item).collect();
+    let items: Vec<ImportingCipher> = header
+        .accounts
+        .into_iter()
+        .flat_map(|account| account.items.into_iter().flat_map(parse_item))
+        .collect();
```
Hinton: **question:** Why are we changing this to header? The input from Apple is accounts.

Hinton (changes requested):

crates/bitwarden-exporters/src/cxf/import.rs
abergs: The json file starts with the header info, right? https://github.com/bitwarden/sdk-internal/pull/361/files#diff-4c4eca692a72e9b9bac1667196afc5849a5a26dfce4260540bcacc0722047870R2

abergs: I see what you mean. It's used differently between the Apple call site and the cxf_example.json I guess?

crates/bitwarden-exporters/src/cxf/import.rs
abergs: yes, adressed.

crates/bitwarden-exporters/Cargo.toml
```diff
@@ -20,7 +20,7 @@ uniffi = ["dep:uniffi", "bitwarden-core/uniffi"] # Uniffi bindings
 wasm = [
     "bitwarden-vault/wasm",
     "dep:tsify",
-    "dep:wasm-bindgen"
+    "dep:wasm-bindgen",
```
Hinton: You need to run cargo sort here.

crates/bitwarden-exporters/src/cxf/import.rs
```diff
@@ -22,6 +27,24 @@ pub(crate) fn parse_cxf(payload: String) -> Result<Vec<ImportingCipher>, CxfErro
...
+    let header: Header = serde_json::from_str(&payload)?;
+
+    let items: Vec<ImportingCipher> = header
+        .accounts
+        .into_iter()
+        .flat_map(|account| account.items.into_iter().flat_map(parse_item))
+        .collect();
+
+    Ok(items)
+}
```
Hinton: **suggestion:** Move into the text cfg. We can lift it out when it becomes relevant.

crates/bitwarden-exporters/Cargo.toml
abergs: > Finished: Cargo.toml for bitwarden-exporters is sorted already, no changes made

🤷

crates/bitwarden-exporters/src/cxf/import.rs
abergs: Moved.

crates/bitwarden-exporters/Cargo.toml
abergs: done

crates/bitwarden-exporters/resources/cxf_example.json
```diff
@@ -0,0 +1,599 @@
...
+                    "id": "Z4cFmc21Q5-vCVwd1wJx1g",
+                    "creationAt": 1705142400,
+                    "modifiedAt": 1705228800,
+                    "title": "SSH Key",
+                    "subtitle": "GitHub",
+                    "credentials": [
+                        {
+                            "type": "ssh-key",
+                            "keyType": "ssh-rsa",
+                            "privateKey": "MC4CAQAwBQYDK2VwBCIEID-U9VakauO4Fsv4b_znpDHcdYg74U68siZjnWLPn7Q1",
```
Hinton, line 285: We tweaked this again, let's see which ends up in the final example.

Hinton (dismissed):

crates/bitwarden-exporters/src/cxf/identity.rs
```diff
@@ -0,0 +1,292 @@
...
+) -> (Option<String>, Option<String>) {
+    full_name.as_ref().map_or((None, None), |name| {
+        let parts: Vec<&str> = name.value.0.split_whitespace().collect();
+        match parts.as_slice() {
+            [] => (None, None),
+            [first] => (Some(first.to_string()), None),
+            [first, rest @ ..] => (Some(first.to_string()), Some(rest.join(" "))),
+        }
+    })
+}
```
Hinton, line 249: **suggestion:** Can we add some unit tests to this one in this file?

crates/bitwarden-exporters/src/cxf/login.rs
```diff
@@ -18,17 +19,42 @@ use crate::{Fido2Credential, Field, Login, LoginUri};
...
+/// This ensures we use the exact same encoding and formatting as Bitwarden's core implementation
+fn totp_credential_to_totp(cxf_totp: &TotpCredential) -> Totp {
+    let algorithm = match cxf_totp.algorithm {
+        OTPHashAlgorithm::Sha1 => TotpAlgorithm::Sha1,
+        OTPHashAlgorithm::Sha256 => TotpAlgorithm::Sha256,
+        OTPHashAlgorithm::Sha512 => TotpAlgorithm::Sha512,
+        OTPHashAlgorithm::Unknown(ref algo) if algo == "steam" => TotpAlgorithm::Steam,
+        OTPHashAlgorithm::Unknown(_) | _ => TotpAlgorithm::Sha1, /* Default to SHA1 for unknown
+                                                                  * algorithms */
+    };
```
Hinton: **question (non blocking):** Do you want to extract this. into it's own function and implement unit tests for all braces? I don't think it's strictly required but might be nice.

crates/bitwarden-exporters/src/cxf/import.rs
```diff
@@ -30,7 +40,7 @@ fn convert_date(ts: Option<u64>) -> DateTime<Utc> {
         .unwrap_or(Utc::now())
 }
 
-fn parse_item(value: Item) -> Vec<ImportingCipher> {
+pub(super) fn parse_item(value: Item) -> Vec<ImportingCipher> {
```
Hinton, line 45: **suggestion:** I believe this is only used in a test case within `note.rs` now, if we move that test case into this file we can make this private again.

crates/bitwarden-exporters/src/cxf/import.rs
```diff
@@ -181,6 +238,42 @@ mod tests {
 
     use super::*;
 
+    /**
+     * Parse CXF payload in the format compatible with the CXF specification (At the
+     * Header-level).
+     */
```
Hinton: **issue:** Wrong comments, rust almost always uses `//`, and `///` when it should be exposed as documentation.

Hinton (changes requested):

crates/bitwarden-exporters/src/cxf/import.rs
abergs: The CXF sample tests depend on this method. I've moved them to their own file now, so this visibility is still needed.

crates/bitwarden-exporters/src/cxf/import.rs
abergs: TIL. Thanks, fixed.

crates/bitwarden-exporters/src/cxf/identity.rs
abergs: Yes, I've added some basic tests with none/empty values as well as more common strings.

crates/bitwarden-exporters/src/cxf/login.rs
abergs: Yes, done.

crates/bitwarden-exporters/src/cxf/import.rs
```diff
@@ -40,85 +51,96 @@ fn parse_item(value: Item) -> Vec<ImportingCipher> {
...
+    ]
+    .into_iter()
+    .flatten()
+    .for_each(|(identity, custom_fields)| {
+        add_item(CipherType::Identity(Box::new(identity)), custom_fields);
+    });
+
+    // Standalone Note credentials -> Secure Note (only if no other credentials exist)
+    if !grouped.note.is_empty() && output.is_empty() {
+        let standalone_note_content = grouped.note.first().map(extract_note_content);
```
Hinton, line 151: This should probably use add_item?

abergs: That was my first thought too, but it can't because we run into a problem with the borrow checker on the output array which is capture in the conditional (I think that's why, but I'm no expert). It wouldn't compile when I used the add_item. 

Perhaps it could be refactored to work? I know to little rust to know in what way.

Hinton: Probably. Let's handle it later.

crates/bitwarden-exporters/src/cxf/import_sample_tests.rs
```diff
@@ -0,0 +1,378 @@
...
+        let note_cipher = items
+            .iter()
+            .find(|cipher| cipher.name == "Home alarm")
+            .expect("Should find Home alarm note item");
+
+        // Validate it's a SecureNote cipher
+        match &note_cipher.r#type {
+            CipherType::SecureNote(_) => (), // Successfully identified as SecureNote
+            _ => panic!("Expected SecureNote for standalone note credential"),
+        }
```
Hinton, line 113: We generally recommend using if let when you only care about a single branch. (This applies to most tests in this file)

```suggestion
        let identity = if let CipherType::Identity(identity) = &address_cipher.r#type {
            identity
        } else {
            panic!("Expected Identity cipher for address")
        };
```

crates/bitwarden-exporters/src/cxf/import_sample_tests.rs
```diff
@@ -0,0 +1,378 @@
...
+
+        let issuing_authority = identity_document_cipher
+            .fields
+            .iter()
+            .find(|f| f.name.as_deref() == Some("Issuing Authority"))
+            .expect("Should have Issuing Authority");
+        assert_eq!(
+            issuing_authority.value,
+            Some("Department of State".to_string())
+        );
```
Hinton: **suggestion:** Compare fields directly which checks more than just the value.
```suggestion
        let expected_fields = vec![
            Field {
                name: Some("Issuing Country".to_string()),
                value: Some("US".to_string()),
                r#type: FieldType::Text as u8,
                linked_id: None,
            },
            Field {
                name: Some("Nationality".to_string()),
                value: Some("American".to_string()),
                r#type: FieldType::Text as u8,
                linked_id: None,
            },
            Field {
                name: Some("Birth Date".to_string()),
                value: Some("1990-04-15".to_string()),
                r#type: FieldType::Text as u8,
                linked_id: None,
            },
            Field {
                name: Some("Birth Place".to_string()),
                value: Some("New York, USA".to_string()),
                r#type: FieldType::Text as u8,
                linked_id: None,
            },
            Field {
                name: Some("Sex".to_string()),
                value: Some("F".to_string()),
                r#type: FieldType::Text as u8,
                linked_id: None,
            },
            Field {
                name: Some("Issue Date".to_string()),
                value: Some("2020-01-01".to_string()),
                r#type: FieldType::Text as u8,
                linked_id: None,
            },
            Field {
                name: Some("Expiry Date".to_string()),
                value: Some("2030-01-01".to_string()),
                r#type: FieldType::Text as u8,
                linked_id: None,
            },
            Field {
                name: Some("Issuing Authority".to_string()),
                value: Some("Department of State".to_string()),
                r#type: FieldType::Text as u8,
                linked_id: None,
            },
        ];

        assert_eq!(identity_document_cipher.fields, expected_fields);
```

Hinton (dismissed):

crates/bitwarden-exporters/Cargo.toml
```diff
@@ -30,9 +30,10 @@ bitwarden-core = { workspace = true }
 bitwarden-crypto = { workspace = true }
 bitwarden-error = { workspace = true }
 bitwarden-fido = { workspace = true }
+bitwarden-ssh = { workspace = true }
 bitwarden-vault = { workspace = true }
 chrono = { workspace = true, features = ["std"] }
-credential-exchange-format = ">=0.1, <0.2"
+credential-exchange-format = { git = "https://github.com/bitwarden/credential-exchange", rev = "38e8a013c13644f832c457555baaa536fe481b77" }
```
dani-garcia, line 36: Question: Are we going to make a new release of this crate?

crates/bitwarden-exporters/src/cxf/identity.rs
```diff
@@ -0,0 +1,461 @@
...
+
+/// Convert address credentials to Identity (no custom fields needed for address)
+/// According to the mapping specification:
+/// - streetAddress: EditableField<"string"> → Identity::address1
+/// - city: EditableField<"string"> → Identity::city
+/// - territory: EditableField<"subdivision-code"> → Identity::state
+/// - country: EditableField<"country-code"> → Identity::country
+/// - tel: EditableField<"string"> → Identity::phone
+/// - postalCode: EditableField<"string"> → Identity::postal_code
+pub(super) fn address_to_identity(address: &AddressCredential) -> (Identity, Vec<Field>) {
```
dani-garcia: Non-blocking thought: With all the amount of clones we're doing in this function, I feel like it would be easier to take `AddressCredential` by value and just clone the whole struct once when it's passed in. Similar thing for the other conversion functions here.

crates/bitwarden-exporters/src/cxf/identity.rs
```diff
@@ -0,0 +1,461 @@
...
+        person_name.surname_prefix.as_ref(),
+        person_name.surname.as_ref(),
+        person_name.surname2.as_ref(),
+    ]
+    .into_iter()
+    .flatten()
+    .map(|field| field.value.0.clone())
+    .collect::<Vec<_>>()
+    .into_iter()
+    .reduce(|acc, part| format!("{} {}", acc, part));
```
dani-garcia: This will be a warning in the latest version of rust:
```suggestion
    .reduce(|acc, part| format!("{acc} {part}"));
```

crates/bitwarden-ssh/src/import.rs
```diff
@@ -32,6 +32,12 @@ pub fn import_key(
     }
 }
 
+/// Import a DER encoded private key, and returns a decoded [SshKeyView]. This is primarily used for
+/// importing SSH keys from other Credential Managers through Credential Exchange.
+pub fn import_pkcs8_der(encoded_key: &[u8]) -> Result<SshKeyView, SshKeyImportError> {
```
dani-garcia: Why not make `import_der_key` public, instead of having this wrapper that just calls it?

crates/bitwarden-exporters/Cargo.toml
Hinton: Yes. I wanted to explore if we needed anything else since we'll need a major bump again 😞 .

crates/bitwarden-exporters/src/cxf/identity.rs
Hinton: Done, much simpler code.

dani-garcia (approved):

PR #369, opened 2025-07-30
[PM-24243] Add load_flags method to wasm PlatformClient

shane-melton:
## 🎟️ Tracking

[PM-24243](https://bitwarden.atlassian.net/browse/PM-24243)

## 📔 Objective

Add `load_flags` method to the wasm `PlatformClient` to match parity with the uniffi `PlatformClient` so that the TS clients can set the SDK flags dynamically.

Related Clients PR: https://github.com/bitwarden/clients/pull/15855

~Added a `FlagsInput` struct that mimics the internal `Flags` struct but with tsify bindings.~

~Another potential option is to make the internal `Flags` struct public with wasm bindings to avoid duplicating the struct.~

## ⏰ Reminders before review

- CI builds passed
  team

## 🦮 Reviewer guidelines

[PM-24243]: https://bitwarden.atlassian.net/browse/PM-24243?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/crates/bitwarden-core/src/client/internal.rs
+++ b/crates/bitwarden-core/src/client/internal.rs
@@ -118,6 +118,12 @@ impl InternalClient {
 
     #[allow(missing_docs)]
     #[cfg(feature = "internal")]
+    pub fn set_flags(&self, flags: &Flags) {
+        *self.flags.write().expect("RwLock is not poisoned") = flags.clone();
+    }
+
+    #[allow(missing_docs)]
+    #[cfg(feature = "internal")]
     pub fn get_flags(&self) -> Flags {
         self.flags.read().expect("RwLock is not poisoned").clone()
     }
--- a/crates/bitwarden-core/src/client/mod.rs
+++ b/crates/bitwarden-core/src/client/mod.rs
@@ -19,6 +19,7 @@ mod flags;
 
 pub use client::Client;
 pub use client_settings::{ClientSettings, DeviceType};
+pub use flags::Flags;
 
 #[allow(missing_docs)]
 #[cfg(feature = "internal")]
--- a/crates/bitwarden-core/src/client/mod.rs
+++ b/crates/bitwarden-core/src/client/mod.rs
@@ -14,8 +14,8 @@ pub mod login_method;
 #[cfg(feature = "secrets")]
 pub(crate) use login_method::ServiceAccountLoginMethod;
 pub(crate) use login_method::{LoginMethod, UserLoginMethod};
-#[cfg(feature = "internal")]
-mod flags;
+#[allow(missing_docs)]
+pub mod flags;
 
 pub use client::Client;
 pub use client_settings::{ClientSettings, DeviceType};
--- a/crates/bitwarden-core/src/client/mod.rs
+++ b/crates/bitwarden-core/src/client/mod.rs
@@ -14,11 +14,12 @@ pub mod login_method;
 #[cfg(feature = "secrets")]
 pub(crate) use login_method::ServiceAccountLoginMethod;
 pub(crate) use login_method::{LoginMethod, UserLoginMethod};
-#[allow(missing_docs)]
-pub mod flags;
+#[cfg(feature = "internal")]
+mod flags;
 
 pub use client::Client;
 pub use client_settings::{ClientSettings, DeviceType};
+#[cfg(feature = "internal")]
 pub use flags::Flags;
 
 #[allow(missing_docs)]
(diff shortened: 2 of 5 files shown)


crates/bitwarden-wasm-internal/src/platform/mod.rs
```diff
@@ -1,10 +1,28 @@
...
+#[tsify(into_wasm_abi, from_wasm_abi)]
+pub struct FlagsInput {
+    #[serde(rename = "enableCipherKeyEncryption")]
+    pub enable_cipher_key_encryption: Option<bool>,
+}
+
+impl From<FlagsInput> for std::collections::HashMap<String, bool> {
+    fn from(flags: FlagsInput) -> Self {
+        let js_value = serde_wasm_bindgen::to_value(&flags)
+            .expect("FlagsInput should always serialize successfully");
```
dani-garcia: We should try to avoid expect, as it would produce a panic. I don't think it would happen in this case but better safe than sorry.
```suggestion
        let js_value = serde_wasm_bindgen::to_value(&flags).unwrap_or_default();
```

crates/bitwarden-wasm-internal/src/platform/mod.rs
```diff
@@ -1,10 +1,28 @@
...
-use wasm_bindgen::prelude::wasm_bindgen;
+use tsify::{serde_wasm_bindgen, Tsify};
+use wasm_bindgen::prelude::*;
 
 mod repository;
 pub mod token_provider;
 
+#[derive(serde::Serialize, serde::Deserialize, Tsify)]
+#[tsify(into_wasm_abi, from_wasm_abi)]
+pub struct FlagsInput {
```
dani-garcia: Rather than duplicate the Flags struct, we should tag the Flags struct in bitwarden_core with:
```
#[cfg_attr(
    feature = "wasm",
    derive(tsify::Tsify),
)]
```

Otherwise we'll have to ensure we keep them both in sync, which is error prone.

shane-melton: Agreed! Thanks @dani-garcia! Updated in [9a7ca68](https://github.com/bitwarden/sdk-internal/pull/369/commits/9a7ca686cbacde6fea04679360669ddcb688d930)

I also had to make the `Flags` struct in `bitwarden_core` public and remove the `#[cfg(feature="internal")]` attribute in [8f2680b](https://github.com/bitwarden/sdk-internal/pull/369/commits/8f2680bb5e40a793c8f9885a856d87f5d846fc37). If there's another or better way to go about it please let me know!

crates/bitwarden-wasm-internal/src/platform/mod.rs
shane-melton: Good to know! But this should no longer be needed after [9a7ca68](https://github.com/bitwarden/sdk-internal/pull/369/commits/9a7ca686cbacde6fea04679360669ddcb688d930)

crates/bitwarden-core/src/client/mod.rs
```diff
@@ -14,11 +14,12 @@ pub mod login_method;
 #[cfg(feature = "secrets")]
 pub(crate) use login_method::ServiceAccountLoginMethod;
 pub(crate) use login_method::{LoginMethod, UserLoginMethod};
-#[cfg(feature = "internal")]
-mod flags;
+#[allow(missing_docs)]
+pub mod flags;
```
dani-garcia:
```suggestion
mod flags;
```

You don't need to make this module pub, as you're already doing `pub use flags::Flags` a few lines after it.

Also both lines should be able to be tagged with the internal flag, the wasm crate is enabling it.
```
#[cfg(feature = "internal")]
mod flags;

#[cfg(feature = "internal")]
pub use flags::Flags;
```

shane-melton: Gotcha, thanks! Fixed in [1484c3c](https://github.com/bitwarden/sdk-internal/pull/369/commits/1484c3c25972ba5d6715887188781d1b9cd5fb6d)

justindbaur (approved):

crates/bitwarden-core/src/client/mod.rs
Hinton: We intentionally didn't make flags public for mobile in order to avoid removal of flags from being a breaking change.

crates/bitwarden-core/src/client/internal.rs
```diff
@@ -116,6 +116,12 @@ impl InternalClient {
         *self.flags.write().expect("RwLock is not poisoned") = Flags::load_from_map(flags);
     }
 
+    #[allow(missing_docs)]
+    #[cfg(feature = "internal")]
+    pub fn set_flags(&self, flags: &Flags) {
+        *self.flags.write().expect("RwLock is not poisoned") = flags.clone();
+    }
```
Hinton, line 123: We already have `load_flags` above, and `#[allow(missing_docs)]` is not allowed on new functions.

PR #372, opened 2025-07-31
[PM-24263] Pin protected key envelope unlock

quexten:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-24263

## 📔 Objective

Adds enrollment functionality for Password(PIN)Protected user key envelope. Both crypto initialization via the init request, *and* a function exposing the raw key material are provided. The latter is required since unlock is not yet done via the init methods on WASM/`clients`.

## ⏰ Reminders before review

- CI builds passed
  team

## 🦮 Reviewer guidelines


coroiu (dismissed):
Looks good to me, but I don't have a ton of experience with the key store. Let me know if there's anything you'd like me to take a deeper look at!

quexten:
Looks like we had a deadlock on the keystore lock, caught by the CI / tests which should be fixed in the latest revision.

crates/bitwarden-core/src/client/encryption_settings.rs
```diff
@@ -48,6 +48,9 @@ pub enum EncryptionSettingsError {
 
     #[error(transparent)]
     UserIdAlreadySetError(#[from] UserIdAlreadySetError),
+
+    #[error("Wrong Pin")]
+    WrongPin,
```
Hinton, line 50: **question:** What error do we return if you pass in the wrong password? It would seem reasonable that we return a similar error for both.

I suspect we derive the master key and then attempt to decrypt the user key which would fail and you'd get a decrypt error? Maybe it would be more elegant to handle that with a better error message in which case both wrong password and pin could use the same error?

crates/bitwarden-crypto/src/safe/password_protected_key_envelope.rs
Hinton, line 1: Can we lift this into the previous PR or is there a good reason this is handled here?

mzieniukbw (dismissed):

crates/bitwarden-core/src/key_management/crypto_client.rs
```diff
@@ -80,6 +82,31 @@ impl CryptoClient {
...
+    /// used to initialize another client instance by using the PIN and the PIN key with
+    /// `initialize_user_crypto`.
+    pub fn enroll_pin(&self, pin: String) -> Result<EnrollPinResponse, CryptoClientError> {
+        enroll_pin(&self.client, pin)
+    }
+
+    /// Decrypts a `PasswordProtectedKeyEnvelope`, returning the user key, if successful.
+    /// This is a stop-gap solution, until initialization of the SDK is used.
+    #[cfg(feature = "wasm")]
+    pub fn unseal_password_protected_key_envelope(
```
mzieniukbw, line 116: 💡 Shall we add some unit test coverage ?

crates/bitwarden-core/src/client/internal.rs
```diff
@@ -315,6 +318,31 @@ impl InternalClient {
...
+        pin_protected_user_key_envelope: PasswordProtectedKeyEnvelope,
+        key_state: UserKeyState,
+    ) -> Result<(), EncryptionSettingsError> {
+        let decrypted_user_key = {
+            // Note: This block ensures ctx is dropped. Otherwise it would cause a deadlock when
+            // initializing the user crypto
+            use crate::key_management::SymmetricKeyId;
+            let ctx = &mut self.key_store.context_mut();
+            let decrypted_user_key_id = pin_protected_user_key_envelope
+                .unseal(SymmetricKeyId::Local("tmp_unlock_pin"), &pin, ctx)
```
Hinton, line 328: @dani-garcia did we have any convention for local IDs? Or is it just the wild west.

crates/bitwarden-crypto/src/error.rs
```diff
@@ -69,6 +69,9 @@ pub enum CryptoError {
 
     #[error("Encoding error, {0}")]
     EncodingError(#[from] EncodingError),
+
+    #[error("Password protected key envelope error, {0}")]
+    PasswordProtectedKeyEnvelopeError(#[from] PasswordProtectedKeyEnvelopeError),
```
Hinton: **nit:** We try and avoid having `Error` suffixes in the variant names.
```suggestion
    PasswordProtectedKeyEnvelope(#[from] PasswordProtectedKeyEnvelopeError),
```

crates/bitwarden-crypto/examples/protect_key_with_password.rs
Hinton, line 1: Lift to previous PR?

crates/bitwarden-core/src/key_management/non_generic_wrappers.rs
Hinton, line 1: **question:** What happens if we want to sent another generic of `PasswordProtectedKeyEnvelope`? We've reserved the name of the generic now.

crates/bitwarden-crypto/src/error.rs
Hinton: Is this required to live in `CryptoError` or could we lift it up. I'm a bit concerned about the size of `CryptoError` since it's intended as a low level error.

crates/bitwarden-core/src/key_management/crypto.rs
```diff
@@ -315,6 +334,40 @@ pub(super) fn update_password(
...
+#[cfg_attr(feature = "uniffi", derive(uniffi::Record))]
+#[cfg_attr(feature = "wasm", derive(Tsify), tsify(into_wasm_abi, from_wasm_abi))]
+pub struct EnrollPinResponse {
+    /// [UserKey] protected by PIN
+    pin_protected_user_key_envelope: PasswordProtectedKeyEnvelope,
+    /// PIN protected by [UserKey]
+    user_key_encrypted_pin: EncString,
+}
+
+pub(super) fn enroll_pin(
```
matt-livefront, line 350: ❓ Does this need to be exposed to uniffi? I'm not seeing in the Swift bindings.

crates/bitwarden-core/src/client/internal.rs
quexten: I think the approach here is to push the other work dani proposed ( https://bitwarden.atlassian.net/browse/PM-18102), so that we don't have to deal with naming collisions..

crates/bitwarden-core/src/client/encryption_settings.rs
quexten: With master-key decryption you cannot know if:

- Your email is different to the one used on creation
- Your kdf settings are different to the one used on creation
- Your password is different "wrong" to the one used on creation
- (Your master-key-wrapped-user-key was tampered with)

and all of these would result in the same error. In *most cases* the true reason would be "wrong password" but there are reasonable cases such as the KDF settings being out of sync, that would lead to the same error path.

With the PasswordProtectedKeyEnvelope, it's only:
- Your password is different "wrong"
- (Your key envelope was tampered with)
so for me it feels more appropriate to use the "WrongPin" error.

(I'm OK changing both to return "WrongPassword", but then we should document that this could also mean that KDF/salt(email) can be wrong for masterkey encrypted items).

crates/bitwarden-core/src/key_management/non_generic_wrappers.rs
quexten: I think in this specific case it's OK. The reason is that the generic parameter is `KeyIds`, of which only one instance is ever supposed to exist as far as I'm aware - the one defined in core, describing all Bitwarden keys in the key hierarchy.

I would have preferred to not even make the key envelope generic in the first place, but had to do so in order to have the public interface take a KeyStoreContext and key ids instead of literal keys.


For cases where we have actually varying generic parameters, we should make the names of the non-generic wrappers such that they don't collide. I.e "CoseKeyBytes". "LegacyBitwardenKeyBytes", etc.

quexten: Note: I've lifted this into the previous PR too, because it's not specific to pin unlock.

crates/bitwarden-crypto/src/safe/password_protected_key_envelope.rs
quexten: GitHub hid what this references, but I've lifted anything that I found to be non-pin-generic to the previous PR, which should make the diff a lot cleaner.

crates/bitwarden-core/src/key_management/crypto.rs
quexten: Thanks for catching this! It seems we still separate out clients which is confusing, but it should be fixed in the latest revision of the PR, and I have verified this gets emitted to the swift bindings locally.

crates/bitwarden-core/src/key_management/crypto_client.rs
quexten: Added test for the enroll -> re-enroll with encrypted -> unseal flow.

crates/bitwarden-crypto/src/error.rs
quexten: Would it be more acceptable to be a variant of `CryptoClientError`? In that case we can lift it up to there.

quexten: (Lifted for now)

crates/bitwarden-core/src/client/internal.rs
dani-garcia: Yeah, I'd like to see that merged so we don't have to care about naming in situations like this.

crates/bitwarden-core/src/key_management/crypto.rs
```diff
@@ -315,6 +337,38 @@ pub(super) fn update_password(
...
+    client: &Client,
+    pin: String,
+) -> Result<EnrollPinResponse, CryptoClientError> {
+    let key_store = client.internal.get_key_store();
+    let mut ctx = key_store.context_mut();
+
+    let key_envelope =
+        PasswordProtectedKeyEnvelope(bitwarden_crypto::safe::PasswordProtectedKeyEnvelope::seal(
```
Hinton, line 358: **suggestion:** We already import `PasswordProtectedKeyEnvelope`, use the short path.
```suggestion
        PasswordProtectedKeyEnvelope(PasswordProtectedKeyEnvelope::seal(
```

Hinton (dismissed):

crates/bitwarden-core/src/key_management/crypto.rs
quexten: Ah, my bad I should have not merged this. The imported PasswordProtectedKeyEnvelope is the non-generic wrapper. The contained envelope that uses the long path is the generic struct from crypto. So they are different structs.

Maybe they should have different naming after all..
I've undone the merge now.

Hinton: Yea that naming is unfortunate.

Hinton (dismissed):

dani-garcia (dismissed):

quexten:
Had to resolve some conflicts from the `B64` work

dani-garcia (approved):

quexten:
Merging since there was no significant change since the last set of reviews, so 1 review seems enough.

PR #373, opened 2025-07-31
UniFFI don't panic on failed lift

dani-garcia:
## 📔 Objective

UniFFI uses the `custom_type!` macro to convert request/response types. The request conversion is allowed to fail which should produce an error in the apps, but there are cases where it would produce a panic instead. The cases are:
- If the FFI function doesn't return a result, then UniFFI produces a panic.
- If the Error type of the function is different than the Error type in `custom_type!`, then UniFFI produces a panic.

One way to fix this is to have the `custom_type!` calls return `bitwarden_uniffi::error::BitwardenError`, but this isn't possible as core and crypto can't depend on uniffi due to circular dependency issues.

This PR makes the uniffi crate register an error conversion function during client initialization, thanks to the use of a new `uniffi-error` crate. It also updates all the calls to `custom_type!` to use this new error conversion.

## ⏰ Reminders before review

- CI builds passed
  team

## 🦮 Reviewer guidelines


addisonbeck (dismissed):

crates/bitwarden-uniffi-error/src/lib.rs
```diff
@@ -0,0 +1,38 @@
...
+use std::sync::OnceLock;
+
+#[allow(clippy::type_complexity)]
+static ERROR_TO_UNIFFI_ERROR: OnceLock<
+    Box<dyn Fn(Box<dyn std::error::Error + Send + Sync>) -> anyhow::Error + Send + Sync + 'static>,
+> = OnceLock::new();
+
+pub use anyhow::Error;
+
+/// Configure an error converter to conver errors in calls to [`uniffi::custom_type!`] into the main
```
Thomas-Avery:
```suggestion
/// Configure an error converter to convert errors in calls to [`uniffi::custom_type!`] into the main
```

Thomas-Avery (dismissed):
Changes look good from the KM perspective.

Thomas-Avery (dismissed):
Approving for KM

dani-garcia:
Added support for the latest bitwarden-encoding crate created in #404

Thomas-Avery (approved):

addisonbeck (approved):

PR #375, opened 2025-08-01
[PM-23653] Add support for ssh keys

Hinton:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-23653

## 📔 Objective

Adds support for importing ssh keys using CXF.

## ⏰ Reminders before review

- CI builds passed
  team

## 🦮 Reviewer guidelines


crates/bitwarden-exporters/src/cxf/ssh.rs
```diff
@@ -0,0 +1,99 @@
...
+    let encoded_key = import_der_key(&encoded_key)?;
+
+    let ssh = SshKey {
+        private_key: encoded_key.private_key,
+        public_key: encoded_key.public_key,
+        fingerprint: encoded_key.fingerprint,
+    };
+
+    let fields = [
+        credential.key_comment.as_ref().map(|comment| Field {
```
quexten, line 22: Note (non-actionable): In the agent, we treat the name of the cipher as the comment, and advertise it as such. We may want to research (not here) how other clients treat comment vs name, to not disrupt user experience. If someone comes over from another product which uses a separate key comment field, suddenly all their ciphers may be named something different.

quexten (dismissed):
After the removal of `expect` this looks good to me. One note, but it's not actionable here.

crates/bitwarden-exporters/src/cxf/ssh.rs
Hinton: I think we will have a lot of follow up work as other credential managers implements credential exchange and we discover edge cases 😰

crates/bitwarden-exporters/src/cxf/import.rs
```diff
@@ -126,6 +127,16 @@ pub(super) fn parse_item(value: Item) -> Vec<ImportingCipher> {
         add_item(CipherType::Identity(Box::new(identity)), custom_fields);
     });
 
+    // SSH Key credentials
+    if let Some(ssh) = grouped.ssh.first() {
+        match to_ssh(ssh) {
+            Ok((ssh_key, fields)) => add_item(CipherType::SshKey(Box::new(ssh_key)), fields),
+            Err(_) => {
+                // Include information about the failed items, or import as note?
+            }
```
abergs, line 136: Yeah we need to figure out a strategy to handle errors in a more structured, actionable way

abergs: We don't need to change the approach in this PR, but I do think we should opt for the "don't loose any data" approach and save it to a note, if we can encode the value in a meaningful way.

crates/bitwarden-ssh/src/import.rs
```diff
@@ -32,6 +32,12 @@ pub fn import_key(
     }
 }
 
+/// Import a DER encoded private key, and returns a decoded [SshKeyView]. This is primarily used for
+/// importing SSH keys from other Credential Managers through Credential Exchange.
+pub fn import_pkcs8_der(encoded_key: &[u8]) -> Result<SshKeyView, SshKeyImportError> {
+    import_der_key(encoded_key)
+}
+
```
abergs, line 40: question: Should these `import_pkcs8_der`and `import_der_key` live in ssh.rs ?

abergs (approved):
Just smaller notes. Is there any update we can/should do to the cxf_sample tests?

crates/bitwarden-ssh/src/import.rs
Hinton: No, we support importing keys outside cxf so we can't move it into the exporters crate.

PR #376, opened 2025-08-05
[PM-24051] MasterPasswordUnlockData model with response mapping and adds it to identity success response model

mzieniukbw:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-24051

## 📔 Objective

Adds `MasterPasswordUnlockData` model with `MasterPasswordUnlockResponseModel` mapping. This wi
Adds `UserDecryptionOptionsResponseModel` into the identity success token response. This model includes the master password unlock field.
The `/sync` response model is autogenerated by OpenApi and already includes user decryption option and the master password unlock fields.
Includes Uniffi and WASM bindings for `MasterPasswordUnlock`.

In further PR's we plan to use the `MasterPasswordUnlock` in `InitUserCryptoMethod` enum - this is still in discussion. 

## ⏰ Reminders before review

- CI builds passed
  team

## 🦮 Reviewer guidelines


crates/bitwarden-core/src/key_management/master_password.rs
```diff
@@ -0,0 +1,353 @@
...
+    tsify(into_wasm_abi, from_wasm_abi)
+)]
+pub struct MasterPasswordUnlockData {
+    pub kdf: Kdf,
+    pub master_key_wrapped_user_key: EncString,
+    pub salt: String,
+}
+
+impl MasterPasswordUnlockData {
+    pub fn process_response(
```
quexten: Could we do `impl TryFrom<MasterPasswordUnlockResponseModel> for MasterPasswordUnlockData` instead?

crates/bitwarden-core/src/key_management/master_password.rs
```diff
@@ -0,0 +1,353 @@
...
+#[cfg_attr(feature = "uniffi", derive(uniffi::Record))]
+#[cfg_attr(
+    feature = "wasm",
+    derive(tsify::Tsify),
+    tsify(into_wasm_abi, from_wasm_abi)
+)]
+pub struct MasterPasswordUnlockData {
+    pub kdf: Kdf,
+    pub master_key_wrapped_user_key: EncString,
+    pub salt: String,
```
quexten: We have an opaque string type `MasterPasswordSalt` in `clients`: https://github.com/bitwarden/clients/blob/61cd0c4f51d9fe390c93f61f2f55076de33b57cf/libs/common/src/key-management/master-password/types/master-password.types.ts#L15. 

Ideally we would expose this from rust as we do for EncString https://github.com/bitwarden/sdk-internal/blob/481349272195b330c52610f64b4b7f3559b0e25b/crates/bitwarden-crypto/src/enc_string/symmetric.rs#L18 and replace the TS version with the SDK version so that we have type safety across the FFI boundary.

This one is optional, if we don't do it here it is tech debt we should record.

crates/bitwarden-core/src/key_management/master_password.rs
```diff
@@ -0,0 +1,353 @@
+#![allow(missing_docs)]
```
quexten: Please remove this and add comments. We require comments on everything that is publicly exposed.

crates/bitwarden-core/src/key_management/master_password.rs
```diff
@@ -0,0 +1,353 @@
...
+            EncString::from_str(master_key_encrypted_user_key.as_str())
+                .map_err(|e: CryptoError| MasterPasswordError::from(e))?;
+
+        let salt = require!(response.salt);
+
+        Ok(MasterPasswordUnlockData {
+            kdf,
+            master_key_wrapped_user_key,
+            salt,
+        })
```
quexten:
```suggestion
         let master_key_encrypted_user_key = require!(response.master_key_encrypted_user_key);

        let salt = require!(response.salt);

        Ok(MasterPasswordUnlockData {
            kdf,
            master_key_wrapped_user_key: master_key_encrypted_user_key.as_str().parse()?,
            salt,
        })
```

crates/bitwarden-core/src/key_management/user_decryption.rs
```diff
@@ -0,0 +1,142 @@
...
+    feature = "wasm",
+    derive(tsify::Tsify),
+    tsify(into_wasm_abi, from_wasm_abi)
+)]
+pub struct UserDecryptionData {
+    pub master_password_unlock: Option<MasterPasswordUnlockData>,
+}
+
+impl UserDecryptionData {
+    pub fn process_response(
```
quexten: Same here with `TryFrom`

crates/bitwarden-core/src/key_management/user_decryption.rs
```diff
@@ -0,0 +1,142 @@
+#![allow(missing_docs)]
```
quexten: Same comment here about missing docs

crates/bitwarden-wasm-internal/src/key_management/master_password.rs
```diff
@@ -0,0 +1,160 @@
+#![allow(missing_docs)]
```
quexten: Same comment here about missing docs

crates/bitwarden-wasm-internal/src/key_management/master_password.rs
```diff
@@ -0,0 +1,160 @@
...
+use serde::{Deserialize, Serialize};
+use tsify::Tsify;
+use wasm_bindgen::prelude::*;
+
+// WASM-compatible wrapper for the auto-generated MasterPasswordUnlockKdfResponseModel
+#[derive(Clone, Debug, PartialEq, Serialize, Deserialize, Tsify)]
+#[tsify(into_wasm_abi, from_wasm_abi)]
+#[serde(rename_all = "camelCase")]
+pub struct MasterPasswordUnlockKdfResponseModel {
+    #[serde(alias = "KdfType")]
```
quexten: Non-actionable note: We should probably look into automating this to add the macro to the auto-generated models...

crates/bitwarden-wasm-internal/src/key_management/master_password.rs
```diff
@@ -0,0 +1,160 @@
...
+            }),
+            master_key_encrypted_user_key: wasm_model.master_key_encrypted_user_key,
+            salt: wasm_model.salt,
+        }
+    }
+}
+
+/// WASM-exposed function to process a MasterPasswordUnlockResponse
+#[wasm_bindgen]
+pub fn process_master_password_unlock_response(
```
quexten: I'm not sure if we are ok with exporting this globally without any kind of namespacing. It feels like this should either live on a client, or some construct like `PureCrypto`. 

Question: Is mobile intended to use this for parsing requests too, or are they expected to do the parsing on mobile? If the answer is that they should use it too, then this should be moved and uniffi bindings added.

crates/bitwarden-core/src/key_management/master_password.rs
```diff
@@ -0,0 +1,353 @@
...
+#[serde(rename_all = "camelCase", deny_unknown_fields)]
+#[cfg_attr(feature = "uniffi", derive(uniffi::Record))]
+#[cfg_attr(
+    feature = "wasm",
+    derive(tsify::Tsify),
+    tsify(into_wasm_abi, from_wasm_abi)
+)]
+pub struct MasterPasswordUnlockData {
+    pub kdf: Kdf,
+    pub master_key_wrapped_user_key: EncString,
```
quexten: Do these fields need to be `pub`?

crates/bitwarden-core/src/key_management/user_decryption.rs
```diff
@@ -0,0 +1,142 @@
...
+impl UserDecryptionData {
+    pub fn process_response(
+        response: UserDecryptionResponseModel,
+    ) -> Result<UserDecryptionData, UserDecryptionError> {
+        let master_password_unlock = match response.master_password_unlock {
+            Some(master_password_unlock) => Some(MasterPasswordUnlockData::process_response(
+                master_password_unlock.as_ref().clone(),
+            )?),
+            None => None,
+        };
```
quexten:
```suggestion
        let master_password_unlock = response
            .master_password_unlock
            .map(|mpu| MasterPasswordUnlockData::process_response(*mpu))
            .transpose()?;
```
(or the try from equivalent)

crates/bitwarden-core/src/key_management/user_decryption.rs
```diff
@@ -0,0 +1,142 @@
...
+#[derive(Serialize, Deserialize, Debug)]
+#[serde(rename_all = "camelCase", deny_unknown_fields)]
+#[cfg_attr(feature = "uniffi", derive(uniffi::Record))]
+#[cfg_attr(
+    feature = "wasm",
+    derive(tsify::Tsify),
+    tsify(into_wasm_abi, from_wasm_abi)
+)]
+pub struct UserDecryptionData {
+    pub master_password_unlock: Option<MasterPasswordUnlockData>,
```
quexten: Does this field need to be `pub`?

crates/bitwarden-core/src/key_management/master_password.rs
```diff
@@ -0,0 +1,353 @@
...
+    pub fn process_response(
+        response: MasterPasswordUnlockResponseModel,
+    ) -> Result<MasterPasswordUnlockData, MasterPasswordError> {
+        let kdf = match response.kdf.kdf_type {
+            KdfType::PBKDF2_SHA256 => Kdf::PBKDF2 {
+                iterations: NonZeroU32::new(response.kdf.iterations as u32)
+                    .ok_or(MissingFieldError(stringify!(response.kdf.iterations)))?,
+            },
+            KdfType::Argon2id => Kdf::Argon2id {
+                iterations: NonZeroU32::new(response.kdf.iterations as u32)
```
quexten: Optional: A private helper function along the lines of:
```
fn parse_nonzero_u32(value: impl TryInto<u32>, field_name: &'static str) -> Result<NonZeroU32, MissingFieldError> {
    let num: u32 = value.try_into().map_err(|_| MissingFieldError(field_name))?;
    NonZeroU32::new(num).ok_or(MissingFieldError(field_name))
}
```

would make this more readable.

quexten (changes requested):
Some initial comments, some of these are required, some are idiomatic (but probably required since the quality bar for SDK seems to be set much higher).

I'd like a review from someone with more rust experience such as @Hinton or @dani-garcia too. I believe specifically the request parsing is something that has not existed in the SDK so far, and there may be a better way to do this that we're unaware of.

crates/bitwarden-core/src/key_management/master_password.rs
mzieniukbw: Fixed in 70ad9d37

crates/bitwarden-core/src/key_management/master_password.rs
mzieniukbw: Fixed in 70ad9d37

crates/bitwarden-core/src/key_management/user_decryption.rs
mzieniukbw: Fixed in 70ad9d37

crates/bitwarden-wasm-internal/src/key_management/master_password.rs
mzieniukbw: Fixed in 70ad9d37.
Wrapped it around `UserDecryption` struct

crates/bitwarden-core/src/key_management/user_decryption.rs
mzieniukbw: Fixed in 70ad9d37

crates/bitwarden-core/src/key_management/master_password.rs
mzieniukbw: Fixed in 70ad9d37

crates/bitwarden-core/src/key_management/user_decryption.rs
mzieniukbw: We are eventually going to use those fields, outside of WASM. 
How about i make them `pub(crate)` ? Which we need for the unit tests anyway.
Or your thinking is that we will assess what we need, once we start using them ?

crates/bitwarden-core/src/key_management/master_password.rs
mzieniukbw: From what i understand, in cases `EncString`, we have an equivalent type in Rust SDK, that is different during Serialization: `EncString` becomed `String` type, hence why we export it in TS as string.

This is different here, since we don't have equivalent in our Rust code.
If we want one, then i think either of the two will be ok:
```
#[cfg_attr(feature = "wasm", tsify::declare(tsify::Tsify))]
pub type MasterPasswordSalt = String;
```
Or change the `salt` field to have different type in TS.
```
    #[cfg_attr(
        feature = "wasm",
        tsify(type = "MasterPasswordSalt")
    )]
    pub salt: String,
```

crates/bitwarden-core/src/key_management/user_decryption.rs
mzieniukbw: Fixed in 3bf880f8

crates/bitwarden-wasm-internal/src/key_management/master_password.rs
mzieniukbw: No longer required with 70ad9d37

crates/bitwarden-core/src/key_management/master_password.rs
quexten: I think my main concern here is that we want the tagged type, so that the typesafety stretches to TS.

A bit more verbose than I'd like, but this does the job, and is the same approach I use for the new PinProtectedKeyEnvelope (PasswordProtectedkeyEnvelope):
```
#[derive(Serialize, Deserialize)]
#[serde(transparent)]
pub struct MasterPasswordSalt(
    #[cfg_attr(
        feature = "wasm",
        tsify(type = r#"Tagged<string, "MasterPasswordSalt">"#)
    )]
   String,
);
```

But again, we can do that later if we don't want to do in in this PR, I don't want to scope creep it.

crates/bitwarden-wasm-internal/src/key_management/master_password.rs
quexten: I think you did that now, so closing.

crates/bitwarden-core/src/key_management/user_decryption.rs
quexten: If they are intended to be used from the clients, they can be pub, apologies. Resolving.

crates/bitwarden-core/src/key_management/master_password.rs
quexten: I assume these are intended for use outside of the SDK too, so they can be pub.

crates/bitwarden-core/src/key_management/user_decryption.rs
quexten: I think you accidentally resolved the wrong comment, this one still has no comments

crates/bitwarden-core/src/key_management/master_password.rs
```diff
@@ -0,0 +1,380 @@
...
+/// Error for master password related operations.
+#[allow(missing_docs)]
+#[bitwarden_error(flat)]
+#[derive(Debug, thiserror::Error)]
+pub enum MasterPasswordError {
+    #[error(transparent)]
+    Crypto(#[from] CryptoError),
+    #[error(transparent)]
+    MissingField(#[from] MissingFieldError),
+}
```
quexten:
```suggestion
#[bitwarden_error(flat)]
#[derive(Debug, thiserror::Error)]
pub enum MasterPasswordError {
    /// The wrapped encryption key could not be parsed
    #[error(transparent)]
    Crypto(#[from] CryptoError),
    /// The KDF data is malformed
    #[error(transparent)]
    MissingField(#[from] MissingFieldError),
}
```

I would actually suggest making these errors more descriptive. If I understand the rest of the file correctly, missingfield happens exactly when the KDF is broken, and Crypto happens exactly when the encstring parsing fails, so you could just skip wrapping the errors and make it more clear to the caller what actually failed:

```suggestion
#[bitwarden_error(flat)]
#[derive(Debug, thiserror::Error)]
pub enum MasterPasswordError {
    /// The wrapped encryption key could not be parsed because the encstring is malformed
    EncryptionKeyMalformed,
    /// The KDF data could not be parsed, because it is missing values or has an invalid value
    KdfMalformed
}
```

crates/bitwarden-core/src/key_management/master_password.rs
```diff
@@ -0,0 +1,380 @@
...
+    pub kdf: Kdf,
+    pub master_key_wrapped_user_key: EncString,
+    pub salt: String,
+}
+
+impl TryFrom<MasterPasswordUnlockResponseModel> for MasterPasswordUnlockData {
+    type Error = MasterPasswordError;
+
+    fn try_from(response: MasterPasswordUnlockResponseModel) -> Result<Self, Self::Error> {
+        let kdf = match response.kdf.kdf_type {
```
quexten, line 45: If you do apply the suggestion from above, then this would require mapping to the different error types

crates/bitwarden-core/src/key_management/user_decryption.rs
```diff
@@ -0,0 +1,149 @@
...
+use bitwarden_api_api::models::UserDecryptionResponseModel;
+use bitwarden_error::bitwarden_error;
+use serde::{Deserialize, Serialize};
+#[cfg(feature = "wasm")]
+use wasm_bindgen::prelude::*;
+
+use crate::key_management::master_password::{MasterPasswordError, MasterPasswordUnlockData};
+
+/// Error for master user decryption related operations.
+#[allow(missing_docs)]
```
quexten:
```suggestion
```

crates/bitwarden-core/src/key_management/user_decryption.rs
```diff
@@ -0,0 +1,149 @@
...
+#[bitwarden_error(flat)]
+#[derive(Debug, thiserror::Error)]
+pub enum UserDecryptionError {
+    #[error(transparent)]
+    MasterPasswordError(#[from] MasterPasswordError),
+}
+
+/// Represents data required to decrypt user's vault.
+/// Currently, this is only used for master password unlock.
+#[allow(missing_docs)]
```
quexten:
```suggestion
```

crates/bitwarden-wasm-internal/src/key_management/user_decryption.rs
```diff
@@ -0,0 +1,24 @@
+use bitwarden_api_api::models::{MasterPasswordUnlockResponseModel, UserDecryptionResponseModel};
+use bitwarden_core::key_management::{
+    master_password::{MasterPasswordError, MasterPasswordUnlockData},
+    user_decryption::{UserDecryptionData, UserDecryptionError},
+};
+use wasm_bindgen::prelude::wasm_bindgen;
+
+#[wasm_bindgen]
```
quexten:
```suggestion
/// This is a stop-gap solution to allow parsing of response models via the SDK while the full decryption functionality is not yet moved over to the SDK
#[wasm_bindgen]
```

quexten (commented):
Looking pretty nice now! Some more smaller items that improve usability by teams unfamiliar with the code (docs / error names)

Hinton:
This tightly couples consumers of the SDK to the server generated bindings which is somewhat dangerous. We've previously addressed this by wrapping requests/responses in a manner that more accurately represents the action from the SDK's perspective. There be dragons here and we should consider this a bit.

Do we expect to do similar things to other request/responses? If not I think we should simply wrap the logic with our own struct, while it's a few lines of boilerplate it reduces the complexity.

>  Includes Uniffi and WASM support.

There is no UniFFI support here.

crates/bitwarden-wasm-internal/src/key_management/user_decryption.rs
Hinton, line 1: **question:** Why are we implementing logic in wasm-internal? This would generally belong to our feature crates.

support/openapi-template/Cargo.mustache
```diff
@@ -90,3 +110,5 @@ mockall = ["dep:mockall"]
 bon = ["dep:bon"]
 {{/useBonBuilder}}
 {{/reqwestTrait}}
+tsify = { workspace = true, optional = true, features = ["js"], default-features = false }
+wasm-bindgen = { workspace = true, optional = true, features = ["serde-serialize"] }
```
Hinton: We'd need some way to configure this in runtime, since this would now also apply to identity.

crates/bitwarden-api-api/src/models/master_password_unlock_response_model.rs
```diff
@@ -9,24 +9,31 @@
...
+    feature = "wasm",
+    derive(tsify::Tsify),
+    tsify(into_wasm_abi, from_wasm_abi)
+)]
 pub struct MasterPasswordUnlockResponseModel {
     #[serde(rename = "kdf")]
     pub kdf: Box<models::MasterPasswordUnlockKdfResponseModel>,
     #[serde(rename = "masterKeyEncryptedUserKey")]
-    pub master_key_encrypted_user_key: Option<String>,
+    pub master_key_encrypted_user_key: String,
```
Hinton: Platform hasn't made any strides into making fields required as it has compatibility implications. This will throw an error should these fields not be available in the past, present or future. I.e. can we guarantee forever that this endpoint in the past and future has and always will include `master_key_encrypted_user_key` and `salt`? And if these are missing are we fine with it throwing?

This is related to https://bitwarden.atlassian.net/wiki/spaces/DEV/pages/760971277/Improve+API+bindings.

mzieniukbw: Reverted with [3e6a46a](https://github.com/bitwarden/sdk-internal/pull/376/commits/3e6a46a79d9d3651445156c57feb2d57455623f4)

support/openapi-template/Cargo.mustache
mzieniukbw: Reverted with [3e6a46a](https://github.com/bitwarden/sdk-internal/pull/376/commits/3e6a46a79d9d3651445156c57feb2d57455623f4)

crates/bitwarden-wasm-internal/src/key_management/user_decryption.rs
mzieniukbw: Reverted with [3e6a46a](https://github.com/bitwarden/sdk-internal/pull/376/commits/3e6a46a79d9d3651445156c57feb2d57455623f4)

crates/bitwarden-wasm-internal/src/key_management/user_decryption.rs
mzieniukbw: Reverted with [3e6a46a](https://github.com/bitwarden/sdk-internal/pull/376/commits/3e6a46a79d9d3651445156c57feb2d57455623f4)

crates/bitwarden-core/src/key_management/master_password.rs
mzieniukbw: Reverted the WASM bindings with [3e6a46a](https://github.com/bitwarden/sdk-internal/pull/376/commits/3e6a46a79d9d3651445156c57feb2d57455623f4).
Do we still want this struct ?

quexten: If this is referring to masterpasswordsalt, we can just move that out of this task, I don't want to scope creep.

crates/bitwarden-core/src/key_management/master_password.rs
mzieniukbw: Fixed with [9676069](https://github.com/bitwarden/sdk-internal/pull/376/commits/9676069424d62c6257e546f7fbf6d6f7445579dc)

crates/bitwarden-core/src/key_management/user_decryption.rs
mzieniukbw: Fixed with [9676069](https://github.com/bitwarden/sdk-internal/pull/376/commits/9676069424d62c6257e546f7fbf6d6f7445579dc)

crates/bitwarden-core/src/auth/api/response/identity_success_response.rs
```diff
@@ -1,5 +1,6 @@
 use std::{collections::HashMap, num::NonZeroU32};
 
+use bitwarden_api_api::models::UserDecryptionResponseModel;
```
Hinton: **question (blocking):** Is this actually `UserDecryptionResponseModel` or  is it supposed to be `UserDecryptionOptions`? Looking at the server they are not the same, and `UserDecryptionResponseModel` is only used for sync.

crates/bitwarden-core/src/key_management/master_password.rs
```diff
@@ -0,0 +1,409 @@
...
+                        .kdf
+                        .memory
+                        .ok_or(MasterPasswordError::KdfMalformed)?,
+                )?,
+                parallelism: kdf_parse_nonzero_u32(
+                    response
+                        .kdf
+                        .parallelism
+                        .ok_or(MasterPasswordError::KdfMalformed)?,
+                )?,
```
Hinton: **question:** Is there a reason for not just using `require!`?
```suggestion
                memory: kdf_parse_nonzero_u32(require!(response.kdf.memory))?,
                parallelism: kdf_parse_nonzero_u32(require!(response.kdf.parallelism))?,
```

crates/bitwarden-core/src/key_management/master_password.rs
```diff
@@ -0,0 +1,409 @@
...
+            },
+        };
+
+        let master_key_encrypted_user_key = require!(response.master_key_encrypted_user_key);
+        let salt = require!(response.salt);
+
+        Ok(MasterPasswordUnlockData {
+            kdf,
+            master_key_wrapped_user_key: master_key_encrypted_user_key
+                .as_str()
```
Hinton: **issue:** Parse is implemented for String ...
```suggestion
```

crates/bitwarden-core/src/key_management/master_password.rs
```diff
@@ -0,0 +1,409 @@
...
+        })
+    }
+}
+
+fn kdf_parse_nonzero_u32(value: impl TryInto<u32>) -> Result<NonZeroU32, MasterPasswordError> {
+    let num: u32 = value
+        .try_into()
+        .map_err(|_| MasterPasswordError::KdfMalformed)?;
+    NonZeroU32::new(num).ok_or(MasterPasswordError::KdfMalformed)
+}
```
Hinton: **suggestion:** You can rewrite this in a more functional manner.
```suggestion
fn 32(value: i32) -> Result<NonZeroU32, MasterPasswordError> {
    value
        .try_into()
        .ok()
        .and_then(NonZeroU32::new)
        .ok_or(MasterPasswordError::KdfMalformed)
}
```

crates/bitwarden-core/src/key_management/mod.rs
```diff
@@ -18,8 +18,13 @@ mod crypto_client;
 #[cfg(feature = "internal")]
 pub use crypto_client::CryptoClient;
 
+#[cfg(feature = "internal")]
+pub mod master_password;
```
Hinton: **suggestion:** It's generally a good idea to not make deep modules public. It leaks the modules to the consumers and makes renaming the module a breaking action.

```suggestion
#[cfg(feature = "internal")]
mod master_password;
#[cfg(feature = "internal")]
pub use master_password:*;
```

This also lets you remove the mostly pointless comments in the modules.

crates/bitwarden-core/src/key_management/user_decryption.rs
```diff
@@ -0,0 +1,139 @@
...
+        };
+
+        let result = UserDecryptionData::try_from(response);
+        assert!(matches!(
+            result,
+            Err(UserDecryptionError::MasterPasswordError(
+                MasterPasswordError::MissingField(_)
+            ))
+        ));
+    }
```
Hinton: **issue:** This is re-testing the `TryFrom<MasterPasswordUnlockResponseModel> for MasterPasswordUnlockData` implementation from `master_password.rs`. We don't need two layers of tests for the same functionality as it only makes it harder to ever change anything.

If you are worried about test coverage that will be addressed when the logic is actually consumed somewhere.

crates/bitwarden-core/src/key_management/master_password.rs
```diff
@@ -0,0 +1,409 @@
...
+        let result = MasterPasswordUnlockData::try_from(response);
+        assert!(result.is_ok());
+        let data = result.unwrap();
+
+        match data.kdf {
+            Kdf::PBKDF2 { iterations } => {
+                assert_eq!(iterations.get(), 600_000);
+            }
+            _ => panic!("Expected PBKDF2 KDF"),
+        }
```
Hinton: **issue:** use `let if` when you only care about one arm of a match. https://contributing.bitwarden.com/contributing/code-style/rust#if-let

```suggestion
        if let Kdf::PBKDF2 { iterations } = data.kdf {
            assert_eq!(iterations.get(), 600_000);
        } else {
            panic!("Expected PBKDF2 KDF");
        }
```

crates/bitwarden-core/src/key_management/master_password.rs
```diff
@@ -0,0 +1,409 @@
...
+            }
+            _ => panic!("KDF types don't match"),
+        }
+
+        assert_eq!(
+            data.master_key_wrapped_user_key.to_string(),
+            deserialized.master_key_wrapped_user_key.to_string()
+        );
+        assert_eq!(data.salt, deserialized.salt);
+    }
```
Hinton: **issue:** Testing that serde serializes and deserializes the same is fairly pointless as serde has extensive tests for this.

**suggestion:** Remove the tests, and/or replace them with just deserializing from a fixed string which would confirm the fields are properly named.

crates/bitwarden-core/src/key_management/master_password.rs
```diff
@@ -0,0 +1,409 @@
...
+                iterations: 600_000,
+                memory: None,
+                parallelism: None,
+            }),
+            master_key_encrypted_user_key: Some(TEST_USER_KEY.to_string()),
+            salt: Some(TEST_SALT.to_string()),
+        };
+
+        let result = MasterPasswordUnlockData::try_from(response);
+        assert!(result.is_ok());
```
Hinton: **suggestion:** You can remove this since you're calling `unwrap` immediately afterwards which will panic if it's not successful.

crates/bitwarden-core/src/key_management/master_password.rs
```diff
@@ -0,0 +1,409 @@
...
+
+    #[test]
+    fn test_try_from_master_password_unlock_response_model_pbkdf2_success() {
+        let response = MasterPasswordUnlockResponseModel {
+            kdf: Box::new(MasterPasswordUnlockKdfResponseModel {
+                kdf_type: KdfType::PBKDF2_SHA256,
+                iterations: 600_000,
+                memory: None,
+                parallelism: None,
+            }),
```
Hinton: **observation:** You could make some helper functions for building the responses as it's fairly verbose.

```rust
create_pbkdf2_response(TEST_USER_KEY, TEST_SALT, 600_000)
```

Hinton (changes requested):
**issue:** The PR title mentions sync response, there is nothing touching sync here.

We should be careful about introducing public objects that are not consumed. Once something is public it's technically a breaking change to flip the visibility and if you don't know fully what needs to be public it's easy to accidentally leave public remnants that should be private.

In these case just making items private and going back later and flipping the visibility where necessary can be useful. Which can also assist with documentation since documentation should be written from the consumers standpoint which is hard when you don't have the interfaces making it public.

crates/bitwarden-core/src/key_management/master_password.rs
mzieniukbw: @quexten pointed out, that it might be better to consolidate all KDF related verification into the `KdfMalformed` error. Using `require!` would provide non-KDF specific error.

crates/bitwarden-core/src/key_management/master_password.rs
mzieniukbw: Fixed in 5a4f3140

crates/bitwarden-core/src/key_management/master_password.rs
mzieniukbw: Fixed in 5a4f3140

crates/bitwarden-core/src/key_management/master_password.rs
mzieniukbw: Removed those tests in 5a4f3140

crates/bitwarden-core/src/key_management/master_password.rs
mzieniukbw: Fixed in 5a4f3140

crates/bitwarden-core/src/key_management/user_decryption.rs
mzieniukbw: Removed in 5a4f3140

crates/bitwarden-core/src/key_management/mod.rs
mzieniukbw: Fixed in 5a4f3140

crates/bitwarden-core/src/key_management/master_password.rs
mzieniukbw: Fixed in 5a4f3140

crates/bitwarden-core/src/key_management/master_password.rs
mzieniukbw: Fixed in 5a4f3140

crates/bitwarden-core/src/auth/api/response/identity_success_response.rs
mzieniukbw: You are right, this two are separate models. Fixed in 7973f12a

mzieniukbw: To avoid any confusion, i renamed  the new response to `IdentityUserDecryptionOptionsResponseModel` with 7e425ca5 and added mapping to `UserDecryptionData` in c9aaa8cb

quexten (dismissed):
Looks good from my side.

crates/bitwarden-core/src/key_management/master_password.rs
Hinton: It would, but it's the error you would get if the server side bindings were more accurate. (Using an enum or https://swagger.io/docs/specification/v3_0/data-models/oneof-anyof-allof-not/#oneof on the serverside)

mzieniukbw: Changed to `require!` in 77c2d68

quexten (dismissed):

crates/bitwarden-core/src/auth/api/response/identity_user_decryption_options_response.rs
```diff
@@ -0,0 +1,11 @@
+use bitwarden_api_api::models::MasterPasswordUnlockResponseModel;
+use serde::{Deserialize, Serialize};
+
+#[derive(Serialize, Deserialize, Debug, PartialEq)]
+pub struct IdentityUserDecryptionOptionsResponseModel {
```
Hinton: **question:** This is called `UserDecryptionOptions` in server, and contains several fields that are missing here.

I think it's helpful to document that this is `UserDecryptionOptions` if we don't want to use that name. And we should document if this is only partially implemented.

crates/bitwarden-core/src/key_management/user_decryption.rs
```diff
@@ -0,0 +1,56 @@
...
+        let master_password_unlock = response
+            .master_password_unlock
+            .map(MasterPasswordUnlockData::try_from)
+            .transpose()?;
+
+        Ok(UserDecryptionData {
+            master_password_unlock,
+        })
+    }
+}
```
Hinton: **suggestion:** Use `MasterPasswordError` directly instead of wrapping it? This allows you to get rid of a unnecessary error type.

```suggestion
/// Represents data required to decrypt user's vault.
/// Currently, this is only used for master password unlock.
#[derive(Serialize, Deserialize, Debug)]
#[serde(rename_all = "camelCase", deny_unknown_fields)]
pub struct UserDecryptionData {
    /// Optional master password unlock data.
    pub master_password_unlock: Option<MasterPasswordUnlockData>,
}

impl TryFrom<UserDecryptionResponseModel> for UserDecryptionData {
    type Error = MasterPasswordError;

    fn try_from(response: UserDecryptionResponseModel) -> Result<Self, Self::Error> {
        let master_password_unlock = response
            .master_password_unlock
            .map(|response| MasterPasswordUnlockData::try_from(*response))
            .transpose()?;

        Ok(UserDecryptionData {
            master_password_unlock,
        })
    }
}

impl TryFrom<IdentityUserDecryptionOptionsResponseModel> for UserDecryptionData {
    type Error = MasterPasswordError;

    fn try_from(response: IdentityUserDecryptionOptionsResponseModel) -> Result<Self, Self::Error> {
        let master_password_unlock = response
            .master_password_unlock
            .map(MasterPasswordUnlockData::try_from)
            .transpose()?;

        Ok(UserDecryptionData {
            master_password_unlock,
        })
    }
}
```

crates/bitwarden-core/src/key_management/master_password.rs
```diff
@@ -0,0 +1,316 @@
...
+
+/// Represents the data required to unlock with the master password.
+#[derive(Serialize, Deserialize, Debug)]
+#[serde(rename_all = "camelCase", deny_unknown_fields)]
+#[cfg_attr(feature = "uniffi", derive(uniffi::Record))]
+#[cfg_attr(
+    feature = "wasm",
+    derive(tsify::Tsify),
+    tsify(into_wasm_abi, from_wasm_abi)
+)]
```
Hinton: **question:**. We're planning on exposing `MasterPasswordUnlockData` to wasm and mobile?

crates/bitwarden-core/src/key_management/user_decryption.rs
```diff
@@ -0,0 +1,56 @@
...
+
+/// Represents data required to decrypt user's vault.
+/// Currently, this is only used for master password unlock.
+#[derive(Serialize, Deserialize, Debug)]
+#[serde(rename_all = "camelCase", deny_unknown_fields)]
+pub struct UserDecryptionData {
+    /// Optional master password unlock data.
+    pub master_password_unlock: Option<MasterPasswordUnlockData>,
+}
```
Hinton: **question:**. We're **not** planning on exposing `UserDecryptionData` to wasm and mobile?

Hinton (commented):
The previous top level comments still apply.

- The PR description mentions sync, nothing here touches sync. 
- You *MUST* go back after the next PR and confirm everything public is used since rust will not warn about unused public items. (Since it's public the consumer can be another crate)

Hinton:
Related to this PR. The auth API models being `pub` annoyed me so I've opened https://github.com/bitwarden/sdk-internal/pull/403 which marks them as internal to within the `auth` namespace.

crates/bitwarden-core/src/auth/api/response/identity_user_decryption_options_response.rs
mzieniukbw: Changed it back to what it's called in server in ad732ce6.
Added documentation in 5e6f1dd0

crates/bitwarden-core/src/key_management/user_decryption.rs
mzieniukbw: Fixed in d11f6535

crates/bitwarden-core/src/key_management/master_password.rs
mzieniukbw: We don't have to right now. Removed in 0e104bba

crates/bitwarden-core/src/key_management/user_decryption.rs
mzieniukbw: We don't have to right now.

mzieniukbw:
@Hinton reduced the visibility with 70dee0c6

crates/bitwarden-core/src/auth/mod.rs
```diff
@@ -9,6 +9,8 @@ use crate::{NotAuthenticatedError, VaultLockedError, WrongPasswordError};
 mod access_token;
 // API is intentionally not visible outside of `auth` as these should be considered private.
 mod api;
+#[allow(unused_imports)]
+pub(crate) use api::response::user_decryption_options_response;
```
Hinton: **question:** Why are we importing and not using the request here? if it's because the code is unused, you can use `#[allow(dead_code)]`

quexten (dismissed):

crates/bitwarden-core/src/auth/mod.rs
mzieniukbw: Fixed in 6861006e

quexten (dismissed):

crates/bitwarden-core/src/key_management/mod.rs
```diff
@@ -19,13 +19,17 @@ mod crypto_client;
 #[cfg(feature = "internal")]
 pub use crypto_client::CryptoClient;
 
+#[cfg_attr(feature = "internal", allow(dead_code))]
```
Hinton: **issue:** This now conditionally applies `allow(dead_code)` for feature `internal`. I don't believe that's the intention?

**suggestion:** Apply `allow(dead_code)` on the types inside this file. That also makes it clear which types are used and which aren't.

**question:** I removed the `allow(dead_code)` and clippy seems fine?

crates/bitwarden-core/src/auth/mod.rs
```diff
@@ -9,6 +9,7 @@ use crate::{NotAuthenticatedError, VaultLockedError, WrongPasswordError};
 mod access_token;
 // API is intentionally not visible outside of `auth` as these should be considered private.
 mod api;
+pub(crate) use api::response::user_decryption_options_response;
```
Hinton: **nit:** In this case I would expose the response model directly as the `UserDecryptionOptionsResponseModel` is the public interface and probably not the module `user_decryption_options_response`?

```suggestion
pub(crate) use api::response::user_decryption_options_response::UserDecryptionOptionsResponseModel;
```

mzieniukbw: Fixed in 6b98aaf0

crates/bitwarden-core/src/key_management/mod.rs
mzieniukbw: Fixed in 6b98aaf0

Hinton (approved):

quexten (approved):

justindbaur (approved):

PR #378, opened 2025-08-11
Update wasm to use the same method as mobile for feature flags

Hinton:
## 📔 Objective

Updates wasm to use the same implementation as mobile, i.e. `load_flags`. This uses a hashmap instead of the internal `Flags` enum as we want to prevent additions or removal of flags to be a breaking change.

Consumers should pass in the full flag object they get from the config endpoint and let the SDK handle the rest.

## ⏰ Reminders before review

- CI builds passed
  team

## 🦮 Reviewer guidelines

### DIFF
--- a/crates/bitwarden-wasm-internal/src/platform/mod.rs
+++ b/crates/bitwarden-wasm-internal/src/platform/mod.rs
@@ -1,6 +1,6 @@
-use bitwarden_core::{Client, Flags};
+use bitwarden_core::Client;
 use bitwarden_vault::{Cipher, Folder};
-use wasm_bindgen::prelude::*;
+use wasm_bindgen::prelude::wasm_bindgen;
 
 mod repository;
 pub mod token_provider;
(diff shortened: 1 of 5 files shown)


shane-melton (dismissed):

shane-melton:
One thing to note is cipher key encryption feature flag from the config endpoint is of the form `cipher-key-encryption` while the SDK uses `enableCipherKeyEncryption`. We'll need to transform that somewhere, at least for that flag. Android currently does something similar [here.](https://github.com/bitwarden/android/blob/9bfbe0c08724b08f6225777df5d7ace23a5d4ac3/app/src/main/kotlin/com/x8bit/bitwarden/data/platform/manager/FeatureFlagManagerImpl.kt#L21-L24)

Hinton:
@shane-melton that should be addressed now. I added a serde alias for the existing value and added default handling to use `kebab-case`.

With this change we should be able to pass in the flags directly from the config endpoint.

shane-melton (dismissed):

crates/bitwarden-wasm-internal/src/platform/mod.rs
```diff
@@ -21,8 +32,9 @@ impl PlatformClient {
     }
 
     /// Load feature flags into the client
-    pub fn load_flags(&self, flags: Flags) -> Result<(), JsValue> {
-        self.0.internal.set_flags(&flags);
+    #[wasm_bindgen(js_name = load_flags)]
```
dani-garcia: Do we need to set the name? this one should be the default right?

dani-garcia (dismissed):

dani-garcia (approved):

PR #383, opened 2025-08-13
[PM-24683] Add updateKdf function

quexten:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-24683

## 📔 Objective

Exposes functionality to update the KDF, with the new masterpassword unlock data, and masterpassword authentication data models. These can be directly passed to the server models.

## ⏰ Reminders before review

- CI builds passed
  team

## 🦮 Reviewer guidelines


crates/bitwarden-vault/src/collection_client.rs
```diff
@@ -3,6 +3,7 @@ use bitwarden_collections::{
     tree::{NodeItem, Tree},
 };
 use bitwarden_core::Client;
+#[cfg(feature = "wasm")]
```
quexten: Note to reviewers, this is tracking a change in `main` that is not yet in the target PR: https://github.com/bitwarden/sdk-internal/pull/385

crates/bitwarden-exporters/Cargo.toml
```diff
@@ -19,13 +19,14 @@ keywords.workspace = true
...
     "bitwarden-vault/wasm",
+    "bitwarden-collections/wasm",
     "dep:tsify",
     "dep:wasm-bindgen"
 ] # WebAssembly bindings
 
 [dependencies]
 base64 = ">=0.22.1, <0.23"
-bitwarden-collections = { workspace = true, features = ["wasm"] }
+bitwarden-collections = { workspace = true }
```
quexten: Note to reviewers, this is tracking a change in `main` that is not yet in the target PR: https://github.com/bitwarden/sdk-internal/pull/385

crates/bitwarden-crypto/src/keys/master_key.rs
```diff
@@ -62,10 +62,12 @@ impl MasterKey {
     }
 
     /// Derive the master key hash, used for local and remote password validation.
-    pub fn derive_master_key_hash(&self, password: &[u8], purpose: HashPurpose) -> Result<String> {
-        let hash = util::pbkdf2(self.inner_bytes().as_slice(), password, purpose as u32);
-
-        Ok(STANDARD.encode(hash))
+    pub fn derive_master_key_hash(&self, password: &[u8], purpose: HashPurpose) -> String {
```
quexten: Note to reviewers: This saved an expect / error type for an error case that never occurs

crates/bitwarden-core/src/key_management/crypto.rs
```diff
@@ -265,6 +270,63 @@ pub(super) async fn get_user_encryption_key(client: &Client) -> Result<String, C
...
+    #[allow(deprecated)]
+    let user_key = ctx.dangerous_get_symmetric_key(SymmetricKeyId::User)?;
+
+    let login_method = client
+        .internal
+        .get_login_method()
+        .ok_or(NotAuthenticatedError)?;
+    let email = match login_method.as_ref() {
+        LoginMethod::User(UserLoginMethod::Username { email, .. })
+        | LoginMethod::User(UserLoginMethod::ApiKey { email, .. }) => email,
```
dani-garcia: Small non-blocking nit: You can match both userloginmethods in the same branch:
```suggestion
        LoginMethod::User(
            UserLoginMethod::Username { email, .. } | UserLoginMethod::ApiKey { email, .. },
        ) => email,
```

crates/bitwarden-core/src/key_management/crypto.rs
```diff
@@ -265,6 +270,63 @@ pub(super) async fn get_user_encryption_key(client: &Client) -> Result<String, C
...
+#[derive(Serialize, Deserialize, Debug)]
+#[serde(rename_all = "camelCase", deny_unknown_fields)]
+#[cfg_attr(feature = "uniffi", derive(uniffi::Record))]
+#[cfg_attr(feature = "wasm", derive(Tsify), tsify(into_wasm_abi, from_wasm_abi))]
+pub struct UpdateKdfResponse {
+    /// The authentication data for the new KDF setting
+    master_password_authentication_data: MasterPasswordAuthenticationData,
+    /// The unlock data for the new KDF setting
+    master_password_unlock_data: MasterPasswordUnlockData,
+    /// The authentication data for the prior to the KDF change
```
dani-garcia: I feel like this comment needs to be reworded
```suggestion
    /// The authentication data for the KDF setting prior to the change
```

crates/bitwarden-core/src/key_management/crypto.rs
```diff
@@ -265,6 +270,63 @@ pub(super) async fn get_user_encryption_key(client: &Client) -> Result<String, C
...
+        .get_login_method()
+        .ok_or(NotAuthenticatedError)?;
+    let email = match login_method.as_ref() {
+        LoginMethod::User(UserLoginMethod::Username { email, .. })
+        | LoginMethod::User(UserLoginMethod::ApiKey { email, .. }) => email,
+        #[cfg(feature = "secrets")]
+        LoginMethod::ServiceAccount(_) => return Err(NotAuthenticatedError)?,
+    };
+
+    let authentication_data = MasterPasswordAuthenticationData::derive(password, new_kdf, email)
```
dani-garcia, line 328: These derive functions will only fail if the KDF settings are insufficient right? If that's the case we should rename the error variant to make it more clear, because the error is unrelated to the master password.

crates/bitwarden-core/src/key_management/master_password.rs
```diff
@@ -44,6 +50,27 @@ pub struct MasterPasswordUnlockData {
...
+    pub(crate) fn derive(
+        password: &str,
+        kdf: &Kdf,
+        salt: &str,
+        user_key: &SymmetricCryptoKey,
+    ) -> Result<Self, MasterPasswordError> {
+        let master_key = MasterKey::derive(password, salt, kdf)
+            .map_err(|_| MasterPasswordError::InvalidKdfConfiguration)?;
+        let master_key_wrapped_user_key = master_key
+            .encrypt_user_key(user_key)
```
dani-garcia, line 65: Just a future improvement idea, but I can't think of any cases where `encrypt_user_key` would return an error in real life, if we could remove the Result from that function we won't need the `Crypto` variant here either.

dani-garcia (commented):
Looks pretty good to me, just some comments about error handling.

As a future improvement, I've noticed that a lot of the `derive` functions return errors only in cases of invalid KDF parameters. I wonder if we could improve this by making these functions take an already validated KDF params object and doing the validation only once. We probably still need the bare `Kdf` enum for FFI, but we could have a valid wrapper.
```rust
pub struct ValidKdf(Kdf);
impl TryFrom<Kdf> for ValidKdf {
 ...
}
```

crates/bitwarden-core/src/key_management/crypto_client.rs
```diff
@@ -80,6 +80,16 @@ impl CryptoClient {
...
+
+    /// Update the user's kdf settings, which will re-encrypt the user's encryption key with the new
+    /// kdf settings. This returns the new encrypted user key and the new password hash.
+    pub fn update_kdf(
+        &self,
+        password: String,
+        kdf: Kdf,
+    ) -> Result<UpdateKdfResponse, CryptoClientError> {
+        update_kdf(&self.client, &password, &kdf)
+    }
```
mzieniukbw: The name is a bit confusing, since it does not update (state), but make's the update kdf (response).
```suggestion
    /// Update the user's kdf settings, which will re-encrypt the user's encryption key with the new
    /// kdf settings. This returns the new encrypted user key and the new password hash.
    pub fn make_update_kdf(
        &self,
        password: String,
        kdf: Kdf,
    ) -> Result<UpdateKdfResponse, CryptoClientError> {
        make_update_kdf(&self.client, &password, &kdf)
    }
```

quexten: Hmm, I think your suggestion definitely is the better naming. However, this would break convention with what's existing in the file where we use update_password which also does not actually update state. In my mental list of priorities, consistency within a file ranks higher than (without context) better naming. 

So, how do we want to proceed:
- Go with make_update_kdf , but at the same time add make_update_password that just routes through to update_password, with a note / ticket for mobile to update to the new function, after which we will remove update_password
- Just go with update_kdf and accept the naming is somewhat confusing, but gain concistency
- Anything else?

mzieniukbw: I would vote for `make_update_kdf` and `make_update_password`

crates/bitwarden-uniffi/src/crypto.rs
```diff
@@ -84,14 +91,14 @@ impl CryptoClient {
 
     /// Update the user's kdf settings, which will re-encrypt the user's encryption key with the new
```
mzieniukbw: Do we need to update the docs too ?

crates/bitwarden-uniffi/src/crypto.rs
```diff
@@ -45,9 +45,16 @@ impl CryptoClient {
     /// Update the user's password, which will re-encrypt the user's encryption key with the new
     /// password. This returns the new encrypted user key and the new password hash.
     pub fn update_password(&self, new_password: String) -> Result<UpdatePasswordResponse> {
+        self.make_update_password(new_password)
+    }
+
+    /// Update the user's password, which will re-encrypt the user's encryption key with the new
```
mzieniukbw: Do we need to update the docs too ?

crates/bitwarden-core/src/key_management/crypto_client.rs
```diff
@@ -101,11 +101,23 @@ impl CryptoClient {
...
+    /// Note: This function is deprecated and will be replaced by `make_update_password`
     pub fn update_password(
         &self,
         new_password: String,
     ) -> Result<UpdatePasswordResponse, CryptoClientError> {
-        update_password(&self.client, new_password)
+        self.make_update_password(new_password)
+    }
+
+    /// Update the user's password, which will re-encrypt the user's encryption key with the new
```
mzieniukbw: Do we need to update the docs too ?

crates/bitwarden-core/src/key_management/crypto_client.rs
```diff
@@ -101,11 +101,23 @@ impl CryptoClient {
 
     /// Update the user's password, which will re-encrypt the user's encryption key with the new
```
mzieniukbw: Do we need to update the docs too ?

crates/bitwarden-core/src/key_management/crypto_client.rs
```diff
@@ -83,12 +83,12 @@ impl CryptoClient {
 
     /// Update the user's kdf settings, which will re-encrypt the user's encryption key with the new
```
mzieniukbw: Do we need to update the docs too ?

crates/bitwarden-vault/src/collection_client.rs
mzieniukbw: Merged `main` in base branch, this is no longer needed

crates/bitwarden-exporters/Cargo.toml
mzieniukbw: Merged `main` in base branch, this is no longer needed

crates/bitwarden-uniffi/src/crypto.rs
quexten: Updated

crates/bitwarden-uniffi/src/crypto.rs
quexten: Updated

crates/bitwarden-core/src/key_management/crypto_client.rs
quexten: Updated

crates/bitwarden-core/src/key_management/crypto_client.rs
quexten: Updated

crates/bitwarden-core/src/key_management/crypto_client.rs
quexten: Updated

crates/bitwarden-core/src/key_management/crypto.rs
quexten: That makes sense. The variant is now just "InvalidKdfSettings", and discards the returned `MasterPasswordError`

crates/bitwarden-core/src/key_management/master_password.rs
quexten: Leaving this up for future improvement, but I agree. Looking at the internal impl's the reason is that `stretch_key` returns result, because `hkdf_expand` returns a Result, because the inputs to `hkdf_expand` could have invalid sizes. We should probably just `expect` in `stretch_key`, because there we know the input sizes to `hkdf_expand` are safe and won't error.

quexten:
@dani-garcia Yeah, the valid kdf newtype makes sense. We enforce the same correct values on e.g. keys too with custom types that can only be validly created, so there is no reason not to do that with KDF.

Not adding it to this PR, but I'd definitely also like to see this as a future improvement.

crates/bitwarden-core/src/key_management/crypto_client.rs
```diff
@@ -99,13 +100,28 @@ impl CryptoClient {
         get_user_encryption_key(&self.client).await
     }
 
-    /// Update the user's password, which will re-encrypt the user's encryption key with the new
-    /// password. This returns the new encrypted user key and the new password hash.
+    /// Create the data necessary to update the user's password. The user's encryption key is
+    /// re-encrypted with the new
+    /// password. This returns the new encrypted user key and the new password hash but does not
```
mzieniukbw: line breaks seems to be off

crates/bitwarden-core/src/key_management/crypto_client.rs
```diff
@@ -99,13 +100,28 @@ impl CryptoClient {
...
         &self,
         new_password: String,
     ) -> Result<UpdatePasswordResponse, CryptoClientError> {
-        update_password(&self.client, new_password)
+        self.make_update_password(new_password)
+    }
+
+    /// Create the data necessary to update the user's password. The user's encryption key is
+    /// re-encrypted with the new
+    /// password. This returns the new encrypted user key and the new password hash but does not
```
mzieniukbw: line breaks seems to be off

crates/bitwarden-uniffi/src/crypto.rs
```diff
@@ -42,12 +42,24 @@ impl CryptoClient {
             .map_err(Error::MobileCrypto)?)
     }
 
-    /// Update the user's password, which will re-encrypt the user's encryption key with the new
-    /// password. This returns the new encrypted user key and the new password hash.
+    /// Create the data necessary to update the user's password. The user's encryption key is
+    /// re-encrypted with the new
+    /// password. This returns the new encrypted user key and the new password hash but does not
```
mzieniukbw: line breaks seems to be off

crates/bitwarden-uniffi/src/crypto.rs
```diff
@@ -42,12 +42,24 @@ impl CryptoClient {
...
+    /// update sdk state.
+    ///
+    /// Note: This is deprecated and `make_update_password` should be used instead
     pub fn update_password(&self, new_password: String) -> Result<UpdatePasswordResponse> {
+        self.make_update_password(new_password)
+    }
+
+    /// Create the data necessary to update the user's password. The user's encryption key is
+    /// re-encrypted with the new
+    /// password. This returns the new encrypted user key and the new password hash but does not
```
mzieniukbw: line breaks seems to be off

mzieniukbw (changes requested):

crates/bitwarden-core/src/key_management/crypto_client.rs
quexten: Whoops. I had assumed `cargo fmt` would take care of this, but it seems it only re-wraps too-long lines, not too short ones.

mzieniukbw (dismissed):

dani-garcia (dismissed):

crates/bitwarden-exporters/Cargo.toml
```diff
@@ -20,6 +20,7 @@ uniffi = ["dep:uniffi", "bitwarden-core/uniffi"] # Uniffi bindings
 wasm = [
     "bitwarden-collections/wasm",
     "bitwarden-vault/wasm",
+    "bitwarden-collections/wasm",
```
Hinton: Duplicative.

crates/bitwarden-uniffi/src/crypto.rs
```diff
@@ -42,12 +42,22 @@ impl CryptoClient {
             .map_err(Error::MobileCrypto)?)
     }
 
-    /// Update the user's password, which will re-encrypt the user's encryption key with the new
-    /// password. This returns the new encrypted user key and the new password hash.
+    /// Create the data necessary to update the user's password. The user's encryption key is
+    /// re-encrypted with the new password. This returns the new encrypted user key and the new
+    /// password hash but does not update sdk state.
+    ///
+    /// Note: This is deprecated and `make_update_password` should be used instead
```
Hinton, line 50: **question:** Why not mark this as deprecated?

crates/bitwarden-uniffi/src/crypto.rs
```diff
@@ -81,4 +91,18 @@ impl CryptoClient {
...
     }
+
+    /// Create the data necessary to update the user's kdf settings. The user's encryption key is
+    /// re-encrypted for the password under the new kdf settings. This returns the new encrypted
+    /// user key and the new password hash but does not update sdk state.
+    pub fn make_update_kdf(
+        &self,
+        password: String,
+        kdf: Kdf,
+    ) -> Result<bitwarden_core::key_management::crypto::UpdateKdfResponse> {
```
Hinton: **nit:** The other functions here uses the shorthand path.

```suggestion
    ) -> Result<UpdateKdfResponse> {
```

crates/bitwarden-core/src/key_management/crypto_client.rs
```diff
@@ -80,6 +80,17 @@ impl CryptoClient {
     ) -> Result<UserCryptoV2KeysResponse, StatefulCryptoError> {
         get_v2_rotated_account_keys(&self.client)
     }
+
+    /// Create the data necessary to update the user's kdf settings. The user's encryption key is
+    /// re-encrypted for the password under the new kdf settings. This returns the new encrypted
+    /// user key and the new password hash but does not update sdk state.
```
Hinton:
```suggestion
    /// re-encrypted for the password under the new kdf settings. This returns the re-encrypted
    /// user key and the new password hash but does not update sdk state.
```

crates/bitwarden-core/src/key_management/crypto_client.rs
```diff
@@ -89,13 +100,26 @@ impl CryptoClient {
         get_user_encryption_key(&self.client).await
     }
 
-    /// Update the user's password, which will re-encrypt the user's encryption key with the new
-    /// password. This returns the new encrypted user key and the new password hash.
+    /// Create the data necessary to update the user's password. The user's encryption key is
+    /// re-encrypted with the new password. This returns the new encrypted user key and the new
+    /// password hash but does not update sdk state.
+    ///
+    /// Note: This is deprecated and `make_update_password` should be used instead
```
Hinton: **question:** Can we just remove this function here? Or is it actively used in web?

crates/bitwarden-exporters/Cargo.toml
quexten: Left over from the upstream changes that were needed to unblock the PR. I've undone the duplicate import now.

crates/bitwarden-uniffi/src/crypto.rs
quexten: This does not seem to play nice with uniffi / clippy, and ends up failing the CI because a warning gets emitted. I tried `#[allow(deprecated)]` on it, but it did not seem to help.

crates/bitwarden-core/src/key_management/crypto_client.rs
quexten: Removed!

crates/bitwarden-uniffi/src/crypto.rs
quexten: Cleaned up!

crates/bitwarden-core/src/auth/api/response/identity_user_decryption_options_response.rs
```diff
@@ -0,0 +1,11 @@
+use bitwarden_api_api::models::MasterPasswordUnlockResponseModel;
+use serde::{Deserialize, Serialize};
+
+#[derive(Serialize, Deserialize, Debug, PartialEq)]
+pub struct IdentityUserDecryptionOptionsResponseModel {
```
dani-garcia: Question: Why not use `#[serde(rename_all = "camelCase")]` rather than renaming the field?

crates/bitwarden-core/src/platform/get_user_api_key.rs
```diff
@@ -78,7 +78,7 @@ fn build_secret_verification_request(
         let master_password_hash = input
             .master_password
             .as_ref()
-            .map(|p| {
+            .map(|p| -> Result<String, CryptoError> {
```
dani-garcia: Is this need here? We're already typing `Ok::<String, CryptoError>(` below, seems like the compiler should be able to infer it.

dani-garcia (dismissed):
LGTM, just two small non-blocking nits

crates/bitwarden-uniffi/src/crypto.rs
```diff
@@ -43,12 +43,22 @@ impl CryptoClient {
             .map_err(Error::MobileCrypto)?)
     }
 
-    /// Update the user's password, which will re-encrypt the user's encryption key with the new
-    /// password. This returns the new encrypted user key and the new password hash.
+    /// Create the data necessary to update the user's password. The user's encryption key is
+    /// re-encrypted with the new password. This returns the new encrypted user key and the new
+    /// password hash but does not update sdk state.
+    ///
+    /// Note: This is deprecated and `make_update_password` should be used instead
```
Hinton, line 50: Do. we have a task for tracking the removal? And ideally a task to assign to mobile to clean up their implementation?

crates/bitwarden-core/src/platform/get_user_api_key.rs
Hinton: If we keep it we should remove the type hint for `Ok`.

crates/bitwarden-core/src/auth/api/response/identity_user_decryption_options_response.rs
Hinton: **question:** Can we use `pub(crate)` here? Most API request/responses were updated to only be create public.


```suggestion
pub(crate) struct IdentityUserDecryptionOptionsResponseModel {
```

crates/bitwarden-core/src/auth/api/response/mod.rs
```diff
@@ -4,6 +4,7 @@ mod identity_success_response;
 mod identity_token_fail_response;
 mod identity_token_response;
 mod identity_two_factor_response;
+pub(crate) mod identity_user_decryption_options_response;
```
Hinton: **question:** Could we just make this mod and use `use` below?

Hinton (commented):
A couple of minor things and we should be good. Only the two API comments are blocking.

crates/bitwarden-uniffi/src/crypto.rs
jlf0dev: Created the following Tech Debt tasks:
2. https://bitwarden.atlassian.net/browse/PM-25651
3. https://bitwarden.atlassian.net/browse/PM-25652

crates/bitwarden-core/src/auth/api/response/identity_user_decryption_options_response.rs
jlf0dev: This response model was actually leftover from a previous upstream PR. Removed since its not needed.

crates/bitwarden-core/src/platform/get_user_api_key.rs
jlf0dev: I removed this and didn't get any compiler issues so I'm assuming its able to infer.

crates/bitwarden-core/src/auth/api/response/identity_user_decryption_options_response.rs
jlf0dev: This response model was actually leftover from a previous upstream PR. Removed since its not needed.

crates/bitwarden-core/src/auth/api/response/mod.rs
jlf0dev: This response model was actually leftover from a previous upstream PR. Removed since its not needed.

crates/bitwarden-core/src/key_management/master_password.rs
```diff
@@ -74,6 +109,43 @@ fn kdf_parse_nonzero_u32(value: impl TryInto<u32>) -> Result<NonZeroU32, MasterP
         .ok_or(MasterPasswordError::KdfMalformed)
 }
 
+/// Represents the data required to unlock with the master password.
```
mzieniukbw:
```suggestion
/// Represents the data required to authenticate with the master password.
```

crates/bitwarden-core/src/key_management/master_password.rs
```diff
@@ -3,39 +3,74 @@ use std::num::NonZeroU32;
...
+        user_key: &SymmetricCryptoKey,
+    ) -> Result<Self, MasterPasswordError> {
+        let master_key = MasterKey::derive(password, salt, kdf)
+            .map_err(|_| MasterPasswordError::InvalidKdfConfiguration)?;
+        let master_key_wrapped_user_key = master_key
+            .encrypt_user_key(user_key)
+            .map_err(MasterPasswordError::Crypto)?;
+
+        Ok(Self {
+            kdf: kdf.clone(),
```
mzieniukbw: (nit) Since it's a reference, we should use `to_owned()` instead of `clone()`
```suggestion
            kdf: kdf.to_owned(),
```

crates/bitwarden-core/src/key_management/master_password.rs
```diff
@@ -74,6 +109,43 @@ fn kdf_parse_nonzero_u32(value: impl TryInto<u32>) -> Result<NonZeroU32, MasterP
...
+    ) -> Result<Self, MasterPasswordError> {
+        let master_key = MasterKey::derive(password, salt, kdf)
+            .map_err(|_| MasterPasswordError::InvalidKdfConfiguration)?;
+        let hash = master_key.derive_master_key_hash(
+            password.as_bytes(),
+            bitwarden_crypto::HashPurpose::ServerAuthorization,
+        );
+
+        Ok(Self {
+            kdf: kdf.clone(),
```
mzieniukbw: (nit) Since it's a reference, we should use `to_owned()` instead of `clone()`
```suggestion
            kdf: kdf.to_owned(),
```

mzieniukbw (changes requested):

mzieniukbw (approved):

Hinton (approved):

dani-garcia (approved):

PR #384, opened 2025-08-14
[Vault] Cipher Versioning POC 

gbubemismith:
## ⏰ Reminders before review

- CI builds passed
  team

## 🦮 Reviewer guidelines

### DIFF
--- a/crates/bitwarden-api-api/src/models/cipher_request_model.rs
+++ b/crates/bitwarden-api-api/src/models/cipher_request_model.rs
@@ -51,6 +51,12 @@ pub struct CipherRequestModel {
     pub secure_note: Option<Box<models::CipherSecureNoteModel>>,
     #[serde(rename = "sshKey", skip_serializing_if = "Option::is_none")]
     pub ssh_key: Option<Box<models::CipherSshKeyModel>>,
+    /// The version of the cipher data. Default is 1 for backward compatibility.
+    #[serde(rename = "version", skip_serializing_if = "Option::is_none")]
+    pub version: Option<i32>,
+    /// Opaque data for versioned clients.
+    #[serde(rename = "data", skip_serializing_if = "Option::is_none")]
+    pub data: Option<String>,
     #[serde(
         rename = "lastKnownRevisionDate",
         skip_serializing_if = "Option::is_none"
@@ -79,6 +85,8 @@ impl CipherRequestModel {
             identity: None,
             secure_note: None,
             ssh_key: None,
+            version: None,
+            data: None,
             last_known_revision_date: None,
         }
     }
--- a/crates/bitwarden-vault/src/cipher/migrations/registry.rs
+++ b/crates/bitwarden-vault/src/cipher/migrations/registry.rs
@@ -1,0 +1,64 @@
+use bitwarden_core::key_management::{KeyIds, SymmetricKeyId};
+use bitwarden_crypto::KeyStoreContext;
+
+use crate::{
+    migrations::versions::{V1ToV2Migration, V2ToV3Migration},
+    CipherError,
+};
+
+pub trait Migration {
+    fn from_version(&self) -> u32;
+    fn to_version(&self) -> u32;
+    fn migrate(
+        &self,
+        cipher_data: &mut serde_json::Value,
+        ctx: Option<&mut KeyStoreContext<KeyIds>>,
+        cipher_key: Option<SymmetricKeyId>,
+    ) -> Result<(), CipherError>; // this can be migration error
+}
+
+pub struct MigrationRegistry {
+    migrations: Vec<Box<dyn Migration>>,
+}
+
+impl MigrationRegistry {
+    pub fn new() -> Self {
+        let mut registry = Self {
+            migrations: Vec::new(),
+        };
+
+        // something like this
+        registry.register(Box::new(V1ToV2Migration));
+        registry.register(Box::new(V2ToV3Migration));
+
+        registry
+    }
+
+    pub fn register(&mut self, migration: Box<dyn Migration>) {
+        self.migrations.push(migration);
+    }
+
+    pub fn migrate(
+        &self,
+        cipher_data: &mut serde_json::Value,
+        from_version: u32,
+        to_version: u32,
+        mut ctx: Option<&mut KeyStoreContext<KeyIds>>,
+        cipher_key: Option<SymmetricKeyId>,
+    ) -> Result<(), CipherError> {
+        let mut current_version = from_version;
+
+        while current_version < to_version {
+            let migration = self
+                .migrations
+                .iter()
+                .find(|m| m.from_version() == current_version)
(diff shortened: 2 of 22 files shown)


crates/bitwarden-vault/src/cipher/migrations/registry.rs
```diff
@@ -0,0 +1,64 @@
...
+        let mut current_version = source_version;
+
+        while current_version < target_version {
+            let migration = self
+                .migrations
+                .iter()
+                .find(|m| m.source_version() == current_version)
+                .ok_or(CipherError::UnsupportedCipherVersion(current_version))?;
+
+            migration.migrate(cipher_data, ctx.as_deref_mut(), cipher_key)?;
```
nick-livefront, line 58: This is hard for me to conceptualize when or why this would happen, but I'll throw it out there. 

What would be the experience if a migration were to fail? I would imagine that there would be some ciphers that have migrated and some that have not. It's hard to say why this would happen and maybe we don't know that it ever would.

gbubemismith: Yeah, good question. I thought about this, but didn't capture it in this POC. We might have migration failures in cases where we need to decrypt or encrypt and then encounter a failure. The way I'll address this is to create a back up before migration and then if we encounter an error we restore the backup on failure

nick-livefront: 👍 Good idea on the backup! That probably is the most straight forward approach. Being in a partial state could spin up any number of edge cases.

crates/bitwarden-vault/src/cipher/migrations/versions/v2_to_v3.rs
```diff
@@ -0,0 +1,118 @@
...
+            for fido2_credential in fido2_credentials {
+                if let Some(credential_id_str) = fido2_credential
+                    .get("credentialId")
+                    .and_then(|v| v.as_str())
+                {
+                    let enc_string: EncString = credential_id_str.parse()?;
+                    let dec_credential_id: String = enc_string.decrypt(ctx, ciphers_key)?;
+                    let b64_credential_id = BASE64_STANDARD.encode(&dec_credential_id);
+                    let enc_credential_id: EncString =
+                        b64_credential_id.encrypt(ctx, ciphers_key)?;
```
coroiu, line 43: issue: I think this migration does a good job as an example of how one would modify a field, but the what it actually does (since this is a real example) is wrong. Probably don't need to fix it in the PoC, but this should not be merged to production as-is.

**What it does:**
It takes the UUID as a string directly and encodes the Unicode bytes as B64 i.e. `UUID -> B64`.
E.g. the UUID standard format `0000 ...` will be converted to the unicode bytes `0x30 0x30 0x30 0x30 ...` and saved as B64

**What it should do:**
Take the UUID standard format and convert it into raw format e.g. `0000 ...` means `0x00 0x00` and then save that as B64. There are already existing examples and functions for this so you don't need to write your own :)

gbubemismith: Oh nice, Thanks for pointing that out

crates/bitwarden-api-api/src/models/cipher_request_model.rs
```diff
@@ -51,6 +51,12 @@ pub struct CipherRequestModel {
     pub secure_note: Option<Box<models::CipherSecureNoteModel>>,
     #[serde(rename = "sshKey", skip_serializing_if = "Option::is_none")]
     pub ssh_key: Option<Box<models::CipherSshKeyModel>>,
+    /// The version of the cipher data. Default is 1 for backward compatibility.
+    #[serde(rename = "version", skip_serializing_if = "Option::is_none")]
+    pub version: Option<i32>,
```
nikwithak, line 56: 🤔 Is there any important distinction between `version: Some(1)` and `version: None`? If not then you could use `serde(default = )`to default it to 1 and ditch the `Option`.

This would mean writing `version: 1` on save - would that matter for older clients, or do they already safely ignore unknown fields?

crates/bitwarden-vault/src/cipher/migrations/versions/v1_to_v2.rs
```diff
@@ -0,0 +1,54 @@
...
+            if !obj.contains_key("SecurityQuestions") {
+                obj.insert(
+                    "SecurityQuestions".to_string(),
+                    serde_json::Value::Array(vec![]),
+                );
+            }
+        }
+
+        Ok(())
+    }
```
nikwithak, line 33: 💭 I'm still digging through the feature docs elsewhere, but here are some brainstorming / initial thoughts:

Where is the spec defined for each version - should we maintain a spec for each version inside the SDK code?

From this code it looks like each version will have a structured schema that we want to be able to enforce / verify - Rust's typing system is really good for this type of migration. I would suggest it defining each version as its own structured data type.

A few benefits of this would be:

* make the code a little bit safer & error-resistant for invalid or mispelled keys.
* Give us a place in the SDK code where the version schemas are strictly defined, making interfacing with the data elsewhere in the SDK easier and more maintainable.
* We could potentially create an `enum CipherData { .... }`, and consolidate some split-versioning logic to that struct.
    * Perhaps adding getters for consistent data access between versions.
* We could skip the need for these isolated `Migration` structs, in favor of something like `impl From<CipherDataV1> for CipherDataV2 {...}` (or `TryFrom`, if we anticipate the possibility of errors - ref: @nick-livefront's [comment above](https://github.com/bitwarden/sdk-internal/pull/384/files#r2277666176))

If we have a need for unstructured elements in the version as well, we could use `#[serde(flatten)]` to create a catch-all for unmatched values, as shown in [this example](https://github.com/serde-rs/serde/issues/941#issuecomment-375486935)

crates/bitwarden-vault/src/cipher/migrations/versions/v1_to_v2.rs
```diff
@@ -0,0 +1,54 @@
...
+            "Uris": [],
+            "Totp": null
+        });
+
+        V1ToV2Migration.migrate(&mut data, None, None).unwrap();
+
+        assert!(data.get("SecurityQuestions").is_some());
+        assert_eq!(data["SecurityQuestions"], serde_json::json!([]));
+    }
+}
```
nikwithak, line 54: ⛏️ Would it be benficial to store some `.json` files with some actual test-data to verify the before/after for various cipher types?

crates/bitwarden-api-api/src/models/cipher_request_model.rs
nikwithak, line 56: Since the `version` field indicates the version of the data itself, would it make sense to store this as part of the `data` object itself?

gbubemismith: Yeah, that's what we have decided to go with. I have that changed stashed locally

PR #386, opened 2025-08-14
Added not wasm config for get_ancestors

jrmccannon:
## 🎟️ Tracking
[PM-25167](https://bitwarden.atlassian.net/browse/PM-25167)
## 📔 Objective
Adds conditional code for when wasm feature isn't enabled for get_ancestors.

## ⏰ Reminders before review

- CI builds passed
  team

## 🦮 Reviewer guidelines

[PM-25167]: https://bitwarden.atlassian.net/browse/PM-25167?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


crates/bitwarden-vault/src/collection_client.rs
```diff
@@ -80,6 +83,15 @@ impl CollectionViewNodeItem {
                 map
             })
     }
+
+    #[cfg(not(feature = "wasm"))]
+    pub fn get_ancestors(&self) -> HashMap<String, String> {
```
Hinton: **suggestion:** This is a public function and having different return types depending on features will shift the burden of handling this to consumers which is generally not advised.

If any rust code calls `get_ancestors` they now need to use the wasm flag themselves and ensure they handle both Map and HashMap since you can get both return types. (This is because flags are enabled globally)

I would advice to name them differently. Another alternative is to make a `NewType` which implements tsify and contains the ancestors, similar to how i handle flags in https://github.com/bitwarden/sdk-internal/pull/378/files#diff-0824a57e7a6afdb1d2c5991d91eb46376e1d9ecee456f7e503ed69bc025adb16R13. That would allow a single function to work in both wasm and non wasm environments.

jrmccannon: Like this?
https://github.com/bitwarden/sdk-internal/pull/386/commits/4cb672d614ca8fb1223b787076d42c54841ba6a4

crates/bitwarden-vault/src/collection_client.rs
```diff
@@ -1,10 +1,16 @@
...
     collection::{Collection, CollectionView},
     tree::{NodeItem, Tree},
 };
 use bitwarden_core::Client;
+use serde::{Deserialize, Serialize};
 #[cfg(feature = "wasm")]
-use bitwarden_error::js_sys::Map;
+use tsify::Tsify;
+#[cfg(feature = "uniffi")]
+use uniffi;
```
Hinton: **issue:** You shouldn't need to use `uniffi` here, since you use the full path below.

```suggestion
```

crates/bitwarden-vault/src/collection_client.rs
```diff
@@ -56,6 +62,13 @@ pub struct CollectionViewNodeItem {
     node_item: NodeItem<CollectionView>,
 }
 
+#[cfg_attr(feature = "wasm", derive(Tsify, Serialize, Deserialize))]
+#[cfg_attr(feature = "wasm", tsify(into_wasm_abi, from_wasm_abi))]
```
Hinton: We usually combine ´cfg_attr` with the same condition into a single.

```suggestion
#[cfg_attr(
    feature = "wasm",
    derive(Tsify, Serialize, Deserialize),
    tsify(into_wasm_abi, from_wasm_abi)
)]
```

crates/bitwarden-vault/src/collection_client.rs
```diff
@@ -56,6 +62,13 @@ pub struct CollectionViewNodeItem {
...
+#[cfg_attr(feature = "uniffi", derive(uniffi::Record))]
+pub struct AncestorMap {
+    pub ancestors: HashMap<Uuid, String>,
```
Hinton: **question:** Should this be `Uuid` or `CollectionId`?

jrmccannon: CollectionId in this case

dani-garcia (dismissed):

Hinton (dismissed):

crates/bitwarden-uniffi/src/vault/collections.rs
```diff
@@ -40,25 +40,34 @@ pub struct CollectionViewTree {
...
 #[uniffi::export]
 impl CollectionViewTree {
-    pub fn get_item_by_id(&self, collection_id: Uuid) -> Option<Arc<CollectionViewNodeItem>> {
+    pub fn get_item_by_id(
+        &self,
+        collection_view: CollectionView,
+    ) -> Option<Arc<CollectionViewNodeItem>> {
         self.tree
-            .get_item_by_id(collection_id)
+            .get_item_by_id(collection_view.id.unwrap_or_default().into())
```
jaasen-livefront, line 54: ❓ Are we ok with swallowing an empty or invalid `id` here? Is that the goal of this?

crates/bitwarden-uniffi/src/vault/collections.rs
```diff
@@ -40,25 +40,34 @@ pub struct CollectionViewTree {
...
     node_item: NodeItem<CollectionView>,
 }
 
 #[uniffi::export]
 impl CollectionViewTree {
-    pub fn get_item_by_id(&self, collection_id: Uuid) -> Option<Arc<CollectionViewNodeItem>> {
+    pub fn get_item_by_id(
+        &self,
+        collection_view: CollectionView,
+    ) -> Option<Arc<CollectionViewNodeItem>> {
```
jaasen-livefront: 🎨 now that we're passing in `collection_view` does the method name still make sense here? perhaps `get_item_for_view` or `find_node_for_collection`?

jrmccannon: good call :+1:

crates/bitwarden-uniffi/src/vault/collections.rs
Hinton: Long term I think we would split `CollectionView` similar to folders so that ID is required, and you have a separate DTO for creating them.

I wouldn't be particularly concerned about ID being `None` here.

dani-garcia (approved):

Hinton (approved):

jaasen-livefront (approved):

PR #399, opened 2025-08-19
[PM-24051] Init crypto and update kdf with MasterPasswordUnlock

mzieniukbw:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-24051

## 📔 Objective

- Initialise user crypto with `MasterPasswordUnlock` - added `initialize_user_crypto_master_password_unlock` function into `InternalClient`'s crypto.
- Auth's login methods (password, and api_key specifically) use `MasterPasswordUnlockData` when present in response and uses new function.
  - KM team have decided that the double KDF derivation issue is something we are ok with for login.
  - Moved `set_login_method` to be the after the `initialize_user_crypto` function in execution order, since the latter overrides the `set_login_method`. Move everywhere (all login methods) for consistency.
  - Removed `kdf` as init arg of `password.rs`, consistently use `prelogin` now.
- Update user's KDF settings  with `MasterPasswordUnlock` via `update_user_master_password_unlock` in `InternalClient`, triggered in sync.

Continuation of #376 

## ⏰ Reminders before review

- CI builds passed
  team

## 🦮 Reviewer guidelines


crates/bitwarden-core/src/auth/login/auth_request.rs
```diff
@@ -88,20 +88,11 @@ pub(crate) async fn complete_auth_request(
...
         );
-        client
-            .internal
-            .set_login_method(LoginMethod::User(UserLoginMethod::Username {
-                client_id: "web".to_owned(),
-                email: auth_req.email.to_owned(),
-                kdf: kdf.clone(),
-            }));
 
         let method = match res.master_password_hash {
```
quexten, line 99: We should probably just drop this code. Support for this auth method on the approving side has been removed 3 months ago: https://github.com/bitwarden/clients/pull/14236

(not this PR)

crates/bitwarden-vault/src/sync.rs
```diff
@@ -39,6 +45,17 @@ pub(crate) async fn sync(client: &Client, input: &SyncRequest) -> Result<SyncRes
...
 
+    if let Some(master_password_unlock_response) = sync
+        .user_decryption
+        .as_deref()
+        .and_then(|d| d.master_password_unlock.as_deref())
+    {
+        let master_password_unlock =
+            MasterPasswordUnlockData::try_from(master_password_unlock_response.clone())?;
+
+        client.internal.update_kdf(master_password_unlock.kdf)?;
```
quexten: I'm not sure if we want to do this in this PR, but we should probably store the entire masterpassword unlock blob, and not split it out into kdf / email, so they cannot get desynced.

crates/bitwarden-core/src/client/internal.rs
```diff
@@ -327,6 +346,45 @@ impl InternalClient {
     ) -> Result<(), EncryptionSettingsError> {
         EncryptionSettings::set_org_keys(org_keys, &self.key_store)
     }
+
+    #[cfg(feature = "internal")]
+    pub fn update_kdf(&self, new_kdf: Kdf) -> Result<(), NotAuthenticatedError> {
+        let login_method = self.get_login_method().ok_or(NotAuthenticatedError)?;
```
quexten, line 420: I'm not sure if we want to do this in this PR, but we should probably store the entire masterpassword unlock blob, and not split it out into kdf / email, so they cannot get desynced.

crates/bitwarden-vault/src/sync.rs
mzieniukbw: Fixed in [e5f16dc](https://github.com/bitwarden/sdk-internal/pull/399/commits/e5f16dc7f21ac68c177b9848337dd91d29d303b9)

crates/bitwarden-core/src/client/internal.rs
mzieniukbw: Ideally, but i think that's a bit stretch, without doing major refactor of the login method.

crates/bitwarden-core/src/auth/login/access_token.rs
```diff
@@ -85,6 +85,11 @@ pub(crate) async fn login_access_token(
             r.refresh_token.clone(),
             r.expires_in,
         );
+
+        client
+            .internal
+            .initialize_crypto_single_org_key(organization_id, encryption_key);
```
Thomas-Avery, line 91: 💭  **question:** 

`initialize_crypto_single_org_key` didn't have the side effect of setting the login method? Is this change just to have consistency with the others? I'm a little hesitant because it's SM.

crates/bitwarden-core/src/auth/login/api_key.rs
```diff
@@ -32,34 +33,54 @@ pub(crate) async fn login_api_key(
         let kdf = client.auth().prelogin(email.clone()).await?;
 
         client.internal.set_tokens(
-            r.access_token.clone(),
-            r.refresh_token.clone(),
+            r.access_token.to_owned(),
+            r.refresh_token.to_owned(),
```
Thomas-Avery: 💭  **not a rust expert**

It is weird to me that we are keeping `.clone()` in `crates/bitwarden-core/src/auth/login/access_token.rs` `crates/bitwarden-core/src/auth/login/auth_request.rs` and `crates/bitwarden-core/src/auth/login/password.rs` and only updating to `.to_owned()` here.

From the rust docs https://doc.rust-lang.org/std/borrow/trait.ToOwned.html 

> A generalization of Clone to borrowed data.
> Some types make it possible to go from borrowed to owned, usually by implementing the Clone trait. But Clone works only for going from &T to T. The ToOwned trait generalizes Clone to construct owned data from any borrow of a given type.

It sounds like we should prefer clone when going from `T` to `T`? In this case we want to go from `String` -> `String` and `Option<String>` -> `Option<String>`.

crates/bitwarden-core/src/auth/login/api_key.rs
```diff
@@ -32,34 +33,54 @@ pub(crate) async fn login_api_key(
...
+            .and_then(|user_decryption| user_decryption.master_password_unlock)
+        {
+            client
+                .internal
+                .initialize_user_crypto_master_password_unlock(
+                    input.password.clone(),
+                    master_password_unlock,
+                    user_key_state,
+                )?;
+        } else {
```
Thomas-Avery: :pick: 
Since we care about both branches of logic `match` might be more readable here than`if let Some(`


```
      let master_password_unlock = r
            .user_decryption_options
            .as_ref()
            .map(UserDecryptionData::try_from)
            .transpose()?
            .and_then(|user_decryption| user_decryption.master_password_unlock);

        match master_password_unlock {
            Some(master_password_unlock) => {
                client
                    .internal
                    .initialize_user_crypto_master_password_unlock(
                        input.password.clone(),
                        master_password_unlock,
                        user_key_state,
                    )?;
            }
            None => {
                let user_key = r.key.as_deref();
                let user_key: EncString = require!(user_key).parse()?;
                let master_key = MasterKey::derive(&input.password, &email, &kdf)?;

                client.internal.initialize_user_crypto_master_key(
                    master_key,
                    user_key,
                    user_key_state,
                )?;
            }
        }
```

crates/bitwarden-core/src/auth/login/auth_request.rs
```diff
@@ -130,6 +121,14 @@ pub(crate) async fn complete_auth_request(
             })
             .await?;
 
+        client
+            .internal
+            .set_login_method(LoginMethod::User(UserLoginMethod::Username {
+                client_id: "web".to_owned(),
+                email: auth_req.email.to_owned(),
+                kdf: Kdf::default(),
+            }));
```
Thomas-Avery: ❓ Looking to clarify.

So this will be a change in behavior? Previously this would get overridden by the `initialize_user_crypto` setting the final state to: 

```
 client
        .internal
        .set_login_method(LoginMethod::User(UserLoginMethod::Username {
            client_id: "".to_string(),
            email: req.email,
            kdf: req.kdf_params,
        }));
```

With the difference being `client_id` will be getting set to `web` now? 

Reading through the old flow previously kdf was getting set to default and emails match up.

Odd for initing crypto to have that side effect but I don't understand all the flows.

crates/bitwarden-core/src/auth/login/password.rs
```diff
@@ -41,26 +42,48 @@ pub(crate) async fn login_password(
...
+        let private_key = r.private_key.as_deref();
+        let private_key: EncString = require!(private_key).parse()?;
+
+        let user_key_state = UserKeyState {
+            private_key,
+            signing_key: None,
+            security_state: None,
+        };
+
+        if let Some(master_password_unlock) = r
```
Thomas-Avery: :pick: Same comment about the match for readability.

crates/bitwarden-core/src/client/internal.rs
```diff
@@ -354,11 +371,65 @@ impl InternalClient {
...
+    ) -> Result<(), NotAuthenticatedError> {
+        let new_kdf = master_password_unlock.kdf;
+
+        let login_method = self.get_login_method().ok_or(NotAuthenticatedError)?;
+
+        let kdf = match login_method.as_ref() {
+            LoginMethod::User(UserLoginMethod::Username { kdf, .. }) => kdf,
+            LoginMethod::User(UserLoginMethod::ApiKey { kdf, .. }) => kdf,
+            #[cfg(feature = "secrets")]
+            LoginMethod::ServiceAccount(_) => return Err(NotAuthenticatedError)?,
```
Thomas-Avery:
```suggestion
            LoginMethod::ServiceAccount(_) => return Err(NotAuthenticatedError),
```

I don't think returning the error needs the unwrap.

crates/bitwarden-core/src/client/internal.rs
```diff
@@ -354,11 +371,65 @@ impl InternalClient {
...
+                    email,
+                    ..
+                }) => self.set_login_method(LoginMethod::User(UserLoginMethod::ApiKey {
+                    client_id: client_id.to_owned(),
+                    client_secret: client_secret.to_owned(),
+                    email: email.to_owned(),
+                    kdf: new_kdf,
+                })),
+                #[cfg(feature = "secrets")]
+                LoginMethod::ServiceAccount(_) => return Err(NotAuthenticatedError)?,
```
Thomas-Avery:
```suggestion
                LoginMethod::ServiceAccount(_) => return Err(NotAuthenticatedError),
```

crates/bitwarden-core/src/key_management/crypto.rs
```diff
@@ -1363,4 +1384,50 @@ mod tests {
...
+        let key_store = client.internal.get_key_store();
+        let context = key_store.context();
+        assert!(context.has_symmetric_key(SymmetricKeyId::User));
+        assert!(context.has_asymmetric_key(AsymmetricKeyId::UserPrivateKey));
+        let login_method = client.internal.get_login_method().unwrap();
+        if let LoginMethod::User(UserLoginMethod::Username { email, kdf, .. }) =
+            login_method.as_ref()
+        {
+            assert_eq!(*email, test_account.user.email);
+            assert_eq!(*kdf, test_account.user.kdf_params);
```
Thomas-Avery: Should we add an assert for the `client_id` to be the empty string?

Thomas-Avery (commented):
Mostly just questions for my clarity and a few optional minor changes.

crates/bitwarden-core/src/auth/login/password.rs
```diff
@@ -41,26 +42,48 @@ pub(crate) async fn login_password(
             r.refresh_token.clone(),
             r.expires_in,
         );
+
+        let private_key = r.private_key.as_deref();
```
quexten: Looks like this two-line pattern to convert from Option<&str> to EncString repeats a fair bit in this PR. Could we just implement TryFrom<Option<&str>> for EncString, and then remove all of thees temporary variables? It would make this a bit shorter and more readable.

crates/bitwarden-core/src/auth/login/password.rs
```diff
@@ -41,26 +42,48 @@ pub(crate) async fn login_password(
...
+        {
+            client
+                .internal
+                .initialize_user_crypto_master_password_unlock(
+                    input.password.clone(),
+                    master_password_unlock,
+                    user_key_state,
+                )?;
+        } else {
+            let user_key = r.key.as_deref();
```
quexten: Ref the comment above about TryFrom<Option<&str>>

crates/bitwarden-core/src/auth/login/api_key.rs
```diff
@@ -32,34 +33,54 @@ pub(crate) async fn login_api_key(
...
         client.internal.set_tokens(
-            r.access_token.clone(),
-            r.refresh_token.clone(),
+            r.access_token.to_owned(),
+            r.refresh_token.to_owned(),
             r.expires_in,
         );
 
-        let master_key = MasterKey::derive(&input.password, &email, &kdf)?;
+        let private_key = r.private_key.as_deref();
```
quexten: Ref the other comment about TryFrom<Option<&str>>

crates/bitwarden-core/src/auth/login/api_key.rs
```diff
@@ -32,34 +33,54 @@ pub(crate) async fn login_api_key(
...
+            client
+                .internal
+                .initialize_user_crypto_master_password_unlock(
+                    input.password.clone(),
+                    master_password_unlock,
+                    user_key_state,
+                )?;
+        } else {
+            let user_key = r.key.as_deref();
+            let user_key: EncString = require!(user_key).parse()?;
```
quexten: Ref the comment about TryFrom<Option<&str>>

crates/bitwarden-core/src/auth/login/api_key.rs
```diff
@@ -32,34 +33,54 @@ pub(crate) async fn login_api_key(
...
+        let private_key: EncString = require!(private_key).parse()?;
+
+        let user_key_state = UserKeyState {
+            private_key,
+            signing_key: None,
+            security_state: None,
+        };
+
+        if let Some(master_password_unlock) = r
+            .user_decryption_options
```
quexten, line 41: Suggestion (non-blocking): We have this long chain quite a bit. It may be nice to just implement "master_password_unlock() -> Option<MasterPasswordUnlockData>" on the decryptionData.

Then, you can rewrite as (untested):
```
let decryption_data: UserDecryptionData = r.user_decryption_options.
if let Some(master_password_unlock_data) = decryption_data.master_password_unlock_data() {
            client
                .internal
                .initialize_user_crypto_master_password_unlock(
                    input.password.clone(),
                    master_password_unlock,
                    user_key_state,
                )?;
} else {
           
                client.internal.initialize_user_crypto_master_key(
                    MasterKey::derive(&input.password, &email, &kdf)?,
                    r.key.try_into()?,
                    user_key_state,
                )?;
}

File: crates/bitwarden-core/src/auth/login/api_key.rs (hunk 5 above)
Comment (quexten) [reply crates/bitwarden-core/src/auth/login/api_key.rs]: Agree on this. In this case we are not even going from borrowed data to owned data, so semantically it is confusing saying "to owned" (implying we are *converting* to an owned object (by cloning usually).

We're going from String to String (T to T), so we should clone.

File: crates/bw/src/vault/sync.rs [hunk 17]
```diff
@@ -39,6 +44,18 @@ pub(crate) async fn sync(client: &Client, input: &SyncRequest) -> Result<SyncRes
         .await
         .map_err(|e| SyncError::Api(e.into()))?;
 
+    if let Some(master_password_unlock) = sync
+        .user_decryption
```
quexten, line 51: Ref the comment about implementing the masterpassword function on the decyrption data

crates/bitwarden-core/src/key_management/master_password.rs
```diff
@@ -29,19 +29,27 @@ pub(crate) enum MasterPasswordError {
...
+#[cfg_attr(feature = "uniffi", derive(uniffi::Record))]
+#[cfg_attr(
+    feature = "wasm",
+    derive(tsify::Tsify),
+    tsify(into_wasm_abi, from_wasm_abi)
+)]
+pub struct MasterPasswordUnlockData {
     /// The key derivation function used to derive the master key
-    kdf: Kdf,
+    pub kdf: Kdf,
```
quexten, line 48: Looks like this can be `pub(crate) ` unless I'm missing something?

crates/bitwarden-core/src/auth/login/api_key.rs
```diff
@@ -32,34 +33,54 @@ pub(crate) async fn login_api_key(
...
             r.expires_in,
         );
 
-        let master_key = MasterKey::derive(&input.password, &email, &kdf)?;
+        let private_key = r.private_key.as_deref();
+        let private_key: EncString = require!(private_key).parse()?;
+
+        let user_key_state = UserKeyState {
+            private_key,
+            signing_key: None,
```
quexten, line 36: (Not in this PR) looks like we don't have support for parsing the new keys from the token response here yet. We should make a follow-up ticket somewhere to fix this.

quexten (commented):
Looks mostly good, I have some stylistic / readability improvements noted.

crates/bitwarden-core/src/auth/login/access_token.rs
mzieniukbw: It's just for consistency. There is no risk, since we don't expect one or another to happen in any sequence.

crates/bitwarden-core/src/auth/login/password.rs
mzieniukbw: This would create an "opinionated" function, that actually requires the `&str`, even though it accepts `Option<&str>`. Probably for this reason we have [try_from_optional](https://github.com/bitwarden/sdk-internal/blob/49f84e6094884af64edd2d7d7399e7400ffed35a/crates/bitwarden-crypto/src/enc_string/symmetric.rs#L114), but it does not work here, since we have to clone borrowed reference:
```
        let private_key = EncString::try_from_optional(r.private_key.clone())?;
        let private_key = require!(private_key);
```
I want to retain the user friendly error: `MissingFieldError("private_key")`, so i don't see any better way to avoid the two-line pattern with the `r.private_key` being borrowed.

crates/bitwarden-core/src/auth/login/password.rs
mzieniukbw: See my comment https://github.com/bitwarden/sdk-internal/pull/399#discussion_r2330313242

crates/bitwarden-core/src/auth/login/api_key.rs
mzieniukbw: See my comment https://github.com/bitwarden/sdk-internal/pull/399#discussion_r2330313242

crates/bitwarden-core/src/auth/login/api_key.rs
mzieniukbw: See my comment https://github.com/bitwarden/sdk-internal/pull/399#discussion_r2330313242

crates/bitwarden-core/src/auth/login/api_key.rs
mzieniukbw: This doesn't help much, since we have to get `UserDecryptionData` first:
```
let decryption_data: UserDecryptionData = r
            .user_decryption_options
            .as_ref()
            .map(UserDecryptionData::try_from);
```

crates/bw/src/vault/sync.rs
mzieniukbw: See my comment https://github.com/bitwarden/sdk-internal/pull/399#discussion_r2330427770

crates/bitwarden-core/src/key_management/master_password.rs
mzieniukbw: Fixed in cfa499cb

crates/bitwarden-core/src/key_management/crypto.rs
mzieniukbw: Fixed in ef76b78e

crates/bitwarden-core/src/client/internal.rs
mzieniukbw: Fixed in ef76b78e

crates/bitwarden-core/src/client/internal.rs
mzieniukbw: Fixed in ef76b78e

crates/bitwarden-core/src/auth/login/password.rs
mzieniukbw: Fixed in ef76b78e

crates/bitwarden-core/src/auth/login/auth_request.rs
mzieniukbw: I wanted to replicate original behaviour with the `client_id`, before we started setting the `set_login_method` inside the `initialize_user_crypto`, but it have been over a year with this.
I have removed the manual `set_login_method` in ef76b78e, let's fully rely on `initialize_user_crypto` instead.

crates/bitwarden-core/src/auth/login/api_key.rs
mzieniukbw: Fixed in ef76b78e

crates/bitwarden-core/src/auth/login/api_key.rs
mzieniukbw: You are right, I have not realised the fields are dereferenced when accessed, even though the `r` is borrowed, hence it's `String` type.
Reverted to `clone()` in ef76b78e

Thomas-Avery (dismissed):
Confirmed all my feedback has been addressed. Looks good to me.

crates/bitwarden-core/src/auth/login/api_key.rs
```diff
@@ -37,29 +38,53 @@ pub(crate) async fn login_api_key(
             r.expires_in,
         );
 
-        let master_key = MasterKey::derive(&input.password, &email, &kdf)?;
+        let private_key = r.private_key.as_deref();
+        let private_key: EncString = require!(private_key).parse()?;
```
Hinton: **suggestion:** Can we just use `&` instead of `deref`?


```suggestion
        let private_key: EncString = require!(&r.private_key).parse()?;
```

crates/bitwarden-core/src/auth/login/api_key.rs
```diff
@@ -37,29 +38,53 @@ pub(crate) async fn login_api_key(
...
+                let user_key: EncString = require!(user_key).parse()?;
+                let master_key = MasterKey::derive(&input.password, &email, &kdf)?;
+
+                client.internal.initialize_user_crypto_master_key(
+                    master_key,
+                    user_key,
+                    user_key_state,
+                )?;
+            }
+        }
```
Hinton: **nit** (subjective): I think `if let` would be more idiomatic here.

crates/bitwarden-core/src/auth/login/api_key.rs
```diff
@@ -37,29 +38,53 @@ pub(crate) async fn login_api_key(
...
+                    .internal
+                    .initialize_user_crypto_master_password_unlock(
+                        input.password.clone(),
+                        master_password_unlock,
+                        user_key_state,
+                    )?;
+            }
+            None => {
+                let user_key = r.key.as_deref();
+                let user_key: EncString = require!(user_key).parse()?;
```
Hinton: **suggestion:** Can we just use `&` instead of `deref`?
```suggestion
                let user_key: EncString = require!(&r.key).parse()?;
```

crates/bitwarden-core/src/auth/login/password.rs
```diff
@@ -41,26 +42,51 @@ pub(crate) async fn login_password(
             r.refresh_token.clone(),
             r.expires_in,
         );
+
+        let private_key = r.private_key.as_deref();
+        let private_key: EncString = require!(private_key).parse()?;
```
Hinton:
```suggestion
        let private_key: EncString = require!(&r.private_key).parse()?;
```

crates/bitwarden-core/src/auth/login/password.rs
```diff
@@ -41,26 +42,51 @@ pub(crate) async fn login_password(
...
+                    .internal
+                    .initialize_user_crypto_master_password_unlock(
+                        input.password.clone(),
+                        master_password_unlock,
+                        user_key_state,
+                    )?;
+            }
+            None => {
+                let user_key = r.key.as_deref();
+                let user_key: EncString = require!(user_key).parse()?;
```
Hinton:
```suggestion
                let user_key: EncString = require!(&r.key).parse()?;
```

crates/bitwarden-core/src/client/test_accounts.rs
```diff
@@ -41,8 +41,8 @@ impl Client {
 /// accounts to be used on other servers this protection was explicitly removed from these data
 /// dumps.
 pub struct TestAccount {
-    user: InitUserCryptoRequest,
-    org: Option<InitOrgCryptoRequest>,
+    pub user: InitUserCryptoRequest,
+    pub org: Option<InitOrgCryptoRequest>,
```
Hinton: I'm not convinced we should be leaking these outside this struct. 🤔 I'm wondering if it would be better to implement some helper functions to get some of the data.

CC @dani-garcia

crates/bitwarden-core/src/auth/login/auth_request.rs
```diff
@@ -89,20 +88,11 @@ pub(crate) async fn complete_auth_request(
...
             r.refresh_token.clone(),
             r.expires_in,
         );
-        client
-            .internal
-            .set_login_method(LoginMethod::User(UserLoginMethod::Username {
-                client_id: "web".to_owned(),
-                email: auth_req.email.to_owned(),
-                kdf: kdf.clone(),
-            }));
```
Hinton, line 104: **question:** Set login method and crypto were previously two distinct operations. This feels like muddling the waters a bit. I know there is active work towards lifting out authentication which might change this again. Do we foresee this intermixing being an issue.

CC @dani-garcia @coroiu

crates/bitwarden-core/src/auth/login/auth_request.rs
```diff
@@ -118,8 +108,8 @@ pub(crate) async fn complete_auth_request(
             .crypto()
             .initialize_user_crypto(InitUserCryptoRequest {
                 user_id: None,
-                kdf_params: kdf,
-                email: auth_req.email,
+                kdf_params: Kdf::default(),
```
Hinton: **question:** Looking back at all these `Kdf::default`, what happens if we change the default? Will auth requests stop working until both clients are updated`?

I don't think we need to necessarily fix it but it's an observation.

crates/bitwarden-core/src/auth/login/password.rs
```diff
@@ -41,26 +42,51 @@ pub(crate) async fn login_password(
             r.refresh_token.clone(),
             r.expires_in,
         );
+
```
Hinton, line 48: **question:** On line 33-34 we create a MasterKey from the email and KDF we get from prelogin. Why are we not able to just grab the salt in prelogin and generate the master key once? That eliminates the distinction from these two flows.

Or are we fine with the MP Hash using a completely different KDF from Master Key.

Hinton (commented):
Partial review.

shane-melton (dismissed):
Vault owned change looks good

crates/bitwarden-core/src/auth/login/auth_request.rs
coroiu: If possible I think it would be beneficial to keep crypto and auth separate, but the exact design of a standalone auth client is still unknown so I have a hard time giving any more concrete advice

crates/bitwarden-core/src/auth/login/api_key.rs
mzieniukbw: I can change it back to `if let` if that's desired. https://github.com/bitwarden/sdk-internal/pull/399#discussion_r2320301154. I don't have strong opinion on this one.

coroiu (dismissed):
Don't have much to add, most of this is squarely in KM territory 🤷‍♂️

crates/bitwarden-core/src/auth/login/api_key.rs
mzieniukbw: When `r.private_key` is None, we will get `MissingFieldError("&response.salt")` error instead (with the `&`).
I specifically wanted to avoid this, hence why there is 2nd variable.

crates/bitwarden-core/src/auth/login/auth_request.rs
mzieniukbw: ~~It would break, but if we change it, we would change the default and all places that use it, for backward compatibility.
Maybe we should have an explicit function for this, instead of using default, but that's something i would consider in the future.~~
Actually, this might be wrong statement. If user logins with auth request (uses other trusted device to login) and that user have master password, then the Kdf settings have to match user's actual Kdf. This is likely a bug then ? Thoughts ?

crates/bitwarden-core/src/auth/login/password.rs
mzieniukbw: The other flow `initialize_user_crypto_master_key` will probably get deprecated and removed at some point. Having two is intermediary, but necessary step.

Hinton: If we remove the other flow then you still calculate the MP hash using the KDF settings from prelogin, but for decryption key you use completely different KDF settings.

Have we disconnected the MP KDF from decryption KDF?

crates/bitwarden-core/src/auth/login/api_key.rs
Hinton: This would be an internal error though? Or are we expected to propagate the actual error to the UI?

The `require!` macro is designed as a quick helper for throwing a generic `MissingFieldError`, while it sets a inner value the value should be considered internal and only for debugging purposes. If you want a "stable" error message that is more helpful I think you want to craft the inner value yourself.

crates/bitwarden-core/src/auth/login/password.rs
mzieniukbw: Fixed in 85eba26f

crates/bitwarden-core/src/auth/login/password.rs
mzieniukbw: Fixed in 85eba26f

crates/bitwarden-core/src/auth/login/api_key.rs
mzieniukbw: Fixed in 85eba26f

crates/bitwarden-core/src/auth/login/api_key.rs
mzieniukbw: Fixed in 85eba26f

crates/bitwarden-core/src/auth/login/password.rs
mzieniukbw: It's still the same from server perspective (as of the same source of KDF values) and yes, it's disconnected.

Maybe the prelogin should be responsibility of `login_password` in here, instead of being part of input's `PasswordLoginRequest` ?

But you are right. We have to construct `MasterKey` twice (once to get the password hash for authentication and second time for decryption), which is rather expensive operation. Maybe we have to rethink it. Opinions ?

Hinton: The first sentence is a contradiction. If it's the same from the server perspective i.e. a single source of truth then the KDF for login and decryption is connected.

If that's the case we can safely re-use `MasterKey` and probably should.

crates/bitwarden-core/src/auth/login/auth_request.rs
Hinton: I think so, this should probably be the users KDF? Initially I thought the kdf was related to the auth request but I don't think that's the case?

crates/bitwarden-core/src/client/test_accounts.rs
Hinton: Circling back to this, after discussion with @dani-garcia we both agree we don't want to make these fields public. I think for the purpose of the sync test you should use explicit keys and not rely on the `TestAccount` struct since it's more there for when you don't care about the account keys.

crates/bitwarden-core/src/auth/login/api_key.rs
Hinton: No strong opinion either.

crates/bitwarden-core/src/client/test_accounts.rs
mzieniukbw: Reverted in 65ec578a

crates/bitwarden-core/src/auth/login/password.rs
mzieniukbw: Fixed in dc2aa72b. Decided not to introduce new MasterPasswordUnlock method

mzieniukbw: Also removed `kdf` from `PasswordLoginRequest`, instead use `prelogin()` inside `login_password` for consistency

crates/bitwarden-core/src/auth/login/auth_request.rs
mzieniukbw: Fixed in 3f63f518

crates/bitwarden-core/src/auth/login/password.rs
```diff
@@ -41,26 +42,55 @@ pub(crate) async fn login_password(
...
+                    master_key,
+                    master_password_unlock.master_key_wrapped_user_key,
+                    user_key_state,
+                )?;
+
+                client
+                    .internal
+                    .set_login_method(LoginMethod::User(UserLoginMethod::Username {
+                        client_id: "web".to_owned(),
+                        email: master_password_unlock.salt,
```
jlf0dev: Just a heads up, Auth is looking to disconnect the email from the salt soon. I don't know if we need to do anything now, but eventually we'll need to pass in the email.

mzieniukbw: Thank you for heads up. The good information is that the `LoginMethod`'s email is only used for KDF derivation (As salt), so it should not be a big deal to change it. Maybe it would be good idea to change it as follow up to this pr ?

crates/bitwarden-core/src/auth/login/password.rs
```diff
@@ -41,26 +42,55 @@ pub(crate) async fn login_password(
...
+
+                client
+                    .internal
+                    .set_login_method(LoginMethod::User(UserLoginMethod::Username {
+                        client_id: "web".to_owned(),
+                        email: input.email.clone(),
+                        kdf,
+                    }));
+            }
+        };
```
Hinton: **question:** Should we have a TODO here to clean this up eventually?

crates/bitwarden-core/src/client/internal.rs
```diff
@@ -378,4 +436,94 @@ mod tests {
...
+            .unwrap()
+            .clone();
+        #[allow(deprecated)]
+        let organization_keys = context
+            .dangerous_get_symmetric_key(SymmetricKeyId::Organization(organization_id))
+            .unwrap()
+            .clone();
+
+        (user_key, organization_keys)
+    }
```
Hinton: **question:** What is the plan once these methods are removed? I don't believe we intend to have a way to extract keys out of the store eventually. Would the tests need to be completely rewritten, and should we do it now rather than increase the cost of the inevitable?

crates/bitwarden-core/src/client/internal.rs
```diff
@@ -354,11 +354,69 @@ impl InternalClient {
...
+                    email: email.to_owned(),
+                    kdf: new_kdf,
+                })),
+                #[cfg(feature = "secrets")]
+                LoginMethod::ServiceAccount(_) => return Err(NotAuthenticatedError),
+            };
+        }
+
+        Ok(())
+    }
```
Hinton: (non blocking): This is a decent chunk of logic which probably doesn't really belong in `InternalClient`, but we also already have a fair bit of logic related to keys ...

We should do something about it soon.

crates/bw/src/vault/sync.rs
Hinton, line 1: You are aware this code is all unused at the moment and we have no direct plans to use it?

**issue (non-blocking):** The auth logic included here doesn't belong to vault and I think this motivates lifting it out, the problem is we don't have a good place to store it right now.

Hinton (commented):
**question:** There is currently no way to interact with the added logic, do you intend to wire up web and mobile bindings?

mzieniukbw:
Moved to draft, until KM team decides whether to go back to `InitUserCryptoMethod::MasterPasswordUnlock`.

crates/bitwarden-core/src/client/internal.rs
mzieniukbw: Fixed in 73277859

crates/bw/src/vault/sync.rs
mzieniukbw: I am having hard time with this future proofing. There are several code paths that this PR touches, that i got to know are not used or/and have no plans on using it.
Unless this code is removed, marked as deprecated or marked with a comment somehow, i can't see why we should not change it.
Do you have any proposal ? Reverse this change, since it's not going to be used anyway ?

crates/bitwarden-core/src/auth/login/password.rs
mzieniukbw: Raised https://bitwarden.atlassian.net/browse/PM-26585 to clean this up.

mzieniukbw: Cleaned up already from `password.rs` and `api_key.rs`, since those two does not seem to be used anywhere, except testing

crates/bitwarden-core/src/client/internal.rs
```diff
@@ -390,11 +390,85 @@ impl InternalClient {
...
     }
+
+    #[cfg(feature = "internal")]
+    pub(crate) fn initialize_user_crypto_master_password_unlock(
+        &self,
+        password: String,
+        master_password_unlock: MasterPasswordUnlockData,
+        key_state: UserKeyState,
+    ) -> Result<(), EncryptionSettingsError> {
+        let master_key = MasterKey::derive(
```
quexten, line 401: Thought for follow-up tech debt:  Should these live on the masterPasswordUnlockData? I.e
```
let user_key = MasterPasswordUnlockData::decrypt(&my_password);
```
or even
```
MasterPasswordUnlockData::decrypt(&my_password, SymmKeyId::User, &ctx);
```

(If we agree this is a good change, let's make it a *follow-up* tech debt, since this PR is already quite large)

crates/bw/src/vault/sync.rs
quexten: > , the problem is we don't have a good place to store it right now.

Is there a reason we can't delete it entirely? It causes extensive maintenance burden as seen by the large changes needed in this PR, and we can get it back in git history (granted, it would then need updating).

crates/bitwarden-core/src/client/internal.rs
```diff
@@ -390,11 +390,85 @@ impl InternalClient {
...
+            &password,
+            &master_password_unlock.salt,
+            &master_password_unlock.kdf,
+        )?;
+        let user_key =
+            master_key.decrypt_user_key(master_password_unlock.master_key_wrapped_user_key)?;
+        self.initialize_user_crypto_decrypted_key(user_key, key_state)
+    }
+
+    /// Updates user's KDF for the master password unlock login method.
```
quexten: Can we clarify this comment a bit? "Updates the user's KDF" suggests to me that it sends a request to the server, or at the very least returns data that can be used to update the KDF with the server. But this really is just setting the in-memory state on sync?

quexten (commented):
Only minor things from my side. I wish we could skip all of the unused code changes. Other than these I'm happy to approve.

crates/bw/src/vault/sync.rs
mzieniukbw: Moved `sync.rs` into `bw` testing cli 9e289716

crates/bitwarden-core/src/client/internal.rs
mzieniukbw: Feels reasonable to me. I would go with option 1

crates/bitwarden-core/src/client/internal.rs
mzieniukbw: Changed to 'set' and updated docs in 7444ee61

quexten (approved):

crates/bitwarden-core/src/client/internal.rs
quexten: https://bitwarden.atlassian.net/browse/PM-26906

coroiu (approved):

crates/bitwarden-core/src/auth/login/password.rs
```diff
@@ -29,38 +28,55 @@ pub(crate) async fn login_password(
...
+                    input.password.clone(),
+                    master_password_unlock.clone(),
+                    user_key_state,
+                )?;
+
+            client
+                .internal
+                .set_login_method(LoginMethod::User(UserLoginMethod::Username {
+                    client_id: "web".to_owned(),
+                    email: master_password_unlock.salt,
```
JaredSnider-Bitwarden, line 76: ❓ : Auth is in process of breaking down the separation of the user's email and salt. If we allow users to set new salt's in the future on key rotation (which per my understanding is the plan), then this would break.  Should we not use `&input.email` here instead?

crates/bitwarden-core/src/auth/login/auth_request.rs
```diff
@@ -113,12 +106,27 @@ pub(crate) async fn complete_auth_request(
...
+            .map(|mpu| mpu.salt.clone())
+            .unwrap_or_else(|| auth_req.email.clone());
+
         client
             .crypto()
             .initialize_user_crypto(InitUserCryptoRequest {
                 user_id: None,
                 kdf_params: kdf,
-                email: auth_req.email,
+                email: salt,
```
JaredSnider-Bitwarden, line 129: Same comment as https://github.com/bitwarden/sdk-internal/pull/399#discussion_r2433382858

crates/bitwarden-core/src/auth/login/api_key.rs
```diff
@@ -23,44 +23,44 @@ pub(crate) async fn login_api_key(
...
+                    master_password_unlock.clone(),
+                    user_key_state,
+                )?;
+
+            client
+                .internal
+                .set_login_method(LoginMethod::User(UserLoginMethod::ApiKey {
+                    client_id: input.client_id.clone(),
+                    client_secret: input.client_secret.clone(),
+                    email: master_password_unlock.salt,
```
JaredSnider-Bitwarden, line 60: Same comment as https://github.com/bitwarden/sdk-internal/pull/399#discussion_r2433382858

crates/bitwarden-core/src/auth/login/password.rs
mzieniukbw: The `set_login_method` function does not really work for neither of the plans (master password authentication and unlock abstraction and separation of email and salt) and i think everyone is in understanding that it needs to change. This is out of scope of this PR.
From Auth perspective, salt is used to derive the master password hash, used to authenticate.
From KM perspective, salt is used to derive Master Key, which is used to decrypt the User Key.
The salt that is used right now (email) is used in the same way for both scenarios, but there is no reason to keep them linked in the future.
The `set_login_method` is function used for unlock, not for authentication.
Hence i think it is fair to use the salt from `MasterPasswordUnlockData` for this scenario, even if the naming seems confusing and misplaced.

JaredSnider-Bitwarden: Ah, I pulled down the code and took a better look at how the email string is used - it's actually used as the salt in other scenarios. That makes more sense now. Thank you for the explanation. I look forward to when we can make this more clear!

JaredSnider-Bitwarden (approved):

shane-melton (approved):
Vault changes and the removal of the outdated sync look good!

PR #409, opened 2025-08-22
Use the new B64 in bitwarden-core and bitwarden-crypto

Hinton:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-25121

## 📔 Objective

Migrates most usages of B64 into the new `B64` struct.

## ⏰ Reminders before review

- CI builds passed
  team

## 🦮 Reviewer guidelines


crates/bitwarden-core/src/auth/login/access_token.rs
```diff
@@ -130,7 +129,8 @@ fn load_tokens_from_state(
             let organization_id: OrganizationId = organization_id
                 .parse()
                 .map_err(|_| LoginError::InvalidOrganizationId)?;
-            let encryption_key = SymmetricCryptoKey::try_from(client_state.encryption_key)?;
+            let encryption_key =
+                SymmetricCryptoKey::try_from(client_state.encryption_key.to_string())?;
```
Hinton: @quexten I think we want to implement `From<B64>` for `SymmetricCryptoKey`?

crates/bitwarden-core/src/auth/access_token.rs
```diff
@@ -67,11 +65,11 @@ impl FromStr for AccessToken {
...
         };
 
-        let encryption_key = STANDARD_INDIFFERENT.decode(encryption_key)?;
-        let encryption_key = Zeroizing::new(encryption_key.try_into().map_err(|e: Vec<_>| {
+        let encryption_key: B64 = encryption_key.parse()?;
+        let encryption_key = Zeroizing::new(encryption_key.as_ref().try_into().map_err(|_| {
             AccessTokenInvalidError::InvalidBase64Length {
                 expected: 16,
-                got: e.len(),
+                got: encryption_key.as_ref().len(),
```
dani-garcia: I wonder if we should implement `Deref<[u8]> for B64`, then you could remove the `as_ref()`:
```rust
        let encryption_key = Zeroizing::new((*encryption_key).try_into().map_err(|_| {
            AccessTokenInvalidError::InvalidBase64Length {
                expected: 16,
                got: encryption_key.len(),
            }
        })?);
```

crates/bitwarden-core/src/auth/auth_request.rs
```diff
@@ -36,11 +37,10 @@ pub(crate) fn new_auth_request(email: &str) -> Result<AuthRequestResponse, Crypt
     let spki = key.to_public_key().to_der()?;
 
     let fingerprint = fingerprint(email, &spki)?;
-    let b64 = STANDARD.encode(&spki);
 
     Ok(AuthRequestResponse {
-        private_key: STANDARD.encode(key.to_der()?),
-        public_key: b64,
+        private_key: key.to_der()?.as_ref().into(),
+        public_key: spki.as_ref().into(),
```
dani-garcia: Do we want a direct B64 -> Bytes and Bytes -> B64 conversion? It should avoid the need to call as_ref as much

crates/bitwarden-crypto/Cargo.toml
dani-garcia, line 1: I've had to add the `bitwarden-decoding/uniffi` feature to the `uniffi` feature in the Cargo.toml of `bitwarden-crypto`:
```
[features]
default = []
wasm = ["dep:tsify", "dep:wasm-bindgen"] # WASM support

uniffi = ["dep:uniffi", "bitwarden-encoding/uniffi"]  # Uniffi bindings
``` 


Otherwise it fails to compile when running a test from VSCode for me. This is with the `EncString` tests:
```
error[E0277]: the trait bound `bitwarden_encoding::B64: uniffi::TypeId<UniFfiTag>` is not satisfied
  --> crates/bitwarden-crypto/src/rsa.rs:19:17
   |
19 |     pub public: B64,
   |                 ^^^ the trait `uniffi::TypeId<UniFfiTag>` is not implemented for `bitwarden_encoding::B64`
```

Might need that in the `bitwarden-core` toml too?

dani-garcia (commented):
Overall looks pretty good to me, some comments

crates/bitwarden-core/src/auth/auth_request.rs
Hinton: @quexten thoughts?

crates/bitwarden-core/src/auth/access_token.rs
Hinton: Maybe. At the same time I like that we're explicit when we need to go B64 -> slice. In your suggestion there is an ambiguity of what len returns. The length of bytes or length of the b64 encoded string?

dani-garcia: True, but then `as_ref` is not a particularly descriptive trait either. If we wanted to be unambiguous I feel like we should use an `as_slice` function.

crates/bitwarden-core/src/auth/auth_request.rs
quexten: I think that makes sense. 

Note, for some instantiations of `Bytes` such as utf8, (maybe octet stream, pkcs8privatekey) we don't need / want the implementation, so we may add bounds and a marker trait for the variant that we allow the conversion for?

crates/bitwarden-core/src/auth/login/access_token.rs
quexten: I wonder how many more use-cases there are. If you are b64 encoding your unencrypted key, that to me indicates that the use-case is not supported long-term since keys should not leave the SDK, so we'd be making ergonomic improvements for paths we want to discourage.

Are there other areas we'd want to b64 encode a key in?
(Maybe: Never lock? Biometric unlock?)

dani-garcia (approved):

Thomas-Avery (approved):
Nice, this makes things easier to follow and understand.

quexten (approved):
Nice, the marker trait is neat!

PR #421, opened 2025-09-02
Expose counter to support more efficient sync

dereknance:
## 🎟️ Tracking

[PM-22201][PM-22201]

## 📔 Objective

Mobile clients need to know if the FIDO 2 counter property is nonzero in order to determine if sync is required before performing authentication.

Updates `cargo-udeps` in `lint.yml` to version `0.1.57`. Previous version was 9 months old. Locally, I was experiencing a bug where `cargo-udeps` was considering `#[cfg(test)]` to be invalid.

Updates a couple static assertions that were showing up as unused when running `cargo +nightly udeps --workspace --all-features` locally. Unsure why this wasn't previously showing up in CI.

Added an explicit lifetime to fix a `mismatched-lifetime-syntax` error, again when running the `udeps` check locally.

## ⏰ Reminders before review

- CI builds passed
  team

## 🦮 Reviewer guidelines

[PM-22201]: https://bitwarden.atlassian.net/browse/PM-22201?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -87,7 +87,7 @@ jobs:
         run: cargo sort --workspace --check
 
       - name: Install cargo-udeps
-        run: cargo install cargo-udeps --version 0.1.53 --locked
+        run: cargo install cargo-udeps --version 0.1.57 --locked
 
       - name: Cargo udeps
         run: cargo +"${{ steps.nightly-toolchain.outputs.RUST_NIGHTLY_TOOLCHAIN }}" udeps --workspace --all-features
--- a/crates/bitwarden-crypto/src/keys/asymmetric_crypto_key.rs
+++ b/crates/bitwarden-crypto/src/keys/asymmetric_crypto_key.rs
@@ -76,11 +76,9 @@ pub struct AsymmetricCryptoKey {
 
 // Note that RsaPrivateKey already implements ZeroizeOnDrop, so we don't need to do anything
 // We add this assertion to make sure that this is still true in the future
-const _: () = {
+const _: fn() = || {
     fn assert_zeroize_on_drop<T: zeroize::ZeroizeOnDrop>() {}
-    fn assert_all() {
-        assert_zeroize_on_drop::<RsaPrivateKey>();
-    }
+    assert_zeroize_on_drop::<RsaPrivateKey>();
 };
 impl zeroize::ZeroizeOnDrop for AsymmetricCryptoKey {}
 impl CryptoKey for AsymmetricCryptoKey {}
--- a/crates/bitwarden-fido/src/types.rs
+++ b/crates/bitwarden-fido/src/types.rs
@@ -97,9 +97,7 @@ impl Fido2CredentialAutofillView {
                                     .as_ref()
                                     .and_then(|l| l.username.none_whitespace()))
                                 .or(cipher.name.none_whitespace()),
-                            has_counter: c.counter.none_whitespace().is_some_and(|counter_str| {
-                                counter_str.parse::<u64>().is_ok_and(|counter| counter > 0)
-                            }),
+                            has_counter: Self::has_signature_counter(&c.counter),
                         })
                     })
             })
@@ -136,15 +134,18 @@ impl Fido2CredentialAutofillView {
                                     .or(c.user_display_name.none_whitespace())
                                     .or(username.none_whitespace())
                                     .or(cipher.name.none_whitespace()),
-                                // `Fido2CredentialListView` doesn't expose the counter property, so
-                                // assume there is not one.
-                                has_counter: false,
+                                has_counter: Self::has_signature_counter(&c.counter),
                             })
                         })
                 })
                 .collect(),
             _ => Ok(vec![]),
         }
+    }
+
+    fn has_signature_counter(str: &String) -> bool {
+        str.none_whitespace()
+            .is_some_and(|counter_str| counter_str.parse::<u64>().is_ok_and(|counter| counter > 0))
     }
 }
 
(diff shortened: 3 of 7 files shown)


dereknance:
@addisonbeck aside from a unit test or two, this is ready to start looking at. I figured I'd wait to take this PR out of draft until I received a first pass from you.

I went ahead and included fixes for the things that the `udeps` check was failing on locally. I can totally rip those out if that's undesirable! For whatever reason, they haven't been failing the `udeps` check in CI.

.github/workflows/lint.yml
```diff
@@ -87,7 +87,7 @@ jobs:
         run: cargo sort --workspace --check
 
       - name: Install cargo-udeps
-        run: cargo install cargo-udeps --version 0.1.53 --locked
+        run: cargo install cargo-udeps --version 0.1.57 --locked
```
addisonbeck, line 90: Not gonna lie, I expected this to be a bigger version jump considering it was broken for you locally

crates/bitwarden-crypto/src/keys/asymmetric_crypto_key.rs
```diff
@@ -76,11 +76,9 @@ pub struct AsymmetricCryptoKey {
 
 // Note that RsaPrivateKey already implements ZeroizeOnDrop, so we don't need to do anything
 // We add this assertion to make sure that this is still true in the future
-const _: () = {
+const _: fn() = || {
     fn assert_zeroize_on_drop<T: zeroize::ZeroizeOnDrop>() {}
-    fn assert_all() {
-        assert_zeroize_on_drop::<RsaPrivateKey>();
-    }
```
addisonbeck, line 83: Mind explaining this change to me? I don't really get why assert_all is being removed.

addisonbeck (commented):
This looks good to me. I request a review from Dani, since I don't have my name in the commit history of the SDK and he's the expert.

dani-garcia:
I think the issues you were finding with the static assertions and the elided lifetime were likely caused by using a newer `nightly` which has some newer lints, rather than by `udeps` specifically. Seems okay to include them though, they're fairly small anyway.

crates/bitwarden-crypto/src/keys/asymmetric_crypto_key.rs
dereknance: `assert_all` was being picked up as unused because it wasn't ever called. The type assertion was still valid. I could also have added a call to `assert_all()` after its declaration and that would have yielded the same result.

dereknance:
> I think the issues you were finding with the static assertions and the elided lifetime were likely caused by using a newer `nightly` which has some newer lints, rather than by `udeps` specifically. Seems okay to include them though, they're fairly small anyway.

I thought so too, so I tried running the same command with the `nightly-2025-08-18` toolchain (since that's what I understand the `lint.yml` workflow is using) and udeps 0.1.54, but those same unused code bits show up.

At this point I'm curious what it is about my env that is different from CI.

crates/bitwarden-fido/src/types.rs
```diff
@@ -132,6 +136,9 @@ impl Fido2CredentialAutofillView {
                                     .or(c.user_display_name.none_whitespace())
                                     .or(username.none_whitespace())
                                     .or(cipher.name.none_whitespace()),
+                                // `Fido2CredentialListView` doesn't expose the counter property, so
+                                // assume there is not one.
```
dani-garcia: I don't think this is a safe assumtion to make, we should probably add `has_counter` to `Fido2CredentialListView` as well so we can pass it over here.

dereknance: Fixed in df88442!

addisonbeck (approved):

crates/bitwarden-fido/src/types.rs
```diff
@@ -24,6 +24,7 @@ pub struct Fido2CredentialAutofillView {
     pub rp_id: String,
     pub user_name_for_ui: Option<String>,
     pub user_handle: Vec<u8>,
+    pub has_counter: bool,
```
addisonbeck, line 30:
```suggestion
    /// Indicates if this credential uses a signature counter (legacy passkeys).
    /// When true, mobile clients must sync before authentication to ensure
    /// counter values are current. Modern passkeys (counter = 0) can work offline.
    pub has_counter: bool,
```
I realized while I was going over this again that I didn't understand _why_ mobile needed this counter to decide if sync was needed or not in order to support old passkeys. Here's a quote from Anders:

>That was likely the case for passkeys created before 2024-ish, but should not be the case now.
>The reason was because we implemented support for signatureCounter which required us to sync the vault on each passkey assertion.
>We’ve since tweaked our behavior and all ”modern” passkeys should not require vault access.

So, in `mobile` we're complicating the sync-by-default approach, and want to check if the field is used for a given passkey to determine if it's an "old passkey" and needs special treatment (a sync). This is some in-the-weeds Passkeys stuff, and is information that will be lost to time. I think adding a comment would be helpful.

addisonbeck (changes requested):

dani-garcia (approved):

crates/bitwarden-fido/src/types.rs
dereknance: Fixed in [778d1f3](https://github.com/bitwarden/sdk-internal/pull/421/commits/778d1f367ce0c861255ad3fce4f76d57a421aad4)!

addisonbeck (approved):
Great job!

coroiu (approved):
Looks good, nice job 👍

mzieniukbw (approved):

nick-livefront (approved):

PR #424, opened 2025-09-03
[PM-23785] Use actual error types in UniFFI

dani-garcia:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-23785

## 📔 Objective

Currently all the UniFFI errors get serialized as a string. This PR updates UniFFI to the latest main changes to add support for proper errors. This requires a couple of changes:

- I've tagged all the errors with `uniffi::Error` and added support to the bitwarden-error macro for uniffi.
- I've had to add some workaround code to deal with an edge case that caused some compilation issues, more info upstream: https://github.com/mozilla/uniffi-rs/issues/2636
- Updated the gradle wrapper included as the old version was having some issues with my java version

How this looks on the mobile apps:
Android:
```kt
try {
    client.crypto().deriveKeyConnector(request)
} catch (ex: BitwardenException.DeriveKeyConnector) {
    when (ex.v1) {
        is DeriveKeyConnectorException.WrongPassword -> {
            println("That is the wrong password: $ex")
        }
        else -> {
            println("Other bitwarden error: ${ex.v1}")
        }
    }
} catch (ex: BitwardenException) {
    println("Other bitwarden error: $ex")
}
```

iOS:
```swift
do {
    try client.crypto().deriveKeyConnector(request);
} catch(BitwardenError.DeriveKeyConnector(DeriveKeyConnectorError.WrongPassword(let msg))) {
    print("Wrong password:", msg)
} catch {
    print("Other bitwarden error:", error)
}
```

## ⏰ Reminders before review

- CI builds passed
  team

## 🦮 Reviewer guidelines


crates/bitwarden-uniffi/src/platform/repository.rs
```diff
@@ -64,12 +63,7 @@ macro_rules! create_uniffi_repository {
             async fn has(
                 &self,
                 id: String,
-            ) -> Result<bool, $crate::platform::repository::RepositoryError> {
-                match self.get(id).await {
-                    Ok(x) => Ok(x.is_some()),
-                    Err(e) => Err(e),
-                }
-            }
+            ) -> Result<bool, $crate::platform::repository::RepositoryError>;
```
dani-garcia, line 66: The latest UniFFI doesn't allow default trait methods, so I've removed it. We weren't using it anyway.

crates/bitwarden-error-macro/src/full/attribute.rs
```diff
@@ -21,6 +21,7 @@ pub(crate) fn bitwarden_error_full(
 
     quote! {
         #[derive(serde::Serialize)]
+        #[cfg_attr(feature = "uniffi", derive(uniffi::Error))]
```
dani-garcia, line 24: Note that this is different than what WASM is doing. The WASM part is generating the code based on the `wasm` feature flag of the macro crate itself, while the UniFFI is always emmitting the `cfg_attr` code, which means it's the feature flag of the final crate that gets evaluated.

I've needed to do this because I couldn't quite get it working the other way, I think the workspace feature unification was causing issues and generating the UniFFI code all the time, even when not needed (like for the bitwarden-ipc crate).

Doing it this new way has one negative effect, which is that it's not obvious that the final crate has to have a `uniffi` feature, and otherwise you'll get a warning (that I've had to silence in the root cargo.toml because of wasm-only crates like bitwarden-ipc)

dereknance (dismissed):

dereknance (approved):

PR #430, opened 2025-09-05
Add dylint lints for errors

Hinton:
## 📔 Objective

This adds two custom dylint lint.

- `error_suffix`: Requires all types that implements `Error` ends with an `Error` suffix.
- `error_enum`: Forbids ending enum error variants with `Error`.

## ⏰ Reminders before review

- CI builds passed
  team

## 🦮 Reviewer guidelines


Cargo.toml
```diff
@@ -1,6 +1,7 @@
 [workspace]
 resolver = "2"
 members = ["bitwarden_license/*", "crates/*"]
+exclude = ["support/lints"]
```
dani-garcia, line 4: Do you need to explicitly exclude this when it's not part of `members`?

support/lints/rust-toolchain
dani-garcia, line 1: can we rename the file to `rust-toolchain.toml` to have syntax highlighting?

Also would be nice for the version to match the one we use for formatting, to avoid having multiples:
https://github.com/bitwarden/sdk-internal/blob/8447df0c9a2f0c8639a5937c889d20662b18b8b4/rust-toolchain.toml#L8

dani-garcia (commented):
Looks pretty reasonable overall, would be nice to add some instructions on how to run them  on the readme, we already have a lint section

Cargo.toml
Hinton: We do, if not we get warnings.

support/lints/rust-toolchain
Hinton: I did attempt that but I got compile errors. I suspect I would have to bump the clippy version as well and I'm unsure if we can do that easily with dylint.

dani-garcia: That makes sense, they seem to be using clippy and rust internals and probably need to be updated in lockstep. we'll need to make sure that renovate doesn't trigger on them

dani-garcia (dismissed):

support/lints/error_suffix/src/lib.rs
```diff
@@ -16,10 +16,15 @@ dylint_linting::declare_late_lint! {
...
+        let ident = match item.kind {
+            ItemKind::Enum(ident, ..) | ItemKind::Struct(ident, ..) => ident,
+            _ => return,
+        };
+
+        let item_name = ident.name.as_str();
 
         match &item.kind {
-            ItemKind::Enum(_, _) | ItemKind::Struct(_, _) => {
+            ItemKind::Enum(_, _, _) | ItemKind::Struct(_, _, _) => {
```
dani-garcia: In these cases we could probably match against `ItemKind::Enum(..)` which would work even if the field count was updated again, but not a big deal either way

dani-garcia (dismissed):

dani-garcia (approved):

PR #433, opened 2025-09-10
PM-25012: Cipher versioning data types

nikwithak:
## 🎟️ Tracking

[PM-25012](https://bitwarden.atlassian.net/browse/PM-25012)

## 📔 Objective

Moves the deserialization of the `data` object into the SDK - as part of running cipher migrations, the CipherType objects will be extracted instead of being passed in from the client.

## ⏰ Reminders before review

- CI builds passed
  team

## 🦮 Reviewer guidelines

[PM-25012]: https://bitwarden.atlassian.net/browse/PM-25012?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


crates/bitwarden-vault/src/cipher/login.rs
```diff
@@ -279,7 +279,7 @@ impl Decryptable<KeyIds, SymmetricKeyId, Fido2CredentialFullView> for Fido2Crede
 
 #[allow(missing_docs)]
 #[derive(Serialize, Deserialize, Debug, Clone)]
-#[serde(rename_all = "camelCase", deny_unknown_fields)]
+#[serde(rename_all = "camelCase")]
```
nikwithak, line 283: Are there any concerns / issues with allowing (and ignoring) unknown fields during deserialization? Removing this attribute from all the CipherType types drastically simplifies the extraction process.

gbubemismith: I think we always want `deny_unknown_fields` to keep things in sync

crates/bitwarden-vault/src/cipher/cipher.rs
```diff
@@ -455,6 +454,24 @@ impl Cipher {
             .map(|kind| kind.get_copyable_fields(Some(self)))
             .unwrap_or_default()
     }
+
+    /// Extracts and sets the CipherType-specific fields from the opaque `data` field.
```
gbubemismith, line 458: can we add some more documentation here ?

crates/bitwarden-vault/src/cipher/cipher.rs
```diff
@@ -455,6 +454,24 @@ impl Cipher {
...
+    pub(crate) fn populate_cipher_types(&mut self) {
+        let Ok(data) =
+            serde_json::from_str::<serde_json::Value>(self.data.as_deref().unwrap_or("{}"))
```
gbubemismith: ❓ looks like we are doing a double deserialization. Would it be better if we did this directly on the target type

crates/bitwarden-vault/src/cipher/cipher.rs
```diff
@@ -455,6 +454,24 @@ impl Cipher {
...
+            // If we can't deserialize the data, then we'll return the cipher as-is.
+            return;
+        };
+
+        match &self.r#type {
+            crate::CipherType::Login => self.login = serde_json::from_value(data).ok(),
+            crate::CipherType::SecureNote => self.secure_note = serde_json::from_value(data).ok(),
+            crate::CipherType::Card => self.card = serde_json::from_value(data).ok(),
+            crate::CipherType::Identity => self.identity = serde_json::from_value(data).ok(),
+            crate::CipherType::SshKey => self.ssh_key = serde_json::from_value(data).ok(),
```
gbubemismith: ❓ what happens if deserialization fails?

crates/bitwarden-vault/src/cipher/card.rs
```diff
@@ -12,7 +12,7 @@ use super::cipher::CipherKind;
 use crate::{cipher::cipher::CopyableCipherFields, Cipher, VaultParseError};
 
 #[derive(Serialize, Deserialize, Debug, Clone)]
-#[serde(rename_all = "camelCase", deny_unknown_fields)]
```
gbubemismith, line 15: we shouldn't remove the deny_unknown_fields

nikwithak: > I think we always want deny_unknown_fields to keep things in sync

What are we trying to keep in sync - is it that we don't get additional data from the server that we don't expect?

To me that makes sense if we need to make sure that the server isn't giving us additional data that we aren't expecting. But with this change, we are no longer using the server response for these fields, and are instead extracting it ourselves from the `data` field on the client-side - with that change, these values simply become windowed views into the `data`field, and I don't think we are gaining value from rejecting unknown fields anymore.

crates/bitwarden-vault/src/cipher/cipher.rs
nikwithak: If the deserialization fails, then we set the value to None. The cases I can think of where serialization fails are that `data` doesn't contain the required values for that CipherType.

Is it preferable to fail entirely and return an error in that case? It would mean that the data stored in the cipher's`.data` is inconsistent with the schema we've defined.

crates/bitwarden-vault/src/cipher/cipher.rs
nikwithak: Are you referring to the fact that we're doing `from_str` here, and `from_value` below? 

Since we're only worried about the CipherType deserializations in this function, I think I can do away with the `from_str` call on this line, and only deserialize inside the match (with from_str instead of from_value).

However it sounds like I may need to refactor this anyway - if we need to keep `#[deny_unknown_fields]` like called out below, then we can't use serde to deserialize and I'll need to manually extract each of the values from `data` here - see [discussion below](https://github.com/bitwarden/sdk-internal/pull/433#discussion_r2356288864).

gbubemismith:
@nikwithak Nice work so far. I think we want to target the `vault/feature/cipher-versioning` branch,

crates/bitwarden-vault/src/cipher/cipher.rs
gbubemismith: > Since we're only worried about the CipherType deserializations in this function, I think I can do away with the from_str call on this line, and only deserialize inside the match (with from_str instead of from_value).

Yeah this is what I am referring to

crates/bitwarden-vault/src/cipher/card.rs
nikwithak: I synced with @gbubemismith offline on this, and right now we agreed that there's no need to keep `deny_unknown_fields`

crates/bitwarden-vault/src/cipher/cipher.rs
nikwithak: @gbubemismith I added error handling - I added some additional error types to the existing `VaultParseError` and `CipherError` types in the new revision to handle these cases.

nikwithak:
@gbubemismith 

> I think we want to target the vault/feature/cipher-versioning branch,

I changed the target to `vault/feature/cipher-versioning`, but changed it back since it cluttered the diff with all of the API changes from the current target. Do you plant to merge `vault/cipher-versioning-with-data` (which has the generated API code) into the feature branch, or should I nix those commits from my branch entirely? This change is dependent on the `.data` field added by that branch.

gbubemismith:
> @gbubemismith
> 
> > I think we want to target the vault/feature/cipher-versioning branch,
> 
> I changed the target to `vault/feature/cipher-versioning`, but changed it back since it cluttered the diff with all of the API changes from the current target. Do you plant to merge `vault/cipher-versioning-with-data` (which has the generated API code) into the feature branch, or should I nix those commits from my branch entirely? This change is dependent on the `.data` field added by that branch.

I have a PR that has the api generated code, I can merge mine first

nikwithak:
Converting to draft while I sort out the diff from changing the target branch.

PR #455, opened 2025-09-15
[PM-25818] Migrate Basic Cipher Create, Edit, and Get Operations to SDK

gbubemismith:
## 🎟️ Tracking
https://bitwarden.atlassian.net/browse/PM-25818

## ⏰ Reminders before review

- CI builds passed
  team

## 🦮 Reviewer guidelines


crates/bitwarden-vault/src/cipher/cipher.rs
```diff
@@ -788,6 +797,76 @@ impl From<bitwarden_api_api::models::CipherRepromptType> for CipherRepromptType
...
+            folder_id: cipher.folder_id.map(FolderId::new),
+            collection_ids: vec![], // CipherResponseModel doesn't include collection_ids
+            name: require!(EncString::try_from_optional(cipher.name)?),
+            notes: EncString::try_from_optional(cipher.notes)?,
+            r#type: require!(cipher.r#type).into(),
+            login: cipher.login.map(|l| (*l).try_into()).transpose()?,
+            identity: cipher.identity.map(|i| (*i).try_into()).transpose()?,
+            card: cipher.card.map(|c| (*c).try_into()).transpose()?,
+            secure_note: cipher.secure_note.map(|s| (*s).try_into()).transpose()?,
+            // TODO: add ssh_key
```
nick-livefront: ❓ Are there tickets associated with these `TODO`s?

crates/bitwarden-vault/src/cipher/cipher.rs
```diff
@@ -788,6 +797,76 @@ impl From<bitwarden_api_api::models::CipherRepromptType> for CipherRepromptType
...
+            ssh_key: None,
+            favorite: cipher.favorite.unwrap_or(false),
+            reprompt: cipher
+                .reprompt
+                .map(|r| r.into())
+                .unwrap_or(CipherRepromptType::None),
+            organization_use_totp: cipher.organization_use_totp.unwrap_or(true),
+            edit: cipher.edit.unwrap_or(true),
```
nick-livefront: ❓ Probably an edge case, the current CipherView [defaults these values](https://github.com/bitwarden/clients/blob/fc41d4237cc8a9cadcf70ace42c1c166a35bb05e/libs/common/src/vault/models/view/cipher.view.ts#L36) to `false` rather than `true`. Should this logic follow suit or is the deviation intentional?

crates/bitwarden-vault/src/cipher/cipher_client.rs
```diff
@@ -174,6 +181,68 @@ impl CiphersClient {
...
+
+    /// Get all ciphers from state and decrypt them to a list of [CipherView].
+    pub async fn list(&self) -> Result<Vec<CipherView>, GetCipherError> {
+        let key_store = self.client.internal.get_key_store();
+        let repository = self.get_repository()?;
+
+        list_ciphers(key_store, repository.as_ref()).await
+    }
+
+    /// Get [Cipher] by ID from state and decrypt it to a [CipherView].
```
nick-livefront: I did not know you could link to other entities in a comment with `[...]`, TIL!

crates/bitwarden-vault/src/cipher/create.rs
```diff
@@ -0,0 +1,304 @@
...
+            name: encrypted_cipher.name.to_string(),
+            notes: encrypted_cipher.notes.map(|n| n.to_string()),
+            fields: encrypted_cipher
+                .fields
+                .map(|f| f.into_iter().map(|f| f.into()).collect()),
+            password_history: encrypted_cipher
+                .password_history
+                .map(|ph| ph.into_iter().map(|ph| ph.into()).collect()),
+            attachments: None,
+            attachments2: encrypted_cipher.attachments.map(|a| {
```
nick-livefront: ❓ What is the difference between `attachments` and `attachments2`?

nick-livefront (commented):
Just a handful of questions from me. I'm following this and the implementation but I would say that my rust experience is a little lack luster on if there are improvements to be made on that thread. I will hold off on approving to hopefully solicit some more impactful feedback, but I can if I'm the gateway between moving this forward.

crates/bitwarden-vault/src/cipher/cipher.rs
gbubemismith: @nick-livefront I actually have that change locally, I was waiting to get a different PR merged in first. Moved to draft

crates/bitwarden-vault/src/cipher/create.rs
gbubemismith: On the server, `attachments` is deprecated. The main difference is `attachments2` has a structured value and `attachements` is just a string on the server

crates/bitwarden-vault/src/cipher/get_list.rs
```diff
@@ -0,0 +1,41 @@
...
+
+    Ok(store.decrypt(&cipher)?)
+}
+
+pub(super) async fn list_ciphers(
+    store: &KeyStore<KeyIds>,
+    repository: &dyn Repository<Cipher>,
+) -> Result<Vec<CipherView>, GetCipherError> {
+    let ciphers = repository.list().await?;
+    let views = store.decrypt_list(&ciphers)?;
```
nikwithak: Would it be better to use `decrypt_list_failures` here to avoid blocking when only some of them fail to decrypt?

crates/bitwarden-vault/src/cipher/cipher.rs
```diff
@@ -788,6 +797,76 @@ impl From<bitwarden_api_api::models::CipherRepromptType> for CipherRepromptType
...
+            collection_ids: vec![], // CipherResponseModel doesn't include collection_ids
+            name: require!(EncString::try_from_optional(cipher.name)?),
+            notes: EncString::try_from_optional(cipher.notes)?,
+            r#type: require!(cipher.r#type).into(),
+            login: cipher.login.map(|l| (*l).try_into()).transpose()?,
+            identity: cipher.identity.map(|i| (*i).try_into()).transpose()?,
+            card: cipher.card.map(|c| (*c).try_into()).transpose()?,
+            secure_note: cipher.secure_note.map(|s| (*s).try_into()).transpose()?,
+            // TODO: add ssh_key
+            ssh_key: None,
```
nikwithak: 🤔 Do you plan for these changes to be in use / consumed from the client before the cipher versioning changes in #433 and the associated server changes go in? We might plan to use the `Cipher::populate_cipher_types` introduced in that PR for pulling these from `data` directly.

crates/bitwarden-vault/src/cipher/create.rs
```diff
@@ -0,0 +1,304 @@
...
+    #[error(transparent)]
+    Crypto(#[from] CryptoError),
+    #[error(transparent)]
+    Api(#[from] ApiError),
+    #[error(transparent)]
+    VaultParse(#[from] VaultParseError),
+    #[error(transparent)]
+    MissingField(#[from] MissingFieldError),
+    #[error(transparent)]
+    RepositoryError(#[from] RepositoryError),
```
nikwithak: ⛏️ You might need to remove the `Error` naming from the enum variant, per #430

crates/bitwarden-vault/src/cipher/get_list.rs
nikwithak: Thanks for adding `list_ciphers_with_failures`! Do you think we still need the `list_ciphers` for any use cases, or could we remove it altogether?

crates/bitwarden-vault/src/cipher/cipher.rs
```diff
@@ -692,6 +692,15 @@ impl Decryptable<KeyIds, SymmetricKeyId, CipherListView> for Cipher {
...
 
+#[cfg(feature = "wasm")]
+impl wasm_bindgen::__rt::VectorIntoJsValue for CipherView {
+    fn vector_into_jsvalue(
+        vector: wasm_bindgen::__rt::std::boxed::Box<[Self]>,
+    ) -> wasm_bindgen::JsValue {
+        wasm_bindgen::__rt::js_value_vector_into_jsvalue(vector)
+    }
+}
+
```
coroiu, line 695: question: interesting, haven't seen this before. Did you have issues returning `Vec<CipherView` without this?

crates/bitwarden-vault/src/cipher/get_list.rs
coroiu, line 1: question: why are putting this code in a separate file? Is it because we want to keep `ciphers_client.rs` from growing? Does seem like we're gaining much by it?

coroiu: If you want to make `ciphers_client` smaller maybe you can have `impl CiphersClient` blocks in multiple files instead, e.g. 
- `ciphers_client_create.rs`
- `ciphers_client_get_list.rs`

or maybe even add them in a subfolder to make it easier to access the internal fields of `CiphersClient`

- `ciphers_client/mod.rs`
- `ciphers_client/create.rs`
- `ciphers_client/get_list.rs`

crates/bitwarden-vault/src/cipher/cipher.rs
dani-garcia: This is because tsify isn't adding this derive, which wasm-bindgen expects for types returning lists. Oscar and I had the same problem with folders:
https://github.com/bitwarden/sdk-internal/blob/main/crates/bitwarden-vault/src/folder/folder_models.rs#L44-L51

Reported upstream:
https://github.com/madonoharu/tsify/issues/66

coroiu: Ah, nice, thanks!

crates/bitwarden-vault/src/cipher/get_list.rs
gbubemismith: Yeah, exactly that. I like the sub folders approach

gbubemismith: Changes here a0bca6b85e548b489788eac442cfa7f4dfd4d146

crates/bitwarden-vault/src/cipher/cipher_client/edit.rs
```diff
@@ -0,0 +1,353 @@
...
+    Repository(#[from] RepositoryError),
+    #[error(transparent)]
+    Uuid(#[from] uuid::Error),
+}
+
+async fn edit_cipher<R: Repository<Cipher> + ?Sized>(
+    key_store: &KeyStore<KeyIds>,
+    api_config: &bitwarden_api_api::apis::configuration::Configuration,
+    repository: &R,
+    cipher_id: &str,
```
Hinton: **suggestion:** Please use the type safe IDs.
```suggestion
    cipher_id: CipherId,
```

crates/bitwarden-vault/src/cipher/cipher_client/edit.rs
```diff
@@ -0,0 +1,353 @@
...
+    request: CipherAddEditRequest,
+) -> Result<CipherView, EditCipherError> {
+    repository
+        .get(cipher_id.to_owned())
+        .await?
+        .ok_or(ItemNotFoundError)?;
+
+    let cipher_request = key_store.encrypt(request)?;
```
Hinton: **suggestion:** This would be a good spot to calculate password history as this is a complex scenario currently handled in the UI. If we don't want to do this in this unit of work we should name this method `edit_no_history` so that a future `edit` function won't introduce unexpected behaviour.

crates/bitwarden-vault/src/cipher/cipher_client/create.rs
```diff
@@ -0,0 +1,323 @@
...
+}
+
+/// Request to add or edit a cipher.
+#[derive(Serialize, Deserialize, Debug)]
+#[serde(rename_all = "camelCase")]
+#[cfg_attr(feature = "uniffi", derive(uniffi::Record))]
+#[cfg_attr(feature = "wasm", derive(Tsify), tsify(into_wasm_abi, from_wasm_abi))]
+pub struct CipherAddEditRequest {
+    /// The cipher view data to be added or edited.
+    pub cipher: CipherView,
```
Hinton: **suggestion:** `CipherView` contains additional data that can't be set on the creation.

- ID: None
- Key: None
- CreatationDate, DeletedDate, RevisionDate, ArchivedDate: None

Which means that `CipherView` is probably a poor fit here.

Hinton (commented):
Nice!

Some high level observations.

crates/bitwarden-vault/src/cipher/get_list.rs
Hinton: @coroiu we need to update the folder behaviour because it also follows this pattern.

coroiu: Weird, I added [a comment](https://github.com/bitwarden/sdk-internal/commit/a0bca6b85e548b489788eac442cfa7f4dfd4d146#r166111279) but apparently outside of the PR somehow :shrug: 

Anyway, my suggestion is to merge the two functions, we don't need to have one function call another one. e.g:

```rs
// Remove all together
// async fn create_cipher<R: Repository<Cipher> + ?Sized>(

impl CiphersClient {
    /// Create a new [Cipher] and save it to the server.
    pub async fn create(
        &self,
        mut request: CipherAddEditRequest,
    ) -> Result<CipherView, CreateCipherError> {
        let config = self.client.internal.get_api_configurations().await;
        let repository = self.get_repository()?;

        request.encrypted_for = self.client.internal.get_user_id();

        let cipher_request = key_store.encrypt(request)?;
        let resp = ciphers_api::ciphers_post(api_config, Some(cipher_request))
        .await
        .map_err(ApiError::from)?;

        let cipher: Cipher = resp.try_into()?;

        repository
        .set(require!(cipher.id).to_string(), cipher.clone())
        .await?;

        Ok(key_store.decrypt(&cipher)?)
    }
}
```

crates/bitwarden-vault/src/cipher/cipher_client/edit.rs
Hinton: I fixed the folder instances in https://github.com/bitwarden/sdk-internal/pull/464. Seems they were missed in the initial migration.

crates/bitwarden-vault/src/cipher/get_list.rs
gbubemismith: @coroiu I tried merging the functions, but ran into repository access issues during testing. `CiphersClient::get_repository()` call requires full client state initialization, while the separate function makes it possible to inject `MemoryRepository directly::<Cipher>::default()`, is there a way to mock the repository when testing through the client?

coroiu: @gbubemismith I'm sorry, I hadn't noticed that you were simply using an already established pattern here. I think it was a bit unfair of me to request these changes in this PR like this. I think the work you've done splitting these functions into separate files has already improved the structure a lot and I think we can go ahead an merge as-is. 

I created an example https://github.com/bitwarden/sdk-internal/pull/467 which shows how `create` can be merged, but I don't like how the `Client` needs to be constructed with the `mock_server.address()` manually like this. Platform needs to make this process easier. I'll look into it and and we can revisit this pattern at a later date :)

crates/bitwarden-vault/src/cipher/cipher_client/create.rs
gbubemismith: That makes sense. 0f35f92a05d53ec06ca9b6cca4824cee44c1d8b7

crates/bitwarden-vault/src/cipher/cipher_client/edit.rs
gbubemismith: Added here 0f35f92a05d53ec06ca9b6cca4824cee44c1d8b7

crates/bitwarden-vault/src/cipher/cipher_client/create.rs
```diff
@@ -0,0 +1,328 @@
...
+    Api(#[from] ApiError),
+    #[error(transparent)]
+    VaultParse(#[from] VaultParseError),
+    #[error(transparent)]
+    MissingField(#[from] MissingFieldError),
+    #[error(transparent)]
+    Repository(#[from] RepositoryError),
+}
+
+/// Request to add or edit a cipher.
```
Hinton:
```suggestion
/// Request to add a cipher.
```

crates/bitwarden-vault/src/cipher/cipher_client/create.rs
```diff
@@ -0,0 +1,328 @@
...
+#[derive(Serialize, Deserialize, Debug)]
+#[serde(rename_all = "camelCase")]
+#[cfg_attr(feature = "uniffi", derive(uniffi::Record))]
+#[cfg_attr(feature = "wasm", derive(Tsify), tsify(into_wasm_abi, from_wasm_abi))]
+pub struct CipherCreateRequest {
+    pub encrypted_for: Option<UserId>,
```
Hinton: **question:** Do we want to document `encrypted_for` at least? Since it might not be completely obvious.

crates/bitwarden-vault/src/cipher/cipher_client/create.rs
```diff
@@ -0,0 +1,328 @@
...
+    pub name: String,
+    pub notes: Option<String>,
+    pub r#type: CipherType,
+    pub favorite: bool,
+    pub reprompt: CipherRepromptType,
+    pub login: Option<LoginView>,
+    pub identity: Option<IdentityView>,
+    pub card: Option<CardView>,
+    pub secure_note: Option<SecureNoteView>,
+    pub ssh_key: Option<SshKeyView>,
```
Hinton: I wonder if now is a good time to just extract a proper enum of the cipher type rather than keeping this flattened structure.

Originally we did it this way in an effort to mirror the existing implementations as much as possible but now that most of the cipher logic has moved to the SDK it would be nice to make the interface more idiomatic.

```rust
pub enum CipherViewType {
    Login(LoginView),
    SecureNote(SecureNoteView),
    Card(CardView),
    Identity(IdentityView),
    SshKey(SshKeyView),
}
```

crates/bitwarden-vault/src/cipher/cipher_client/create.rs
```diff
@@ -0,0 +1,328 @@
...
+    pub fields: Option<Vec<FieldView>>,
```
Hinton: **suggestion:** Maybe we should make this non optional?

```suggestion
    pub fields: Vec<FieldView>,
```

crates/bitwarden-vault/src/cipher/cipher_client/create.rs
```diff
@@ -0,0 +1,328 @@
...
+        key: SymmetricKeyId,
+    ) -> Result<CipherRequestModel, CryptoError> {
+        let cipher_request = CipherRequestModel {
+            encrypted_for: self.encrypted_for.map(|id| id.into()),
+            r#type: Some(self.r#type.into()),
+            organization_id: self.organization_id.map(|id| id.to_string()),
+            folder_id: self.folder_id.map(|id| id.to_string()),
+            favorite: Some(self.favorite),
+            reprompt: Some(self.reprompt.into()),
+            key: None,
```
Hinton: **issue:** I believe we should support cipher key encryption here ideally?

crates/bitwarden-vault/src/cipher/cipher_client/create.rs
```diff
@@ -0,0 +1,328 @@
...
+                totp: None,
+                autofill_on_page_load: None,
+                fido2_credentials: None,
+            }),
+            identity: None,
+            card: None,
+            secure_note: None,
+            ssh_key: None,
+            fields: None,
+        }
```
Hinton, line 310: **suggestion:** We could derive `Default` for `CipherCreateRequest` and simplify this into


```suggestion
        CipherCreateRequest {
            name: "Test Login".to_string(),
            notes: Some("Test notes".to_string()),
            r#type: CipherType::Login,
            login: Some(LoginView {
                username: Some("test@example.com".to_string()),
                password: Some("password123".to_string()),
                password_revision_date: None,
                uris: None,
                totp: None,
                autofill_on_page_load: None,
                fido2_credentials: None,
            }),
            ..Default::default()
        }
```

crates/bitwarden-vault/src/cipher/cipher_client/create.rs
```diff
@@ -0,0 +1,328 @@
...
+    use wiremock::{matchers, Mock, Request, ResponseTemplate};
+
+    use super::*;
+    use crate::{CipherId, CipherRepromptType, CipherType, LoginView};
+
+    const TEST_CIPHER_ID: &str = "5faa9684-c793-4a2d-8a12-b33900187097";
+
+    fn generate_test_cipher_create_request() -> CipherCreateRequest {
+        CipherCreateRequest {
+            encrypted_for: None,
```
Hinton: **question:** I think `encrypted_for` is supposed to be required and we should error if it's missing?

crates/bitwarden-vault/src/cipher/cipher_client/edit.rs
```diff
@@ -0,0 +1,623 @@
...
+}
+
+/// Request to edit a cipher.
+#[derive(Serialize, Deserialize, Debug)]
+#[serde(rename_all = "camelCase")]
+#[cfg_attr(feature = "uniffi", derive(uniffi::Record))]
+#[cfg_attr(feature = "wasm", derive(Tsify), tsify(into_wasm_abi, from_wasm_abi))]
+pub struct CipherEditRequest {
+    /// The cipher view data to be added or edited.
+    pub cipher: CipherView,
```
Hinton: **question:** I wonder if `CipherView` is a poor fit here as well? Not everything is editable by the UI such as `key`, `organization_use_totp`, `edit`, `permissions`, `view_password`, `attachments`, `password_history`, `creation_date`, `deleted_date`, `revision_date` and `archived_date`?

You should probably only be able to modify these fields through separate methods?

crates/bitwarden-vault/src/cipher/cipher_client/edit.rs
```diff
@@ -0,0 +1,623 @@
...
+    /// The user ID for whom this cipher is encrypted (internal use only).
+    pub encrypted_for: Option<UserId>,
```
Hinton: **question:** Should this be required?

crates/bitwarden-vault/src/cipher/cipher_client/edit.rs
```diff
@@ -0,0 +1,623 @@
...
+}
+
+impl CipherEditRequest {
+    pub fn update_password_history(&mut self, original_cipher: &CipherView) {
```
Hinton: **question:** I don't think we want this method to be public?

```suggestion
   fn update_password_history(&mut self, original_cipher: &CipherView) {
```

crates/bitwarden-vault/src/cipher/cipher_client/get.rs
```diff
@@ -0,0 +1,82 @@
...
+}
+
+impl CiphersClient {
+    /// Get all ciphers from state and decrypt them to a list of [CipherView].
+    pub async fn list(&self) -> Result<Vec<CipherView>, GetCipherError> {
+        let key_store = self.client.internal.get_key_store();
+        let repository = self.get_repository()?;
+
+        list_ciphers(key_store, repository.as_ref()).await
+    }
```
Hinton: **question:** Do we need this or will we always use `list_with_failures`?

crates/bitwarden-vault/src/cipher/cipher_client/mod.rs
```diff
@@ -11,6 +14,10 @@ use crate::{
     DecryptError, EncryptError, Fido2CredentialFullView,
 };
 
+pub mod create;
+pub mod edit;
+pub mod get;
```
Hinton: **suggestion:** Can we please keep these private and just re-export if we need anything outside?

```suggestion
mod create;
mod edit;
mod get;
```

crates/bitwarden-vault/src/cipher/cipher.rs
```diff
@@ -793,6 +802,75 @@ impl From<bitwarden_api_api::models::CipherRepromptType> for CipherRepromptType
...
+impl TryFrom<CipherResponseModel> for Cipher {
+    type Error = VaultParseError;
+
+    fn try_from(cipher: CipherResponseModel) -> Result<Self, Self::Error> {
+        Ok(Self {
+            id: cipher.id.map(CipherId::new),
+            organization_id: cipher.organization_id.map(OrganizationId::new),
+            folder_id: cipher.folder_id.map(FolderId::new),
+            collection_ids: vec![], // CipherResponseModel doesn't include collection_ids
+            name: require!(EncString::try_from_optional(cipher.name)?),
```
Hinton: I think we can write it as this which provides a slightly nicer error if the field is missing

```suggestion
            require!(cipher.name).parse()?
```

crates/bitwarden-vault/src/password_history.rs
```diff
@@ -78,3 +78,44 @@ impl TryFrom<CipherPasswordHistoryModel> for PasswordHistory {
...
+        }
+    }
+
+    pub fn into_history_entry(self) -> PasswordHistoryView {
+        PasswordHistoryView {
+            password: self.display_name,
+            last_used_date: self.changed_at,
+        }
+    }
+}
```
Hinton: **question:** Why create a new struct for this rather than using `PasswordHistory`?

crates/bitwarden-vault/src/cipher/cipher_client/edit.rs
```diff
@@ -0,0 +1,623 @@
...
+            changes.extend(login_changes);
+        }
+
+        changes.extend(self.detect_hidden_field_changes(original_cipher));
+
+        for change in changes.into_iter().rev() {
+            history.insert(0, change.into_history_entry());
+        }
+
+        Self::limit_history_length(&mut history);
```
Hinton: **suggestion:** We can get rid of the mutability and write this more idiomatic using functional programming.

- Use `chain` to combine login passwords and fields.
- Avoid the need of an intermediary vector `changes` by just using the iterator directly.
- Remove mutable on history by doing everything within an iterator.
- Use `take` to just get the correct number of items from the iterato.

```suggestion
        let changes = self
            .detect_login_password_changes(original_cipher)
            .unwrap_or_default()
            .into_iter()
            .chain(self.detect_hidden_field_changes(original_cipher));

        let history: Vec<_> = changes
            .rev()
            .map(|change| change.into_history_entry())
            .chain(original_cipher.password_history.iter().flatten().cloned())
            .take(MAX_PASSWORD_HISTORY_ENTRIES)
            .collect();
```

crates/bitwarden-vault/src/cipher/cipher_client/edit.rs
```diff
@@ -0,0 +1,623 @@
...
+                        (FieldType::Hidden, Some(name), Some(value))
+                            if !name.is_empty() && !value.is_empty() =>
+                        {
+                            Some((name.clone(), value.clone()))
+                        }
+                        _ => None,
+                    })
+                    .collect()
+            })
+            .unwrap_or_default()
```
Hinton:
```suggestion
        fields
            .iter()
            .flatten()
            .filter_map(|f| match (&f.r#type, &f.name, &f.value) {
                (FieldType::Hidden, Some(name), Some(value))
                    if !name.is_empty() && !value.is_empty() =>
                {
                    Some((name.clone(), value.clone()))
                }
                _ => None,
            })
            .collect()
```

crates/bitwarden-vault/src/cipher/cipher_client/edit.rs
```diff
@@ -0,0 +1,623 @@
...
+
+        Self::limit_history_length(&mut history);
+
+        self.cipher.password_history = (!history.is_empty()).then_some(history);
+    }
+
+    fn detect_login_password_changes(
+        &mut self,
+        original_cipher: &CipherView,
+    ) -> Option<Vec<PasswordChange>> {
```
Hinton: **suggestion:** If we always return a vec it would simplify some consuming logic.

crates/bitwarden-vault/src/cipher/cipher_client/edit.rs
Hinton: **suggestion:** Should we test this more extensively? Password history is a somewhat complex portion that needs to work really good.

Hinton:
Should we also expose these to mobile? I think it would be useful to get their perspective on the bindings.

crates/bitwarden-vault/src/cipher/cipher_client/create.rs
nikwithak: It's far from being working or review-ready, but I started down this path as part of the cipher versioning work a few weeks ago: https://github.com/bitwarden/sdk-internal/blob/vault/wip__cipher_view_data/crates/bitwarden-vault/src/cipher/cipher_view_data.rs

crates/bitwarden-vault/src/cipher/cipher_client/create.rs
nikwithak: On the server-side, we verify that `encrypted_for` is the same as the current user, but it doesn't error if it's not provided - so while we should generally include it in our requests, it's not blocking if it goes through as `None` since the server accepts a null value.

That said,[ it *is* getting set ](https://github.com/bitwarden/sdk-internal/pull/455/files#diff-1807043f0e96ea095924fef9f11d81a1ffd6a78d55e72d422336a4f44ddb34c2R171-R172)for all requests through the client, and the `create_cipher` function that it gets passed too won't get called outside of the that path.

shane-melton: @nikwithak The server allows it to be optional to provide backwards compatibility with clients that didn't know about the `encrypted_for`. We do ultimately want to make it required and since this is the first time we're introducing creation logic it to the SDK we should be safe to make it required here.

crates/bitwarden-vault/src/cipher/cipher_client/create.rs
shane-melton: Yeah, we need to add support for cipher key encryption here. We should be able to mimic the [implementation for CipherView](https://github.com/bitwarden/sdk-internal/blob/31b18f02ec8f5713133ea67d42993c12c5923782/crates/bitwarden-vault/src/cipher/cipher.rs#L296-L297). We'll also need to generate the cipher key at the client level in `create_cipher` method. ([see here for an example](https://github.com/bitwarden/sdk-internal/blob/a0bca6b85e548b489788eac442cfa7f4dfd4d146/crates/bitwarden-vault/src/cipher/cipher_client/mod.rs#L38-L49))

Admittedly, it isn't ideal that we need to duplicate this logic in multiple places; but I'm not certain how we can re-use the logic with different structs.

crates/bitwarden-vault/src/cipher/cipher_client/create.rs
```diff
@@ -0,0 +1,328 @@
...
+    pub fields: Option<Vec<FieldView>>,
+}
+
+impl CompositeEncryptable<KeyIds, SymmetricKeyId, CipherRequestModel> for CipherCreateRequest {
+    fn encrypt_composite(
+        &self,
+        ctx: &mut KeyStoreContext<KeyIds>,
+        key: SymmetricKeyId,
+    ) -> Result<CipherRequestModel, CryptoError> {
+        let cipher_request = CipherRequestModel {
```
shane-melton, line 121: ⚠️ We are not generating the URI checksums like we do when encrypting a `CipherView`. 
https://github.com/bitwarden/sdk-internal/blob/31b18f02ec8f5713133ea67d42993c12c5923782/crates/bitwarden-vault/src/cipher/cipher.rs#L305

shane-melton (changes requested):

crates/bitwarden-vault/src/cipher/cipher_client/create.rs
Hinton: The problem is it's featured flagged and feature flags aren't allowed this deep. If we always used cipher flags this would be trivial but we don't.

crates/bitwarden-vault/src/cipher/cipher_client/create.rs
nikwithak: I added a commentdoc here: https://github.com/bitwarden/sdk-internal/pull/455/files#diff-1807043f0e96ea095924fef9f11d81a1ffd6a78d55e72d422336a4f44ddb34c2R46-R48

crates/bitwarden-vault/src/cipher/cipher_client/create.rs
nikwithak: In order to derive `Default` I had to add `Default` implementations for [CipherType](https://github.com/bitwarden/sdk-internal/pull/455/files#diff-a09f3c6f83362c95e3eed722783dbbbae136f953a19ddc90ffa15e03c1d52cf5R70-R71)(defaults to `Login`) and [CipherRepromptType](https://github.com/bitwarden/sdk-internal/pull/455/files#diff-a09f3c6f83362c95e3eed722783dbbbae136f953a19ddc90ffa15e03c1d52cf5R84-R86) (defaults to `None`). I don't think this is a concern, but want to call it out. The alternative is a manual `impl Default for CipherCreateRequest`, which I can also do if we don't want defaults for the other two.

crates/bitwarden-vault/src/cipher/cipher_client/edit.rs
nikwithak: Incorporated your suggestions [here](https://github.com/bitwarden/sdk-internal/pull/455/files#diff-a00d3aa62970621e7b6048e02038aaa4dfa5124ea05915df90677d23300db9d9R62-R73)

crates/bitwarden-vault/src/cipher/cipher_view_type.rs
```diff
@@ -0,0 +1,164 @@
...
+
+    fn as_secure_note_view(&self) -> Option<&SecureNoteView> {
+        self.as_ref()
+            .map(|inner| inner.as_secure_note_view())
+            .flatten()
+    }
+
+    fn as_ssh_key_view(&self) -> Option<&SshKeyView> {
+        self.as_ref().map(|inner| inner.as_ssh_key_view()).flatten()
+    }
```
nikwithak: Just some helper functions to make ergonomics easier with `Option<CipherViewType>`

crates/bitwarden-vault/src/cipher/cipher_client/create.rs
nikwithak: I added this to the Create/Edit request types here: [223cfe1](https://github.com/bitwarden/sdk-internal/pull/455/commits/223cfe1756230a1e57f8952d24d3824e57900c39) and [3f07956](https://github.com/bitwarden/sdk-internal/pull/455/commits/3f079569dbb6dd8e42695bb1757aa78fd4764f05) and

crates/bitwarden-vault/src/cipher/cipher_client/create.rs
nikwithak: I added this logic to `CiphersClient::create()`, where we *do* have access to the feature flag. The actual encryption logic just checks for the existence of a key (following the CipherView example). 

[f8dfeea](github.com/bitwarden/sdk-internal/pull/455/commits/f8dfeeaa5e968279e2616efc44ed0987b2d1c8cd)

crates/bitwarden-vault/src/cipher/cipher_client/create.rs
nikwithak: I ended up removing this from the client-facing `CipherCreateRequest`, after running it by @shane-melton , and it's now set automatically based on the currently logged-in user when calling `create_cipher()`. 

[bde6dde](https://github.com/bitwarden/sdk-internal/pull/455/commits/bde6ddeb32dd8a9ac5fd97e05273e087e0905822)

crates/bitwarden-vault/src/cipher/cipher_client/edit.rs
nikwithak: I refactored this struct pretty heavily, so now it has all the data fields directly instead of just wrapping a `CipherView` - along with a `From<CipherView>` implementation for ease of use. I ended up duplicating some of the bits of logic in CipherView, but I think that was inevitable. 

[4593cd3](https://github.com/bitwarden/sdk-internal/pull/455/commits/4593cd3d7080a0bdd7a4667402301779faa92b28)

crates/bitwarden-vault/src/cipher/cipher_client/edit.rs
nikwithak: Addressed here, using the same approach for  `CipherCreateRequest` and `CipherEditRequest`: https://github.com/bitwarden/sdk-internal/pull/455#discussion_r2380624380

crates/bitwarden-vault/src/cipher/cipher_client/create.rs
```diff
@@ -0,0 +1,387 @@
...
+}
+
+/// Request to add a cipher.
+#[derive(Serialize, Deserialize, Debug, Default)]
+#[serde(rename_all = "camelCase")]
+#[cfg_attr(feature = "uniffi", derive(uniffi::Record))]
+#[cfg_attr(feature = "wasm", derive(Tsify), tsify(into_wasm_abi, from_wasm_abi))]
+pub struct CipherCreateRequest {
+    /// The ID of the user that is encrypting the cipher - this should always match the user
+    /// calling the API.
```
Hinton: This was probably for the now removed `encrypted_for`?
```suggestion
```

crates/bitwarden-vault/src/cipher/cipher_client/create.rs
Hinton: Yea that's a bummer. I'm a bit less happy with that solution, does it make semantic sense that `CipherType` has a default value? I think `CipherRepromptType` is fine.

With the introduction of the. enum we've now reduced the field count and we might not need default any longer?

Another completely different way of solving this would be to just have a test function that returns a `CipherCreateRequest`, i.e.
```rust
fn default_create_request() {
  CipherCreateRequest. { .... }
}

CipherCreateRequest {
  name: "Overriden".to_string(),
  ..default_create_requst()
}
```

crates/bitwarden-vault/src/cipher/cipher_client/edit.rs
Hinton: Yes duplication is probably inevitable. Once we've updated the clients to use this we should re-visit `CipherView` and clean it up since we don't need a way to encrypt it, simplify it since we know this now represents a created cipher.

crates/bitwarden-vault/src/cipher/cipher_client/create.rs
```diff
@@ -0,0 +1,387 @@
...
+                .map(|s| s.encrypt_composite(ctx, cipher_key))
+                .transpose()?
+                .map(|s| Box::new(s.into())),
+            ssh_key: self
+                .type_data
+                .as_ssh_key_view()
+                .as_ref()
+                .map(|s| s.encrypt_composite(ctx, cipher_key))
+                .transpose()?
+                .map(|s| Box::new(s.into())),
```
Hinton: Messy, but I suppose we're stuck with it until the cipher is just a blob and we can serialize this manually somehow.

crates/bitwarden-vault/src/cipher/cipher_client/create.rs
```diff
@@ -0,0 +1,387 @@
...
+    pub organization_id: Option<OrganizationId>,
+    pub folder_id: Option<FolderId>,
+    pub name: String,
+    pub notes: Option<String>,
+    pub r#type: CipherType,
+    pub favorite: bool,
+    pub reprompt: CipherRepromptType,
+    pub type_data: Option<CipherViewType>,
+    pub fields: Vec<FieldView>,
+    pub key: Option<EncString>,
```
Hinton: **question:** Should we document this? "INTERNAL, do not set from clients."

crates/bitwarden-vault/src/cipher/cipher_client/create.rs
```diff
@@ -0,0 +1,387 @@
...
+
+impl CipherCreateRequest {
+    /// Generate a new key for the cipher, re-encrypting internal data, if necessary, and stores the
+    /// encrypted key to the cipher data.
+    pub fn generate_cipher_key(
+        &mut self,
+        ctx: &mut KeyStoreContext<KeyIds>,
+        key: SymmetricKeyId,
+    ) -> Result<(), CryptoError> {
+        let old_key = Cipher::decrypt_cipher_key(ctx, key, &self.key)?;
```
Hinton, line 85: **question:** Since we're creating a cipher key will never exist right?

crates/bitwarden-vault/src/cipher/cipher_client/create.rs
```diff
@@ -0,0 +1,387 @@
...
+
+        const NEW_KEY_ID: SymmetricKeyId = SymmetricKeyId::Local("new_cipher_key");
+
+        let new_key = ctx.generate_symmetric_key(NEW_KEY_ID)?;
+        self.type_data
+            .as_login_view_mut()
+            .map(|l| l.reencrypt_fido2_credentials(ctx, old_key, new_key))
+            .transpose()?;
```
Hinton: **thought:** This raises a good question, how would we create a new cipher with a passkey? They are intentionally encrypted but would that even be viable generally?

I think temporarily we can pass in already encrypted data in the web clients, but this might not work on mobile. Long term having a different flow for creating items with passkeys might be required? CC @coroiu  since you have more insight into our sdk passkey flows.

crates/bitwarden-vault/src/cipher/cipher_client/create.rs
```diff
@@ -0,0 +1,387 @@
...
+    pub reprompt: CipherRepromptType,
+    pub type_data: Option<CipherViewType>,
+    pub fields: Vec<FieldView>,
+    pub key: Option<EncString>,
+}
+
+impl CipherCreateRequest {
+    /// Generate a new key for the cipher, re-encrypting internal data, if necessary, and stores the
+    /// encrypted key to the cipher data.
+    pub fn generate_cipher_key(
```
Hinton:
```suggestion
   fn generate_cipher_key(
```

I think this would be internal to this create?

crates/bitwarden-vault/src/cipher/cipher_client/create.rs
```diff
@@ -0,0 +1,387 @@
...
+        let cipher_key = Cipher::decrypt_cipher_key(ctx, key, &self.key)?;
+
+        let cipher_request = CipherRequestModel {
+            encrypted_for: None,
+            r#type: Some(self.r#type.into()),
+            organization_id: self.organization_id.map(|id| id.to_string()),
+            folder_id: self.folder_id.map(|id| id.to_string()),
+            favorite: Some(self.favorite),
+            reprompt: Some(self.reprompt.into()),
+            key: None,
```
Hinton: **issue:** This should be the newly created cipher key right? Encrypted by the org or user key.

crates/bitwarden-vault/src/cipher/cipher_client/edit.rs
```diff
@@ -0,0 +1,718 @@
...
+
+/// Request to edit a cipher.
+#[derive(Clone, Serialize, Deserialize, Debug)]
+#[serde(rename_all = "camelCase")]
+#[cfg_attr(feature = "uniffi", derive(uniffi::Record))]
+#[cfg_attr(feature = "wasm", derive(Tsify), tsify(into_wasm_abi, from_wasm_abi))]
+pub struct CipherEditRequest {
+    pub id: CipherId,
+
+    pub r#type: CipherType,
```
Hinton: **question**: Do we need both `type` and `type_data`?

crates/bitwarden-vault/src/cipher/cipher_client/edit.rs
```diff
@@ -0,0 +1,718 @@
...
+            notes: value.notes,
+            fields: value.fields.unwrap_or_default(),
+            password_history: value.password_history,
+            attachments: value.attachments,
+            type_data,
+            revision_date: value.revision_date,
+            archived_date: value.archived_date,
+        })
+    }
+}
```
Hinton, line 104: **question:** It seems a bit weird going from `CipherView` to `CipherEditRequest`?

crates/bitwarden-vault/src/cipher/cipher_client/edit.rs
```diff
@@ -0,0 +1,718 @@
...
+                if current_value != Some(&original_value) {
+                    Some(PasswordChange::new_field(&field_name, &original_value))
+                } else {
+                    None
+                }
+            })
+            .collect()
+    }
+
+    fn extract_hidden_fields(fields: &Vec<FieldView>) -> HashMap<String, String> {
```
Hinton: **suggestion:** Use a slice rather than a reference to a vector. It's more generic.
```suggestion
    fn extract_hidden_fields(fields: &[FieldView]) -> HashMap<String, String> {
```

crates/bitwarden-vault/src/cipher/cipher_client/edit.rs
```diff
@@ -0,0 +1,718 @@
...
+
+        self.password_history = (!history.is_empty()).then_some(history);
+    }
+
+    fn detect_login_password_changes(
+        &mut self,
+        original_cipher: &CipherView,
+    ) -> Vec<PasswordChange> {
+        if self.r#type != CipherType::Login || original_cipher.r#type != CipherType::Login {
+            return Default::default();
```
Hinton: **suggestion:** I think returning `vec![]` explicitly improves clarity. (same for the other instances in this function)

crates/bitwarden-vault/src/cipher/cipher_client/edit.rs
```diff
@@ -0,0 +1,718 @@
...
+        } else {
+            // Password changed - update revision date and track change
+            current_login.password_revision_date = Some(Utc::now());
+            vec![PasswordChange::new_password(original_password)]
+        }
+    }
+
+    fn detect_hidden_field_changes(&self, original_cipher: &CipherView) -> Vec<PasswordChange> {
+        let original_fields =
+            Self::extract_hidden_fields(original_cipher.fields.as_ref().unwrap_or(&vec![]));
```
Hinton: With the change to accept a slice you can update this to use default.
```suggestion
            Self::extract_hidden_fields(original_cipher.fields.as_ref().unwrap_or_default();
```

Hinton (commented):
Overall good, there are a couple of things we should consider.

crates/bitwarden-vault/src/cipher/cipher_client/create.rs
nikwithak: It shouldn't, in which case  `decrypt_cipher_key` will return the existing key. I included this just in case we *do* get a `CipherCreateRequest` that already has a key, which would imply the Fido2 credentials were already encrypted with that key. Happy to remove if we're making the assumption that won't happen.

crates/bitwarden-vault/src/cipher/cipher_client/edit.rs
nikwithak: The intent here was that a client will have a CipherView locally, make changes to it, and then go to save those changes by converting it into an Edit request.

Something along the lines of:

```rust
// Start with a decrypted cipher
let cipher_view: CipherView = ciphers_client.decrypt(cipher)?;

// Make some local changes to the CipherView
cipher_view.name = "New Cipher Name";
cipher_view.login = Some(LoginView {
    username: Some("new_login".to_string()),
    password: Some("password".to_string()),
    ..Default::default(),
});

// Send the changes upstream
ciphers_client.edit(cipher_view.into());
```

crates/bitwarden-vault/src/cipher/cipher_client/create.rs
nikwithak: You're right - with the enum (and removal of `r#type`) we no longer need a default, so I was able to remove that derive from `CipherType`.

crates/bitwarden-vault/src/cipher/cipher_client/create.rs
nikwithak: Great catch! Yes, updated to actually pass the cipher key along.

crates/bitwarden-vault/src/cipher/cipher_client/edit.rs
nikwithak: Not anymore - removed `type` and now pull this data from `CipherViewType`

crates/bitwarden-vault/src/cipher/cipher_client/create.rs
nikwithak: I added a comment - I tried setting it as non-pub, but that doesn't affect the TS types that are generated, and didn't seem to add much benefit in this particular use case.

crates/bitwarden-vault/src/cipher/cipher_client/create.rs
nikwithak: I initially fixed this in edit.rs, but missed the create.rs instance. Updated [here](https://github.com/bitwarden/sdk-internal/pull/455/files#diff-1807043f0e96ea095924fef9f11d81a1ffd6a78d55e72d422336a4f44ddb34c2R98).

crates/bitwarden-vault/src/cipher/cipher_client/create.rs
coroiu: On `clients` we don't create ciphers with passkeys. The cipher is created first by the popout, which returns the new `cipherId` to `background` which then reads out the cipher and attaches a new passkey to it. It's a bad way of doing it but I think it saves us from having to worry about creation with passkeys in this PR.

`mobile` on the other hand does create ciphers with passkeys but creation and encryption is encapsulated in the SDK and could use the `Repository<Cipher>` instead of going through the `CipherClient`

coroiu (dismissed):

crates/bitwarden-vault/src/cipher/cipher_client/get.rs
```diff
@@ -0,0 +1,65 @@
...
+        successes,
+        failures: failures.into_iter().cloned().collect(),
+    })
+}
+
+impl CiphersClient {
+    /// Get all ciphers from state and decrypt them, returning both successes and failures.
+    /// This method will not fail when some ciphers fail to decrypt, allowing for graceful
+    /// handling of corrupted or problematic cipher data.
+    pub async fn list_with_failures(&self) -> Result<DecryptCipherListResult, GetCipherError> {
```
Hinton: **suggestion (non blocking):** I think we can just call this `list`?

crates/bitwarden-vault/src/cipher/cipher_client/edit.rs
Hinton: Hmm, that does seem like a nicer rust interface although I'd caution against modifying a `CipherView` ever, and you should probably convert to a `CipherEditRequest` before modifying it. The conversion to `CipherEditRequest` is lossly and might not save passkeys for example.

crates/bitwarden-vault/src/cipher/cipher_client/create.rs
Hinton: I think with the new APIs you would need to create the cipher first before adding passkeys. We'll see what troubles appears when we start migrating mobile to use it.

crates/bitwarden-vault/src/cipher/cipher_client/create.rs
```diff
@@ -0,0 +1,408 @@
...
+#[cfg_attr(feature = "uniffi", derive(uniffi::Record))]
+#[cfg_attr(feature = "wasm", derive(Tsify), tsify(into_wasm_abi, from_wasm_abi))]
+pub struct CipherCreateRequest {
+    pub organization_id: Option<OrganizationId>,
+    pub folder_id: Option<FolderId>,
+    pub name: String,
+    pub notes: Option<String>,
+    pub favorite: bool,
+    pub reprompt: CipherRepromptType,
+    pub type_data: Option<CipherViewType>,
```
Hinton: **question:** Is this actually optional? I don't think you can create ciphers without a type.
```suggestion
    pub type_data: CipherViewType,
```

Hinton: **question (non blocking):** Should we rename this to `type`?

crates/bitwarden-vault/src/cipher/cipher_client/edit.rs
```diff
@@ -0,0 +1,721 @@
...
+    pub id: CipherId,
+
+    pub organization_id: Option<OrganizationId>,
+    pub folder_id: Option<FolderId>,
+    pub favorite: bool,
+    pub reprompt: CipherRepromptType,
+    pub name: String,
+    pub notes: Option<String>,
+    pub fields: Vec<FieldView>,
+    pub password_history: Option<Vec<PasswordHistoryView>>,
```
Hinton: **question:** I think we can drop this field? It should be calculated by the SDK.

crates/bitwarden-vault/src/cipher/cipher_client/edit.rs
```diff
@@ -0,0 +1,721 @@
...
+    pub attachments: Option<Vec<AttachmentView>>,
```
Hinton: I want to say we should drop this as well and refer people to dedicated functions but maybe we want to keep it around for now until we have those APIs?

Hinton (commented):
This is mostly looking great. There are some last comments, and some previous comments that were missed.

- https://github.com/bitwarden/sdk-internal/pull/455#discussion_r2371922551
- https://github.com/bitwarden/sdk-internal/pull/455#discussion_r2372148041

crates/bitwarden-vault/src/cipher/cipher.rs
shane-melton, line 1: 👍 Thanks for moving the uri checksum generation and fido2 re-encryption logic to `Login`!

crates/bitwarden-vault/src/cipher/cipher_view_type.rs
shane-melton, line 1: 👍 Very nice!

crates/bitwarden-vault/src/cipher/cipher_client/edit.rs
shane-melton: Agreed! We should be able to remove it at let it be calculated internally.

crates/bitwarden-vault/src/cipher/cipher_client/edit.rs
shane-melton: We should be able to remove it now and avoid the possibility of mistakenly trying to update attachments directly. At least for the TS clients, attachment creation/management is through its own methods that can be migrated to the SDK at a later time.

shane-melton (commented):
Changes really good from my perspective, nice work on this! I wanted to echo/agree with a few of Oscar's comments that I think we should address before merge.

crates/bitwarden-vault/src/cipher/cipher_client/edit.rs
nikwithak: Makes sense to me - is it worth keeping in place or do you think I should remove it?

Cargo.toml
nikwithak, line 1: Just formatting changes caused by the pre-commit hooks when merging from main.

crates/bitwarden-vault/Cargo.toml
nikwithak, line 1: Just formatting changes caused by the pre-commit hooks when merging from main.

crates/bitwarden-vault/src/cipher/cipher_client/edit.rs
nikwithak: I removed this from the client-facing request type in favor of calculating internally.

The previous way still had it calculated internally, it just squashed whatever the user passed in with the calculated value. Since it's not part of that request, I had to finagle it a bit to encrypt it separately and attach it to the server request: https://github.com/bitwarden/sdk-internal/pull/455/files#diff-a00d3aa62970621e7b6048e02038aaa4dfa5124ea05915df90677d23300db9d9R304-R311

crates/bitwarden-vault/src/password_history.rs
nikwithak: I'm not sure if there was a specific reason @gbubemismith originally created a new struct. I removed `PasswordChange` and just attached these functions to `PasswordHistoryView`: https://github.com/bitwarden/sdk-internal/pull/455/files#diff-9e2048cc067713f88b6ba27e33ca7a7806ac0eb39d6021c32c90d3ad900c9c84R81-R106

crates/bitwarden-vault/src/cipher/cipher_client/edit.rs
nikwithak: I added one more test to make sure we preserve the history when the password unchanged, and that the `last_used_date ` date values are preserved [here](https://github.com/bitwarden/sdk-internal/pull/455/files#diff-a00d3aa62970621e7b6048e02038aaa4dfa5124ea05915df90677d23300db9d9R617-R683).  Are there any additional / critical test cases you think are missing?

nikwithak:
Okay I think I got everything @Hinton @shane-melton  - a bit of a refactor with the recent commits, so please look carefully to make sure I didn't miss anything

Hinton:
I think you only updated edit and forgot about create?

crates/bitwarden-vault/src/cipher/cipher_client/edit.rs
```diff
@@ -331,10 +301,14 @@ async fn edit_cipher<R: Repository<Cipher> + ?Sized>(
     let original_cipher_view: CipherView = key_store.decrypt(&original_cipher)?;
 
     // Update password history
-    request.update_password_history(&original_cipher_view);
+    let password_history = request.generate_password_history(&original_cipher_view);
+    let enc_password_history =
+        password_history.encrypt_composite(&mut key_store.context(), request.key_identifier())?;
```
Hinton: I think if you want a nicer interface for this you can create a internal struct that encapsulates the outside data.

nikwithak: If we were re-using it elsewhere I think that would make sense, but since the need for it is isolated to this one function, I'd prefer to leave it as-is if you're fine with it.

nikwithak:
> I think you only updated edit and forgot about create?

Nice catch - I updated `type_data: Option<CipherViewType>` -> `r#type: CipherViewType`. I did away with the Default implementation as well in favor of just updating the test generator function, to avoid making assumptions about the CipherType.

I also added in the `wasm_bindgen` attributes to the `CiphersClient::edit` and `CiphersClient::create` with this update.

crates/bitwarden-vault/src/cipher/cipher_client/edit.rs
Hinton: The problem is that both `CipherEditRequest` and `CompositeEncryptable` are public. Which means that anyone (both internal to the vault crate but also externally) can call `encrypt_composite` on it. With the current logic you get back a `CipherEditRequest` which someone might believe they can post to our API. Without knowing that it lacks several required fields.

If it was just internal misuse I would be a bit less concerned but I could imagine someone calling encrypt on `CipherEditRequest`.

Encapsulating it in an internal struct prevents this sort of misuse. It's not the biggest thing but encoding this into the types prevents this from ever occurring.

crates/bitwarden-vault/src/password_history.rs
```diff
@@ -78,3 +78,29 @@ impl TryFrom<CipherPasswordHistoryModel> for PasswordHistory {
...
+impl From<PasswordHistory> for CipherPasswordHistoryModel {
+    fn from(history: PasswordHistory) -> Self {
+        Self {
+            password: history.password.to_string(),
+            last_used_date: history.last_used_date.to_rfc3339(),
+        }
+    }
+}
+
+#[allow(missing_docs)]
```
Hinton: **issue:** Should the methods here be public?

If yes: Please document them.
If no: Please lower the visibility and remove the allow(missing_docs)

I think no since we don't want people to manually modify the history.

crates/bitwarden-vault/src/cipher/cipher_view_type.rs
```diff
@@ -0,0 +1,166 @@
...
+            CipherViewType::Identity(_) => crate::CipherType::Identity,
+            CipherViewType::SecureNote(_) => crate::CipherType::SecureNote,
+            CipherViewType::SshKey(_) => crate::CipherType::SshKey,
+        }
+    }
+}
+
+/// Extension trait to provide type-safe accessors for the different cipher view types.
+#[allow(private_bounds)]
+pub trait CipherViewTypeExt
```
Hinton: **question:** Does this need to be public? If not you can lower the visibility and remove the allow. You can probably also get rid of the private trait.

Hinton (commented):
I think the two visbility issues are the only blockers now.,

crates/bitwarden-vault/src/cipher/cipher_client/edit.rs
nikwithak: Yes, makes perfect sense. I added a `CipherEditRequestInternal`, and modified the password history function to update it directly (rather than return the result). I also changed the password history tests to test on the output of `CipherEditRequestInternal::new` to make sure it gets set properly during the conversion.

crates/bitwarden-vault/src/cipher/cipher_view_type.rs
nikwithak: I did away with this trait altogether - since we no longer have `Option<CipherViewType>` in the structs, it sin't needed.  I just implemented those functions on `CipherViewType` directly.

crates/bitwarden-vault/src/password_history.rs
nikwithak: Updated to `pub(crate)`

crates/bitwarden-vault/src/cipher/cipher_client/edit.rs
nikwithak: @Hinton I took this approach for `CipherCreateRequest` too, since we had a prior discussion about flagging `key` as something the consumer shouldn't set - with this new approach, we have the same protections around that field on create.

crates/bitwarden-vault/src/cipher/cipher_client/create.rs
```diff
@@ -55,11 +55,40 @@ pub struct CipherCreateRequest {
...
+    organization_id: Option<OrganizationId>,
+    folder_id: Option<FolderId>,
+    name: String,
+    notes: Option<String>,
+    favorite: bool,
+    reprompt: CipherRepromptType,
+    r#type: CipherViewType,
+    fields: Vec<FieldView>,
+    key: Option<EncString>,
+}
```
Hinton, line 66: **suggestion:** I think we can just wrap the public request?

```suggestion
struct CipherCreateRequestInternal {
    request: CipherCreateRequest,
    key: Option<EncString>,
}
```

crates/bitwarden-vault/src/cipher/cipher_client/edit.rs
```diff
@@ -101,11 +100,46 @@ impl TryFrom<CipherView> for CipherEditRequest {
...
+    favorite: bool,
+    reprompt: CipherRepromptType,
+    name: String,
+    notes: Option<String>,
+    fields: Vec<FieldView>,
+    r#type: CipherViewType,
+    revision_date: DateTime<Utc>,
+    archived_date: Option<DateTime<Utc>>,
+    password_history: Vec<PasswordHistoryView>,
+    key: Option<EncString>,
```
Hinton: **suggestion:** Wrap with the request and you should be able to avoid some of the duplicate code?

Hinton (dismissed):
Looks good to me, I think you can simplify the internal structs and just wrap the request.

crates/bitwarden-vault/src/cipher/cipher_client/create.rs
nikwithak: Good idea - updated to just wrap the public structs.

crates/bitwarden-vault/src/cipher/cipher_client/edit.rs
```diff
@@ -0,0 +1,779 @@
...
+        edit_cipher(
+            key_store,
+            &config.api_client,
+            repository.as_ref(),
+            user_id,
+            request,
+        )
+        .await
+    }
+}
```
shane-melton: ❓ I think we will want a `generate_cipher_key()` method call here (similar to `create`) when the cipher key feature flag is enabled. Otherwise, existing ciphers will never be migrated to use cipher keys.

shane-melton (commented):
Changes look really good from my perspective! Just one question on how we want to handle cipher key encryption when editing existing ciphers.

crates/bitwarden-vault/src/cipher/cipher_client/edit.rs
nikwithak: I added this to the EditRequest, and moved some logic around for re-use in the other places we generate a cipher key.

[f565d340](https://github.com/bitwarden/sdk-internal/pull/455/commits/f565d3402626eaf5515e6d3e9cf5fff52b35ff5b)

shane-melton (approved):
🎉 Looks great!

Hinton (approved):

PR #484, opened 2025-10-01
[PM-22591] Unify client repository initialization

dani-garcia:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-22591

## 📔 Objective

Currently, all app crates (wasm, uniffi) need to declare all the client managed repositories that they allow clients to implement. This means that all the crates need to be updated when adding more repositories, which adds more steps to implementations and can lead to things being out of sync. This PR unifies the declaration of client managed repositories so that there's only one place where they are defined. 

While I was at it, I've changed slightly the way we register the repositories. Rather than provide one function for each (`register_cipher_repository`/`register_folder_repository`) I provide a single function that takes an object that contains all the repos. This is easier to generate and should make it easier to know if any of the repos are missing (you should get a compile error). I left the previous ones for backwards-compat

Sadly both wasm and uniffi have a lot of compile time type shenanigans and macros which don't make this easy, so this solution contains multiple parts:
- In `bitwarden_wasm` and `bitwarden_uniffi` we already have a macro that creates the repository implementations based on a type. This has been expanded to take multiple types and also create the `Repositories` struct which is just a grouping of all of them. Now, instead of manually calling these macros manually for each type, they instead get passed to the `bitwarden_pm::create_client_managed_repositories!` macro.
- The `bitwarden_pm::create_client_managed_repositories!` macro's only purpose is to call the passed macro repeatedly with all the types that need to be created. This is the only way I found to lazily define the types in `bitwarden-pm` so that they would be usable in macros in the other crates. The definition is fairly repetitive with the types, because we don't have easy ways to change the case of identifiers or merge them together. We could use the `paste` crate to avoid the repetition but didn't seem worth it at the moment

Note that in the past we placed the sdk-managed definitions in a shared `bitwarden-state-migrations` crate, but now that we have the new `bitwarden-pm` crate as a shared crate focused on password manager functionality I don't think `bitwarden-state-migrations` is needed. Instead this PR moves the sdk-managed definitions to `bitwarden-pm` (next to the new client managed repositories macro) and deletes the `bitwarden-state-migrations` crate.

## ⏰ Reminders before review

- CI builds passed
  team

## 🦮 Reviewer guidelines


crates/bitwarden-state/README.md
```diff
@@ -39,21 +39,20 @@ need to define some functions in `bitwarden-wasm-internal` and `bitwarden-uniffi
...
+macro_rules! create_client_managed_repositories {
+    ($container_name:ident, $macro:ident) => {
+        $macro! {
+            $container_name;
+            // List any SDK-managed repositories here. The format is:
+            // <fully qualified path to the item>, <item type idenfier>, <field name>, <name of the repository implementation>
+            ::bitwarden_vault::Cipher, Cipher, cipher, CipherRepository;
+            ::bitwarden_vault::Folder, Folder, folder, FolderRepository;
+        }
+    };
```
coroiu, line 55: praise: Nice idea! Using macros do define the repositories in a place where we're not actually able to depend on them 👍

coroiu (approved):

PR #485, opened 2025-10-01
[PM-26354] Add methods to un-/wrap asymmetric keys in keystore context (1/3)

iinuwa:
## 🎟️ Tracking

[PM-26354](https://bitwarden.atlassian.net/browse/PM-26354)

## 📔 Objective

In order to use support PRF keys in the mobile clients, we need to generate rotateable keysets in . This adds the necessary methods to the keystore needed for creating the rotateable keysets.

## ⏰ Reminders before review

- CI builds passed
  team

## 🦮 Reviewer guidelines

[PM-26354]: https://bitwarden.atlassian.net/browse/PM-26354?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/crates/bitwarden-crypto/src/store/context.rs
+++ b/crates/bitwarden-crypto/src/store/context.rs
@@ -8,11 +8,11 @@ use zeroize::Zeroizing;
 
 use super::KeyStoreInner;
 use crate::{
-    AsymmetricCryptoKey, BitwardenLegacyKeyBytes, ContentFormat, CryptoError, EncString, KeyId,
-    KeyIds, PublicKeyEncryptionAlgorithm, Result, RotatedUserKeys, Signature, SignatureAlgorithm,
-    SignedObject, SignedPublicKey, SignedPublicKeyMessage, SigningKey, SymmetricCryptoKey,
-    UnsignedSharedKey, derive_shareable_key, error::UnsupportedOperationError, signing,
-    store::backend::StoreBackend,
+    AsymmetricCryptoKey, AsymmetricPublicCryptoKey, BitwardenLegacyKeyBytes, ContentFormat,
+    CryptoError, EncString, KeyId, KeyIds, Pkcs8PrivateKeyBytes, PublicKeyEncryptionAlgorithm,
+    Result, RotatedUserKeys, Signature, SignatureAlgorithm, SignedObject, SignedPublicKey,
+    SignedPublicKeyMessage, SigningKey, SpkiPublicKeyBytes, SymmetricCryptoKey, UnsignedSharedKey,
+    derive_shareable_key, error::UnsupportedOperationError, signing, store::backend::StoreBackend,
 };
 
 /// The context of a crypto operation using [super::KeyStore]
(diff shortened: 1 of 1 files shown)


crates/bitwarden-crypto/src/store/context.rs
```diff
@@ -241,6 +241,97 @@ impl<Ids: KeyIds> KeyStoreContext<'_, Ids> {
...
+    }
+
+    /// Decrypt an asymmetric public key into the context using an already existing symmetric key
+    ///
+    /// # Arguments
+    ///
+    /// * `wrapping_key` - The key id used to decrypt the `wrapped_key`. It must already exist in
+    ///   the context
+    /// * `wrapped_key` - The key to decrypt
+    pub fn unwrap_public_key(
```
quexten, line 300: Imo this is not a public API that we want to offer to teams. We should either deprecate it with a note that this is only to be used for rotateable key sets, or we should move the rotateable keyset in the follow-up PR into the crypto crate and remove this method (or make it pub(crate).

The same applies for `wrap_public_key`.

quexten (commented):
Changes look mostly good, but I don't think wrapping public keys should be a supported public API, and other teams (aside from what's necessary for the rotateable keyset) should not use it.

Any thoughts on moving the rotateable key set to the crypto crate entirely? We probably don't even need to implement things on the key context then and can go with the keys directly as they are ephemeral for interacting with the key set anyways and shouldn't be held in the key context longer than needed.

If that's not possible I'm OK with making it depracated with the deprecation note stating it's meant for use with the rotateable key set only.

Hinton:
In my perspective KM owns the public interface of the keystore, please have a discussion with KM what level of support is required here. And if it can be made generic and in line with existing and future requirements.

Also, splitting this up into 3 PRs does not make the review process easier IMHO, since you need all three PRs to understand the scope.

iinuwa:
> In my perspective KM owns the public interface of the keystore, please have a discussion with KM what level of support is required here. And if it can be made generic and in line with existing and future requirements.

Sorry about that. I agree this warrants further discussion; I'll mark this as a draft until we've talked further.

> Also, splitting this up into 3 PRs does not make the review process easier IMHO, since you need all three PRs to understand the scope.

Yeah, I see that now. I can drop these and merge them into a single PR after discussion with the KM team.

crates/bitwarden-crypto/src/store/context.rs
iinuwa: > We probably don't even need to implement things on the key context then and can go with the keys directly as they are ephemeral for interacting with the key set anyways and shouldn't be held in the key context longer than needed.

I had a question about that: does the context received from `store.context()` live as long as the `store` or as long as the reference that is returned?

> I don't think wrapping public keys should be a supported public API, and other teams (aside from what's necessary for the rotateable keyset) should not use it.
> 
> Any thoughts on moving the rotateable key set to the crypto crate entirely?

That's fine by me! I wasn't sure about where ownership between `bitwarden-crypto` and `bitwarden_core::key_management` lies, so I put the methods closest to where they would be used. I initially implemented the rotateable key set without the keystore, but I had to use a bunch of things marked as deprecated. Are those just intended to mark dangerous/internal-but-has-to-be-pub methods?

iinuwa:
Closed in favor of #494.

PR #488, opened 2025-10-02
[PM-26534] Remove non-generic wrapper for PasswordProtectedKeyEnvelope

quexten:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-26534

## 📔 Objective

The initial implementation assumed that the struct of of the PasswordProtectedKeyEnvelope in crypto had to be generic, because the key context is generic w.r.t. the keys used, and the keys are only then provided by core. Because uniffy / wasm-bindgen can't handle generics, I added a non-generic wrapper struct in core, leaving us with two different "PasswordProtectedKeyEnvelopes".

This understanding was not correct. You don't actually need to make the struct generic, having just two specific functions on the impl generic to the key id's suffices, allowing us to drop the "non generic wrapper" entirely, vastly simplifing our code here. The changes are mostly removing code and moving the uniffi mapping to the crypto crate.

## ⏰ Reminders before review

- CI builds passed
  team

## 🦮 Reviewer guidelines


crates/bitwarden-crypto/src/safe/password_protected_key_envelope.rs
```diff
@@ -447,6 +442,29 @@ impl From<TryFromIntError> for PasswordProtectedKeyEnvelopeError {
...
+
+#[cfg(feature = "wasm")]
+impl wasm_bindgen::describe::WasmDescribe for PasswordProtectedKeyEnvelope {
+    fn describe() {
+        <String as wasm_bindgen::describe::WasmDescribe>::describe();
+    }
+}
+
+#[cfg(feature = "wasm")]
+impl FromWasmAbi for PasswordProtectedKeyEnvelope {
```
Thomas-Avery, line 459: concern `FromWasmAbi` is marked as unstable. https://docs.rs/wasm-bindgen/0.2.104/wasm_bindgen/convert/trait.FromWasmAbi.html#-unstable

crates/bitwarden-crypto/src/safe/password_protected_key_envelope.rs
```diff
@@ -447,6 +442,29 @@ impl From<TryFromIntError> for PasswordProtectedKeyEnvelopeError {
...
+    type Abi = <String as FromWasmAbi>::Abi;
+
+    unsafe fn from_abi(abi: Self::Abi) -> Self {
+        let s = unsafe { String::from_abi(abi) };
+        Self::from_str(&s).expect("Invalid PasswordProtectedKeyEnvelope")
```
Thomas-Avery: I don't understand all the magic that happens behind wasm_bindgen. Will this panic be caught and morphed into a JS exception?

crates/bitwarden-crypto/src/safe/password_protected_key_envelope.rs
quexten: Previously, because we had the newtype wrapper, we could use tsify to auto-generate it. For just working with the type directly, I did not find a good way to use tsify other than creating a newtype in the same file, which did not seem like the right solution.

That said, tsify just generates FromWasmAbi. https://github.com/madonoharu/tsify/blob/ab399fe86761e94f99bdb57a0b18a78d54b08671/tsify-macros/src/wasm_bindgen.rs#L202

crates/bitwarden-crypto/src/safe/password_protected_key_envelope.rs
quexten: Sorry, expect was just not correct here, and `unwrap_throw` should have been used instead. Updated.

crates/bitwarden-crypto/src/safe/password_protected_key_envelope.rs
quexten: Wasm bindgen explicitly documents this also here: https://wasm-bindgen.github.io/wasm-bindgen/contributing/design/rust-type-conversions.html#from-js-to-rust.

Thomas-Avery: Thank you for the context. I guess if it breaks in the future we can just follow whatever tsify comes up with to solve for it. I would like someone with more wasm bindgen experience than me to review.

quexten: Leaving this comment open until the review from @bitwarden/team-sdk-sme

Hinton (dismissed):

Thomas-Avery (dismissed):

crates/bitwarden-crypto/src/safe/password_protected_key_envelope.rs
Hinton: Yea it's unstable but a lot of things depends on it and we'll know if it stops compiling.

addisonbeck (dismissed):

Thomas-Avery (approved):

Hinton (approved):

crates/bitwarden-crypto/src/safe/password_protected_key_envelope.rs
dani-garcia: Sadly TSify doesn't support this type of string type-alias pattern we're doing, so manual implementation seems like the way to go. If we find ourselves doing it a lot we could consider introducing a macro for that.

We might also want to implement `IntoWasmAbi` for symmetry, even if it's not used yet.
```rust
impl IntoWasmAbi for PasswordProtectedKeyEnvelope {
    type Abi = <String as IntoWasmAbi>::Abi;
    fn into_abi(self) -> Self::Abi {
        let string: String = self.into();
        string.into_abi()
    }
}
```

dani-garcia (approved):

PR #492, opened 2025-10-06
[PM-26542] Add listing items functionality with session-based authentication with environment variable support for API key login

fer:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-26542

## 📔 Objective

Implements session-based authentication for the Rust Bitwarden CLI, enabling stateless command execution similar to the Node.js CLI. Users can now authenticate once with `bw login api-key` and reuse the session for subsequent commands.

demo

Added environment variable support (`BW_CLIENTID`, `BW_CLIENTSECRET`, `BW_PASSWORD`) to the `bw login api-key` command and made it return a base64-encoded session key (like the Node.js CLI, **although this one is reaally long**) that can be potentially used with --session (or `BW_SESSION`) for subsequent vault operations.

### Session Management

- **Session Export**: Added `export_session()` to serialize full client state
- **Session Import**: Added `import_session()` to restore complete authenticated state from session string
- **Session Format**: Base64-encoded JSON containing all crypto keys and OAuth tokens needed for vault operations

### Environment Variable Support
Added support for environment variables in `bw login api-key`:
- `BW_CLIENTID` (or `BW_CLIENT_ID`): API client ID
- `BW_CLIENTSECRET` (or `BW_CLIENT_SECRET`): API client secret
- `BW_PASSWORD`: Master password
- Environment variables take precedence over interactive prompts

### List Command Implementation
- Implemented `bw list items` with filtering options:
  - `--search`: Text search across items
  - `--folderid`: Filter by folder
  - `--collectionid`: Filter by collection
  - `--organizationid`: Filter by organization
  - `--trash`: Show deleted items
- Returns decrypted vault items as JSON

### Session Usage
The global `--session` flag and `BW_SESSION` environment variable now work across all commands, automatically restoring client state before execution.

## Usage Example
```bash
# Set credentials via environment variables
export BW_CLIENTID="user.xxx"
export BW_CLIENTSECRET="xxx"
export BW_PASSWORD="xxx"

# Login and capture session
SESSION=$(bw login api-key)

# Use session for subsequent commands
bw list items --session "$SESSION"

# Or via environment variable
export BW_SESSION="$SESSION"
bw list items
```

## ⏰ Reminders before review

- CI builds passed
  team

## 🦮 Reviewer guidelines

### DIFF
--- a/crates/bitwarden-cli/src/lib.rs
+++ b/crates/bitwarden-cli/src/lib.rs
@@ -15,3 +15,29 @@ pub fn text_prompt_when_none(prompt: &str, val: Option<String>) -> InquireResult
         Text::new(prompt).prompt()?
     })
 }
+
+/// Try to get a value from CLI arg, then from environment variables, then prompt
+///
+/// Checks multiple environment variable names in order (e.g., BW_CLIENTID, BW_CLIENT_ID)
+pub fn text_or_env_prompt(
+    prompt: &str,
+    cli_val: Option<String>,
+    env_var_names: &[&str],
+) -> InquireResult<String> {
+    // First check if provided via CLI
+    if let Some(val) = cli_val {
+        return Ok(val);
+    }
+
+    // Then check environment variables
+    for env_var in env_var_names {
+        if let Ok(val) = std::env::var(env_var) {
+            if !val.is_empty() {
+                return Ok(val);
+            }
+        }
+    }
+
+    // Finally, prompt the user
+    Text::new(prompt).prompt()
+}
--- a/crates/bitwarden-cli/src/lib.rs
+++ b/crates/bitwarden-cli/src/lib.rs
@@ -19,7 +19,7 @@ pub fn text_prompt_when_none(prompt: &str, val: Option<String>) -> InquireResult
 /// Try to get a value from CLI arg, then from environment variables, then prompt
 ///
 /// Checks multiple environment variable names in order (e.g., BW_CLIENTID, BW_CLIENT_ID)
-pub fn text_or_env_prompt(
+pub fn resolve_user_input_value(
     prompt: &str,
     cli_val: Option<String>,
     env_var_names: &[&str],
--- a/crates/bw/src/auth/login.rs
+++ b/crates/bw/src/auth/login.rs
@@ -1,4 +1,4 @@
-use bitwarden_cli::text_prompt_when_none;
+use bitwarden_cli::{text_or_env_prompt, text_prompt_when_none};
 use bitwarden_core::{
     Client,
     auth::login::{
@@ -93,11 +93,24 @@ pub(crate) async fn login_api_key(
     client: Client,
     client_id: Option<String>,
     client_secret: Option<String>,
-) -> Result<()> {
-    let client_id = text_prompt_when_none("Client ID", client_id)?;
-    let client_secret = text_prompt_when_none("Client Secret", client_secret)?;
+) -> Result<String> {
+    let client_id = text_or_env_prompt("Client ID", client_id, &["BW_CLIENTID", "BW_CLIENT_ID"])?;
+    let client_secret = text_or_env_prompt(
+        "Client Secret",
+        client_secret,
+        &["BW_CLIENTSECRET", "BW_CLIENT_SECRET"],
+    )?;
 
-    let password = Password::new("Password").without_confirmation().prompt()?;
+    // Check for password in environment variable first
+    let password = if let Ok(pwd) = std::env::var("BW_PASSWORD") {
+        if !pwd.is_empty() {
+            pwd
+        } else {
+            Password::new("Password").without_confirmation().prompt()?
+        }
+    } else {
+        Password::new("Password").without_confirmation().prompt()?
+    };
 
     let result = client
         .auth()
@@ -110,7 +123,10 @@ pub(crate) async fn login_api_key(
 
     debug!("{result:?}");
 
-    Ok(())
+    // Export the user key as a session string
+    let session = client.internal.export_user_key_as_session()?;
+
+    Ok(session)
 }
 
 pub(crate) async fn login_device(
--- a/crates/bw/src/auth/login.rs
+++ b/crates/bw/src/auth/login.rs
@@ -94,7 +94,8 @@ pub(crate) async fn login_api_key(
     client_id: Option<String>,
     client_secret: Option<String>,
 ) -> Result<String> {
-    let client_id = resolve_user_input_value("Client ID", client_id, &["BW_CLIENTID", "BW_CLIENT_ID"])?;
+    let client_id =
+        resolve_user_input_value("Client ID", client_id, &["BW_CLIENTID", "BW_CLIENT_ID"])?;
     let client_secret = resolve_user_input_value(
         "Client Secret",
         client_secret,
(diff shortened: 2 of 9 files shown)


crates/bitwarden-cli/src/lib.rs
```diff
@@ -15,3 +15,29 @@ pub fn text_prompt_when_none(prompt: &str, val: Option<String>) -> InquireResult
         Text::new(prompt).prompt()?
     })
 }
+
+/// Try to get a value from CLI arg, then from environment variables, then prompt
+///
+/// Checks multiple environment variable names in order (e.g., BW_CLIENTID, BW_CLIENT_ID)
+pub fn text_or_env_prompt(
```
addisonbeck:
```suggestion
pub fn resolve_user_input_value(
```

Just a suggestion. This name seems a little vague.

addisonbeck (dismissed):
This looks good to me. Will re-approve if you decide to rename `text_or_env_prompt`, but that's not blocking.

crates/bitwarden-cli/src/lib.rs
fer: thank you for this feedback! I've replaced the function name and usages in my last commit.

Also wondering about renaming `text_prompt_when_none` for a more proper name, don't hesitate to point out your suggestion please:

https://github.com/bitwarden-labs/sdk-internal/blob/9236ae5a0b98453598926aeed9a175588596732d/crates/bw/src/auth/login.rs#L1

crates/bw/src/vault/list.rs
```diff
@@ -0,0 +1,111 @@
+use bitwarden_core::Client;
+use bitwarden_vault::{CipherListView, SyncRequest, VaultClientExt};
+use color_eyre::eyre::{Result, bail};
+
+use crate::render::CommandOutput;
+
+#[derive(Debug)]
+pub struct ListOptions {
+    pub object: String,
```
nick-livefront: ⛏️ With their only being a finite amount of options an enum can narrow the available options here. 

```rust
pub enum ObjectType {
    #[value(name = "items")]
    Items,
    #[value(name = "folders")]
    Folders,
    #[value(name = "collections")]
    Collections,
    #[value(name = "organizations")]
    Organizations,
    #[value(name = "org-collections")]
    OrgCollections,
    #[value(name = "org-members")]
    OrgMembers,
}
```

crates/bw/src/vault/list.rs
```diff
@@ -0,0 +1,111 @@
...
+    pub search: Option<String>,
+    pub folderid: Option<String>,
+    pub collectionid: Option<String>,
+    pub organizationid: Option<String>,
+    pub trash: bool,
+}
```
nick-livefront, line 28: ❓ Archive is near future feature that was recently implemented in the CLI, should that be considered here?

https://github.com/bitwarden/clients/pull/16502

crates/bw/src/vault/list.rs
```diff
@@ -0,0 +1,111 @@
...
+        if let Some(ref org_id) = options.organizationid {
+            if item.organization_id.as_ref().map(|id| id.to_string()) != Some(org_id.clone()) {
+                return false;
+            }
+        }
+
+        // Search filter (case-insensitive search in name)
+        if let Some(ref search_term) = options.search {
+            let search_lower = search_term.to_lowercase();
+            if !item.name.to_lowercase().contains(&search_lower) {
```
nick-livefront: ❌ The current implementation of search within the CLI checks the `id`, `name`, `subtitle` and `uris`.  [searchService.searchCiphersBasic](https://github.com/bitwarden/clients/blob/91a661a02560b0840d6d858a2939a0c5cca5bb93/libs/common/src/vault/services/search.service.ts#L313-L333), I would think we'd want parity to that here.

nick-livefront (changes requested):
Let me know if there is future work slated to flush out the implementation of the vault related items. If `list` is just being used as an easy implementation to test against, the changes are good with me.

crates/bw/src/vault/list.rs
fer: Changes:
- Added `ObjectType` enum with `ValueEnum` derive for clap integration
- Updated `ListOptions.object` field from `String` to `ObjectType`
- Replaced string matching with enum pattern matching in list()
- Removed manual error handling for invalid object types

The enum uses kebab-case naming to maintain CLI compatibility.

fer: The `ListOptions` struct is still needed because it contains the filtering parameters (`search`, `folderid`, `collectionid`, `organizationid`, `trash`) that apply to list operations. 

The CLI now provides type-safe validation and shows available options in help output.

crates/bw/src/vault/list.rs
fer: I've added an unimplemented archive feature to match Node.js CLI:

- Add `archive` command with todo!() placeholder
- Add `--archived` flag to list command
- Update `ListOptions` struct with archived field

fer: Updated to match `searchService.searchCiphersBasic` - now searches ID (8+ chars), name, subtitle, and URIs with case-insensitive matching.

Please let me know if tests are needed.

nick-livefront (approved):
Thank you for addressing my feedback!

fer:
> This looks good to me. Will re-approve if you decide to rename `text_or_env_prompt`, but that's not blocking.

Please don't hesitate to propose more changes if needed, happy to help!

crates/bw/src/auth/login.rs
```diff
@@ -108,7 +124,19 @@ pub(crate) async fn login_api_key(
...
 
-    Ok(())
+    // Sync vault data after successful login
+    let sync_result = client
+        .vault()
+        .sync(&SyncRequest {
+            exclude_subdomains: Some(true),
+        })
+        .await?;
+    info!("Synced {} ciphers", sync_result.ciphers.len());
```
Patrick-Pimentel-Bitwarden, line 134: ❓ Do we have any guidance on how and when we are to be logging?

Hinton (commented):
Do we need to add this into the core crate? I was imaging we could temporarily add some glue code to the cli expecially since this implementation isn't really compliant to the current cli.

fer:
> Do we need to add this into the core crate? I was imaging we could temporarily add some glue code to the cli expecially since this implementation isn't really compliant to the current cli.

@Hinton I do believe it probably should be somewhere else, and agreed that the direct key extraction/injection might violate KM architecture. 

Please don't hesitate to point me out on how to refactor this situation. My current idea is to move export/import to `bitwarden_crypto` (❓) context file and then in `internal.rs` just simply wrap.

crates/bw/src/auth/login.rs
fer: Since it's outputting non-sensitive data, this should be fine. 

But I can also remove it if uncomfortable.

Patrick-Pimentel-Bitwarden: As discussed, seems to be fine per our recommendations. I am thinking we should have out logging intention defined somewhere so I'll make a task for that.

Hinton:
Not sure where you got `bitwarden-crypto` from. I specifically asked for the cli.

I hacked together https://github.com/bitwarden-labs/sdk-internal/pull/1 which ilustrates how we can implement this in the least invasive way possible to other crates.

Additionally after discussions with other people we don't feel comfortable calling this session, because it works in a fundamentally different way from session in the regular cli. Rather than providing an encryption key the session is just a b64 encoded dump of very sensitive data.

To that end, we think renaming `login api-key` to `login temporary-insecure-login api-key` and `session` to `temporary-insecure-session` will avoid any confusion about this down the road.

JaredSnider-Bitwarden:
@fer , can we please close this PR since the associated ticket is abandoned?

PR #486, opened 2025-10-01
[PM-26177] Add methods to create rotateable key set (2/3)

iinuwa:
## 🎟️ Tracking

[PM-26177](https://bitwarden.atlassian.net/browse/PM-26177)

## 📔 Objective

In order to set up PRF passkeys on mobile clients, we need to create a rotateable key set. This adds an internal method to the SDK that can be used by a future PRF derivation method.

## ⏰ Reminders before review

- Contributor guidelines followed
- All formatters and local linters executed and passed
- Written new unit and / or integration tests where applicable
- Protected functional changes with optionality (feature flags)
- Used internationalization (i18n) for all UI strings
- CI builds passed
- Communicated to DevOps any deployment requirements
- Updated any necessary documentation (Confluence, contributing docs) or informed the documentation
  team

## 🦮 Reviewer guidelines

- 👍 (`:+1:`) or similar for great changes
- 📝 (`:memo:`) or ℹ️ (`:information_source:`) for notes or general info
- ❓ (`:question:`) for questions
- 🤔 (`:thinking:`) or 💭 (`:thought_balloon:`) for more open inquiry that's not quite a confirmed
  issue and could potentially benefit from discussion
- 🎨 (`:art:`) for suggestions / improvements
- ❌ (`:x:`) or ⚠️ (`:warning:`) for more significant problems or concerns needing attention
- 🌱 (`:seedling:`) or ♻️ (`:recycle:`) for future improvements or indications of technical debt
- ⛏ (`:pick:`) for minor or nitpick changes

[PM-26177]: https://bitwarden.atlassian.net/browse/PM-26177?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #494, opened 2025-10-06
[PM-26354] Add methods to create rotateable key sets from PRF

iinuwa:
## 🎟️ Tracking
[PM-26354](https://bitwarden.atlassian.net/browse/PM-26354)

## 📔 Objective

In order to set up unlock passkeys on mobile clients, this PR adds a method to create a rotateable key set derived from a PRF value.

This is based on existing code in the TypeScript library and web vault:
- https://github.com/bitwarden/clients/blob/main/libs/common/src/auth/services/webauthn-login/webauthn-login-prf-key.service.ts#L15-L17
- https://github.com/bitwarden/clients/blob/main/apps/web/src/app/auth/core/services/rotateable-key-set.service.ts

## ⏰ Reminders before review

- CI builds passed
  team

## 🦮 Reviewer guidelines

[PM-26177]: https://bitwarden.atlassian.net/browse/PM-26177?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[PM-26354]: https://bitwarden.atlassian.net/browse/PM-26354?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


crates/bitwarden-crypto/src/keys/prf.rs
```diff
@@ -0,0 +1,44 @@
...
+/// Takes the output of a PRF and derives a symmetric key
+pub fn derive_symmetric_key_from_prf(prf: &[u8]) -> Result<SymmetricCryptoKey, CryptoError> {
+    let (secret, _) = prf
+        .split_at_checked(32)
+        .ok_or_else(|| CryptoError::InvalidKeyLen)?;
+    let secret: [u8; 32] = secret.try_into().unwrap();
+    if secret.iter().all(|b| *b == b'\0') {
+        return Err(CryptoError::ZeroNumber);
+    }
+    Ok(SymmetricCryptoKey::Aes256CbcHmacKey(stretch_key(
```
iinuwa, line 14: I have seen in some places that we want to begin preferring XChaCha20-Poly1305 keys over AES256-CBC-HMAC. I based this method on the existing code which only derives AES keys, but if this is a direction we're going, then I could parameterize and adjust the callers of this method.

crates/bitwarden-crypto/src/store/key_rotation.rs
```diff
@@ -45,6 +49,113 @@ pub fn dangerous_get_v2_rotated_account_keys<Ids: KeyIds>(
     })
 }
 
+/// A set of keys where a given `EncryptionKey` is protected by an encrypted public/private
```
iinuwa: I'm not sure if `EncryptionKey` is the best description for this, I'm open to suggestions.

crates/bitwarden-crypto/src/store/key_rotation.rs
```diff
@@ -45,6 +49,113 @@ pub fn dangerous_get_v2_rotated_account_keys<Ids: KeyIds>(
...
+/// and the `PublicKey` is protected by the `EncryptionKey`. This setup allows:
+///
+///   - Access to `EncryptionKey` by knowing the `ExternalKey`
+///   - Rotation to a `NewEncryptionKey` by knowing the current `UserKey`, without needing access to
+///     the `ExternalKey`
+#[derive(Serialize, Deserialize, Debug)]
+#[serde(rename_all = "camelCase", deny_unknown_fields)]
+#[cfg_attr(feature = "uniffi", derive(uniffi::Record))]
+#[cfg_attr(feature = "wasm", derive(Tsify), tsify(into_wasm_abi, from_wasm_abi))]
+pub struct RotateableKeySet {
```
iinuwa: I did not make `RotateableKeySet` generic over `<KeyIds>`, since the UniFFI macro was complaining about it. I made the impl functions generic instead, but happy to discuss better options.

crates/bitwarden-crypto/src/store/key_rotation.rs
```diff
@@ -45,6 +49,113 @@ pub fn dangerous_get_v2_rotated_account_keys<Ids: KeyIds>(
...
+    wrapped_decapsulation_key: EncString,
+}
+
+impl RotateableKeySet {
+    /// Create a set of keys to allow access to the user key via the provided
+    /// symmetric wrapping key while allowing the user key to be rotated.
+    pub fn new<Ids: KeyIds>(
+        ctx: &KeyStoreContext<Ids>,
+        wrapping_key: &SymmetricCryptoKey,
+        key_to_wrap: Ids::Symmetric,
```
iinuwa: I chose to restrict the `EncryptionKey` to be `Symmetric` rather than allowing any asymmetric keys, since I couldn't wrap my head around how an asymmetric key would work in this context while providing cryptographic integrity.

Copying from the [previous discussion](https://github.com/bitwarden/sdk-internal/pull/486#discussion_r2398216249):

Does it make sense to be generic over `KeyId` and not use `SymmetricKeyId` directly? Can we provide the same guarantees with asymmetric keys? 

```
e, d = en-/decapsulation key pair
s = symmetric external key
k = key to wrap (symmetric key, or private key)

KeySet = {
  SymmetricWrap(s, d),
  Wrap(k, e),  // symmetric or asymmetric, depending on type of `k`
  Encaps(e, k),
}
```

In order for `Wrap(k, e)` to not be tampered by the server, `Wrap` must provide integrity, which RSA alone does not, right? Would we also have to sign the ciphertext with a separate signing key?

crates/bitwarden-crypto/src/store/key_rotation.rs
```diff
@@ -45,6 +49,113 @@ pub fn dangerous_get_v2_rotated_account_keys<Ids: KeyIds>(
...
+            .encrypt_with_key(&key_to_wrap_instance)?;
+
+        Ok(RotateableKeySet {
+            encapsulated_encryption_key,
+            encrypted_encapsulation_key,
+            wrapped_decapsulation_key,
+        })
+    }
+
+    fn unlock<Ids: KeyIds>(
```
iinuwa: This method is currently unused since we don't yet have plans to use the rotateable key set on mobile, only to create them. However, it implements functionality that is [currently done in TypeScript](https://github.com/bitwarden/clients/blob/main/libs/auth/src/common/login-strategies/webauthn-login.strategy.ts#L84-L94) that could be moved to the SDK.

I can remove it until we decide to either migrate the browser apps to this SDK method or start using it in the mobile apps.

crates/bitwarden-crypto/src/store/key_rotation.rs
quexten: I agree with this approach.

During review of your previous PR chain I noticed we did the same with PasswordProtectedKeyEnvelope, which required a newtype around the generic type, so I now made a PR to clean that up with the same approach as you describe above.

crates/bitwarden-crypto/src/keys/prf.rs
quexten: We've only done that for user's account symmetric keys "userkey" so far since we needed a clean cryptographic break from AES256-CBC/AES256-CBC-HMAC. We use this to then require existence of certain new security features (SignedSecurityState / signature keypairs).

I don't think we need to switch in other places just yet, and would probably make this PR more complex than it needs to be, since the call-sites for the decrypt path would first need to figure out which key to derive, then call the correct function here.

So from KM's side feel free to skip xchacha20-poly1305.

crates/bitwarden-crypto/src/store/key_rotation.rs
quexten: Thought/proposal:
```
ExternalKey -> UpstreamKey
EncryptionKey -> DownstreamKey
```
that should make it clearer. Both are symmetric.

crates/bitwarden-crypto/src/store/key_rotation.rs
quexten: > I chose to restrict the EncryptionKey to be Symmetric rather than allowing any asymmetric keys, since I couldn't wrap my head around how an asymmetric key would work in this context while providing cryptographic integrity.

In general we've been locking down asymmetric key usage since these have much worse security properties without careful deliberation [1]. Our only use-case currently is protecting a user's user-key anyways so that is covered. Unless we have a specific use-case for it, I would recommend making it symmetric only, and expanding it only if we actually need to.

[1] Missing sender authentication since we don't do e.g. signcryption.


> In order for Wrap(k, e) to not be tampered by the server, Wrap must provide integrity, which RSA alone does not, right? Would we also have to sign the ciphertext with a separate signing key?

In our SDK usage, we have:
```
Encaps = Encrypt a Symmetric Key with a Public Key
Wrap = Encrypt a Symmetric / Private Key with a Symmetric Key
```
So I'm a bit confused about wrap / encaps in your example above.

I believe the current state is:
```
e, d = en-/decapsulation key pair
s = symmetric external key (upstream)
k = key to wrap symmetric (downstream)

KeySet = {
  Wrap(s, d),
  Encrypt(k, e), 
  Encaps(e, k),
}
```

As to the question about RSA integrity, currently RSA-OAEP is used: https://en.wikipedia.org/wiki/Optimal_asymmetric_encryption_padding. An adversary cannot alter existing encrypted messages.

However, since it's a public key encryption system, and we don't do signcryption or else, there is no sender authenticity and an adversary can create new ciphertexts given that the have the public key.

quexten: > In order for Wrap(k, e) to not be tampered by the server, Wrap must provide integrity, which RSA alone does not, right? Would we also have to sign the ciphertext with a separate signing key?


If this is asking whether in the existing implementation an adversary with access to the TLS channel or the Bitwarden server can forge a ciphertext here; They cannot because they do not know the public key (encapsulation key). RSA does not provide this as a guarantee, but supposedly it is not possible to recover the public key from the ciphertext. However, ideally, yes we would use signcryption here. That would however require a re-design of the rotateable key set, which is not yet planned.

crates/bitwarden-crypto/src/store/key_rotation.rs
```diff
@@ -45,6 +48,124 @@ pub fn dangerous_get_v2_rotated_account_keys<Ids: KeyIds>(
...
+        wrapping_key: &SymmetricCryptoKey,
+        key_to_wrap: Ids::Symmetric,
+    ) -> Result<Self, CryptoError> {
+        let key_pair = AsymmetricCryptoKey::make(crate::PublicKeyEncryptionAlgorithm::RsaOaepSha1);
+
+        // This uses this deprecated method and other methods directly on the other keys
+        // rather than the key store context because we don't want the keys to
+        // wind up being stored in the borrowed context.
+        #[allow(deprecated)]
+        let key_to_wrap_instance = ctx.dangerous_get_symmetric_key(key_to_wrap)?;
```
quexten: Thought: (Not for this PR to address): We (km) should probably make a separate function here that is crate internal that is not deprecated. It's perfectly fine and intended to read the symmetric key within the crypto crate, so the deprecation notices are a bit annoying.

crates/bitwarden-crypto/src/keys/prf.rs
```diff
@@ -0,0 +1,44 @@
+use crate::{CryptoError, SymmetricCryptoKey, utils::stretch_key};
+
+/// Takes the output of a PRF and derives a symmetric key
+pub fn derive_symmetric_key_from_prf(prf: &[u8]) -> Result<SymmetricCryptoKey, CryptoError> {
+    let (secret, _) = prf
+        .split_at_checked(32)
+        .ok_or_else(|| CryptoError::InvalidKeyLen)?;
+    let secret: [u8; 32] = secret.try_into().unwrap();
+    if secret.iter().all(|b| *b == b'\0') {
```
quexten, line 11: Nit: Could we add a comment what this check does? I assume it is a check for putting in an uninitialized PRF (null value) accidentally, which we want to block?

crates/bitwarden-crypto/src/keys/prf.rs
```diff
@@ -0,0 +1,44 @@
...
+        let prf = [
+            0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+            0, 0, 0,
+        ];
+        let err = derive_symmetric_key_from_prf(&prf).unwrap_err();
+        assert!(matches!(err, CryptoError::ZeroNumber));
+    }
+    #[test]
+    fn test_short_prf_fails() {
+        let prf = [0, 1, 2, 3, 4, 5, 6, 7, 8];
```
quexten: Could we add a test for the >32 byte case?

crates/bitwarden-crypto/src/store/key_rotation.rs
```diff
@@ -45,6 +48,124 @@ pub fn dangerous_get_v2_rotated_account_keys<Ids: KeyIds>(
...
+    new_encryption_key_id: Ids::Symmetric,
+) -> Result<RotateableKeySet, CryptoError> {
+    let pub_key_bytes = ctx.decrypt_data_with_symmetric_key(
+        old_encryption_key_id,
+        &key_set.encrypted_encapsulation_key,
+    )?;
+    let pub_key = SpkiPublicKeyBytes::from(pub_key_bytes);
+    let encapsulation_key = AsymmetricPublicCryptoKey::from_der(&pub_key)?;
+    // TODO: There is no method to store only the public key in the store, so we
+    // have pull out the encryption key to encapsulate it manually.
```
quexten: Note: I think this needs a KM-owned tech debt follow-up ticket. I'll make a jira ticket when this merges, but adding the APIs for this feels out of scope for an external PR.

quexten (commented):
Changes look mostly good, thank you!
Only some minor questions / nits.

One more (optional) suggestion here would be to take test vectors from the TS clients and feed them into the unit tests so we know the implementation works the same way.

crates/bitwarden-crypto/src/store/key_rotation.rs
iinuwa: Thank you for the context! I misunderstood some pseudocode you sent on the previous PR, where I thought you were implying that we should be generic over `KeyId` rather than specifically using `SymmetricKeyId`. I think that was just shorthand, and the confusion was mine. I agree that we should just use `SymmetricKeyId`.

crates/bitwarden-crypto/src/keys/prf.rs
iinuwa: yes, precisely! I added a comment.

crates/bitwarden-crypto/src/keys/prf.rs
iinuwa: Yes, I added a test to confirm that the PRF is truncated to 32 bytes, assuming that a truncated PRF has the same randomness as the full output, and that 32 bytes is sufficient for HKDF-SHA256. I also added a doc comment for the method. I'm not sure if we want to only accept exactly 32 bytes and require the caller to do the truncation themselves.

iinuwa:
Thank you for the review!

> One more (optional) suggestion here would be to take test vectors from the TS clients and feed them into the unit tests so we know the implementation works the same way.

I looked around the clients codebase, but I don't see a full test vector for an end-to-end (PRF, User Key) -> RotateableKeySet and back. There are mocked values that don't verify the actual cryptography in https://github.com/bitwarden/clients/blob/main/libs/auth/src/common/login-strategies/webauthn-login.strategy.spec.ts and https://github.com/bitwarden/clients/blob/main/libs/common/src/auth/services/webauthn-login/webauthn-login.service.spec.ts. The Rust types guarantee some of the other invariants tested in. Did you have some specific test vectors in mind?

abergs:
@iinuwa just a ping that the lint [failed](https://github.com/bitwarden/sdk-internal/actions/runs/18321744291/job/52176313590?pr=494) and needs a `cargo +"nightly-2025-08-18" fmt`

iinuwa:
> @iinuwa just a ping that the lint [failed](https://github.com/bitwarden/sdk-internal/actions/runs/18321744291/job/52176313590?pr=494) and needs a `cargo +"nightly-2025-08-18" fmt`

Thanks. I'm using `jj` which doesn't run git hooks, and my IDE formats according to Rust stable formatting rules, but doesn't format comments the same way. I'll work that out in my workflow.

crates/bitwarden-crypto/src/store/key_rotation.rs
quexten, line 1: Note: I'd recommend moving RotateableKeySet into it's own file, since it is a different cryptographic construction. `key_rotation.rs` is really just meant for "User Key Rotation", which may interact with RotateableKeySets, but is a different use-case.

(As a KM note, `key_rotation.rs` should probably be renamed to `user_key_rotation.rs`)

crates/bitwarden-crypto/src/keys/prf.rs
quexten: I'm OK with either as long as it's documented.

iinuwa:
I rebased and incorporated the changes mentioned by Bernd, squashed the lint fixes, and added some other commits to make CI all happy.

This is ready for review. (I updated the Jira ticket to [PM-26354](https://bitwarden.atlassian.net/browse/PM-26354) to separate it from the mobile work.)

[PM-26354]: https://bitwarden.atlassian.net/browse/PM-26354?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

quexten (dismissed):

quexten:
Looks like there are some new merge conflicts, other than that LGTM.

crates/bitwarden-crypto/src/keys/prf.rs
```diff
@@ -0,0 +1,56 @@
+use crate::{CryptoError, SymmetricCryptoKey, utils::stretch_key};
+
+/// Takes the output of a PRF and derives a symmetric key.
+///
+/// The PRF output must be at least 32 bytes long.
+pub fn derive_symmetric_key_from_prf(prf: &[u8]) -> Result<SymmetricCryptoKey, CryptoError> {
+    let (secret, _) = prf.split_at_checked(32).ok_or(CryptoError::InvalidKeyLen)?;
```
dani-garcia: We should probably mention in the docs that only the first 32 bytes of the PRF output are processed:

```rust
/// The PRF output must be at least 32 bytes long. If longer, only the first 32 bytes are used and the remainder is discarded.
pub fn derive_symmetric_key_from_prf(prf: &[u8]) -> Result<SymmetricCryptoKey, CryptoError> {
```

iinuwa: Yes, I missed clarifying that. Thank you!

iinuwa:
Addressed feedback, and opened #537 to address failing test coverage step in CI.

quexten (approved):

abergs:
Seems this PR remains un-merged, any blockers?

quexten:
@abergs I think you / @iinuwa have to reach out to vault / platform to get their reviews on this. KM looks good still.

Jingo88 (approved):

Thomas-Avery (approved):
KM changes LGTM

addisonbeck (approved):

PR #499, opened 2025-10-09
[PM-24468] Introduce CipherRiskClient

shane-melton:
## 🎟️ Tracking

[PM-24468](https://bitwarden.atlassian.net/browse/PM-24468)

## 📔 Objective

Implement the cipher risk evaluation logic in the Vault SDK  as a separate `CipherRiskClient` so that it can be re-used in mobile and gain potential performance improvements. 

The `CipherRiskClient` contains logic to check if a multiple login ciphers' passwords are exposed (via HIBP), weak (via zxcvbn), or reused.

## ⏰ Reminders before review

- CI builds passed
  team

## 🦮 Reviewer guidelines

[PM-24468]: https://bitwarden.atlassian.net/browse/PM-24468?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


crates/bitwarden-vault/src/cipher/cipher.rs
```diff
@@ -486,6 +486,23 @@ impl CipherView {
...
+        if let Some(login) = &self.login {
+            if let Some(password) = &login.password {
+                return Some(crate::cipher::cipher_risk::CipherLoginDetails {
+                    id: self.id,
+                    password: password.clone(),
+                    username: login.username.clone(),
+                });
+            }
+        }
+        None
```
nikwithak: suggestion: Rather than the nested `if let Some()` blocks, you can use the `?` operator on Option to short-circuit these:

```suggestion
        let login = self.login.as_ref()?;
        Some(crate::cipher::cipher_risk::CipherLoginDetails {
            id: self.id,
            password: login.password.clone()?,
            username: login.username.clone(),
        })
```

crates/bitwarden-vault/src/cipher/cipher_risk_client.rs
```diff
@@ -0,0 +1,778 @@
...
+        login_details: Vec<CipherLoginDetails>,
+    ) -> Result<PasswordReuseMap, CipherRiskError> {
+        let mut map = HashMap::new();
+        for details in login_details {
+            if !details.password.is_empty() {
+                *map.entry(details.password).or_insert(0) += 1;
+            }
+        }
+        Ok(PasswordReuseMap { map })
+    }
```
nikwithak, line 98: ⛏️ Since this doesn't use `self`, would it make more sense as a constructor function on `PasswordReuseMap`, e.g. `PasswordReuseMap::new(login_details) -> Self { ... }`?

crates/bitwarden-vault/src/cipher/cipher_risk_client.rs
```diff
@@ -0,0 +1,778 @@
...
+                } else {
+                    ExposedPasswordResult::NotChecked
+                };
+
+                // Check reuse from provided map
+                let reuse_count = if let Some(map) = &password_map {
+                    map.map.get(&details.password).copied()
+                } else {
+                    None
+                };
```
nikwithak: :art: You could use option chaining methods here.

```suggestion
               let reuse_count = password_map.as_ref()
                                                 .and_then(|m| m.map.get(&details.password).copied());
```

crates/bitwarden-vault/src/cipher/cipher_risk_client.rs
```diff
@@ -0,0 +1,778 @@
...
+        } else {
+            // Username: split on non-alphanumeric
+            username
+                .trim()
+                .to_lowercase()
+                .split(|c: char| !c.is_alphanumeric())
+                .filter(|s| !s.is_empty())
+                .map(str::to_owned)
+                .collect()
+        }
```
nikwithak: ⛏️ You could use Option::map_or() here to avoid duplicating the string processing chain. 

```suggestion
        username
            // Check if it's email-like (contains @)
            .split_once('@')
            // Email: extract local part tokens
            .map_or(username, |(local_part, _domain)| local_part)
            .trim()
            .to_lowercase()
            .split(|c: char| !c.is_alphanumeric())
            .filter(|s| !s.is_empty())
            .map(str::to_owned)
            .collect()
```

crates/bitwarden-vault/src/cipher/cipher_risk_client.rs
```diff
@@ -0,0 +1,778 @@
...
+    /// Returns the number of times the password appears in breaches (0 if not found).
+    fn parse_hibp_response(response: &str, target_suffix: &str) -> u32 {
+        for line in response.lines() {
+            if let Some((hash_suffix, count_str)) = line.split_once(':') {
+                if hash_suffix.eq_ignore_ascii_case(target_suffix) {
+                    return count_str.trim().parse().unwrap_or(0);
+                }
+            }
+        }
+        0
```
nikwithak: 🎨 Could use iterator methods
```suggestion
        response
            .lines()
            .filter_map(|l| l.split_once(':'))
            .find(|(hash_suffix, _)| hash_suffix.eq_ignore_ascii_case(target_suffix))
            .and_then(|(_, count_str)| count_str.trim().parse().ok())
            .unwrap_or(0)
```

crates/bitwarden-vault/src/cipher/cipher_risk_client.rs
```diff
@@ -0,0 +1,778 @@
...
+                    None
+                };
+
+                CipherRisk {
+                    id: details.id,
+                    password_strength,
+                    exposed_result,
+                    reuse_count,
+                }
+            }
```
nikwithak: :pick: Could this be moved to a separate function `async fn to_cipher_risk(..)` or similar?

nikwithak (commented):
Looks good! A few minor suggestions.

crates/bitwarden-vault/src/cipher/cipher.rs
shane-melton: Thanks for the suggestion! [1ea9850](https://github.com/bitwarden/sdk-internal/pull/499/commits/1ea98508fd1069a84e047325bab6f67e6c8c2ef5)

crates/bitwarden-vault/src/cipher/cipher_risk_client.rs
shane-melton: [1ea9850](https://github.com/bitwarden/sdk-internal/pull/499/commits/1ea98508fd1069a84e047325bab6f67e6c8c2ef5)

crates/bitwarden-vault/src/cipher/cipher_risk_client.rs
shane-melton: 👍 [1ea9850](https://github.com/bitwarden/sdk-internal/pull/499/commits/1ea98508fd1069a84e047325bab6f67e6c8c2ef5)

crates/bitwarden-vault/src/cipher/cipher_risk_client.rs
shane-melton: [1ea9850](https://github.com/bitwarden/sdk-internal/pull/499/commits/1ea98508fd1069a84e047325bab6f67e6c8c2ef5)

crates/bitwarden-vault/src/cipher/cipher_risk_client.rs
shane-melton: Good point, added a constructor function here [5cc8069](https://github.com/bitwarden/sdk-internal/pull/499/commits/5cc806956e3ed0cd48f41f26e748ca5050e1607a)

nikwithak: Thanks! I also just noticed this function returns a `Result<>`, but doesn't have any risk of Error - could it be simplified to just return the `PasswordReuseMap`?

shane-melton: Another good point! In the past I recall running into issues removing the `Result<>` type from functions with WASM bindings. However, I gave it a try here and it appears to have worked without issue.

However, I did notice that by removing `&self` and making the method static, it also forces us to use the static method in the TS clients which circumvents the `SdkService` and requires use of the `SdkLoadService.Ready` which is [recommended only for advanced scenarios](https://github.com/bitwarden/clients/blob/ce5a5e36498eddcf3e10222278a7f8b372a1fed2/libs/common/src/platform/abstractions/sdk/sdk-load.service.ts#L16-L30). I think I'm going to revert making this a static method for now and perhaps we can revisit if necessary (keeping the constructor function though).

nikwithak: Makes sense - one more thing I'm wondering now:

❓ Since we already have the `Vec<CipherLoginDetails>` - would it make sense to calculate the password_map internally (inside of `compute_risk`) and save the caller a step? Or is the expectation that the password_map is generated from the full set of LoginDetails, and we're only comparing a subset?

shane-melton: We may only be checking a single cipher at times, so we need to let the caller pass in an existing password map from all ciphers they want to compare against. 

e.g. Org admins in Access Intelligence will want to check against all ciphers in the org. While an individual may only need to check the cipher they just used for autofill against all of their personal ciphers.

nikwithak: Makes sense! Thanks for the clarification

crates/bitwarden-vault/src/cipher/cipher_risk_client.rs
shane-melton: Done in [b7d3077](https://github.com/bitwarden/sdk-internal/pull/499/commits/b7d3077a902ca2530410c4d5bb753c3e05099f9f)

Hinton:
Will this replace the existing `password_strength` under the auth namespace? https://github.com/bitwarden/sdk-internal/blob/main/crates/bitwarden-core/src/auth/password/strength.rs#L5-L17

shane-melton:
@Hinton I originally planned on utilizing their implementation, however it is adding [Bitwarden specific inputs](https://github.com/bitwarden/sdk-internal/blob/8d5a4e702de19ed020797bbe3a51d3f2e91442d7/crates/bitwarden-core/src/auth/password/strength.rs#L3) as it was being used to specifically check the user's Bitwarden password. It also assumed there was always an email (a requirement for Bitwarden accounts).

Not opposed to making this more generic so it can be reused by Auth. However, at that point I'm not sure the Vault crate is the proper spot for a generic password strength checking utility, especially if it means the Auth crate would depend on Vault.

crates/bitwarden-vault/Cargo.toml
```diff
@@ -55,6 +56,7 @@ uniffi = { workspace = true, optional = true }
 uuid = { workspace = true }
 wasm-bindgen = { workspace = true, optional = true }
 wasm-bindgen-futures = { workspace = true, optional = true }
+zxcvbn = ">=3.0.1, <4.0"
```
Hinton: **issue:** These should be moved to the root workspace and referenced as workspace dependencies since they are consumed by multiple crates.

crates/bitwarden-vault/src/error.rs
```diff
@@ -31,3 +31,12 @@ pub enum VaultParseError {
...
 }
+
+/// Error type for cipher risk evaluation operations
+#[allow(missing_docs)]
+#[bitwarden_error(flat)]
+#[derive(Debug, Error)]
+pub enum CipherRiskError {
+    #[error(transparent)]
+    Reqwest(#[from] reqwest::Error),
+}
```
Hinton: **suggestion:** Having a root `error.rs` is a bit of an anti pattern we're slowly moving away from. Defining it where it's used will result in less file jumping.

crates/bitwarden-vault/src/cipher/cipher.rs
```diff
@@ -486,6 +486,19 @@ impl CipherView {
         }
     }
 
+    /// Extract login details for risk evaluation (login ciphers only).
+    ///
+    /// Returns `Some(CipherLoginDetails)` if this is a login cipher with a password,
+    /// otherwise returns `None`.
+    pub fn to_login_details(&self) -> Option<crate::cipher::cipher_risk::CipherLoginDetails> {
```
Hinton: **question:** Is this used anywhere?

crates/bitwarden-vault/src/cipher/cipher_risk.rs
```diff
@@ -0,0 +1,107 @@
...
+    Error(String),
+}
+
+/// Login cipher data needed for risk evaluation.
+#[derive(Serialize, Deserialize, Debug, Clone)]
+#[cfg_attr(feature = "uniffi", derive(uniffi::Record))]
+#[cfg_attr(feature = "wasm", derive(Tsify), tsify(into_wasm_abi, from_wasm_abi))]
+pub struct CipherLoginDetails {
+    /// Cipher ID to identify which cipher in results.
+    pub id: Option<CipherId>,
```
Hinton: **question:** Can `CipherId` be empty?

crates/bitwarden-vault/src/cipher_risk/types.rs
Hinton, line 1: **suggestion:** This feature seems completely decoupled from ciphers. Maybe move it to a separate dedicated module, and/or crate.

crates/bitwarden-vault/src/cipher_risk/cipher_risk_client.rs
Hinton, line 1: **issue:** We generally only use clients to expose a public interface, this file seems to currently do a lot of various things.

**suggestion:**

1. Extract two sub-modules, HIBP and reuse detection. These can be unit tested in isolation quite nicely which will also significantly decrease the filesize.
2. Consider what public interface you need, is it sufficient to just accept a single struct containing a list of ciphers to check? If so everything except this struct and response can be made internal to the module or crate.

crates/bitwarden-vault/src/cipher/cipher.rs
shane-melton: Nope, it was a leftover. Removed [d201755](https://github.com/bitwarden/sdk-internal/pull/499/commits/d201755f51c780a0ae0fb0841b30a4332e97789f)

crates/bitwarden-vault/src/cipher/cipher_risk.rs
shane-melton: Good point, it really doesn't make sense for it to ever be null in this context. Made required in [9b8b001](https://github.com/bitwarden/sdk-internal/pull/499/commits/9b8b001a5000beaa1df739976908226d75e8c6bb)

crates/bitwarden-vault/src/cipher_risk/types.rs
shane-melton: Good idea, moved to a separate `cipher_risk` module in [5d0e1d2](https://github.com/bitwarden/sdk-internal/pull/499/commits/5d0e1d24d4bc39207ad28713c9a563ad8c353356)

crates/bitwarden-vault/src/error.rs
shane-melton: Done! [5d0e1d2](https://github.com/bitwarden/sdk-internal/pull/499/commits/5d0e1d24d4bc39207ad28713c9a563ad8c353356)

crates/bitwarden-vault/Cargo.toml
shane-melton: Moved to the root workspace in [c039308](https://github.com/bitwarden/sdk-internal/pull/499/commits/c039308e553478808ed90036840816cbabcd0ac2)

crates/bitwarden-vault/src/cipher_risk/cipher_risk_client.rs
shane-melton: Refactored the HIBP and password strength into their own sub-modules to reduce the complexity of the `cipher_risk_client`. (the reuse detection is fairly small and didn't seem worth the same overhead).

I also cleaned up the public interface to only export the required input/output structs, the error, and the client itself.

[ac5c2f0](https://github.com/bitwarden/sdk-internal/pull/499/commits/ac5c2f058ca0e2a81d52397ff9512af3c1d2d144)

crates/bitwarden-vault/src/cipher_risk/hibp.rs
```diff
@@ -0,0 +1,169 @@
...
+
+use super::CipherRiskError;
+
+/// Default base URL for the Have I Been Pwned (HIBP) Pwned Passwords API.
+pub(super) const HIBP_DEFAULT_BASE_URL: &str = "https://api.pwnedpasswords.com";
+
+/// Hash password with SHA-1 and split into prefix/suffix for k-anonymity.
+///
+/// Returns a tuple of (prefix: first 5 chars, suffix: remaining chars).
+pub(super) fn hash_password_for_hibp(password: &str) -> (String, String) {
```
nikwithak: ⛏️ It looks like these are only called from this mod in `check_password_exposed`, so I think you can do away with the `pub(super)` here and on `parse_hibp_response()`

crates/bitwarden-vault/src/cipher_risk/password_strength.rs
```diff
@@ -0,0 +1,82 @@
...
+    zxcvbn::zxcvbn(password, &inputs_refs).score().into()
+}
+
+/// Extract meaningful tokens from username/email for password penalization.
+///
+/// Handles both email addresses and plain usernames:
+/// - For emails: extracts and tokenizes the local part (before @)
+/// - For usernames: tokenizes the entire string
+/// - Splits on non-alphanumeric characters and converts to lowercase
+pub(super) fn extract_user_inputs(username: &str) -> Vec<String> {
```
nikwithak: nit: This is only used in `calculate_password_strength`, so you can ditch the `pub(super)`

nikwithak (dismissed):
A couple of non-blocking observations, but otherwise looks good!

crates/bitwarden-vault/src/cipher_risk/hibp.rs
shane-melton: Nice catch! Removed in [bc1808f](https://github.com/bitwarden/sdk-internal/pull/499/commits/bc1808f983ab591013e191c1fb47768a95d08e24)

crates/bitwarden-vault/src/cipher_risk/password_strength.rs
shane-melton: Removed in [bc1808f](https://github.com/bitwarden/sdk-internal/pull/499/commits/bc1808f983ab591013e191c1fb47768a95d08e24)

nikwithak (dismissed):

dereknance (dismissed):

Cargo.toml
```diff
@@ -85,6 +86,7 @@ wasm-bindgen = { version = ">=0.2.91, <0.3", features = ["serde-serialize"] }
 wasm-bindgen-futures = "0.4.41"
 wasm-bindgen-test = "0.3.45"
 wiremock = ">=0.6.0, <0.7"
+zxcvbn = ">=3.0.1, <4.0"
```
Hinton, line 89: **issue (blocker):** Please update `bitwarden-core` to reference the workspace version.

crates/bitwarden-vault/src/cipher_risk/cipher_risk_client.rs
```diff
@@ -0,0 +1,545 @@
...
+#[cfg_attr(feature = "wasm", wasm_bindgen)]
+impl CipherRiskClient {
+    /// Build password reuse map for a list of login ciphers.
+    ///
+    /// Returns a map where keys are passwords and values are the number of times
+    /// each password appears in the provided list. This map can be passed to `compute_risk()`
+    /// to enable password reuse detection.
+    pub fn password_reuse_map(&self, login_details: Vec<CipherLoginDetails>) -> PasswordReuseMap {
+        PasswordReuseMap::new(login_details)
+    }
```
Hinton, line 46: **observation (non-blocking)**: This is likely a pattern we will want to avoid in the future. Part of the reason for moving towards cipher list views is to decrease the amount of data available in the renderer. Sending a plain text list of passwords to JS is therefore undesired. (Decrypting multiple full ciphers on the renderer in order to get the passwords and sending them to the sdk, would likewise be undesired in the future).

You might be able to send a pointer which would keep the memory on the rust side and ensure the rust de-allocator zeroes the memory.

crates/bitwarden-vault/src/cipher_risk/types.rs
Hinton, line 1: non-blocking: I generally prefer to keep the related data models close to where they are used, (which in this case is the `cipher_risk_client`.

Hinton (changes requested):
One blocker.

Cargo.toml
shane-melton: Fixed [02f5614](https://github.com/bitwarden/sdk-internal/pull/499/commits/02f56148ce0d84cde35d0c6615a7de3eee682feb)

crates/bitwarden-vault/src/cipher_risk/cipher_risk_client.rs
shane-melton: Agreed! Once we have more of the Cipher domain (w/ repositories) migrated to the SDK, I anticipate we'll be able to deprecate this pattern and let the cipher risk client pull from the cipher repository internally. Perhaps accepting a list of CipherIds instead in case callers only wish to map a subset of ciphers (e.g. for AC and reporting)

dereknance (approved):

nikwithak (approved):

Hinton (approved):

PR #514, opened 2025-10-15
[PM-27638|BEEEP] Add key id to passwordprotectedkeyenvelope headers

quexten:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-27638

## 📔 Objective

One of the goals of the safe primitives is to make them server validateable. Encrypted objects contain an "encrypted by" key id. Similarly, objects should include the contained key id, so that the server can match these and validate them.

This adds the contained key id to the password protected key envelope. We don't need to consider the backward compatibility case because currently all users are V1 users with AES keys who do not have a key id, and the only use-case is PIN keys.  

## ⏰ Reminders before review

- CI builds passed
  team

## 🦮 Reviewer guidelines


crates/bitwarden-crypto/src/cose.rs
```diff
@@ -19,17 +19,25 @@ use crate::{
...
+// Custom labels for COSE headers
+// NOTE: Any label below -65536 is reserved for private use in the IANA allocations and can be used
+// freely.
 pub(crate) const ARGON2_SALT: i64 = -71001;
 pub(crate) const ARGON2_ITERATIONS: i64 = -71002;
 pub(crate) const ARGON2_MEMORY: i64 = -71003;
 pub(crate) const ARGON2_PARALLELISM: i64 = -71004;
+/// Indicates for any object containing a key (wrapped key, password protected key envelope) which
+/// key ID that contained key has
+pub(crate) const CONTAINED_KEY_ID: i64 = -71005;
```
Thomas-Avery, line 40: not required

💭  It might be helpful if we keep expanding this to convert over to an enum to add some type safety. For example accidentally having two labels using the same number.

Thomas-Avery (approved):
LGTM

PR #519, opened 2025-10-20
[PM-27115] Force icon uri checksum verification on user crypto v2

quexten:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-27115
https://bitwarden.atlassian.net/browse/VULN-185
https://bitwarden.atlassian.net/browse/PM-4185

## 📔 Objective

Forces the icon URI check for crypto v2 users. This adds some plumbing through the key store, which now also includes other non-key cryptographic state. The account security version is expected to be used in many other places that have access to the key store and we do not want to pass through an additional struct to all places.

Please see the tickets for context about what specifically this change achieves.

Further, this fixes the icon uri check to be constant time.

## ⏰ Reminders before review

- CI builds passed
  team

## 🦮 Reviewer guidelines


mzieniukbw (dismissed):

crates/bitwarden-crypto/src/store/mod.rs
```diff
@@ -100,6 +100,7 @@ pub struct KeyStore<Ids: KeyIds> {
     // We use an Arc<> to make it easier to pass this store around, as we can
     // clone it instead of passing references
     inner: Arc<RwLock<KeyStoreInner<Ids>>>,
+    security_state_version: Arc<RwLock<u64>>,
```
dani-garcia: Should this instead be a part of `KeyStoreInner` to reuse the existing lock?

Having two locks seems wasteful and I feel like could lead to accidental deadlocks. In fact the current code in `context_mut()` smells like it could cause a deadlock to me, as we get a write lock for the `global_keys` and then we try to read the security state while holding the lock. If any other part of the codebase read those two in the opposite order, we'd have issues.

quexten: Should be fixed now

crates/bitwarden-crypto/src/store/mod.rs
```diff
@@ -170,11 +178,13 @@ impl<Ids: KeyIds> KeyStore<Ids> {
     ///     - [KeyStoreContext::encapsulate_key_unsigned]
     ///     - [KeyStoreContext::derive_shareable_key]
     pub fn context(&'_ self) -> KeyStoreContext<'_, Ids> {
+        let data = self.inner.read().expect("RwLock is poisoned");
         KeyStoreContext {
             global_keys: GlobalKeys::ReadOnly(self.inner.read().expect("RwLock is poisoned")),
```
dani-garcia: Nitpick: You can probably restructure this in the same way of context_mut to avoid calling inner.read() twice. it won't ever deadlock so this isn't a blocking problem, but it's mildly wasteful to do it twice if not needed.

crates/bitwarden-crypto/src/store/mod.rs
```diff
@@ -138,6 +140,12 @@ impl<Ids: KeyIds> KeyStore<Ids> {
         keys.signing_keys.clear();
     }
 
+    /// Sets the security state version for this store.
+    pub fn set_security_state_version(&self, version: u64) {
+        let mut data = self.inner.write().expect("RwLock is poisoned");
+        data.security_state_version = version;
```
dani-garcia, line 146: Non-blocking question: Would we want to add some rollback protection here in the future? Something like:
```rust
if version < data.security_state_version {
    return Err(...);
}
```

quexten: Yes! However, I do think this should be considered when sync and at least the crypto state is managed in SDK.

crates/bitwarden-crypto/src/store/mod.rs
quexten: Fixed in [178fe43](https://github.com/bitwarden/sdk-internal/pull/519/commits/178fe437afa7abbcf8ebd900edbfd33348050848)

dani-garcia (approved):

mzieniukbw (approved):

gbubemismith (approved):

PR #521, opened 2025-10-21
Accept SecureNote with Type 1 and set to 0

Jingo88:
## 🎟️ Tracking

[PM-27214](https://bitwarden.atlassian.net/browse/PM-27214)

## 📔 Objective

Some users have Notes with a `Type:1` in their data. This was breaking some vaults as failures were returning. Changes allow for Cipher Notes to have Type:1 and setting them to 0

## ⏰ Reminders before review

- CI builds passed
  team

## 🦮 Reviewer guidelines

[PM-27214]: https://bitwarden.atlassian.net/browse/PM-27214?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/crates/bitwarden-vault/src/cipher/secure_note.rs
+++ b/crates/bitwarden-vault/src/cipher/secure_note.rs
@@ -4,8 +4,8 @@ use bitwarden_core::{
     require,
 };
 use bitwarden_crypto::{CompositeEncryptable, CryptoError, Decryptable, KeyStoreContext};
-use serde::{Deserialize, Serialize};
-use serde_repr::{Deserialize_repr, Serialize_repr};
+use serde::{Deserialize, Deserializer, Serialize};
+use serde_repr::Serialize_repr;
 #[cfg(feature = "wasm")]
 use tsify::Tsify;
 #[cfg(feature = "wasm")]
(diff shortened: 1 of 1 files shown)


shane-melton (dismissed):

Hinton (commented):
Do you know how the mobile apps currently handles this? We haven't had any issues there and they've used this logic since those apps were built. I suspect they handle the failures on their mapping side.

> // This allows the SDK to handle future secure note types gracefully

I don't quite agree on this, a future type can not generally be mapped to an existing type without potential data loss.

Jingo88:
That's fair about the comment, it can be a little misleading. I'll remove it. 

Currently iOS will default to 0. Android throws an exception.

Hinton:
Are we concerned about the different behaviour between mobile and clients? Mobile never goes through the json layer and the 0/1 types are handled in the mapping layer.

Jingo88:
> Are we concerned about the different behaviour between mobile and clients? Mobile never goes through the json layer and the 0/1 types are handled in the mapping layer.

@Hinton No real concerns, the behavior is already different between Mobile and Clients and this is bringing the behavior more in line with the iOS mapping that already automatically maps 1>0

shane-melton (approved):

Hinton (approved):
I still have reservations around mobile handling this in the conversion layer and web handles this by loosely parsing serialized json it's inconsistent. As we shift the response parsing into the SDK this will stop working since we won't manually parse the json and will need to handle this in the `impl From<SecureNoteResponse> for SecureNote`.

Once we have data migrations one of the first migrations should be to get rid of notes with values other than 1 since this will cause headache should we ever want another note type.

Also please ensure Android is updated to handle this gracefully as I understand iOS parses unknown to 1, but Android throws errors.

PR #539, opened 2025-10-29
Add PRF passkey authenticator support and external passkeys

iinuwa:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-27582
https://bitwarden.atlassian.net/browse/PM-26177

## 📔 Objective

This adds the ability to use PRF for passkeys and authenticate using key material tied to a device, not the user key. (We're calling this the "device auth key.") This will effectively allow native clients to act as login+unlock passkeys without requiring vault unlock.

- Incorporates upstream passkey-rs changes
- Adds separate "vault authenticator" and "device authenticator" to distinguish between device-bound flow
- Allows encrypting passkey using a provided external key (device binding key)
- Adds hmacSecret field to FIDO2 credential models

Depends on #494.

# Breaking Changes

- This renames the current `authenticator()` method to `vault_authenticator()` to clarify that this authenticator will operate on vault items, as opposed to the new `device_authenticator()` method, which operates on items stored on the device with a caller-provided encryption key.

## ⏰ Reminders before review

- CI builds passed
  team

## 🦮 Reviewer guidelines

### DIFF
--- a/crates/bitwarden-fido/src/authenticator.rs
+++ b/crates/bitwarden-fido/src/authenticator.rs
@@ -1,6 +1,6 @@
 use std::sync::Mutex;
 
-use bitwarden_core::Client;
+use bitwarden_core::{Client, UserId};
 use bitwarden_crypto::{CompositeEncryptable, CryptoError, SymmetricCryptoKey};
 use bitwarden_vault::{CipherError, CipherView, EncryptionContext};
 use itertools::Itertools;
@@ -382,6 +382,40 @@ pub(super) struct CredentialStoreImpl<'a> {
     authenticator: &'a Fido2Authenticator<'a>,
     create_credential: bool,
 }
+
+impl CredentialStoreImpl<'_> {
+    async fn encrypt_and_store(
+        &self,
+        selected: CipherView,
+        user_id: UserId,
+    ) -> Result<(), CredentialStoreError> {
+        let key_store = self.authenticator.client.internal.get_key_store();
+        self.authenticator
+            .selected_cipher
+            .lock()
+            .expect("Mutex is not poisoned")
+            .replace(selected.clone());
+
+        // Encrypt with the specified encryption key if given, otherwise, use default key.
+        let cipher = if let Some(encryption_key) = &self.authenticator.encryption_key {
+            let mut ctx = key_store.context();
+            let key_id = ctx.add_local_symmetric_key(encryption_key.clone());
+            selected.encrypt_composite(&mut ctx, key_id)?
+        } else {
+            key_store.encrypt(selected)?
+        };
+
+        self.authenticator
+            .credential_store
+            .save_credential(EncryptionContext {
+                cipher,
+                encrypted_for: user_id,
+            })
+            .await?;
+        Ok(())
+    }
+}
+
 pub(super) struct UserValidationMethodImpl<'a> {
     authenticator: &'a Fido2Authenticator<'a>,
 }
@@ -485,23 +519,6 @@ impl passkey::authenticator::CredentialStore for CredentialStoreImpl<'_> {
         rp: passkey::types::ctap2::make_credential::PublicKeyCredentialRpEntity,
         options: passkey::types::ctap2::get_assertion::Options,
     ) -> Result<(), StatusCode> {
-        #[derive(Debug, Error)]
-        enum InnerError {
-            #[error("Client User Id has not been set")]
-            MissingUserId,
-            #[error(transparent)]
-            FillCredential(#[from] FillCredentialError),
-            #[error(transparent)]
-            Cipher(#[from] CipherError),
-            #[error(transparent)]
-            Crypto(#[from] CryptoError),
-            #[error(transparent)]
-            Fido2Callback(#[from] Fido2CallbackError),
-
-            #[error("No selected credential available")]
-            NoSelectedCredential,
-        }
-
         // This is just a wrapper around the actual implementation to allow for ? error handling
         async fn inner(
             this: &mut CredentialStoreImpl<'_>,
@@ -509,13 +526,13 @@ impl passkey::authenticator::CredentialStore for CredentialStoreImpl<'_> {
             user: passkey::types::ctap2::make_credential::PublicKeyCredentialUserEntity,
             rp: passkey::types::ctap2::make_credential::PublicKeyCredentialRpEntity,
             options: passkey::types::ctap2::get_assertion::Options,
-        ) -> Result<(), InnerError> {
+        ) -> Result<(), CredentialStoreError> {
             let user_id = this
                 .authenticator
                 .client
                 .internal
                 .get_user_id()
-                .ok_or(InnerError::MissingUserId)?;
+                .ok_or(CredentialStoreError::MissingUserId)?;
             let cred = try_from_credential_full(cred, user, rp, options)?;
 
             // Get the previously selected cipher and add the new credential to it
@@ -525,11 +542,10 @@ impl passkey::authenticator::CredentialStore for CredentialStoreImpl<'_> {
                 .lock()
                 .expect("Mutex is not poisoned")
                 .clone()
-                .ok_or(InnerError::NoSelectedCredential)?;
-
-            let key_store = this.authenticator.client.internal.get_key_store();
+                .ok_or(CredentialStoreError::NoSelectedCredential)?;
 
             {
+                let key_store = this.authenticator.client.internal.get_key_store();
                 let mut ctx = key_store.context();
                 let key_id = this
                     .authenticator
@@ -540,29 +556,8 @@ impl passkey::authenticator::CredentialStore for CredentialStoreImpl<'_> {
             }
 
             // Store the updated credential for later use
-            this.authenticator
-                .selected_cipher
-                .lock()
-                .expect("Mutex is not poisoned")
-                .replace(selected.clone());
-
-            let cipher = if let Some(encryption_key) = &this.authenticator.encryption_key {
-                let mut ctx = key_store.context();
-                let key_id = ctx.add_local_symmetric_key(encryption_key.clone());
-                selected.encrypt_composite(&mut ctx, key_id)?
-            } else {
-                // Encrypt the updated cipher before sending it to the clients to be stored
-                key_store.encrypt(selected)?
-            };
-
-            this.authenticator
-                .credential_store
-                .save_credential(EncryptionContext {
-                    cipher,
-                    encrypted_for: user_id,
-                })
-                .await?;
-
+            // Encrypt the updated cipher before sending it to the clients to be stored
+            this.encrypt_and_store(selected, user_id).await?;
             Ok(())
         }
 
@@ -575,37 +570,17 @@ impl passkey::authenticator::CredentialStore for CredentialStoreImpl<'_> {
     }
 
     async fn update_credential(&mut self, cred: Passkey) -> Result<(), StatusCode> {
-        #[derive(Debug, Error)]
-        enum InnerError {
-            #[error("Client User Id has not been set")]
-            MissingUserId,
-            #[error(transparent)]
-            InvalidGuid(#[from] InvalidGuidError),
-            #[error("Credential ID does not match selected credential")]
-            CredentialIdMismatch,
-            #[error(transparent)]
-            FillCredential(#[from] FillCredentialError),
-            #[error(transparent)]
-            Cipher(#[from] CipherError),
-            #[error(transparent)]
-            Crypto(#[from] CryptoError),
-            #[error(transparent)]
-            Fido2Callback(#[from] Fido2CallbackError),
-            #[error(transparent)]
-            GetSelectedCredential(#[from] GetSelectedCredentialError),
-        }
-
         // This is just a wrapper around the actual implementation to allow for ? error handling
         async fn inner(
             this: &mut CredentialStoreImpl<'_>,
             cred: Passkey,
-        ) -> Result<(), InnerError> {
+        ) -> Result<(), CredentialStoreError> {
             let user_id = this
                 .authenticator
                 .client
                 .internal
                 .get_user_id()
-                .ok_or(InnerError::MissingUserId)?;
+                .ok_or(CredentialStoreError::MissingUserId)?;
             // Get the previously selected cipher and update the credential
             let selected = this.authenticator.get_selected_credential()?;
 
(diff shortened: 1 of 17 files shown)


crates/bitwarden-uniffi/src/platform/fido2.rs
```diff
@@ -25,9 +28,38 @@ impl ClientFido2 {
...
+    /// with the given external encryption key, not the user key, so the same
+    /// device-specific key must be given to use credentials created with this
+    /// authenticator.
+    pub fn device_authenticator(
+        &self,
+        user_interface: Arc<dyn Fido2UserInterface>,
+        credential_store: Arc<dyn Fido2CredentialStore>,
+        encryption_key: Vec<u8>,
+    ) -> Result<Arc<ClientFido2Authenticator>, CryptoError> {
+        let key = SymmetricCryptoKey::try_from(&BitwardenLegacyKeyBytes::from(encryption_key))?;
```
iinuwa, line 51: Two questions here:
- Should we use a KDF to derive a key with a key purpose?
  - That would allow us to just store a single "device-binding key" on the OS keychain to use for, say, the device auth key and some future device-bound vault items.
- Should we use a COSE key instead? I didn't see an easy way to create one of those from raw key material.

crates/bitwarden-uniffi/src/platform/fido2.rs
```diff
@@ -58,6 +99,7 @@ pub struct ClientFido2Authenticator(
     pub(crate) bitwarden_fido::ClientFido2,
     pub(crate) Arc<dyn Fido2UserInterface>,
     pub(crate) Arc<dyn Fido2CredentialStore>,
+    pub(crate) Fido2AuthenticatorOptions,
```
dani-garcia, line 102: As a future improvement, we should probably consider moving this to a normal struct, rather than deal with them by tuple index.

crates/bitwarden-uniffi/src/platform/fido2.rs
dani-garcia: Good question! I think KM would have more expertise to answer these, you might want to give them a ping.

dani-garcia (approved):
Changes LGTM, though I'm not an expert on PRF internals

crates/bitwarden-uniffi/src/platform/fido2.rs
iinuwa: @quexten, since I spoke most recently about this PR with you, do you have any thoughts about whether we should use a KDF for this purpose?

quexten: So, if I understand your use-case right, you have some material M from the keychain, and you want to get Keys K1...Kn with it?

I would actually recommend just using keywrap here instead of deriving keys. The layer of indirection / decoupling this helps when we want to ever change M without breaking K1...Kn. The "downside" is a few bytes more that we store, which seems reasonable. And, all KM interfaces are designed for this pattern, not key derivation.

LMK if that works for your use-case.

As to using cose or not; A cose (symmetric) key is just 32 bytes, an algorithm, and a few bytes of key id. We could make a function to get a key from material. Do you have a specific amount of bytes that M is large, or can you store any size? I'm wondering whether this function would need to stretch the material or whether we can just take it directly.

iinuwa: tl;dr: I think I'm overcomplicating this and maybe we should forget about it for now.

> So, if I understand your use-case right, you have some material M from the keychain, and you want to get Keys K1...Kn with it?
>
> I would actually recommend just using keywrap here instead of deriving keys. The layer of indirection / decoupling this helps when we want to ever change M without breaking K1...Kn. The "downside" is a few bytes more that we store, which seems reasonable. And, all KM interfaces are designed for this pattern, not key derivation.

Maybe?

**Device auth key use case**
On the caller side, what we have now is this:
```
RawDeviceKey = Random(64)
ProtectedDeviceKey = KeychainProtect(RawDeviceKey)

// for creation
Fido2CredCiphertext = client_fido2
    .device_authenticator(ui, CredStore(nil), KeychainUnprotect(ProtectedDeviceKey))
    .make_credential(options)
DeviceAuthKeyRecord = KeychainProtect(Fido2CredCiphertext)

// for assertion
Assertion_n = client_fido2
    .device_authenticator(ui, CredStore(KeychainUnprotect(DeviceAuthKeyRecord)), KeychainUnprotect(ProtectedDeviceKey))
    .get_assertion(options_n)
```

The device auth key lives outside of the vault, and we're using a symmetric key (also stored on the device) to encrypt the data in the keychain. This is not really for security, since the keychain provides the proper access controls, but rather to fit this new SDK API. (The existing SDK API encrypted records with the user key, but to make this independent from the user key, I modified it to allow you to choose to encrypt with either the user key or an external key, rather than giving the option between user key encryption and no encryption. That seemed harder to misuse.)

Because we're relying on the keychain for our primary protection rather than the cryptography, we shouldn't need key rotation and therefore, I don't think that key wrapping provides any benefit. Do you agree?

**Other device-bound use cases**
If we were going with only this use case, I think that the above would be fine. However, there are some ideas of making "device bound" ciphers, where a cipher would be encrypted with some combination of the user key and a key stored in the device keychain. In that case, we would probably want key rotation.

However, I think I'm overcomplicating things though for a feature that has not been designed yet. We don't even need the device auth key and whatever key we use to bind ciphers to the device to be the same key.

> As to using cose or not; A cose (symmetric) key is just 32 bytes, an algorithm, and a few bytes of key id. We could make a function to get a key from material. Do you have a specific amount of bytes that M is large, or can you store any size? I'm wondering whether this function would need to stretch the material or whether we can just take it directly.

We can generate and store an arbitrary amount of secure random bytes for M. But the same as above, since on the caller-side, we're relying on OS secure storage rather than the cryptography for security, I don't think that we need to worry about the particular algorithm.

crates/bitwarden-uniffi/src/platform/fido2.rs
```diff
@@ -37,16 +69,25 @@ impl ClientFido2 {
...
+                enable_hmac_secret: false,
+                external_encryption_key: None,
+            },
         )))
     }
 
     pub fn decrypt_fido2_autofill_credentials(
         &self,
         cipher_view: CipherView,
+        encryption_key: Option<Vec<u8>>,
```
quexten, line 82: This needs docs. What happens when an encryption key is passed? When do you pass it? What's the behavior?

Or, much rather, this should just be a separate function, not a mutation of the existing function, at least for the public interface that we expose out of the SDK. The mis-use potential is high here.

(As I gather, this captures two entirely distinct use cases. One is decrypting vault items encrypted by the account encryption key, the other is decrypting a vault item used as a device bound passkey, protected by a keychain-only key.

crates/bitwarden-vault/src/cipher/cipher_client/mod.rs
```diff
@@ -142,9 +142,16 @@ impl CiphersClient {
     pub fn decrypt_fido2_credentials(
         &self,
         cipher_view: CipherView,
+        encryption_key: Option<Vec<u8>>,
     ) -> Result<Vec<crate::Fido2CredentialView>, DecryptError> {
         let key_store = self.client.internal.get_key_store();
-        let credentials = cipher_view.decrypt_fido2_credentials(&mut key_store.context())?;
+        let mut ctx = key_store.context();
```
quexten, line 148: This needs docs. What happens when an encryption key is passed? When do you pass it? What's the behavior?

Or, much rather, this should just be a separate function, not a mutation of the existing function, at least for the public interface that we expose out of the SDK. The mis-use potential is high here.

(As I gather, this captures two entirely distinct use cases. One is decrypting vault items encrypted by the account encryption key, the other is decrypting a vault item used as a device bound passkey, protected by a keychain-only key.

crates/bitwarden-vault/src/cipher/cipher_client/mod.rs
```diff
@@ -182,9 +200,15 @@ impl CiphersClient {
     pub fn decrypt_fido2_private_key(
         &self,
         cipher_view: CipherView,
+        key: Option<Vec<u8>>,
```
quexten, line 203: This needs docs. What happens when an encryption key is passed? When do you pass it? What's the behavior?

Or, much rather, this should just be a separate function, not a mutation of the existing function, at least for the public interface that we expose out of the SDK. The mis-use potential is high here.

(As I gather, this captures two entirely distinct use cases. One is decrypting vault items encrypted by the account encryption key, the other is decrypting a vault item used as a device bound passkey, protected by a keychain-only key.

crates/bitwarden-uniffi/src/platform/fido2.rs
```diff
@@ -25,9 +28,38 @@ impl ClientFido2 {
...
+    ///
+    /// The credentials created with this authenticator are encrypted
+    /// with the given external encryption key, not the user key, so the same
+    /// device-specific key must be given to use credentials created with this
+    /// authenticator.
+    pub fn device_authenticator(
+        &self,
+        user_interface: Arc<dyn Fido2UserInterface>,
+        credential_store: Arc<dyn Fido2CredentialStore>,
+        encryption_key: Vec<u8>,
```
quexten, line 49: @dani-garcia Maybe you can clarify this better but I was under the assumption that we used b64 encoding for uniffi / wasm because uniffi (swift / kotlin) don't handle byte arrays that well. Is this still true?

crates/bitwarden-fido/src/types.rs
```diff
@@ -71,8 +79,9 @@ impl Fido2CredentialAutofillView {
     pub fn from_cipher_view(
         cipher: &CipherView,
         ctx: &mut KeyStoreContext<KeyIds>,
+        key_id: Option<SymmetricKeyId>,
```
quexten, line 82: Needs docs or should be a separate function. Which key id gets used if this is not provided?

crates/bitwarden-fido/src/authenticator.rs
```diff
@@ -342,10 +366,56 @@ impl<'a> Fido2Authenticator<'a> {
...
+            .selected_cipher
+            .lock()
+            .expect("Mutex is not poisoned")
+            .replace(selected.clone());
+
+        // Encrypt with the specified encryption key if given, otherwise, use default key.
+        let cipher = if let Some(encryption_key) = &self.authenticator.encryption_key {
+            let mut ctx = key_store.context();
+            let key_id = ctx.add_local_symmetric_key(encryption_key.clone());
+            selected.encrypt_composite(&mut ctx, key_id)?
```
quexten, line 403: As a general question; If we ever want to upgrade this cipher to a newer version - (since vault is introducing versioned ciphers) - how would this work? Currently the construct is such that you need UV to access the key, and thus you cannot migrate without UV, even if you have access to the vault.

Another construction that would fix this issue would be to have:

UserKey   -> Intermediate Key -> My authenticator item
DeviceKey - ^

But this has it's own trade-offs. I guess my concern is that vault items are not guaranteed to be a stable format, but are treated here as such.

crates/bitwarden-uniffi/src/platform/fido2.rs
quexten: > Because we're relying on the keychain for our primary protection rather than the cryptography, we shouldn't need key rotation and therefore, I don't think that key wrapping provides any benefit. Do you agree?

I do not agree. The key is not always accessible, similar to how a Biometric unlock key is not always accessible without UV, or how the masterkey was not accessible depending on the unlock method. 

This makes it impossible to update the encrypted data - here the vault item that acts as the device bound passkey. Vault items are not a stable format and are expected to change soon with vault item versioning, so I do believe there is migration considerations needed.

I would be much less concerned if this was not tied into using vault items directly.

quexten:
I'm not convinced about the re-use of vault items for this purpose, and not just use a different struct entirely to be honest. It can look the same, but should be a different struct. Vault items are a vault feature, and this initiative currently binds vault items as a format to be permanently stable and supported. 

Vault is currently already working on migrating to a different representation - blob encryption with internal versioning - and it is unclear to me what their backwards compatibility story looks like, but for certain cases, compatibility will be dropped. I'm also sure they'd rather at some point split out the legacy code, which is prevented by keeping a non-rotateable local copy here.

I want to include vault into this discussion because it's their stability contract to offer or not offer / I'm not sure what this looks like.

cc @shane-melton 

It seems there are three paths forward:
- We use a non-vault item stable representation
- Vault agrees to indefinitely support this current vault item representation
- We add an intermediary key that can be used to migrate without UV (or make the UV enforced in software only, rather than by the keychain)

crates/bitwarden-uniffi/src/platform/fido2.rs
dani-garcia: As far as I remember, the problem with `Vec<u8>` is it get's translated as `Data` in iOS, and that doesn't implement equality or hash, or maybe it's only doing referencial equality rather than value equality, not fully sure of the details. The problem appears when we try to use `Vec<u8>` in model types like `Cipher` because the iOS app is expecting to be able to compare `Cipher`'s contents.

In this case, I think using it as a function parameter wouldn't cause an issue either way, but yeah it's probably a good idea to change it to `B64` just for consistency's sake.

crates/bitwarden-uniffi/src/platform/fido2.rs
iinuwa: > This needs docs. What happens when an encryption key is passed? When do you pass it? What's the behavior?

My apologies, I neglected to add docs to all the methods that this parameter flows through.

In this implementation, if no `encryption_key` is specified, the default encryption (`UserKey` or `OrgKey` depending on the cipher) will be used for decryption. If it is specified, then `encryption_key` will be used instead of `UserKey`/`OrgKey`.


> As I gather, this captures two entirely distinct use cases. One is decrypting vault items encrypted by the account encryption key, the other is decrypting a vault item used as a device bound passkey, protected by a keychain-only key.

Yes, except that in the case of device auth key, we are decrypting a record, also stored in the keychain, protected by a keychain-only key.

> Or, much rather, this should just be a separate function, not a mutation of the existing function, at least for the public interface that we expose out of the SDK. The mis-use potential is high here.

Thanks for that feedback. The shape of this API is very much influenced by our wrapper around passkey-rs. Based on your comments, I think that I'll add a new wrapper around passkey-rs that is purpose-built for the device auth key. That requires some unraveling to be done at a higher level. I'll go back to the drawing board on this.

harr1424 (approved):
Changes to `bitwarden-exporters` look good

iinuwa:
After discussion, I'm closing this in favor of a more focused API in #816.

PR #549, opened 2025-11-07
[PM-14922][PM-27061] - LoginClient - Login with password MVP

JaredSnider-Bitwarden:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-14922
https://bitwarden.atlassian.net/browse/PM-27061

## 📔 Objective

To deliver an MVP of login with password in the SDK. This will not include 2FA or new device verification support at first. 

## ⏰ Reminders before review

- CI builds passed
  team

## 🦮 Reviewer guidelines


crates/bitwarden-auth/src/api/enums/two_factor_provider.rs
JaredSnider-Bitwarden, line 1: This was copied over from core b/c we eventually want auth to own these files.

crates/bitwarden-core/src/key_management/mod.rs
```diff
@@ -24,9 +24,11 @@ pub use crypto_client::CryptoClient;
 #[cfg(feature = "internal")]
 mod master_password;
 #[cfg(feature = "internal")]
+pub use master_password::MasterPasswordAuthenticationData;
+#[cfg(feature = "internal")]
 pub use master_password::MasterPasswordError;
 #[cfg(feature = "internal")]
-pub(crate) use master_password::{MasterPasswordAuthenticationData, MasterPasswordUnlockData};
+pub use master_password::MasterPasswordUnlockData;
```
quexten: Nit: These have the same feature compile time flag, and the same package prefix, you can replace with `pub use master_password:{MasterPasswordAuthenticationData, MasterPasswordError, MasterPasswordUnlockData}`

crates/bitwarden-core/src/key_management/master_password.rs
```diff
@@ -126,11 +126,8 @@ pub struct MasterPasswordAuthenticationData {
 }
 
 impl MasterPasswordAuthenticationData {
-    pub(crate) fn derive(
-        password: &str,
-        kdf: &Kdf,
-        salt: &str,
-    ) -> Result<Self, MasterPasswordError> {
+    #[allow(missing_docs)]
```
quexten:
```suggestion
/// Derives the authentication data for registration or unlock for an account, given the password, kdf settings and salt of the user.
```

crates/bitwarden-auth/src/identity/models/webauthn_prf_user_decryption_option.rs
```diff
@@ -0,0 +1,48 @@
...
+#[cfg_attr(
+    feature = "wasm",
+    derive(tsify::Tsify),
+    tsify(into_wasm_abi, from_wasm_abi)
+)]
+pub struct WebAuthnPrfUserDecryptionOption {
+    /// PRF key encrypted private key
+    pub encrypted_private_key: EncString,
+
+    /// Private Key encrypted user key
```
quexten: PublicKeyEncryptedUserKey?

quexten: This is not an encstring, that's just the wrong type. Public-Key encryption in the SDK is UnsignedSharedKey

crates/bitwarden-auth/src/identity/models/webauthn_prf_user_decryption_option.rs
```diff
@@ -0,0 +1,48 @@
...
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_webauthn_prf_conversion() {
+        let api = WebAuthnPrfUserDecryptionOptionApiResponse {
+            encrypted_private_key: "2.fkvl0+sL1lwtiOn1eewsvQ==|dT0TynLl8YERZ8x7dxC+DQ==|cWhiRSYHOi/AA2LiV/JBJWbO9C7pbUpOM6TMAcV47hE=".parse().unwrap(),
+            encrypted_user_key: "2.pMS6/icTQABtulw52pq2lg==|XXbxKxDTh+mWiN1HjH2N1w==|Q6PkuT+KX/axrgN9ubD5Ajk2YNwxQkgs3WJM0S0wtG8=".parse().unwrap(),
```
quexten: I think this test vector is wrong, but succeeds because the type of this entry is wrong. It should start with 4., as it is an UnsignedSharedKey

crates/bitwarden-auth/src/login/models/webauthn_prf_user_decryption_option.rs
```diff
@@ -0,0 +1,48 @@
...
+/// SDK domain model for WebAuthn PRF user decryption option.
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq)]
+#[serde(rename_all = "camelCase")]
+#[cfg_attr(feature = "uniffi", derive(uniffi::Record))]
+#[cfg_attr(
+    feature = "wasm",
+    derive(tsify::Tsify),
+    tsify(into_wasm_abi, from_wasm_abi)
+)]
+pub struct WebAuthnPrfUserDecryptionOption {
```
quexten, line 15: Is there opportunity to just re-use https://github.com/bitwarden/sdk-internal/blob/0794e100645d97d0359551939673b2d09db2204d/crates/bitwarden-crypto/src/keys/rotateable_key_set.rs#L27? I see that here the "userkeyencryptedpublickey" would be missing, but it seems otherwise this is duplicated?

crates/bitwarden-auth/src/login/models/user_decryption_options_response.rs
```diff
@@ -0,0 +1,402 @@
...
+        assert!(domain.master_password_unlock.is_none());
+        assert!(domain.trusted_device_option.is_none());
+        assert!(domain.key_connector_option.is_none());
+        assert!(domain.webauthn_prf_option.is_none());
+    }
+
+    #[test]
+    fn test_user_decryption_options_with_master_password_and_trusted_device() {
+        const TDE_ENCRYPTED_PRIVATE_KEY: &str = "2.pMS6/icTQABtulw52pq2lg==|XXbxKxDTh+mWiN1HjH2N1w==|Q6PkuT+KX/axrgN9ubD5Ajk2YNwxQkgs3WJM0S0wtG8=";
+        const TDE_ENCRYPTED_USER_KEY: &str = "2.kTtIypq9OLzd5iMMbU11pQ==|J4i3hTtGVdg7EZ+AQv/ujg==|QJpSpotQVpIW8j8dR/8l015WJzAIxBaOmrz4Uj/V1JA=";
```
quexten: `4.` this is unsigned shared key not encstring

crates/bitwarden-auth/src/identity/models/user_decryption_options_response.rs
```diff
@@ -0,0 +1,402 @@
...
+        assert!(domain.master_password_unlock.is_some());
+        assert!(domain.trusted_device_option.is_none());
+        assert!(domain.key_connector_option.is_some());
+        assert!(domain.webauthn_prf_option.is_none());
+    }
+
+    #[test]
+    fn test_user_decryption_options_with_master_password_and_webauthn_prf() {
+        const WEBAUTHN_ENCRYPTED_PRIVATE_KEY: &str = "2.fkvl0+sL1lwtiOn1eewsvQ==|dT0TynLl8YERZ8x7dxC+DQ==|cWhiRSYHOi/AA2LiV/JBJWbO9C7pbUpOM6TMAcV47hE=";
+        const WEBAUTHN_ENCRYPTED_USER_KEY: &str = "2.pMS6/icTQABtulw52pq2lg==|XXbxKxDTh+mWiN1HjH2N1w==|Q6PkuT+KX/axrgN9ubD5Ajk2YNwxQkgs3WJM0S0wtG8=";
```
quexten: `4.` (unsigned shared key)

crates/bitwarden-auth/src/identity/models/user_decryption_options_response.rs
```diff
@@ -0,0 +1,402 @@
...
+        assert!(domain.webauthn_prf_option.is_none());
+
+        let kc = domain.key_connector_option.unwrap();
+        assert_eq!(kc.key_connector_url, KEY_CONNECTOR_URL);
+    }
+
+    #[test]
+    fn test_user_decryption_options_with_webauthn_prf_only() {
+        const WEBAUTHN_ENCRYPTED_PRIVATE_KEY: &str = "2.fkvl0+sL1lwtiOn1eewsvQ==|dT0TynLl8YERZ8x7dxC+DQ==|cWhiRSYHOi/AA2LiV/JBJWbO9C7pbUpOM6TMAcV47hE=";
+        const WEBAUTHN_ENCRYPTED_USER_KEY: &str = "2.pMS6/icTQABtulw52pq2lg==|XXbxKxDTh+mWiN1HjH2N1w==|Q6PkuT+KX/axrgN9ubD5Ajk2YNwxQkgs3WJM0S0wtG8=";
```
quexten: `4.` (unsignedSharedKey)

crates/bitwarden-auth/src/login/models/user_decryption_options_response.rs
```diff
@@ -0,0 +1,402 @@
...
+        assert_eq!(
+            webauthn.encrypted_user_key,
+            WEBAUTHN_ENCRYPTED_USER_KEY.parse().unwrap()
+        );
+    }
+
+    #[test]
+    fn test_user_decryption_options_with_trusted_device_only() {
+        const TDE_ENCRYPTED_PRIVATE_KEY: &str = "2.pMS6/icTQABtulw52pq2lg==|XXbxKxDTh+mWiN1HjH2N1w==|Q6PkuT+KX/axrgN9ubD5Ajk2YNwxQkgs3WJM0S0wtG8=";
+        const TDE_ENCRYPTED_USER_KEY: &str = "2.kTtIypq9OLzd5iMMbU11pQ==|J4i3hTtGVdg7EZ+AQv/ujg==|QJpSpotQVpIW8j8dR/8l015WJzAIxBaOmrz4Uj/V1JA=";
```
quexten: `4.` (UnsignedSharedKey)

crates/bitwarden-auth/src/identity/models/user_decryption_options_response.rs
```diff
@@ -0,0 +1,402 @@
...
+    #[test]
+    fn test_user_decryption_options_conversion_with_all_options() {
+        // Test data constants
+        const SALT: &str = "test@example.com";
+        const KDF_ITERATIONS: u32 = 600000;
+        const TDE_ENCRYPTED_PRIVATE_KEY: &str = "2.pMS6/icTQABtulw52pq2lg==|XXbxKxDTh+mWiN1HjH2N1w==|Q6PkuT+KX/axrgN9ubD5Ajk2YNwxQkgs3WJM0S0wtG8=";
+        const TDE_ENCRYPTED_USER_KEY: &str = "2.kTtIypq9OLzd5iMMbU11pQ==|J4i3hTtGVdg7EZ+AQv/ujg==|QJpSpotQVpIW8j8dR/8l015WJzAIxBaOmrz4Uj/V1JA=";
+        const KEY_CONNECTOR_URL: &str = "https://key-connector.bitwarden.com";
+        const WEBAUTHN_ENCRYPTED_PRIVATE_KEY: &str = "2.fkvl0+sL1lwtiOn1eewsvQ==|dT0TynLl8YERZ8x7dxC+DQ==|cWhiRSYHOi/AA2LiV/JBJWbO9C7pbUpOM6TMAcV47hE=";
+        const WEBAUTHN_ENCRYPTED_USER_KEY: &str = "2.pMS6/icTQABtulw52pq2lg==|XXbxKxDTh+mWiN1HjH2N1w==|Q6PkuT+KX/axrgN9ubD5Ajk2YNwxQkgs3WJM0S0wtG8=";
```
quexten: `4.` (UnsignedSharedKey)

crates/bitwarden-auth/src/login/models/user_decryption_options_response.rs
```diff
@@ -0,0 +1,402 @@
...
+        assert!(domain.webauthn_prf_option.is_none());
+    }
+
+    #[test]
+    fn test_user_decryption_options_conversion_with_all_options() {
+        // Test data constants
+        const SALT: &str = "test@example.com";
+        const KDF_ITERATIONS: u32 = 600000;
+        const TDE_ENCRYPTED_PRIVATE_KEY: &str = "2.pMS6/icTQABtulw52pq2lg==|XXbxKxDTh+mWiN1HjH2N1w==|Q6PkuT+KX/axrgN9ubD5Ajk2YNwxQkgs3WJM0S0wtG8=";
+        const TDE_ENCRYPTED_USER_KEY: &str = "2.kTtIypq9OLzd5iMMbU11pQ==|J4i3hTtGVdg7EZ+AQv/ujg==|QJpSpotQVpIW8j8dR/8l015WJzAIxBaOmrz4Uj/V1JA=";
```
quexten: `4.` (UnsignedSharedKey)

crates/bitwarden-auth/src/login/models/trusted_device_user_decryption_option.rs
```diff
@@ -0,0 +1,80 @@
...
+
+    #[test]
+    fn test_trusted_device_conversion() {
+        let api = TrustedDeviceUserDecryptionOptionApiResponse {
+            has_admin_approval: true,
+            has_login_approving_device: false,
+            has_manage_reset_password_permission: true,
+            is_tde_offboarding: false,
+            encrypted_private_key: Some("2.pMS6/icTQABtulw52pq2lg==|XXbxKxDTh+mWiN1HjH2N1w==|Q6PkuT+KX/axrgN9ubD5Ajk2YNwxQkgs3WJM0S0wtG8=".parse().unwrap()),
+            encrypted_user_key: Some("2.kTtIypq9OLzd5iMMbU11pQ==|J4i3hTtGVdg7EZ+AQv/ujg==|QJpSpotQVpIW8j8dR/8l015WJzAIxBaOmrz4Uj/V1JA=".parse().unwrap()),
```
quexten: `4.` (UnsignedSharedKey)

crates/bitwarden-auth/src/login/login_via_password/password_prelogin_response.rs
```diff
@@ -0,0 +1,290 @@
...
+    feature = "wasm",
+    derive(tsify::Tsify),
+    tsify(into_wasm_abi, from_wasm_abi)
+)] // add wasm support
+pub struct PasswordPreloginResponse {
+    /// The Key Derivation Function (KDF) configuration for the user
+    pub kdf: Kdf,
+
+    /// The salt used in the KDF process
+    pub salt: String,
```
quexten, line 23: Thought (not for this PR): we should make salt a newtype, since passing it around like this feels dangerous. You could put in a non-lower-cased email and think it's fine, but it would be wrong.

crates/bitwarden-auth/src/login/login_via_password/password_prelogin.rs
```diff
@@ -0,0 +1,242 @@
...
+/// Error type for password prelogin operations
+#[bitwarden_error(flat)]
+#[derive(Debug, Error)]
+pub enum PasswordPreloginError {
+    /// API error occurred during the prelogin request
+    #[error(transparent)]
+    Api(#[from] ApiError),
+    /// A required field was missing in the response
+    #[error(transparent)]
+    MissingField(#[from] MissingFieldError),
```
quexten: Comment/Question: Errors are public API, and your consumers may depend on it if you include the error like this. Is this intended?

Asked in another way, does a consumer behave differently depending on which type of missing field error they got?

If not, I would recommend making this an opaque error and possibly using tracing to log missing fields, to reduce API complexity and API contract surface. (KM currently violates this in many places in new and inherited code, and it will have to be refactored / changed).

crates/bitwarden-auth/src/login/login_via_password/password_prelogin.rs
```diff
@@ -0,0 +1,242 @@
...
+
+use crate::identity::{LoginClient, login_via_password::PasswordPreloginResponse};
+
+/// Error type for password prelogin operations
+#[bitwarden_error(flat)]
+#[derive(Debug, Error)]
+pub enum PasswordPreloginError {
+    /// API error occurred during the prelogin request
+    #[error(transparent)]
+    Api(#[from] ApiError),
```
quexten, line 16: Same thing here as below. Does a consumer care about *which* API error happened, or do they just care that *any* API error happened. Can they act differently depending on the error?

crates/bitwarden-auth/src/login/login_via_password/password_prelogin.rs
quexten: If the concern here is around getting user error reports, then the much better solution is the FlightRecorder platform is working on. It should capture the traces in case an error ocurred, and so we can focus on providing a good public contract here.

crates/bitwarden-auth/src/login/login_via_password/password_login_error.rs
```diff
@@ -0,0 +1,423 @@
...
+    /// - Malformed server responses
+    /// - Future error types not yet handled by this SDK version
+    ///
+    /// The contained string provides details about what went wrong.
+    ///
+    /// # Forward Compatibility
+    /// This variant ensures the SDK can handle new error types introduced by the server
+    /// without breaking existing client code.
+    #[error("Unknown password login error: {0}")]
+    Unknown(String),
```
quexten, line 52: Is there a reason for passing out the string like this? Why not use tracing + opaque error?

crates/bitwarden-auth/src/login/login_via_password/password_login_error.rs
```diff
@@ -0,0 +1,423 @@
...
+    /// and "wrong password" to prevent user enumeration attacks.
+    #[error("Invalid username or password provided.")]
+    InvalidUsernameOrPassword,
+
+    /// Failed to derive master password authentication data from the provided password and KDF
+    /// settings.
+    ///
+    /// This error occurs during local cryptographic processing before the API call, typically when:
+    /// - The KDF parameters are invalid (e.g., iterations below minimum threshold)
+    /// - The KDF algorithm is unsupported or corrupted
```
quexten, line 33: 2 lines above says "typical"; how is "kdf algorithm unsupported" a typical scenario?

crates/bitwarden-auth/src/identity/login_via_password/password_login_error.rs
```diff
@@ -0,0 +1,423 @@
...
+    /// settings.
+    ///
+    /// This error occurs during local cryptographic processing before the API call, typically when:
+    /// - The KDF parameters are invalid (e.g., iterations below minimum threshold)
+    /// - The KDF algorithm is unsupported or corrupted
+    /// - Memory allocation fails during Argon2id processing
+    ///
+    /// # Common Causes
+    /// - KDF iterations below the minimum security threshold (5000 for PBKDF2)
+    /// - Corrupted prelogin data
```
quexten: Seems these are duplicated with the 3 lines above?

crates/bitwarden-auth/src/api/enums/scope.rs
```diff
@@ -4,10 +4,35 @@ use serde::{Deserialize, Serialize};
...
+        match self {
+            Scope::Api => "api",
+            Scope::OfflineAccess => "offline_access",
+            Scope::ApiSendAccess => "api.send.access",
+        }
+    }
+}
+
+/// Converts a slice of scopes into a space-separated string suitable for OAuth 2.0 requests.
+pub(crate) fn scopes_to_string(scopes: &[Scope]) -> String {
```
quexten, line 32: Why not just implement `impl From<&[Scope]> for String`? This is way more ergonomic to use that a function like this.

crates/bitwarden-auth/src/login/api/response/login_error_api_response.rs
```diff
@@ -0,0 +1,500 @@
...
+use serde::Deserialize;
+
+#[derive(Deserialize, PartialEq, Eq, Debug)]
+#[serde(rename_all = "snake_case")]
+pub enum PasswordInvalidGrantError {
+    /// The username or password provided was invalid.
+    InvalidUsernameOrPassword,
+}
+
+// Actual 2fa rejection response for future use in TwoFactorInvalidGrantError
```
quexten, line 11: Is this meant to live here?

crates/bitwarden-auth/src/login/api/response/login_success_api_response.rs
```diff
@@ -0,0 +1,92 @@
...
+    /// This token can be used to obtain new access tokens when the current one expires.
+    pub refresh_token: Option<String>,
+
+    // Custom Bitwarden connect/token response fields:
+    // We send down uppercase fields today so we have to map them accordingly +
+    // we add aliases for deserialization flexibility.
+    /// The user key wrapped user private key
+    /// Deprecated in favor of the `AccountKeys` field but still present for backward
+    /// compatibility. and we can't expose AccountKeys in our LoginSuccessResponse until we get
+    /// a PrivateKeysResponseModel SDK response model from KM with WASM / uniffi support.
```
quexten, line 34: What specifically is missing here? We have:

https://github.com/bitwarden/sdk-internal/blob/0794e100645d97d0359551939673b2d09db2204d/crates/bitwarden-api-api/src/models/private_keys_response_model.rs#L20
and https://github.com/bitwarden/sdk-internal/blob/0794e100645d97d0359551939673b2d09db2204d/crates/bitwarden-core/src/key_management/account_cryptographic_state.rs#L72

crates/bitwarden-auth/src/identity/login_client.rs
```diff
@@ -0,0 +1,59 @@
...
+use wasm_bindgen::prelude::*;
+
+/// Client for authenticating Bitwarden users.
+///
+/// Handles unauthenticated operations to obtain access tokens from the Identity API.
+/// After successful authentication, use the returned tokens to create an authenticated core client.
+///
+/// # Available Methods
+///
+/// - **Password login**: [`login_via_password`](Self::login_via_password) - See the
```
quexten: Won't these get out of date really easily? Usually you can just discover available methods via your IDE's LSP or rust docs.

crates/bitwarden-auth/src/login/login_via_password/login_via_password_impl.rs
```diff
@@ -0,0 +1,592 @@
...
+use crate::identity::{
+    LoginClient,
+    api::{request::LoginApiRequest, send_login_request},
+    login_via_password::{PasswordLoginApiRequest, PasswordLoginError, PasswordLoginRequest},
+    models::LoginResponse,
+};
+
+#[cfg_attr(feature = "wasm", wasm_bindgen)]
+#[cfg_attr(feature = "uniffi", uniffi::export)]
+impl LoginClient {
```
quexten, line 14: Style: I believe putting `impl` for a struct into a separate file is not common / does not exist in the rest of the SDK. It should rather live in the `login_login_client` file?

If the reason is that you want to split by login strategy, and have those in separate files, then the structuring still seems wrong, since you do not want to have multiple impl blocks for one struct (and i'm not even sure if it works with bindgen / uniffi).

One pattern here would be to have the impl block in loginClient, and then delegate out to functions that you have in separate files, which is the pattern that the sdk uses for clients.

crates/bitwarden-auth/src/identity/login_via_password/login_via_password_impl.rs
```diff
@@ -0,0 +1,592 @@
...
+#[cfg_attr(feature = "uniffi", uniffi::export)]
+impl LoginClient {
+    /// Authenticates a user via email and master password.
+    ///
+    /// Derives the master password hash using KDF settings from prelogin, then sends
+    /// the authentication request to obtain access tokens and vault keys.
+    ///
+    /// # Errors
+    ///
+    /// - [`PasswordLoginError::InvalidUsernameOrPassword`] - Invalid credentials
```
quexten: Style: Generally errors are not documented like this, and you are better off just going to definition of the error in rust docs / your IDE and reading through the cases.

crates/bitwarden-auth/src/identity/login_via_password/mod.rs
```diff
@@ -0,0 +1,94 @@
...
+//! 1. **Prelogin**: Retrieve the user's KDF configuration with
+//!    [`LoginClient::get_password_prelogin`]
+//! 2. **Login**: Authenticate with [`LoginClient::login_via_password`] using the KDF settings
+//!
+//! # Security Model
+//!
+//! The master password is **never sent to the server**. Instead:
+//! - User's KDF settings (PBKDF2 or Argon2id) are fetched during prelogin
+//! - Master password is hashed locally using these settings
+//! - Only the derived hash is transmitted for authentication
```
quexten: This suggests that the output of the KDF is sent to the server. That is not what is happening. 

The KDF output - raw masterkey - is:
1. Stretched into an AES256_CBC_HMAC encryption key to unwrap the userkey
2. Hashed again with a single round of PDKBF2 with the password as salt to get the "serverAuthenticationHash" <- this is sent to the server

crates/bitwarden-auth/src/identity/login_via_password/password_login_api_request.rs
```diff
@@ -0,0 +1,281 @@
...
+pub(crate) struct PasswordLoginApiRequest {
+    /// Bitwarden user email address.
+    ///
+    /// Serialized as `"username"` to match OAuth2 conventions expected by the Identity API.
+    #[serde(rename = "username")]
+    pub email: String,
+
+    /// Derived master password authentication hash.
+    ///
+    /// This is the result of applying the user's KDF (PBKDF2 or Argon2id) to their master
```
quexten: This comment is not correct. Please see the PR comment in the mod's comment.

quexten (changes requested):
Added a set of comments with the first pass.

I'll note that some of these are very general rust things, and some of these (especially error design) are things where the SDK sets not great examples.

crates/bitwarden-core/src/key_management/mod.rs
JaredSnider-Bitwarden: 2c59708e7217a4d2fc12e10181fbb0e90fb2dd60

crates/bitwarden-auth/src/identity/models/webauthn_prf_user_decryption_option.rs
JaredSnider-Bitwarden: Ah, thank you for catching that and letting me know about `UnsignedSharedKey`
c855ec29f602f11d212e5e9fea388d0833798f08

crates/bitwarden-auth/src/identity/models/webauthn_prf_user_decryption_option.rs
JaredSnider-Bitwarden: c855ec29f602f11d212e5e9fea388d0833798f08

crates/bitwarden-auth/src/login/login_via_password/password_prelogin_response.rs
JaredSnider-Bitwarden: Added a ticket for that in code and jira: 62b8b3a609b3c31e5a9ffe2f8b301a24c85a7bc7

crates/bitwarden-auth/src/identity/login_client.rs
JaredSnider-Bitwarden: Fair. Removed:
340af16ef1f73f3de95c365e77f73006eee6ae11

crates/bitwarden-auth/src/login/api/response/login_error_api_response.rs
JaredSnider-Bitwarden: Yes. I left it there to save me the time of having to re-look up the 2FA shape to map into a dedicated 2FA invalid grant error (next PR should add 2FA support).

crates/bitwarden-auth/src/login/login_via_password/password_prelogin.rs
JaredSnider-Bitwarden: These are good questions. I'm not aware of any platform tracing work nor any docs on BW best practice error handling. @dani-garcia and I used a similar pattern in auth's send access work - however, that's not to say it can't be improved. I would ask him to weigh in here.

crates/bitwarden-auth/src/login/models/trusted_device_user_decryption_option.rs
JaredSnider-Bitwarden: Are there any test defaults for these specific data types that I could use or do I just continue to do like I did in https://github.com/bitwarden/sdk-internal/commit/c855ec29f602f11d212e5e9fea388d0833798f08?

crates/bitwarden-auth/src/login/login_via_password/password_login_error.rs
JaredSnider-Bitwarden: Nothing beyond "it's what we did before in send access, and that's what I'm basing my understanding of error handling on"

Can you link to an example of tracing?

crates/bitwarden-auth/src/login/login_via_password/password_login_error.rs
JaredSnider-Bitwarden: f6d89b29bfe8b7618dfce3f5bae0676c1f092e49

crates/bitwarden-auth/src/identity/login_via_password/password_login_error.rs
JaredSnider-Bitwarden: f6d89b29bfe8b7618dfce3f5bae0676c1f092e49

crates/bitwarden-auth/src/api/enums/scope.rs
JaredSnider-Bitwarden: I ran into Orphan rules trying to do that.
```
error[E0117]: only traits defined in the current crate can be implemented for types defined outside of the crate
  --> crates/bitwarden-auth/src/api/enums/scope.rs:31:1
   |
31 | impl From<&[Scope]> for String {
   | ^^^^^--------------^^^^^------
   |      |                  |
   |      |                  `std::string::String` is not defined in the current crate
   |      this is not defined in the current crate because this is a foreign trait
   |
   = note: impl doesn't have any local type before any uncovered type parameters
   = note: for more information see https://doc.rust-lang.org/reference/items/implementations.html#orphan-rules
   = note: define and implement a trait or new type instead
   ```

File: crates/bitwarden-auth/src/login/api/response/login_success_api_response.rs (hunk 21 above)
Comment (JaredSnider-Bitwarden) [reply crates/bitwarden-auth/src/login/api/response/login_success_api_response.rs:34]: So, `PrivateKeysResponseModel` is meant to convert into `WrappedAccountCryptographicState`? 
The reason I left that comment is because I looked for all usages of `PrivateKeysResponseModel` and found no conversions into an SDK domain model.  Would KM be able to provide a conversion trait?

File: crates/bitwarden-auth/src/identity/login_via_password/login_via_password_impl.rs (hunk 24 above)
Comment (JaredSnider-Bitwarden) [reply crates/bitwarden-auth/src/identity/login_via_password/login_via_password_impl.rs]: 340af16ef1f73f3de95c365e77f73006eee6ae11

File: crates/bitwarden-auth/src/login/login_via_password/login_via_password_impl.rs (hunk 23 above)
Comment (JaredSnider-Bitwarden) [reply crates/bitwarden-auth/src/login/login_via_password/login_via_password_impl.rs:14]: I appreciate this feedback. @dani-garcia and I explicitly talked about this during the initial design sessions that we had, and from what I understood, multiple `impl` blocks were supposed to be fine. If this will cause issues with bindgen / uniffi, then I'm glad to know now. Do you have an example of an impl delegating out to functions in separate files?

File: crates/bitwarden-auth/src/identity/login_via_password/password_login_api_request.rs (hunk 26 above)
Comment (JaredSnider-Bitwarden) [reply crates/bitwarden-auth/src/identity/login_via_password/password_login_api_request.rs]: 899b7322b33fb080b54fcaa8f082ff5e3ea18ac4

File: crates/bitwarden-auth/src/identity/login_via_password/mod.rs (hunk 25 above)
Comment (JaredSnider-Bitwarden) [reply crates/bitwarden-auth/src/identity/login_via_password/mod.rs]: Sorry about that. Fixed with: dd0db639b281644f02e8d89eff15609fd13b678c

File: crates/bitwarden-auth/src/login/models/webauthn_prf_user_decryption_option.rs (hunk 5 above)
Comment (JaredSnider-Bitwarden) [reply crates/bitwarden-auth/src/login/models/webauthn_prf_user_decryption_option.rs:15]: I was basing code on the existing clients code which doesn't use that yet. Could you help me understand how to properly map the properties into the `RotateableKeySet`?  Is it ok for the `userkeyencryptedpublickey` to be missing?

Comment (quexten) [reply crates/bitwarden-auth/src/login/models/webauthn_prf_user_decryption_option.rs:15]: Hmm, the RotateableKeySet is not really meant to be split up like this. While we could adjust it in the SDK, I do think the better solution is to just sync the missing field down from the server, because then we can completely opaquely view this cryptographic object. Ideally, auth would not even own the individual fields here but a single field containing the entire object, similar to what we do for masterpasswordunlock/auth data, account cryptographic state, etc.

Otherwise we are carrying around two separate implementations of essentially the same object, one that for some reason can only be used to decrypt the downstream with the upstream, and one that can also be used for rotation.

Maybe make this a follow-up / it needs discussion.

File: crates/bitwarden-auth/src/login/login_via_password/password_prelogin.rs (hunk 14 above)
Comment (quexten) [reply crates/bitwarden-auth/src/login/login_via_password/password_prelogin.rs]: Maybe I should make a write-up here in the SDK channel, to get consensus and some changes in the contributing docs...

File: crates/bitwarden-auth/src/login/api/response/login_success_api_response.rs (hunk 21 above)
Comment (quexten) [reply crates/bitwarden-auth/src/login/api/response/login_success_api_response.rs:34]: Oh, interesting. I made a jira ticket capturing it: https://bitwarden.atlassian.net/browse/PM-30222.

The naming is unfortunate currently, but `PrivateKeysResponseModel` holds a `WrappedAccountCryptographicState`. They are equivalent, with the PrivateKeysResponseModel also holding the public-key parts, and the WrappedAccountCryptographicState just holding the wrapped private parts from which the public parts can be derived.

File: crates/bitwarden-auth/src/login/models/webauthn_prf_user_decryption_option.rs (hunk 5 above)
Comment (trmartin4) [reply crates/bitwarden-auth/src/login/models/webauthn_prf_user_decryption_option.rs:15]: I created [PM-30233](https://bitwarden.atlassian.net/browse/PM-30233) for addressing this.

File: crates/bitwarden-auth/src/login/login_via_password/login_via_password_impl.rs (hunk 23 above)
Comment (quexten) [reply crates/bitwarden-auth/src/login/login_via_password/login_via_password_impl.rs:14]: Sure, it's how all clients are structured. Example: https://github.com/bitwarden/sdk-internal/blob/0794e100645d97d0359551939673b2d09db2204d/crates/bitwarden-core/src/key_management/crypto_client.rs#L50

File: crates/bitwarden-auth/src/login/models/trusted_device_user_decryption_option.rs (hunk 12 above)
Comment (quexten) [reply crates/bitwarden-auth/src/login/models/trusted_device_user_decryption_option.rs]: I don't think we have test defaults here, no.

File: crates/bitwarden-auth/src/login/login_via_password/mod.rs [hunk 27]
```diff
@@ -0,0 +1,97 @@
...
+//!    [`LoginClient::get_password_prelogin`]
+//! 2. **Login**: Authenticate with [`LoginClient::login_via_password`] using the KDF settings
+//!
+//! # Security Model
+//!
+//! The master password is **never sent to the server**. Instead:
+//! - User's KDF settings (PBKDF2 or Argon2id) are fetched during prelogin
+//! - Master password is stretched with KDF to derive the master key
+//! - Master key is stretched again into an AES256-CBC-HMAC key to unwrap the user key
+//! - Master key is hashed with single-round PBKDF2 (using password as salt) to create the server
```
Patrick-Pimentel-Bitwarden, line 15: This comment doesn't sound correct `(using password as salt)`

quexten: Despite this sounding like an unsound construction, it is how it was designed at the time and is correct. https://github.com/bitwarden/clients/blob/3558db0e0cc63cdac9ebcd3db21a0bafa7358517/libs/key-management/src/key.service.ts#L336

Hinton:
`schemars::JsonSchema` is used when we want to expose SM json functions.

rr-bw:
@JaredSnider-Bitwarden 

Here is a summary of recent updates:

- https://github.com/bitwarden/sdk-internal/pull/549/changes/856c0bb2611b5391248f187b44d2f17b0a52f121 - merged main and resolved conflicts
- https://github.com/bitwarden/sdk-internal/pull/549/changes/2cd5f801ec714f318a21646d4ffd129f2121b071 - updated default KDF usages based on changes introduced in https://github.com/bitwarden/sdk-internal/pull/654
  - I followed the pattern used in [prelogin.rs](https://github.com/bitwarden/sdk-internal/blob/main/crates/bitwarden-core/src/auth/login/prelogin.rs#L37-L46) and deleted tests that involve a default fallback (since those cases will just panic now due to `expect()`)
- https://github.com/bitwarden/sdk-internal/pull/549/changes/081538fe3ed259785c1565ed160bcef38aae64d7 - fixed reqwest_middleware related errors following changes introduced in https://github.com/bitwarden/sdk-internal/pull/699
- https://github.com/bitwarden/sdk-internal/pull/549/changes/d7e074b1d5cbe4027428705ba839852dec787479 - changed folder name from `bitwarden-auth/src/identity` to `bitwarden-auth/src/login`, and also updated certain usages of the word "identity" to "login"
- https://github.com/bitwarden/sdk-internal/pull/549/changes/ce90bd308ae5751d8064d4e52b6bb9a0b03de859 - updated test `WEBAUTHN_ENCRYPTED_USER_KEY` usages to UnsignedSharedKey (4.) [cc: @quexten]
- https://github.com/bitwarden/sdk-internal/pull/549/changes/09e8912228c2030ebc6204875f43d151581b308a - updated `TrustedDeviceUserDecryptionOption` encrypted_user_key to be UnsignedSharedKey instead of EncString [cc: @quexten]

crates/bitwarden-auth/src/login/login_via_password/login_via_password_impl.rs
```diff
@@ -0,0 +1,572 @@
...
+    /// Derives the master password hash using KDF settings from prelogin, then sends
+    /// the authentication request to obtain access tokens and vault keys.
+    pub async fn login_via_password(
+        &self,
+        request: PasswordLoginRequest,
+    ) -> Result<LoginResponse, PasswordLoginError> {
+        let master_password_authentication = MasterPasswordAuthenticationData::derive(
+            &request.password,
+            &request.prelogin_response.kdf,
+            &request.email,
```
rr-bw: ❓ - Why not `&request.prelogin_response.salt`? Or will that come later?

crates/bitwarden-auth/src/login/mod.rs
```diff
@@ -0,0 +1,16 @@
+//! Login client module
+//! The LoginClient is used to authenticate a Bitwarden User.
```
rr-bw: These sentences get awkwardly merged in the documentation. See also `login_via_password` below.

Image

crates/bitwarden-auth/src/login/login_via_password/password_prelogin.rs
```diff
@@ -0,0 +1,262 @@
...
+#[cfg_attr(feature = "uniffi", uniffi::export)]
+impl LoginClient {
+    /// Retrieves the data required before authenticating with a password.
+    /// This includes the user's KDF configuration needed to properly derive the master key.
+    ///
+    /// # Arguments
+    /// * `email` - The user's email address
+    ///
+    /// # Returns
+    /// * `PasswordPreloginResponse` - Contains the KDF configuration for the user
```
rr-bw: ⛏️ nit: Our code style docs advise against documenting [arguments](https://contributing.bitwarden.com/contributing/code-style/rust#arguments) and [return](https://contributing.bitwarden.com/contributing/code-style/rust#returns) values.

crates/bitwarden-auth/src/login/login_client.rs
```diff
@@ -0,0 +1,53 @@
...
+pub struct LoginClient {
+    pub(crate) client: Client,
+}
+
+impl LoginClient {
+    /// Creates a new `LoginClient` with the given client settings.
+    ///
+    /// # Arguments
+    ///
+    /// * `settings` - Configuration for API endpoints, user agent, and device information
```
rr-bw: ⛏️ nit: see [here](https://github.com/bitwarden/sdk-internal/pull/549/changes#r2736870526)

crates/bitwarden-auth/src/login/login_via_password/login_via_password_impl.rs
trmartin4: I think I can help with this one.  It's captured in upcoming separate work: https://bitwarden.atlassian.net/browse/PM-26955.

crates/bitwarden-auth/Cargo.toml
```diff
@@ -45,10 +53,16 @@ uuid = { workspace = true }
 wasm-bindgen = { workspace = true, optional = true }
 wasm-bindgen-futures = { workspace = true, optional = true }
 
+[target.'cfg(not(target_arch="wasm32"))'.dependencies]
+# TLS stack for HTTP client - WASM uses browser's fetch API instead
+rustls = { version = "0.23.19", default-features = false }
+rustls-platform-verifier = "0.6.0"
```
dani-garcia: I don't think these two dependencies are needed, you should be inheriting them through your dependency in `bitwaden-core`

crates/bitwarden-auth/src/login/api/request/mod.rs
```diff
@@ -0,0 +1,9 @@
+//! Request models for Identity API endpoints that cannot be auto-generated
+//! (e.g., connect/token endpoints) and are shared across multiple features within the login
+//! client
+//!
+//! For standard controller endpoints, use the `bitwarden-api-identity` crate.
+mod login_api_request;
+// STANDARD_USER_SCOPES is used in tests in password_login_api_request.rs
+#[allow(unused_imports)]
```
dani-garcia: You can add the reason as part of the allow directive, which would make it clearer.
```suggestion
#[allow(
    unused_imports,
    reason = "STANDARD_USER_SCOPES is used in tests in password_login_api_request.rs"
)]
```

crates/bitwarden-auth/src/login/api/response/login_error_api_response.rs
```diff
@@ -0,0 +1,506 @@
...
+#[serde(tag = "error")]
+pub enum OAuth2ErrorApiResponse {
+    /// Invalid request error, typically due to missing parameters for a specific
+    /// credential flow. Ex. `password` is required.
+    InvalidRequest {
+        // we need default b/c we don't want deserialization to fail if error_description is
+        // missing. we want it to be None in that case.
+        #[serde(default)]
+        /// The optional error description for invalid request errors.
+        error_description: Option<String>,
```
dani-garcia: Tiny nit, but I'd move the comment above the serde(default) macro, just to keep the same order we use in other places. Same for the others in this enum.

```suggestion
        /// The optional error description for invalid request errors.
		#[serde(default)]
        error_description: Option<String>,
```

crates/bitwarden-auth/src/login/login_via_password/password_login_api_request.rs
```diff
@@ -0,0 +1,269 @@
...
+            serialized
+        );
+        // Verify scope matches the standard scopes (space becomes + in URL encoding)
+        let expected_scope = scopes_to_string(STANDARD_USER_SCOPES).replace(' ', "+");
+        assert!(
+            serialized.contains(&format!("scope={}", expected_scope)),
+            "Should contain scope={}, got: {}",
+            expected_scope,
+            serialized
+        );
```
dani-garcia, line 193: Some of these can be simplified by putting the variable in the format string directy:
```suggestion
        assert!(
            serialized.contains(&format!("scope={expected_scope}")),
            "Should contain scope={expected_scope}, got: {serialized}",
        );
```

There's a couple of others in this file and `password_login_error.rs` too.

crates/bitwarden-auth/src/login/mod.rs
dani-garcia: The comments are markdown so you can put an empty line between them to ensure they don't get merged in one line:
```
//! Login client module
//!
//! The LoginClient is used to authenticate a Bitwarden User....
```

The problem with the login via password one is that we comment it when we declare the module:
```rust
/// Login via password functionality
pub mod login_via_password;
```
but we also comment it inside the module itself:
```rust
//! Password-based authentication for Bitwarden users.
//!
//! This module implements the password login flow, which requires two steps:
//!
```

Those two get merged awkwardly. I'd recomment removing the module comments from this mod file, and instead relying on the inner comments from each module to show up.

dani-garcia (commented):
Looks pretty good, I've included some small comments. The linter is failing because there's an unformatted file, make sure to run `cargo +nightly fmt` to fix it.

About your outstanding notes:
1- on SchemaRs, this is used by secrets manager to generate the language bindings, so we won't need to deal with it until we touch SM related login flows.
2- Should have been fixed with https://github.com/bitwarden/sdk-internal/pull/637
3- It seems to work for me, I think it could be related to the commercial/oss SDK? If you use the `build:bit` commands, you need to make sure you build the SDK with the commercial license (`./build.sh -b`). If you're running the `build:oss` commands, you need the SDK without the commercial license (`./build.sh`)

crates/bitwarden-auth/src/login/models/user_decryption_options_response.rs
JaredSnider-Bitwarden: @rr-bw  resolved with 09e8912228c2030ebc6204875f43d151581b308a

crates/bitwarden-auth/src/identity/models/user_decryption_options_response.rs
JaredSnider-Bitwarden: @rr-bw resolved with https://github.com/bitwarden/sdk-internal/pull/549/changes/ce90bd308ae5751d8064d4e52b6bb9a0b03de859

crates/bitwarden-auth/src/identity/models/user_decryption_options_response.rs
JaredSnider-Bitwarden: @rr-bw resolved with https://github.com/bitwarden/sdk-internal/commit/ce90bd308ae5751d8064d4e52b6bb9a0b03de859

crates/bitwarden-auth/src/login/models/user_decryption_options_response.rs
JaredSnider-Bitwarden: @rr-bw resolved with https://github.com/bitwarden/sdk-internal/commit/09e8912228c2030ebc6204875f43d151581b308a

crates/bitwarden-auth/src/identity/models/user_decryption_options_response.rs
JaredSnider-Bitwarden: @rr-bw resolved with https://github.com/bitwarden/sdk-internal/pull/549/changes/ce90bd308ae5751d8064d4e52b6bb9a0b03de859

crates/bitwarden-auth/src/login/models/trusted_device_user_decryption_option.rs
JaredSnider-Bitwarden: @rr-bw resolved with 09e8912228c2030ebc6204875f43d151581b308a

crates/bitwarden-auth/src/login/models/user_decryption_options_response.rs
JaredSnider-Bitwarden: @rr-bw resolved with 09e8912228c2030ebc6204875f43d151581b308a

crates/bitwarden-auth/src/login/api/response/login_success_api_response.rs
JaredSnider-Bitwarden: Ok, thank you for the context and the ticket. I've went ahead and added comments that reference that ticket with c0384875cd2e7ae7a4c829a2ad3deeda3a262c8a so we can expose the wrapped account crypto state on the login success response later on.

JaredSnider-Bitwarden:
@Hinton 
> `schemars::JsonSchema` is used when we want to expose SM json functions.

Ok, thank you for that context.

crates/bitwarden-auth/src/login/login_via_password/password_prelogin.rs
JaredSnider-Bitwarden: Good catch. Fixed with 24c2f7104635f7c57a2391e6352d7bb4e437446d

crates/bitwarden-auth/src/login/login_client.rs
JaredSnider-Bitwarden: Fixed with f4c7e43235b231b66e83c8ec00adaa20d116ad14

crates/bitwarden-auth/src/login/login_via_password/login_via_password_impl.rs
trmartin4: This is completely wrong, I apologize.  I misread where this comment was referencing.

What I thought I was looking at was the response to the `password_prelogin` request, but this is the login implementation.  Please disregard my comment.

JaredSnider-Bitwarden: Great catch. Due to the age of the server changes (11/20/25), we can just start referencing the salt from the prelogin response as we will have had 3 major server releases by the time we update clients to call this code.
I've updated this here: fa4557d4dc4962ad18e2deec338155e608386a44

crates/bitwarden-auth/Cargo.toml
JaredSnider-Bitwarden: Ah, thank you! Resolved with 1fd065d042baab9aecd63379c4c1dc15b76fb028

crates/bitwarden-auth/src/login/login_via_password/password_login_api_request.rs
JaredSnider-Bitwarden: Resolved with 1663cd9ca817a26176cade56df102ff35f9ee914

crates/bitwarden-auth/src/login/login_via_password/login_via_password_impl.rs
JaredSnider-Bitwarden: Discussed this with @dani-garcia, and despite this being undocumented so far, this is the new direction we are heading for the SDK. Also, it shouldn't have any issues with bindgen / uniffi per that discussion as well.

crates/bitwarden-auth/src/login/login_via_password/password_login_error.rs
JaredSnider-Bitwarden: Upon coming back to this, yes, we want to expose this error for visibility in the client in toasts / logs.

crates/bitwarden-auth/src/login/mod.rs
JaredSnider-Bitwarden: @rr-bw and @dani-garcia , thank you for the tips here. I've overhauled these docs in a2a5eca81972e86aa4fc4264d4ccc355abdda319

crates/bitwarden-auth/src/login/login_via_password/password_prelogin.rs
JaredSnider-Bitwarden: Having the `ApiError` branch gives us an object with a `message` property which is what our clients `ValidationService` expects. I have consolidated the error branches by working with @dani-garcia : https://github.com/bitwarden/sdk-internal/pull/549/changes/daaf0a93c3f6abf93488ba48bf7b82e60ba1afc6.

crates/bitwarden-auth/src/login/login_via_password/password_prelogin.rs
JaredSnider-Bitwarden: @dani-garcia and I consolidated the missing field error to just map to the Unknown error case: https://github.com/bitwarden/sdk-internal/pull/549/changes/daaf0a93c3f6abf93488ba48bf7b82e60ba1afc6

crates/bitwarden-auth/src/login/login_via_password/login_via_password_impl.rs
quexten: Interesting. Thank you!

crates/bitwarden-auth/Cargo.toml
```diff
@@ -45,10 +53,13 @@ uuid = { workspace = true }
 wasm-bindgen = { workspace = true, optional = true }
 wasm-bindgen-futures = { workspace = true, optional = true }
 
+[target.'cfg(not(target_arch="wasm32"))'.dependencies]
```
quexten: Is this empty dependency block intentional?

crates/bitwarden-auth/src/login/login_via_password/login_via_password_impl.rs
quexten: Actually, a small question about naming. Is the `_impl` naming something we'll establish as a standard @dani-garcia / should do in KM-owned code too?

crates/bitwarden-auth/src/login/login_via_password/password_login_request.rs
```diff
@@ -0,0 +1,26 @@
+use serde::{Deserialize, Serialize};
+
+use crate::login::{login_via_password::PasswordPreloginResponse, models::LoginRequest};
+
+/// Public SDK request model for logging in via password
+#[derive(Serialize, Deserialize)]
+#[serde(rename_all = "camelCase", deny_unknown_fields)]
+#[cfg_attr(feature = "uniffi", derive(uniffi::Record))] // add mobile support
```
quexten, line 8: Super nit and feel free to ignore, but the comments here don't add much. We use the uniffi macro and wasm macro all over the place so these comments are just noise.

quexten (dismissed):
Only a small question left for @dani-garcia about general naming in the SDK.

crates/bitwarden-auth/src/login/login_via_password/login_via_password_impl.rs
dani-garcia: I think in this case the naming is more to avoid having the same name as the parent module, because clippy lints against that specifically. 

I don't think we have any file naming guidelines, and there are also no official Rust guidelines that I can point to. I'd recommend to make sure the names are unique, just so they're generally easier for IDE's to deal with (though Rust makes us deal with a ton of `mod.rs` files anyway). Beyond that, my only recommendations is to not nest modules too deep and keep descriptive names for everything.

I think the `_impl` naming in this case is okay, we could also change it to something like `login_via_password/implementation.rs`, though if we do that then we'll probably end up with a similar problem than all the `mod.rs` files.

crates/bitwarden-auth/Cargo.toml
JaredSnider-Bitwarden: Oops, I missed that in f80ef4bfaaf8848da6eedb57793277bbc2b2071e. Thank you for seeing that. I removed with 4ab99f944caad4239f74dbd6d3cd18fc284dbd2b.

crates/bitwarden-auth/src/login/login_via_password/login_via_password_impl.rs
quexten: thanks!

dani-garcia (dismissed):

crates/bitwarden-auth/src/login/login_via_password/password_login_request.rs
JaredSnider-Bitwarden: Agreed, but I like leaving the example for other auth members for the short term.

crates/bitwarden-auth/src/login/login_via_password/password_prelogin.rs
```diff
@@ -0,0 +1,275 @@
...
+        let mock = Mock::given(matchers::method("POST"))
+            .and(matchers::path("identity/accounts/prelogin/password"))
+            .and(matchers::header(
+                reqwest::header::CONTENT_TYPE.as_str(),
+                "application/json",
+            ))
+            .respond_with(ResponseTemplate::new(200).set_body_json(raw_success));
+
+        let (mock_server, _api_config) = start_api_mock(vec![mock]).await;
+        let identity_client = make_login_client(&mock_server);
```
rr-bw: I missed this when I renamed `IdentityClient` to `LoginClient`. We should update to `login_client`, along with other usages in these tests, and also those in `login_via_password_impl.rs`. (Of course be sure not to accidentally rename the API `identity_client` that comes from `api_configs` earlier in the file.)

crates/bitwarden-auth/src/login/mod.rs
rr-bw: I think you should probably just remove line 1 entirely, otherwise we now get this instead of the description:

Image

quexten (dismissed):

crates/bitwarden-auth/src/login/api/request/login_api_request.rs
```diff
@@ -0,0 +1,354 @@
...
+        // Note: These are None, so they won't appear in the serialization
+        // But we can verify they would use camelCase by checking field omission
+        assert!(
+            !serialized.contains("two_factor_token"),
+            "2FA fields should use camelCase, not snake_case"
+        );
+        assert!(
+            !serialized.contains("twoFactorToken"),
+            "2FA fields should be omitted when None"
+        );
```
rr-bw, line 255: Based on the comment, did you intend to add checks for `two_factor_provider` and `two_factor_remember` here as well?

crates/bitwarden-auth/src/login/login_via_password/password_login_api_request.rs
```diff
@@ -0,0 +1,252 @@
...
+            },
+        }
+    }
+
+    fn make_test_master_password_auth() -> MasterPasswordAuthenticationData {
+        let request = make_test_password_login_request(false);
+        MasterPasswordAuthenticationData::derive(
+            &request.password,
+            &request.prelogin_response.kdf,
+            &request.email,
```
rr-bw: `&request.prelogin_response.salt` ?

crates/bitwarden-auth/src/login/api/response/webauthn_prf_user_decryption_option_api_response.rs
rr-bw, line 1: It looks like a few more properties were added to the server response just a few weeks ago here (UserDecryptionOptions.cs): https://github.com/bitwarden/server/pull/6401/changes#diff-2faec28f69bf069d97712107d42f22bf9f4e19bf63c16501769f4e6d1875523a

crates/bitwarden-auth/src/login/models/webauthn_prf_user_decryption_option.rs
rr-bw, line 1: Same here regarding new properties: https://github.com/bitwarden/sdk-internal/pull/549#discussion_r2785030340

crates/bitwarden-auth/src/login/models/user_decryption_options_response.rs
```diff
@@ -0,0 +1,402 @@
...
+                has_admin_approval: true,
+                has_login_approving_device: false,
+                has_manage_reset_password_permission: false,
+                is_tde_offboarding: false,
+                encrypted_private_key: Some(TDE_ENCRYPTED_PRIVATE_KEY.parse().unwrap()),
+                encrypted_user_key: Some(TDE_ENCRYPTED_USER_KEY.parse().unwrap()),
+            }),
+            key_connector_option: Some(KeyConnectorUserDecryptionOptionApiResponse {
+                key_connector_url: KEY_CONNECTOR_URL.to_string(),
+            }),
```
rr-bw, line 141: Is it really possible to have both `trusted_device_option` and `key_connector_option` on the same response given the comment in `user_decryption_options_api_response.rs` ([lines 27-29](https://github.com/bitwarden/sdk-internal/pull/549/changes#diff-e460730dd73a9990be17c873128a1c3640b4801e51cfea7f132c30503133a045R27-R29)):
```rust
    /// Key Connector Decryption Option.
    /// This option is mutually exlusive with the Trusted Device option as you
    /// must configure one or the other in the Organization SSO configuration.
    #[serde(rename = "KeyConnectorOption", skip_serializing_if = "Option::is_none")]
    pub key_connector_option: Option<KeyConnectorUserDecryptionOptionApiResponse>,
```

crates/bitwarden-auth/src/login/login_via_password/login_via_password_impl.rs
```diff
@@ -0,0 +1,572 @@
...
+        for kdf_type in kdf_types {
+            let raw_success = make_mock_success_response();
+
+            let mock = add_standard_login_headers(
+                Mock::given(matchers::method("POST")).and(matchers::path("identity/connect/token")),
+            )
+            .respond_with(ResponseTemplate::new(200).set_body_json(raw_success));
+
+            let (mock_server, _api_config) = start_api_mock(vec![mock]).await;
+            let identity_client = make_identity_client(&mock_server);
```
rr-bw: Same as [here](https://github.com/bitwarden/sdk-internal/pull/549#discussion_r2778100052) regarding naming, but in this file the function name should also be updated.

crates/bitwarden-auth/src/login/login_via_password/password_prelogin.rs
JaredSnider-Bitwarden: e12d5cd52ec14274541cf32369945a0300e706aa

crates/bitwarden-auth/src/login/mod.rs
JaredSnider-Bitwarden: b00b3931f178382c44f593ea053d56d13ae24718

crates/bitwarden-auth/src/login/login_via_password/login_via_password_impl.rs
JaredSnider-Bitwarden: e12d5cd52ec14274541cf32369945a0300e706aa

crates/bitwarden-auth/src/login/api/request/login_api_request.rs
JaredSnider-Bitwarden: Great catch: fixed with 191df22001d3657f137d12b45f004c8bf761ab40 and I also improved the test assertions.

crates/bitwarden-auth/src/login/login_via_password/password_login_api_request.rs
JaredSnider-Bitwarden: cf0ec03cdfdf31174eaf78d2009d88c8c72a7114

crates/bitwarden-auth/src/login/api/response/webauthn_prf_user_decryption_option_api_response.rs
JaredSnider-Bitwarden: 6d0cc79d0247553075d90f68ab5a9a5285150591

crates/bitwarden-auth/src/login/models/webauthn_prf_user_decryption_option.rs
JaredSnider-Bitwarden: 6d0cc79d0247553075d90f68ab5a9a5285150591

crates/bitwarden-auth/src/login/models/user_decryption_options_response.rs
JaredSnider-Bitwarden: Yes, that is true, so I've added a clarifying comment that the test is just to ensure all options convert properly. 
ac6eb603b2b3bc9f0c9381c1b5cd50777d4cb6b3

dani-garcia (dismissed):
Changes LGTM, you have some lockfile conflicts that need resolving

quexten (dismissed):

rr-bw (approved):

dani-garcia (approved):

quexten (approved):

PR #550, opened 2025-11-07
Replace `log` with `tracing` in WASM crate

neuronull:
## 📔 Objective

Cascading the standardization of `tracing` crate from clients repo into the SDK, that was started per https://github.com/bitwarden/clients/pull/16321

, this change updates the WASM crate to use `tracing`.

## 📸 Screenshots

Screenshot 2025-11-19 at 09 32 04

## ⏰ Reminders before review

- Contributor guidelines followed
- All formatters and local linters executed and passed
- Written new unit and / or integration tests where applicable
- Protected functional changes with optionality (feature flags)
- Used internationalization (i18n) for all UI strings
- CI builds passed
- Communicated to DevOps any deployment requirements
- Updated any necessary documentation (Confluence, contributing docs) or informed the documentation
  team

## 🦮 Reviewer guidelines

- 👍 (`:+1:`) or similar for great changes
- 📝 (`:memo:`) or ℹ️ (`:information_source:`) for notes or general info
- ❓ (`:question:`) for questions
- 🤔 (`:thinking:`) or 💭 (`:thought_balloon:`) for more open inquiry that's not quite a confirmed
  issue and could potentially benefit from discussion
- 🎨 (`:art:`) for suggestions / improvements
- ❌ (`:x:`) or ⚠️ (`:warning:`) for more significant problems or concerns needing attention
- 🌱 (`:seedling:`) or ♻️ (`:recycle:`) for future improvements or indications of technical debt
- ⛏ (`:pick:`) for minor or nitpick changes

PR #555, opened 2025-11-12
[PM-28135] Add crypto tracing logs

quexten:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-28135

## 📔 Objective

Enables `tracing`, but does not yet remove the old logging (follow-up). It adds tracing to the most important crypto paths. Follow-up tickets may add more tracing.

Example log output:
```
ERROR log{cipher_id=Some(CipherId(18c99ec6-af2b-407a-a5d0-b39201386413))}:decrypt: bitwarden_crypto::traits::decryptable: Function returned error error=The cipher's MAC doesn't match the expected value

__wbg_error_a7f8fbb0523dae15 @ bitwarden_wasm_internal_bg.js:4508
$func7041 @ 43dc2d8abaf76eb338eb.module.wasm:0x2c3395
$func3785 @ 43dc2d8abaf76eb338eb.module.wasm:0x297ed0
$func232 @ 43dc2d8abaf76eb338eb.module.wasm:0xa3c75
$func6915 @ 43dc2d8abaf76eb338eb.module.wasm:0x2c23e3
$func2456 @ 43dc2d8abaf76eb338eb.module.wasm:0x2687b1
$func817 @ 43dc2d8abaf76eb338eb.module.wasm:0x1b8e06
$func192 @ 43dc2d8abaf76eb338eb.module.wasm:0x5ce74
$func2995 @ 43dc2d8abaf76eb338eb.module.wasm:0x27f181
...

```

## ⏰ Reminders before review

- CI builds passed
  team

## 🦮 Reviewer guidelines


crates/bitwarden-core/src/client/encryption_settings.rs
```diff
@@ -108,11 +110,13 @@ impl EncryptionSettings {
     }
 
     #[cfg(feature = "internal")]
+    #[log_error]
```
dani-garcia: @quexten Tracing has a similar thing as well here, we can also look into that, as we're planning to migrate to tracing anyway:

https://docs.rs/tracing-attributes/latest/tracing_attributes/attr.instrument.html

```
If the function returns a Result<T, E> and E implements std::fmt::Display, adding err or err(Display) will emit error events when the function returns Err:

#[instrument(err)]
fn my_function(arg: usize) -> Result<(), std::io::Error> {
    Ok(())
}
```

quexten: Ahh, nice, looks like I can delete the custom macro again then, thank you for the note.

quexten:
Note: This has some overlap with https://github.com/bitwarden/sdk-internal/pull/550, as this also enables tracing. The linked PR enables tracing and removes log.

crates/bitwarden-core/src/client/encryption_settings.rs
```diff
@@ -201,6 +205,7 @@ impl EncryptionSettings {
     }
 
     #[cfg(feature = "internal")]
+    #[instrument(err, skip(org_enc_keys, store))]
```
dani-garcia: We're skipping both arguments, so we can probably use skip_all to match the rest of the functions, right?
```suggestion
    #[instrument(err, skip_all)]
```

crates/bitwarden-core/src/client/internal.rs
```diff
@@ -304,6 +304,12 @@ impl InternalClient {
         user_key: SymmetricCryptoKey,
         key_state: UserKeyState,
     ) -> Result<(), EncryptionSettingsError> {
+        let span = tracing::info_span!(
+            "User Crypto Initialization",
+            user_id = ?self.get_user_id(),
+        );
+        let _enter = span.enter();
```
dani-garcia: You can also use the `entered` function, which takes ownership of the span and doesn't require two variable definitions:
```rust
        let _span_guard = tracing::info_span!(
            "User Crypto Initialization",
            user_id = ?self.get_user_id(),
        )
        .entered();
```

https://docs.rs/tracing/latest/tracing/struct.Span.html#method.entered

crates/bitwarden-vault/src/cipher/cipher.rs
```diff
@@ -512,6 +513,7 @@ impl CompositeEncryptable<KeyIds, SymmetricKeyId, Cipher> for CipherView {
 }
 
 impl Decryptable<KeyIds, SymmetricKeyId, CipherView> for Cipher {
+    #[instrument(err, skip(self, ctx, key), fields(cipher_id = ?self.id, org_id = ?self.organization_id, kind = ?self.r#type))]
```
dani-garcia: Same thing here, we're skipping all the arguments and then defining some ourselves. I think `skip_all` would work as well, though I haven't tried

dani-garcia (commented):
Looks pretty reasonable to me, just got some small comments.

Hinton:
Do we have any concerns about runtime performance? I think tracing negatively impacts performance but they have ways to indicate some should be compiled out, https://docs.rs/tracing/latest/tracing/level_filters/index.html#compile-time-filters.

(I suspect tracing performance to be a drop in the bucket for the crypto operations but I think this is a worthwhile discussion to have)

quexten:
@Hinton My biggest concern would be if tracing somehow is blocking with respect to the subscribers, and we would log a lot (possibly having to go to JS every time via the hooking to console it uses), though a quick check suggests that it just appends to a buffer that then gets flushed, and actual logs should not be that frequent with the current implementation. 

Is there specific things that you'd like to see measured here? Vault unlock times for large vaults? Actual logs should only be emitted for errors / rare info events, but obviously the context is still stored into a buffer somewhere.

Hinton:
@quexten Comparing calling decrypt 100'000 times with tracing to 100'000 without in a test maybe? I think it's a drop in the bucket. But I feel like it's probably not necessary to be honest.

I do think it's good to decide if we want any of the `release_max_level_` enabled since they would change behaviour. And anything lower than the value would be excluded from release builds and probably safe to use in abundance.

quexten:
@Hinton Okay, I've not further benchmarked it since I agree that it's most likely negligible. I've set the max log level to `release_max_level_info`, since I don't think anyone would expect trace / debug to be emitted in release.

In case we do want logging in an operation that gets called millions of times, we should consider using debug/trace to have it compiled out in release.

mzieniukbw (dismissed):

coroiu (dismissed):

Jingo88 (dismissed):

quexten:
Apologies, had to fix another clippy error

mzieniukbw (dismissed):

coroiu (dismissed):

mzieniukbw (approved):

coroiu (approved):

Jingo88 (approved):

PR #557, opened 2025-11-13
[PM-25820] - Implement the ability to delete an attachment

theMickster:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-25820

## 📔 Objective

Analyzed the functionality found in both the client and server code regarding deleting of attachments and chose to implement as simply as possible. I found these three steps: removal of the file from storage, removal from the collection of attachments, and an update of the RevisionDate { see [Vault/Services/Implementations/CipherService.cs](https://github.com/bitwarden/server/blob/03118079516f02a7022d8727db3410e6ecc2bc2a/src/Core/Vault/Services/Implementations/CipherService.cs#L887) for details}. 

Unless I am missing something & we are in some unforeseen data state with a Chiper, I do not think we need to concern ourselves with the rest of the payload returned and going through the re-mapping of data that should not have changed. Also, given that the payload from the delete attachments API endpoints is an entity instead of a response model, I feel we only seek to complicate things further by over engineering. 

More than welcome to other perspectives.

## ⏰ Reminders before review

- CI builds passed
  team

## 🦮 Reviewer guidelines

### DIFF
--- a/crates/bitwarden-vault/src/cipher/cipher.rs
+++ b/crates/bitwarden-vault/src/cipher/cipher.rs
@@ -1,5 +1,7 @@
 use bitwarden_api_api::{
-    apis::ciphers_api::{PutShareError, PutShareManyError},
+    apis::ciphers_api::{
+        DeleteAttachmentAdminError, DeleteAttachmentError, PutShareError, PutShareManyError,
+    },
     models::{
         CipherDetailsResponseModel, CipherRequestModel, CipherResponseModel,
         CipherWithIdRequestModel,
@@ -72,6 +74,10 @@ pub enum CipherError {
     Chrono(#[from] chrono::ParseError),
     #[error(transparent)]
     SerdeJson(#[from] serde_json::Error),
+    #[error(transparent)]
+    DeleteAttachment(#[from] bitwarden_api_api::apis::Error<DeleteAttachmentError>),
+    #[error(transparent)]
+    DeleteAttachmentAdmin(#[from] bitwarden_api_api::apis::Error<DeleteAttachmentAdminError>),
 }
 
 /// Helper trait for operations on cipher types.
--- a/crates/bitwarden-vault/src/cipher/cipher.rs
+++ b/crates/bitwarden-vault/src/cipher/cipher.rs
@@ -1,7 +1,5 @@
 use bitwarden_api_api::{
-    apis::ciphers_api::{
-        DeleteAttachmentAdminError, DeleteAttachmentError, PutShareError, PutShareManyError,
-    },
+    apis::ciphers_api::{PutShareError, PutShareManyError},
     models::{
         CipherDetailsResponseModel, CipherRequestModel, CipherResponseModel,
         CipherWithIdRequestModel,
@@ -74,10 +72,6 @@ pub enum CipherError {
     Chrono(#[from] chrono::ParseError),
     #[error(transparent)]
     SerdeJson(#[from] serde_json::Error),
-    #[error(transparent)]
-    DeleteAttachment(#[from] bitwarden_api_api::apis::Error<DeleteAttachmentError>),
-    #[error(transparent)]
-    DeleteAttachmentAdmin(#[from] bitwarden_api_api::apis::Error<DeleteAttachmentAdminError>),
 }
 
 /// Helper trait for operations on cipher types.
--- a/crates/bitwarden-vault/src/cipher/cipher_client/delete_attachment.rs
+++ b/crates/bitwarden-vault/src/cipher/cipher_client/delete_attachment.rs
@@ -1,10 +1,12 @@
 use bitwarden_api_api::{apis::ciphers_api::CiphersApi, models};
-use bitwarden_core::require;
-use bitwarden_state::repository::Repository;
+use bitwarden_core::{ApiError, MissingFieldError, require};
+use bitwarden_error::bitwarden_error;
+use bitwarden_state::repository::{Repository, RepositoryError};
+use thiserror::Error;
 #[cfg(feature = "wasm")]
 use wasm_bindgen::prelude::wasm_bindgen;
 
-use crate::{Cipher, CipherError, CipherId, CiphersClient, VaultParseError};
+use crate::{Cipher, CipherId, CiphersClient, VaultParseError};
 
 /// Standalone function to delete an attachment from a cipher that is extracted for ease of unit
 /// testing.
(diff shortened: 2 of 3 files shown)


crates/bitwarden-vault/src/cipher/cipher_client/delete_attachment.rs
```diff
@@ -0,0 +1,453 @@
...
+        api_client
+            .delete_attachment(cipher_id.into(), attachment_id)
+            .await?
+    };
+
+    let boxed_cipher: Box<models::Cipher> = require!(response.cipher);
+    let cipher_response = *boxed_cipher;
+    let mut cipher = require!(repository.get(cipher_id.to_string()).await?);
+
+    cipher.revision_date = require!(cipher_response.revision_date)
```
dani-garcia: Nit: You don't need to manually deref the cipher to unbox it in this case:

```suggestion
    let cipher_response: Box<models::Cipher> = require!(response.cipher);
    let mut cipher = require!(repository.get(cipher_id.to_string()).await?);

    cipher.revision_date = require!(cipher_response.revision_date)
```

crates/bitwarden-vault/src/cipher/cipher_client/delete_attachment.rs
```diff
@@ -0,0 +1,453 @@
...
+    Ok(cipher)
+}
+
+#[cfg_attr(feature = "wasm", wasm_bindgen)]
+impl CiphersClient {
+    /// Delete an attachment from a cipher
+    pub async fn delete_attachment(
+        &self,
+        cipher_id: CipherId,
+        attachment_id: String,
```
dani-garcia: We can change the attachment id parameters for these two functions to be `&str`, no need to consume a string when a reference to it would be enough

crates/bitwarden-vault/src/cipher/cipher_client/delete_attachment.rs
```diff
@@ -0,0 +1,453 @@
...
+    cipher.revision_date = require!(cipher_response.revision_date)
+        .parse()
+        .map_err(Into::<VaultParseError>::into)?;
+
+    if let Some(ref mut attachments) = cipher.attachments {
+        attachments.retain(|a| a.id.as_deref() != Some(attachment_id));
+    }
+
+    repository
+        .set(cipher_id.to_string(), cipher.clone())
```
dani-garcia, line 44: Non-blocking improvement: The API seems to return the full cipher, so rather than manually modify the cipher we have locally, it might be easier and less error prone to just take the response cipher and stick it into the repository. This might need some conversion function like `impl TryFrom<models::Cipher> for Cipher` to be implemented though, so we can leave that for later.

dani-garcia (commented):
Looks pretty good to me, just some small suggestions and a possible future improvement.

crates/bitwarden-vault/src/cipher/cipher_client/delete_attachment.rs
theMickster: Thank you for the code suggestion @dani-garcia!
First _real_ Rust work had me kinda going around in circles with my borrowing 😵‍💫

crates/bitwarden-vault/src/cipher/cipher_client/delete_attachment.rs
theMickster: Thank you for the suggestion! I _thought_ so, but wasn't positive.

theMickster: Resolved ✅

crates/bitwarden-vault/src/cipher/cipher_client/delete_attachment.rs
theMickster: Resolved ✅

crates/bitwarden-vault/src/cipher/cipher_client/delete_attachment.rs
```diff
@@ -0,0 +1,452 @@
...
+    let cipher_response: Box<models::Cipher> = require!(response.cipher);
+    let mut cipher = require!(repository.get(cipher_id.to_string()).await?);
+
+    cipher.revision_date = require!(cipher_response.revision_date)
+        .parse()
+        .map_err(Into::<VaultParseError>::into)?;
+
+    if let Some(ref mut attachments) = cipher.attachments {
+        attachments.retain(|a| a.id.as_deref() != Some(attachment_id));
+    }
```
nikwithak, line 41: ⚠️ There may be an edge case here - you're updating the revision date, but without syncing other fields from the response . That could cause an issue if another client has updated the cipher since the last sync, which would enable overwriting the changes from that update.

e.g.
Client A -> Changes password on Cipher.
Client B -> Deletes an Attachment
Client B -> Changes name of Cipher
Client A -- Original password change is overwritten (Client B didn't sync the latest password, but did have the correct revision_date)

In line with @dani-garcia's comment below, I think it's best to store the cipher in the response directly, which will also save you these steps of manually updating the cipher.

nikwithak (changes requested):

crates/bitwarden-vault/src/cipher/cipher_client/delete_attachment.rs
theMickster: Ok; I have added a note in Slack for us to chat @nikwithak, @gbubemismith & @shane-melton.

crates/bitwarden-vault/src/cipher/cipher.rs
```diff
@@ -72,6 +74,10 @@ pub enum CipherError {
     Chrono(#[from] chrono::ParseError),
     #[error(transparent)]
     SerdeJson(#[from] serde_json::Error),
+    #[error(transparent)]
+    DeleteAttachment(#[from] bitwarden_api_api::apis::Error<DeleteAttachmentError>),
+    #[error(transparent)]
+    DeleteAttachmentAdmin(#[from] bitwarden_api_api::apis::Error<DeleteAttachmentAdminError>),
```
Hinton: **issue:** `CipherError` was primarily centered around "common" errors that occurs when encrypting ciphers. Adding feature specific errors means that every time you recieve a `CipherError` you have to consider what should I do when I get `DeleteAttachment` and `DeleteAttachmentAdmin` errors? But in reality they can't happen.


**suggestion:** Follow the pattern established in `cipher_client` where more direct error enums are used.

theMickster: Fairly sure that I have implemented this properly.

crates/bitwarden-vault/src/cipher/cipher_client/delete_attachment.rs
theMickster: So it appears to bring back the full Cipher, but it does not It's just the Cipher entity which lacks a few properties.

crates/bitwarden-vault/src/cipher/cipher.rs
Hinton: Yes looks great, thanks!

theMickster:
@copilot 
And 
@claude

How about a battle of the bots here to see who triggers and who fails the respond workflow?

PR #560, opened 2025-11-15
[PM-25821] Migrate Cipher Admin operation API calls to SDK

nikwithak:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-25821

## 📔 Objective

Migrates the logic for orchestrating API calls for several operations related to the Admin API functionality. Adds the following operations to `CiphersClient`, which call the appropriate API endpoints:

* delete
* delete_many
* delete_as_admin
* delete_many_as_admin
* soft_delete
* soft_delete_many
* soft_delete_as_admin
* soft_delete_many_as_admin
* restore
* restore_many
* restore_as_admin
* restore_many_as_admin
* edit_as_admin
* list_org_ciphers
* create_as_admin

**PR Notes**: Sorry for the heft - The line count is high on this PR, but a majority of it is unit tests. This should have been split into multiple tickets, in hindsight.

Note also that this hasn't been tested directly with the client yet, and so additional changes may be needed if bugs are discovered when integrating into the clients (future tickets) - these operations are not currently used anywhere.

## ⏰ Reminders before review

- CI builds passed
  team

## 🦮 Reviewer guidelines


Hinton:
Be mindful that the autogenerated API bindings do not have any graceful handling of malformed data. I.e. an enum outside the valid range errors out. https://bitwarden.atlassian.net/browse/PM-6169 tracks improvements to this, and if this improvement is required reach out to platform to ensure it's prioritized.

CC @trmartin4

crates/bitwarden-vault/src/cipher/cipher_client/create.rs
```diff
@@ -77,7 +84,7 @@ impl From<CipherCreateRequest> for CipherCreateRequestInternal {
 impl CipherCreateRequestInternal {
     /// Generate a new key for the cipher, re-encrypting internal data, if necessary, and stores the
     /// encrypted key to the cipher data.
-    fn generate_cipher_key(
+    pub(crate) fn generate_cipher_key(
```
Hinton, line 88: **question:** Is this change needed? It doesn't look like it's used outside this module.

crates/bitwarden-vault/src/cipher/cipher_client/create.rs
```diff
@@ -218,28 +225,57 @@ async fn create_cipher<R: Repository<Cipher> + ?Sized>(
     repository: &R,
     encrypted_for: UserId,
     request: CipherCreateRequestInternal,
+    collection_ids: Vec<CollectionId>,
```
Hinton: **question:** `CipherCreateRequest` has `organization_id`, is there a reason organization is part of the cipher create request while collections are part of the method? Collections are tightly coupled to an organization so this split does not make much sense.

crates/bitwarden-vault/src/cipher/cipher_client/create.rs
```diff
@@ -218,28 +225,57 @@ async fn create_cipher<R: Repository<Cipher> + ?Sized>(
...
+    } else {
+        cipher = api_client
+            .ciphers_api()
+            .post(Some(cipher_request))
+            .await
+            .map_err(ApiError::from)?
+            .try_into()?;
+        repository
+            .set(require!(cipher.id).to_string(), cipher.clone())
+            .await?;
```
Hinton: **question:** What's going on here? It doesn't really make much sense to have three different endpoints for creating ciphers.

- Is `ciphers/create` not identical to doing post on `ciphers`?
- From a REST perspective having a dedicated admin endpoint is also quite weird.

crates/bitwarden-vault/src/cipher/cipher_client/create.rs
```diff
@@ -270,22 +306,63 @@ impl CiphersClient {
...
+    }
+
+    /// Creates a new [Cipher] for an organization, and saves it to the server.
+    pub async fn create_org_cipher(
+        &self,
+        request: CipherCreateRequest,
+        collection_ids: Vec<CollectionId>,
+    ) -> Result<CipherView, CreateCipherError> {
+        self.create_cipher(request, collection_ids, false).await
+    }
```
Hinton: The contract here is weird. `CipherCreteRequest` allows you to specify an organization id. That implies you can make orgnaizational items.

crates/bitwarden-vault/src/cipher/cipher_client/create.rs
```diff
@@ -270,22 +306,63 @@ impl CiphersClient {
...
+            InitOrgCryptoRequest, InitUserCryptoMethod, InitUserCryptoRequest,
+        },
+    };
+    use bitwarden_crypto::{EncString, Kdf, SymmetricCryptoKey};
+    use bitwarden_test::{MemoryRepository, start_api_mock};
+    use chrono::Utc;
+    use wiremock::{
+        Mock, ResponseTemplate,
+        matchers::{method, path},
+    };
```
Hinton: **issue:** We've stopped using wiremock for api mocking. You can look at `crates/bitwarden-vault/src/folder/edit.rs` for up to date exampels of using `ApiClient::new_mocked`.

crates/bitwarden-vault/src/cipher/cipher_client/delete.rs
```diff
@@ -0,0 +1,482 @@
...
+    Repository(#[from] RepositoryError),
+}
+
+impl<T> From<bitwarden_api_api::apis::Error<T>> for DeleteCipherError {
+    fn from(value: bitwarden_api_api::apis::Error<T>) -> Self {
+        Self::Api(value.into())
+    }
+}
+
+impl CiphersClient {
```
Hinton: **issue:** These behave inconsistently with the other code in this module, since they do not persist anything to state.

**suggestion:** Rename these with a `api_` to make it more clear they ONLY call the API.

crates/bitwarden-vault/src/cipher/cipher_client/mod.rs
```diff
@@ -194,6 +196,10 @@ impl CiphersClient {
             .state()
             .get_client_managed::<Cipher>()?)
     }
+
+    async fn get_api_configurations(&self) -> Arc<ApiConfigurations> {
+        self.client.internal.get_api_configurations().await
+    }
```
Hinton: **question:** This method seems pretty pointless?

You could maybe return `self.client.internal.get_api_configurations().await.api_client.ciphers_api()` since that seems to be the actual api client used.

crates/bitwarden-vault/src/cipher/cipher_client/get.rs
```diff
@@ -16,7 +19,13 @@ pub enum GetCipherError {
     #[error(transparent)]
     Crypto(#[from] CryptoError),
     #[error(transparent)]
+    VaultParse(#[from] VaultParseError),
+    #[error(transparent)]
+    MissingField(#[from] MissingFieldError),
+    #[error(transparent)]
     RepositoryError(#[from] RepositoryError),
+    #[error(transparent)]
+    Api(#[from] ApiError),
```
Hinton: **issue:** I think the admin get operation is fundamentally different from the regular get. One acts on state the other using network.

**suggestion:** Define separate errors.

crates/bitwarden-vault/src/cipher/cipher_client/get.rs
```diff
@@ -55,6 +64,28 @@ impl CiphersClient {
...
+        &self,
+        org_id: OrganizationId,
+        include_member_items: bool,
+    ) -> Result<DecryptCipherListResult, GetCipherError> {
+        let configs = self.get_api_configurations().await;
+        let api = configs.api_client.ciphers_api();
+        let response: CipherMiniDetailsResponseModelListResponseModel = api
+            .get_organization_ciphers(Some(org_id.into()), Some(include_member_items))
+            .await
+            .map_err(Into::<ApiError>::into)?;
```
Hinton: **suggestion:** Use `from` when you know the direct type. We generally only use `Into::into` when you want a generic conversion.

```suggestion
            .map_err(ApiError::from)?;
```

crates/bitwarden-vault/src/cipher/cipher_client/get.rs
```diff
@@ -55,6 +64,28 @@ impl CiphersClient {
...
+        let ciphers = response
+            .data
+            .into_iter()
+            .flatten()
+            .map(TryInto::<Cipher>::try_into)
```
Hinton:
```suggestion
            .map(Cipher::try_into)
```

crates/bitwarden-vault/src/cipher/cipher_client/delete.rs
```diff
@@ -0,0 +1,482 @@
...
+        let configs = self.get_api_configurations().await;
+        let api = configs.api_client.ciphers_api();
+        api.delete_many(Some(CipherBulkDeleteRequestModel {
+            ids: cipher_ids.iter().map(|id| id.to_string()).collect(),
+            organization_id: organization_id.map(|id| id.to_string()),
+        }))
+        .await?;
+
+        for cipher_id in cipher_ids {
+            self.get_repository()?.remove(cipher_id.to_string()).await?;
```
Hinton: @bitwarden/team-platform-dev Do we have a bulk API? Since we use indexeddb at least with inserts there was significant performance degradation performing many operations. I think it's less so with localstorage, but performing up to 500 saves is not great.

crates/bitwarden-vault/src/cipher/cipher_client/delete.rs
```diff
@@ -0,0 +1,482 @@
...
+        }
+        Ok(())
+    }
+
+    async fn process_soft_delete(&self, cipher_id: CipherId) -> Result<(), RepositoryError> {
+        let repository = self.get_repository()?;
+        let cipher: Option<Cipher> = repository.get(cipher_id.to_string()).await?;
+        if let Some(mut cipher) = cipher {
+            cipher.deleted_date = Some(Utc::now());
+            cipher.archived_date = None;
```
Hinton: **suggestion:** Define a method on `Cipher` for this instead.

crates/bitwarden-vault/src/cipher/cipher_client/edit.rs
```diff
@@ -99,7 +108,7 @@ impl TryFrom<CipherView> for CipherEditRequest {
 }
 
 impl CipherEditRequest {
-    fn generate_cipher_key(
+    pub(crate) fn generate_cipher_key(
```
Hinton: **question:** this doesn't seem used outside of this module.

crates/bitwarden-vault/src/cipher/cipher_client/edit.rs
```diff
@@ -45,6 +46,14 @@ pub enum EditCipherError {
     Repository(#[from] RepositoryError),
     #[error(transparent)]
     Uuid(#[from] uuid::Error),
+    #[error(transparent)]
+    Decrypt(#[from] DecryptError),
```
Hinton: **question:** What operation requires decrypt?

crates/bitwarden-vault/src/cipher/cipher_client/edit.rs
```diff
@@ -333,29 +343,53 @@ async fn edit_cipher<R: Repository<Cipher> + ?Sized>(
...
 impl CiphersClient {
     /// Edit an existing [Cipher] and save it to the server.
-    pub async fn edit(
+    pub async fn edit(&self, request: CipherEditRequest) -> Result<CipherView, EditCipherError> {
+        self.edit_internal(request, false).await
+    }
+
+    // putCipherAdmin(id, request: CipherRequest)
+    // ciphers_id_admin_put
+    #[allow(missing_docs)] // TODO: add docs
```
Hinton: **issue:** Add documentation.

crates/bitwarden-vault/src/cipher/cipher.rs
```diff
@@ -63,15 +60,19 @@ pub enum CipherError {
...
-    #[error(transparent)]
-    PutShareMany(#[from] bitwarden_api_api::apis::Error<PutShareManyError>),
-    #[error(transparent)]
     Repository(#[from] RepositoryError),
     #[error(transparent)]
     Chrono(#[from] chrono::ParseError),
     #[error(transparent)]
     SerdeJson(#[from] serde_json::Error),
+    #[error(transparent)]
+    Api(#[from] ApiError),
```
Hinton, line 70: The methods putting `Api` errors in `CipherError` should really be updated to not use cipher error.

crates/bitwarden-vault/src/cipher/cipher.rs
```diff
@@ -1061,6 +1062,138 @@ impl TryFrom<CipherResponseModel> for Cipher {
...
+                .map(|d| d.parse())
+                .transpose()
+                .map_err(Into::<VaultParseError>::into)?,
+            edit: Default::default(),
+            favorite: Default::default(),
+            folder_id: Default::default(),
+            permissions: Default::default(),
+            view_password: Default::default(),
+            local_data: Default::default(),
+            collection_ids: Default::default(),
```
Hinton: **reflection:** Putting default values here can be dangerous, because there is nothing to indicate the Cipher model is incomplete.

Hinton (commented):
I would consider moving all the admin commands into a separate client. While they share some common logic I think they fundamentally behave differently and it might be confusing for consumers to call an admin method and not have the state react?

`CiphersClient` & `CiphersAdminClient`

crates/bitwarden-vault/src/cipher/cipher_client/delete.rs
dani-garcia: There isn't one at the moment, but it should be fairly easy to add for both indexeddb/sqlite and client managed storage. I've added a ticket to track it https://bitwarden.atlassian.net/browse/PM-28926

crates/bitwarden-vault/src/cipher/cipher_client/mod.rs
nikwithak: I added this as a convenience method, since I'm calling `self.client.internal.get_api_configurations()` in just about every function. I originally tried doing what you suggest, but lifetime issues made in unfeasible - should I remove this helper function entirely?

crates/bitwarden-vault/src/cipher/cipher_client/create.rs
nikwithak: Originally I had re-used this for a separate admin file, but I ended up adding it to the same file. I thought I already removed this, will fix.

crates/bitwarden-vault/src/cipher/cipher_client/create.rs
nikwithak: This object ends up mapping directly to the `CreateCipherRequestModel`, which is re-used across the different endpoints to create a cipher (`POST /cipher`, `POST /cipher/create`, `POST /cipher/admin`), even though `POST /cipher` does not take `collection_ids`. I will play around with ways to re-structure these in a way that fits what you're describing.

crates/bitwarden-vault/src/cipher/cipher_client/create.rs
nikwithak: The three endpoints is because of the way the backend is structured today.

* `POST /ciphers` is only used for creating an individual cipher. Even though it accepts an `organization_id` in the request, it does not have anywhere to add `collection_ids`.
* `POST /ciphers/create` is how we create an org cipher, as it accepts the `collection_ids`.
* The admin endpoint has its own uses that do a separate set of permission checks than the standard endpoint.

crates/bitwarden-vault/src/cipher/cipher_client/edit.rs
nikwithak: Since the function input for these requests is plaintext, and the SDK handles encryption before sending to the server, we return the updated object decrypted as well. [example](https://github.com/bitwarden/sdk-internal/blob/a0ba6e3942fdb3733a7965de711c6cfd9957fdd3/crates/bitwarden-vault/src/cipher/cipher_client/edit.rs#L389-L393)

We return a `CryptoError` in other instances (e.g. the `edit_cipher` implementation [here](https://github.com/bitwarden/sdk-internal/blob/a0ba6e3942fdb3733a7965de711c6cfd9957fdd3/crates/bitwarden-vault/src/cipher/cipher_client/edit.rs#L369)), but [this call](https://github.com/bitwarden/sdk-internal/blob/a0ba6e3942fdb3733a7965de711c6cfd9957fdd3/crates/bitwarden-vault/src/cipher/cipher_client/edit.rs#L459) returns a `DecryptError`. I will look into casting it approriately to use the same error.

Tangentially: Do you think it makes sense to use this approach - returning the decrypted `CipherView` object to the consumer, rather than the encrypted `Cipher` (all of it locally only)

crates/bitwarden-vault/src/cipher/cipher_client/create.rs
Hinton: I think we can just leave it for now, but would be nice to clean it up in the future.

crates/bitwarden-vault/src/cipher/cipher_client/create.rs
nikwithak: I left this in (changed to `pub(super)` so it can be re-used by the `CipherAdminClient` since it shares the request types.

crates/bitwarden-vault/src/cipher/cipher_client/create.rs
nikwithak: I updated this so that it only calls two endpoints (`POST /ciphers` or `POST /ciphers/create`), adn moved the third (`/ciphers/admin`) to a separate file / client.

crates/bitwarden-vault/src/cipher/cipher_client/create.rs
nikwithak: Updated - now uses standalone methods, and updated unit tests to test those standalones using `ApiClient::new_mocked` with passed-in dependencies.

crates/bitwarden-vault/src/cipher/cipher_client/mod.rs
nikwithak: Removed in favor of calling `self.client.internal.get_api_configurations()` directly.

crates/bitwarden-vault/src/cipher/cipher.rs
nikwithak: The only one still using CipherError is the `share` operations, which calls existing functions that currently returns `CipherError` already (e.g. https://github.com/bitwarden/sdk-internal/blob/vault/pm-25821/cipher-admin-ops/crates/bitwarden-vault/src/cipher/cipher_client/share_cipher.rs#L180-L184) - I think we can migrate this one to its own error type in the future.

crates/bitwarden-vault/src/cipher/cipher.rs
nikwithak: Unfortunately several API operations don't return the full cipher, only the `CipherMiniResponseModel` - for the Admin operations, since we don't update the repository, the intent is to still return a consistent type the user can handle, if needed - the concern is valid though. I think we can:

1. Merge it with the known existing Cipher data (if it's available),
2. Continue with this approach, knowing it's primarily for the admin operations which don't update local state - if this route, I can move the logic to be private to the admin operations, rather than a blanket `From` implementation,
3. Return the `CipherMiniResponseModel` as-is, or create a `MiniCipherView` type for mapping / returning,which doesn't expose these fields at all.

Do you have a preference? Feel free to ping me on Slack when you're free for a deeper discussion - there are a handful of API operations that currently only return a subset of data like this.

crates/bitwarden-vault/src/cipher/cipher_client/create.rs
nikwithak: I agree here - the `CipherCreateRequest` was built to match the valid fields on the API request model, but when we create an org cipher, the client calls the separate (`POST /ciphers/create`) endpoint, which takes the collection_Ids. The standard `POST /ciphers` endpoint does not have collection_ids in the response model.

I'm thinking of modifying the `CipherCreateRequest` to have both `org_id` and `collection_ids`, and writing the logic to call the `POST /ciphers/create` endpoint iff they are both present on the request; LMK your thoughts?

Hinton: Yea that seems saner imho. Ideally we can combine the server endpoints at some point?

crates/bitwarden-vault/src/cipher/cipher_client/delete.rs
Hinton: I think we'd want it before this turns on due to the performance impact especially with the bump to 4000.

crates/bitwarden-vault/src/cipher/cipher_client/admin/mod.rs
```diff
@@ -0,0 +1,14 @@
+use bitwarden_core::Client;
+use wasm_bindgen::prelude::*;
+
+mod create;
+mod delete;
+mod edit;
+mod get;
+mod restore;
+
+#[allow(missing_docs)]
```
Hinton: **suggestion:** Document the purpose of this client and compare it against the regular CipherClient?

crates/bitwarden-vault/src/cipher/cipher.rs
Hinton: I think we just need to be careful. In some ways It would be nice to make an alias to indicate "hey you can't trust this object", or define a new struct with fields optional.

This is not a blocking.

crates/bitwarden-vault/src/cipher/cipher.rs
Hinton: Yea sounds good.

crates/bitwarden-vault/src/cipher/cipher_client/edit.rs
Hinton: In the long term I would love to have `Cipher` be an internal concern of the SDK and we only expose decrypted views. Now in the meanwhile we have to make tradeoffs to move us forward and I don't think it's an issue.

crates/bitwarden-vault/src/cipher/cipher.rs
nikwithak: I added a new `PartialCipher` trait that can merge with an `Option<Cipher>`, replacing the `TryFrom` implementaiton: [58e01df](https://github.com/bitwarden/sdk-internal/pull/560/commits/58e01df2a7a00783ad7fee2c7c6bc5d7f9153aa9)

crates/bitwarden-vault/src/cipher/cipher_client/admin/delete.rs
```diff
@@ -0,0 +1,226 @@
...
+        )
+        .await
+    }
+
+    /// Deletes all [Cipher] objects with a matching [CipherId] from the server, using the admin
+    /// endpoint. Affects server data only, does not modify local state.
+    pub async fn delete_many(
+        &self,
+        cipher_ids: Vec<CipherId>,
+        organization_id: Option<OrganizationId>,
```
Hinton: **question:** should organization_id be required? Looking at the server code it is.

crates/bitwarden-vault/src/cipher/cipher_client/admin/delete.rs
```diff
@@ -0,0 +1,226 @@
...
+        )
+        .await
+    }
+
+    /// Soft-deletes all [Cipher] objects for the given [CipherId]s from the server, using the admin
+    /// endpoint. Affects server data only, does not modify local state.
+    pub async fn soft_delete_many(
+        &self,
+        cipher_ids: Vec<CipherId>,
+        organization_id: Option<OrganizationId>,
```
Hinton: **question:** should organization_id be required? Looking at the server code it is.

crates/bitwarden-vault/src/cipher/cipher_client/create.rs
```diff
@@ -60,8 +67,8 @@ pub struct CipherCreateRequest {
 /// Used as an intermediary between the public-facing [CipherCreateRequest], and the encrypted
 /// value. This allows us to manage the cipher key creation internally.
 #[derive(Clone, Debug)]
-struct CipherCreateRequestInternal {
-    create_request: CipherCreateRequest,
+pub(super) struct CipherCreateRequestInternal {
+    pub create_request: CipherCreateRequest,
```
Hinton: Is this field pub pub or pub(super) or pub(crate)?

crates/bitwarden-vault/src/cipher/cipher_client/create.rs
```diff
@@ -308,6 +365,65 @@ mod tests {
...
+            }
+        };
+
+        let org_request = InitOrgCryptoRequest {
+            organization_keys: std::collections::HashMap::from([(
+                TEST_ORG_ID.parse().unwrap(),
+                "4.rY01mZFXHOsBAg5Fq4gyXuklWfm6mQASm42DJpx05a+e2mmp+P5W6r54WU2hlREX0uoTxyP91bKKwickSPdCQQ58J45LXHdr9t2uzOYyjVzpzebFcdMw1eElR9W2DW8wEk9+mvtWvKwu7yTebzND+46y1nRMoFydi5zPVLSlJEf81qZZ4Uh1UUMLwXz+NRWfixnGXgq2wRq1bH0n3mqDhayiG4LJKgGdDjWXC8W8MMXDYx24SIJrJu9KiNEMprJE+XVF9nQVNijNAjlWBqkDpsfaWTUfeVLRLctfAqW1blsmIv4RQ91PupYJZDNc8nO9ZTF3TEVM+2KHoxzDJrLs2Q==".parse().unwrap()
+            )])
+        };
```
Hinton: **question:** You have the TestAccounts, if you can re-use them you would not need this boilerplate. `crates/bitwarden-core/src/client/test_accounts.rs`

crates/bitwarden-vault/src/cipher/cipher_client/create.rs
```diff
@@ -437,10 +554,80 @@ mod tests {
...
         .await;
 
         assert!(result.is_err());
         assert!(matches!(result.unwrap_err(), CreateCipherError::Api(_)));
     }
+
+    #[tokio::test]
+    async fn test_create_org_cipher() {
+        let (mock_server, _config) = start_api_mock(vec![
+            Mock::given(method("POST"))
```
Hinton: Do we still need wiremock here?

crates/bitwarden-vault/src/cipher/cipher_client/delete.rs
```diff
@@ -0,0 +1,349 @@
...
+impl<T> From<bitwarden_api_api::apis::Error<T>> for DeleteCipherError {
+    fn from(value: bitwarden_api_api::apis::Error<T>) -> Self {
+        Self::Api(value.into())
+    }
+}
+
+async fn delete_cipher(
+    cipher_id: CipherId,
+    api_client: &bitwarden_api_api::apis::ApiClient,
+    repository: &(impl Repository<Cipher> + ?Sized),
```
Hinton: For folders we use generics instead of anonymous type parameters:

```rust
pub(super) async fn create_folder<R: Repository<Folder> + ?Sized>(
    key_store: &KeyStore<KeyIds>,
    api_client: &bitwarden_api_api::apis::ApiClient,
    repository: &R,
    request: FolderAddEditRequest,
) -> Result<FolderView, CreateFolderError> {
```

Do we want to be consistent? I don't believe there is a performance hit for either pattern though.

crates/bitwarden-vault/src/cipher/cipher_client/edit.rs
Hinton: Upcasting it to a CryptoError is probably better. I think that should be as simple as doing `CryptoError::Decrypt(err)`

crates/bitwarden-vault/src/cipher/cipher_client/get.rs
```diff
@@ -16,7 +18,11 @@ pub enum GetCipherError {
     #[error(transparent)]
     Crypto(#[from] CryptoError),
     #[error(transparent)]
+    VaultParse(#[from] VaultParseError),
+    #[error(transparent)]
     RepositoryError(#[from] RepositoryError),
+    #[error(transparent)]
+    Api(#[from] ApiError),
```
Hinton: **issue:** Remove? Now that you have the admin error these can be removed right?

crates/bitwarden-vault/src/cipher/cipher.rs
```diff
@@ -636,6 +637,11 @@ impl Cipher {
         }
         Ok(())
     }
+
+    pub(crate) fn soft_delete(&mut self) {
```
Hinton, line 645: **suggestion:** Document?


```suggestion
    /// Marks the cipher as soft deleted by setting `deletion_date` to now.
    pub(crate) fn soft_delete(&mut self) {
```

crates/bitwarden-vault/src/cipher/cipher.rs
```diff
@@ -991,6 +1006,13 @@ impl From<bitwarden_api_api::models::CipherRepromptType> for CipherRepromptType
     }
 }
 
+/// A trait for merging partial cipher data into a full cipher.
+/// Used to convert from API response models to full Cipher structs,
+/// without losing local data that may not be present in the API response.
+pub(crate) trait PartialCipher {
+    fn merge_with_cipher(self, cipher: Option<Cipher>) -> Result<Cipher, VaultParseError>;
```
Hinton, line 1016: **question:** Why accept an optional cipher? The function might be nicer if cipher is required and you just do `and_then` on callers.

Hinton (commented):
Most of my issues are resolved and I think this looks generally good. There are couple of smaller things and improvements to be made but once resolved feel free to not request another review from me.

crates/bitwarden-vault/src/cipher/cipher_client/edit.rs
nikwithak: There's no `CryptoError::Decrypt(_)` variant, but there is a `CryptoError::KeyDecrypt` (with no inner value) - Updated logic to map to that variant.

crates/bitwarden-vault/src/cipher/cipher_client/create.rs
nikwithak: I made it pub(super) so it can be shared with the admin client (in a child mod), but not outside of the `cipher_client` mod. Since the struct is private outside (super), the `create_request` attr should be inaccessible outside of that scope anyway, but I added it to be more specific.

crates/bitwarden-vault/src/cipher/cipher_client/get.rs
nikwithak: No, these are still used by the operations, since they make API calls and parse the responses using the non-admin endpoints. The Admin operations only apply in certain use cases.

crates/bitwarden-vault/src/cipher/cipher.rs
nikwithak: We have some situations where we don't necessarily have the original cipher when parsing the response - the intent is to allow the functions to create a `Cipher` object from the details in the response, in that event.

crates/bitwarden-vault/src/cipher/cipher_client/delete.rs
nikwithak: I believe the `impl` syntax is just syntactic sugar for the generics - `impl` is personal preference / habit - but for consistency, I updated to use the generics.

crates/bitwarden-vault/src/cipher/cipher_client/create.rs
nikwithak: I manually created them because there were no functions that allowed me to both init with the test accounts and supply clientsettings, without adding changes to core. However, since removing wiremock the tests no longer use the client directly and call the underlying standalone function, so this is moot now anyway!

crates/bitwarden-vault/src/cipher/cipher_client/create.rs
nikwithak: Nope! I missed this one when updating the others to use the mock ApiClient

crates/bitwarden-vault/src/cipher/cipher_client/get.rs
Hinton: Not that I can see, there is nothing here producing parse or api errors.

nikwithak: My bad - I was looking at the wrong file. Removed!

crates/bitwarden-vault/src/cipher/cipher_client/admin/create.rs
```diff
@@ -0,0 +1,189 @@
...
+    }
+}
+
+/// Wraps the API call to create a cipher using the admin endpoint, for easier testing.
+async fn create_cipher(
+    request: CipherCreateRequestInternal,
+    encrypted_for: UserId,
+    api_client: &bitwarden_api_api::apis::ApiClient,
+    key_store: &KeyStore<KeyIds>,
+) -> Result<CipherView, CreateCipherError> {
```
shane-melton: ⛏️ Should this be `CreateCipherAdminError` instead of `CreateCipherError`?
```suggestion
) -> Result<CipherView, CreateCipherAdminError> {
```

crates/bitwarden-vault/src/cipher/cipher_client/admin/create.rs
```diff
@@ -0,0 +1,189 @@
...
+}
+
+#[cfg_attr(feature = "wasm", wasm_bindgen)]
+impl CipherAdminClient {
+    /// Creates a new [Cipher] for an organization, using the admin server endpoints endpoints.
+    /// Creates the Cipher on the server only, does not store it to local state.
+    pub async fn create(
+        &self,
+        request: CipherCreateRequest,
+    ) -> Result<CipherView, CreateCipherError> {
```
shane-melton: ⛏️ Same here:
```suggestion
    ) -> Result<CipherView, CreateCipherAdminError> {
```

crates/bitwarden-vault/src/cipher/cipher.rs
```diff
@@ -639,6 +640,12 @@ impl Cipher {
         }
         Ok(())
     }
+
+    /// Marks the cipher as soft deleted by setting `deletion_date` to now.
+    pub(crate) fn soft_delete(&mut self) {
+        self.deleted_date = Some(Utc::now());
+        self.archived_date = None;
```
shane-melton: ⚠️ We're no longer clearing `archived_date` on soft deletes. See [PM-29650](https://bitwarden.atlassian.net/browse/PM-29650)

crates/bitwarden-vault/src/cipher/cipher_client/admin/create.rs
```diff
@@ -0,0 +1,189 @@
...
+        }))
+        .await?
+        .merge_with_cipher(None)?;
+
+    Ok(key_store.decrypt(&cipher)?)
+}
+
+#[cfg_attr(feature = "wasm", wasm_bindgen)]
+impl CipherAdminClient {
+    /// Creates a new [Cipher] for an organization, using the admin server endpoints endpoints.
```
shane-melton: ⛏️ typo
```suggestion
    /// Creates a new [Cipher] for an organization, using the admin server endpoints.
```

crates/bitwarden-vault/src/cipher/cipher_client/admin/get.rs
shane-melton, line 1: ♻️ We don't need to add more to this ticket, but I imagine we'll need/want a method to get all _full_ organization ciphers and not just the list views. I know the AC vault uses the [full cipher views currently](https://github.com/bitwarden/clients/blob/4d00d0caa5c4a728952a67efeabc1abf763ec272/apps/web/src/app/admin-console/organizations/collections/vault.component.ts#L335-L342) and the list views might not be sufficient on their own.

crates/bitwarden-vault/src/cipher/cipher_client/admin/get.rs
```diff
@@ -0,0 +1,70 @@
...
+use crate::{
+    VaultParseError,
+    cipher::cipher::{DecryptCipherListResult, PartialCipher},
+    cipher_client::admin::CipherAdminClient,
+};
+
+#[allow(missing_docs)]
+#[bitwarden_error(flat)]
+#[derive(Debug, Error)]
+pub enum GetOrganizationCiphersError {
```
shane-melton: ❓ Should this error type also have a reference to "Admin" somewhere in its name like all the other errors for the Admin client?

crates/bitwarden-vault/src/cipher/cipher_client/mod.rs
```diff
@@ -184,6 +187,13 @@ impl CiphersClient {
         let decrypted_key = cipher_view.decrypt_fido2_private_key(&mut key_store.context())?;
         Ok(decrypted_key)
     }
+
+    /// Returns a new client for performing admin operations.
```
shane-melton, line 191: 🎨 Since this is most discoverable place for the `CipherAdminClient`, it might be beneficial to reiterate that these operations use special admin endpoints and do not modify local state. Similar to what you have on the `CipherAdminClient` documentation itself.
```suggestion
    /// Returns a new client for performing admin operations.
    /// Uses the admin server API endpoints and does not modify local state.
```

shane-melton (changes requested):
Really great work on this! Very clean and easy to follow along. I had a few small nits/suggestions and one blocker regarding archive dates and soft delete.

crates/bitwarden-vault/src/cipher/cipher_client/admin/get.rs
nikwithak: I've added this to my list of SDK improvements and will create tickets for them.

shane-melton (approved):
Looks good! Though for some reason the test runner / workflow seems to be running out of space 😕

PR #563, opened 2025-11-18
[PM-27230] Introduce Account Cryptographic State

quexten:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-27230
https://bitwarden.atlassian.net/browse/PM-27313

Client PR, fixing breaking changes: https://github.com/bitwarden/clients/pull/17488

## 📔 Objective

Account cryptographic state describes the core cryptographic objects making up a user. For a V1 encryption user these are: The RSA Keypair. For a V2 user these are:
- The Public-key Encryption Keypair
- The Signature Keypair
- The Signed public key (used to bind a public-key encryption key pair to signing keypair)
- The Signed security state (used for a safe upgrade path that prevents insecure features)

Note: The wrapped version omits the public key and verifying key for the public-key encryption key-pair and signature key-pair, and only includes the wrapped private / signing keys.

Not included are:
- Keys shared to the user via an organization memebership or emergency access
- Cipher keys
which are not part of the user's cryptography, but do interact with it.

Provided is a function to generate such a cryptographic state for v2 users, and conversion to API request models. Further, this changes SDK initialization to be based on the account cryptographic state.

## 🚨 Breaking Changes

Please note that the public API for initializing the user's account cryptography is updated to instead take an enum. The variants for the enum contain the private key for V1 users, and the private key, signing key, signed public key, signed security state for V2 users. 

Aside from re-packaging into an enum variant, no other changes should be needed on the consuming side.

This both helps prevent inconsistent states from being passed in by enforcing consistent state via type safetey, but also cleans up the primitive obsession anti-pattern (https://contributing.bitwarden.com/architecture/server/#avoid-primitive-obsession) that the crypto initialization was facing and will make future changes much easier.

Note that since so far clients did not store the signed public key, it is optional for now, but will be made mandatory later after clients save it for a sufficient time.

To migrate, simply repack the existing values into the corresponding enum. Note that previously the signed public key was not passed in, now it is passed in (optional for now, but required later on). Please ensure it is saved to state on sync / login.

## ⏰ Reminders before review

- CI builds passed
  team

## 🦮 Reviewer guidelines


crates/bitwarden-core/src/key_management/account_cryptographic_state.rs
```diff
@@ -0,0 +1,425 @@
...
+    }
+
+    /// Creates a new V2 account cryptographic state with fresh keys.
+    pub fn make(
+        store: &KeyStore<KeyIds>,
+        user_id: UserId,
+    ) -> Result<Self, AccountCryptographyInitializationError> {
+        let mut ctx = store.context_mut();
+
+        let user_key = ctx.make_symmetric_key(SymmetricKeyAlgorithm::XChaCha20Poly1305);
```
quexten, line 186: (Not for this PR)

@dani-garcia Any thoughts on moving these functions out to the domain objects? I.e something like:
```rust
let private_key = PrivateKey::make(PublicKeyEncryptionAlgorithm::RsaOaepSha1, &mut ctx)?;
```
similarly, the operations such as wrap would be moved to the respective structs.

so that we only have:
- Delete
- Move
left on the context?

dani-garcia: That sounds reasonable, should also help reduce the amount of functions we expose in a context too

crates/bitwarden-core/src/key_management/account_cryptographic_state.rs
```diff
@@ -0,0 +1,489 @@
...
+                    .unwrap_private_key(user_key, private_key)
+                    .map_err(|_| AccountCryptographyInitializationError::WrongUserKey)?;
+                let signing_key_id = ctx
+                    .unwrap_signing_key(user_key, signing_key)
+                    .map_err(|_| AccountCryptographyInitializationError::WrongUserKey)?;
+
+                let security_state: SecurityState = security_state
+                    .to_owned()
+                    .verify_and_unwrap(&ctx.get_verifying_key(signing_key_id)?)
+                    .map_err(|_| AccountCryptographyInitializationError::CorruptData)?;
```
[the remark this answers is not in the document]
quexten: Sure, replaced with a more specific error.

crates/bitwarden-core/src/key_management/account_cryptographic_state.rs
```diff
@@ -0,0 +1,497 @@
...
+                    signed_public_key
+                        .to_owned()
+                        .verify_and_unwrap(&ctx.get_verifying_key(signing_key_id)?)
+                        .map_err(|_| AccountCryptographyInitializationError::TamperedData)?;
+                }
+
+                let security_state: SecurityState = security_state
+                    .to_owned()
+                    .verify_and_unwrap(&ctx.get_verifying_key(signing_key_id)?)
+                    .map_err(|_| AccountCryptographyInitializationError::TamperedData)?;
```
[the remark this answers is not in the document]
quexten: Sure, when we add tracing.

crates/bitwarden-core/src/auth/tde.rs
```diff
@@ -34,12 +34,10 @@ pub(super) fn make_register_tde_keys(
 
     client.internal.initialize_user_crypto_decrypted_key(
         user_key.0,
-        UserKeyState {
+        // TODO (https://bitwarden.atlassian.net/browse/PM-21771) Signing keys are not supported on registration yet. This needs to be changed as
+        // soon as registration is supported.
+        WrappedUserAccountCryptographicState::V1 {
```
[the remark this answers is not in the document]
quexten: This comment does not make much sense.

crates/bitwarden-core/src/key_management/account_cryptographic_state.rs
```diff
@@ -0,0 +1,497 @@
...
+            object: Some("privateKeys".to_string()),
+            signature_key_pair: match self {
+                WrappedUserAccountCryptographicState::V1 { .. } => None,
+                WrappedUserAccountCryptographicState::V2 { signing_key, .. } => Some(Box::new(
+                    bitwarden_api_api::models::SignatureKeyPairResponseModel {
+                        wrapped_signing_key: Some(signing_key.to_string()),
+                        verifying_key: Some(
+                            B64::from(
+                                verifying_key
+                                    .clone()
```
dani-garcia: You should be able to use `.as_ref()` here to avoid the clone.
```suggestion
                                    .as_ref()
```

crates/bitwarden-crypto/src/signing/verifying_key.rs
```diff
@@ -20,13 +20,15 @@ use crate::{
 
 /// A `VerifyingKey` without the key id. This enum contains a variant for each supported signature
 /// scheme.
+#[derive(Clone)]
```
dani-garcia: With the `.as_ref()` change above (https://github.com/bitwarden/sdk-internal/pull/563/files#r2547084705), you don't need to derive clone here.

crates/bitwarden-core/src/key_management/crypto.rs
```diff
@@ -166,7 +164,7 @@ pub(super) async fn initialize_user_crypto(
         client.internal.init_user_id(user_id)?;
     }
 
-    let key_state = (&req).into();
+    let key_state = req.account_cryptographic_state.clone();
```
dani-garcia: This clone doesn't seem needed
```suggestion
    let key_state = req.account_cryptographic_state;
```

crates/bitwarden-core/src/key_management/crypto.rs
```diff
@@ -1412,9 +1400,12 @@ mod tests {
...
                 },
                 email: "test@bitwarden.com".into(),
-                private_key: enrollment_response.private_key,
-                signing_key: Some(enrollment_response.signing_key),
-                security_state: Some(enrollment_response.security_state),
+                account_cryptographic_state: WrappedUserAccountCryptographicState::V2 {
+                    private_key: enrollment_response.private_key.clone(),
+                    signing_key: enrollment_response.signing_key.clone(),
+                    security_state: enrollment_response.security_state.clone(),
+                    signed_public_key: Some(enrollment_response.signed_public_key.clone()),
```
dani-garcia: Clones also not needed here
```suggestion
                    private_key: enrollment_response.private_key,
                    signing_key: enrollment_response.signing_key,
                    security_state: enrollment_response.security_state,
                    signed_public_key: Some(enrollment_response.signed_public_key),
```

crates/bitwarden-core/src/client/internal.rs
```diff
@@ -292,59 +272,31 @@ impl InternalClient {
...
+            .map_err(|_| EncryptionSettingsError::CryptoInitialization)
     }
 
     #[cfg(feature = "internal")]
     pub(crate) fn initialize_user_crypto_pin(
         &self,
         pin_key: PinKey,
         pin_protected_user_key: EncString,
-        key_state: UserKeyState,
+        key_state: WrappedUserAccountCryptographicState,
```
dani-garcia: Can we rename the parameters to these three functions to also be `account_crypto_state` for consistency?

crates/bitwarden-crypto/src/store/context.rs
```diff
@@ -208,6 +247,156 @@ impl<Ids: KeyIds> KeyStoreContext<'_, Ids> {
...
+    /// Move a symmetric key from one identifier to another within the context
+    ///
+    /// The key value is copied to `to` and the original identifier `from` is removed.
+    /// This operates on either the local or global store depending on whether the ids are
+    /// local identifiers.
+    ///
+    /// # Errors
+    /// Returns an error if the source key does not exist or if setting the destination key
+    /// fails (for example due to read-only global store).
+    pub fn move_symmetric_key(&mut self, from: Ids::Symmetric, to: Ids::Symmetric) -> Result<()> {
```
dani-garcia: The comment in `move_signing_key` says `This method is useful for renaming or migrating keys between local identifiers`. That seems not entirely correct to me, the purpose of this function based on how it's used at the moment seems to be moving local keys to global keys only. 

Now that local keys are using unique guids, I can't imagine needing to move/rename between them, I also can't imagine wanting to move/rename global keys into other global keys. So the only possible move I can think of is local -> global.

Maybe we can rename these methods to make it more clear, and name them something like `promote_symmetric_key` or `persist_symmetric_key`? Any thoughts?

crates/bitwarden-crypto/src/store/context.rs
```diff
@@ -323,16 +512,25 @@ impl<Ids: KeyIds> KeyStoreContext<'_, Ids> {
         self.add_local_symmetric_key(SymmetricCryptoKey::make_aes256_cbc_hmac_key())
     }
 
-    /// Generate a new random xchacha20-poly1305 symmetric key and store it in the context
-    #[cfg(test)]
-    pub(crate) fn make_cose_symmetric_key(
+    /// Generate a new symmetric encryption key using the specified algorithm and store it in the
+    /// context as a local key
+    pub fn make_symmetric_key(&mut self, algorithm: SymmetricKeyAlgorithm) -> Ids::Symmetric {
```
dani-garcia, line 531: Not necesary as part of this PR, but now that we have this `make_symmetric_key` here, and the `move_symmetric_key` functions, we should audit all the uses of the deprecated `set_symmetric_key` and replace them if possible.

I know there are a lot of tests of the form:
```rust
        let symm_key = SymmetricCryptoKey::make_aes256_cbc_hmac_key();

        #[allow(deprecated)]
        store
            .context_mut()
            .set_symmetric_key(TestSymmKey::A(0), symm_key.clone())
            .unwrap();
```

that could be simplified with these new functions.

crates/bitwarden-crypto/src/store/context.rs
```diff
@@ -208,6 +247,156 @@ impl<Ids: KeyIds> KeyStoreContext<'_, Ids> {
...
+        Ok(signing_key.to_verifying_key())
+    }
+
+    /// Return the public key corresponding to an asymmetric (private) key identifier.
+    ///
+    /// This converts the stored private key into its public key representation.
+    ///
+    /// # Errors
+    /// Returns an error if the asymmetric key id does not exist.
+    pub fn get_public_key(
```
dani-garcia, line 406: Nice! I had a task to do something like this but never followed up on it: https://bitwarden.atlassian.net/browse/PM-18110

There's at least one use on `generate_user_fingerprint ` that could benefit from this, though not necessarily as part of this PR.

crates/bitwarden-crypto/src/store/context.rs
```diff
@@ -463,8 +674,27 @@ impl<Ids: KeyIds> KeyStoreContext<'_, Ids> {
         key_id
     }
 
+    /// Get the type of a symmetric key stored in the context.
+    pub fn get_algorithm(&self, key_id: Ids::Symmetric) -> Result<SymmetricKeyAlgorithm> {
```
dani-garcia: Should this be named `get_symmetric_key_algorithm` or something, in case we ever want to add methods for asymmetric and signing keys too?

crates/bitwarden-crypto/src/store/context.rs
```diff
@@ -149,6 +164,30 @@ impl<Ids: KeyIds> KeyStoreContext<'_, Ids> {
         self.local_asymmetric_keys.retain(f);
     }
 
+    fn drop_symmetric_key(&mut self, key_id: Ids::Symmetric) -> Result<()> {
```
dani-garcia, line 167: These as currently written will silently fail to delete the key if passed a non-local key while using a read-only context. We should probably change the implementation to be more like:

```rust
    fn drop_symmetric_key(&mut self, key_id: Ids::Symmetric) -> Result<()> {
        if key_id.is_local() {
            self.local_symmetric_keys.remove(key_id);
        } else {
            self.global_keys.get_mut()?.symmetric_keys.remove(key_id);
        }
        Ok(())
    }
```

crates/bitwarden-core/src/key_management/account_cryptographic_state.rs
```diff
@@ -0,0 +1,497 @@
...
+                signed_public_key: Some(signed_public_key),
+                signing_key: ctx.wrap_signing_key(user_key, signing_key)?,
+                security_state: signed_security_state,
+            },
+        ))
+    }
+
+    /// Set the decrypted account cryptographic state to the context's non-local storage. Note, that
+    /// this drops the context and clears the existing local state.
+    pub fn set_to_context(
```
dani-garcia: This function should indicate in the comment that it requires a mutable context. Also it should be pub(crate) as it's not used outside the crate

dani-garcia (commented):
Looks pretty good overall, thanks for all the added docs and for cleaning up EncryptionSettings!

I've left a bunch of comments with some small improvements.

crates/bitwarden-crypto/src/store/context.rs
quexten: I agree, the comment didn't make much sense. Updated these to be named `persist_symmetric_key`. This now enforces `from` to be local, and `to` to be global.

crates/bitwarden-crypto/src/store/context.rs
quexten: I migrated a bunch of these now. `create_test_crypto_with_user_key` is not migrated yet, as this uses some static test vectors sometimes.

crates/bitwarden-crypto/src/store/context.rs
quexten: Agreed, but leaving out of scope from the PR.

dani-garcia (dismissed):

crates/bitwarden-collections/src/collection.rs
```diff
@@ -162,14 +162,14 @@ mod tests {
...
         let org_id = ORGANIZATION_ID.parse().unwrap();
 
-        #[allow(deprecated)]
-        store
-            .context_mut()
-            .set_symmetric_key(SymmetricKeyId::Organization(org_id), key)
+        let mut ctx = store.context_mut();
+
+        let local_key_id = ctx.make_symmetric_key(SymmetricKeyAlgorithm::Aes256CbcHmac);
+        ctx.persist_symmetric_key(local_key_id, SymmetricKeyId::Organization(org_id))
```
Hinton, line 170: **question:** It seems that these two operations are called fairly often together. Would it make sense to offer some way to create a persisted key immediately? Behind the scene we have to drop and re-create the key in a new memory location.

Is it reasonable that anyone with the knowledge of the key id can persist the key? Presumably people wouldn't do it, but it would be completely valid to do so.

crates/bitwarden-core/src/key_management/account_cryptographic_state.rs
```diff
@@ -0,0 +1,503 @@
...
+    }
+}
+
+/// Any keys / cryptographic protection "downstream" from the account symmetric key (user key).
+/// Private keys are protected by the user key.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+#[cfg_attr(feature = "uniffi", derive(uniffi::Enum))]
+#[cfg_attr(feature = "wasm", derive(Tsify), tsify(into_wasm_abi, from_wasm_abi))]
+#[allow(clippy::large_enum_variant)]
+pub enum WrappedUserAccountCryptographicState {
```
Hinton: **question:** Why is this called `WrappedUserAccountCryptographicState`? The file is called `account_cryptographic_state`. Most functions and objects using it call it `account_cryptographic_state` or `account_crypto_state`.

crates/bitwarden-core/src/key_management/account_cryptographic_state.rs
```diff
@@ -0,0 +1,503 @@
...
+}
+
+impl WrappedUserAccountCryptographicState {
+    /// Converts to a PrivateKeysResponseModel in order to make API requests. Since the
+    /// [WrappedUserAccountCryptographicState] is encrypted, the key store needs to contain the
+    /// user key required to unlock this state.
+    pub fn to_private_keys_request_model(
+        &self,
+        store: &KeyStore<KeyIds>,
+    ) -> Result<PrivateKeysResponseModel, AccountCryptographyInitializationError> {
```
Hinton: **issue:** The function implies you are making request models, but the return type is response models?

crates/bitwarden-core/src/key_management/account_cryptographic_state.rs
```diff
@@ -0,0 +1,503 @@
...
+                            .verify_and_unwrap(vk_ref)
+                            .map_err(|_| AccountCryptographyInitializationError::TamperedData)?
+                            .version() as i32,
+                    }))
+                }
+            },
+        })
+    }
+
+    /// Creates a new V2 account cryptographic state with fresh keys.This does not change the user
```
Hinton:
```suggestion
    /// Creates a new V2 account cryptographic state with fresh keys. This does not change the user
```

crates/bitwarden-core/src/key_management/account_cryptographic_state.rs
```diff
@@ -0,0 +1,503 @@
...
+                signed_public_key: Some(signed_public_key),
+                signing_key: ctx.wrap_signing_key(user_key, signing_key)?,
+                security_state: signed_security_state,
+            },
+        ))
+    }
+
+    /// Set the decrypted account cryptographic state to the context's non-local storage.
+    /// This needs a mutable context passed in that already has a user_key set to a local key slot,
+    /// for which che id is passed in as `user_key`. Note, that this function drops the context
```
Hinton:
```suggestion
    /// for which the id is passed in as `user_key`. Note, that this function drops the context
```

crates/bitwarden-core/src/key_management/account_cryptographic_state.rs
```diff
@@ -0,0 +1,503 @@
...
+    /// and clears the existing local state, after persisting it.
+    pub(crate) fn set_to_context(
+        &self,
+        sdk_security_state: &RwLock<Option<SecurityState>>,
```
Hinton: **question:** What does `sdk_security_state` mean? Is there a non sdk security state?

crates/bitwarden-core/src/key_management/account_cryptographic_state.rs
```diff
@@ -0,0 +1,503 @@
...
+                    .unwrap_signing_key(SymmetricKeyId::User, signing_key)
+                    .map_err(|_| AccountCryptographyInitializationError::WrongUserKey)?;
+                ctx.get_verifying_key(signing_key)
+                    .map(Some)
+                    .map_err(|e| e.into())
+            }
+        }
+    }
+
+    /// Retrieve the verifying key from the wrapped state, if present. This requires the user key to
```
Hinton:
```suggestion
    /// Retrieve the public key from the wrapped state, if present. This requires the user key to
```

crates/bitwarden-core/src/key_management/crypto_client.rs
```diff
@@ -1,4 +1,4 @@
-#[cfg(feature = "wasm")]
+#[cfg(any(feature = "wasm", test))]
```
Hinton: **question:** Is this needed? There doesn't seem to be any other changes to this file.

crates/bitwarden-core/src/key_management/security_state.rs
```diff
@@ -114,6 +114,13 @@ impl TryFrom<&CoseSign1Bytes> for SignedSecurityState {
     }
 }
 
+impl From<&SignedSecurityState> for String {
+    fn from(val: &SignedSecurityState) -> Self {
+        let bytes: CoseSign1Bytes = val.clone().into();
+        B64::from(bytes.as_ref()).to_string()
```
Hinton: **nit:** You already have the owned implementation below. I think you can just clone and use the existing implementation?

crates/bitwarden-crypto/src/store/context.rs
Hinton, line 1: Just an observation for the future.

Multiple of the functions only return a single error variant, we could make the return types more explicit and only return those specific errors.

That does mean a breaking change should we re-add error variants though.

Hinton, line 1: Another observation, since it's up to the caller to specify what algorithm we use the places that consumes the `make_key` should probably be limited so that migrating to a new algorithm doesn't turn out to be too difficult.

Hinton (commented):
Looks good, some thoughts, nothing blocking.

We seem to use multiple names for the same thing `account_crypto_state`, `account_cryptographic_state` and `WrappedUserAccountCryptographicState`. Is there some way to be more consistent? What does `WrappedUser` imply in this case?

To be clear I'm not against having variables be shorter than the full struct name.

crates/bitwarden-collections/src/collection.rs
quexten: If the concern is convenience of the public api, yeah, i think this makes sense. That said, generally, we should not be creating "persisted" keys that often, and the local case should be by far the more common case, outside of tests.

If the concern is performance, then since we don't create keys often, the microseconds don't matter for any use-case I'm aware of.

For the convenience concern, can we move this to future work? I don't want to scope creep the PR more if not needed since it's already large.

crates/bitwarden-core/src/key_management/account_cryptographic_state.rs
quexten: Renamed to `security_state_rwlock`. This is mainly because of a naming conflict with the security_state read from the UserAccountCryptographicState.

crates/bitwarden-core/src/key_management/crypto_client.rs
quexten: Undone.

crates/bitwarden-core/src/key_management/account_cryptographic_state.rs
quexten: The intent here is to communicate to people that the secret parts of this object - in this case keys - are protected (key wrapped). Sometimes it is not clear to consumers and we have gotten questions about "is this X encrypted?", so wrapped is terminology that advertises that.

However, the file as a whole is meant to cover the user's account cryptographic state as a whole and anything relating to it. I.e:
- Generation of a new state and the error relating to this
- Validation of a state
- Setting the state to a key context

So in my mind it made sense to not have the filename focus on the detail that the publicly exposed struct happens to have the secret parts wrapped.

That said, I've removed the "User" part at least since it was very inconsistently used / redundant with "Account".

crates/bitwarden-core/src/key_management/security_state.rs
quexten: Makes sense. Replaced with `to_owned` + re-using the existing impl.

crates/bitwarden-collections/src/collection.rs
dani-garcia: I'd be somewhat reluctant to create convenience functions that modify the global persisted state as it would make it harder to audit what code is modifying the keys and in what way.

I think all usages of the `persist_*_key` functions outside of `account_cryptographic_state.rs` are just for testing, so if we're thinking about convenience what we need is more convenient test-only initialization functions for KeyStore, but I agree that that's a separate body of work from this PR.

 If we expose that test-only utilities that do the initialization, it would also make it easier to search for all the usages of `persist_symmetric_key` in production code.

crates/bitwarden-core/src/key_management/account_cryptographic_state.rs
quexten: Sorry, this had the wrong model. We have some models which are really the same thing, which is also the case here, causing a bit of confusion. Replaced with "AccountKeysRequestModel" now.

crates/bitwarden-crypto/src/store/context.rs
quexten: Would this mean per-function error enums / error structs since you can't just use a variant as a return type? I'm not exactly sure how this looks in practice.

quexten: Agreed. The pattern generally is being moved to either:
- Use a `safe` struct like dataenvelope / passwordprotectedkeyenvelope / etc. that internally handles keygen
- Have a high level KM api that other teams consume that does not expose the specific alg outside of the KM-owned code in `core` (should probably be it's own crate)

But there is still code to migrate to these patterns.

crates/bitwarden-core/src/key_management/account_cryptographic_state.rs
```diff
@@ -0,0 +1,515 @@
...
+
+                if let Some(signed_public_key) = signed_public_key {
+                    signed_public_key
+                        .to_owned()
+                        .verify_and_unwrap(&ctx.get_verifying_key(signing_key_id)?)
+                        .map_err(|_| AccountCryptographyInitializationError::TamperedData)?;
+                }
+
+                let security_state: SecurityState = security_state
+                    .to_owned()
```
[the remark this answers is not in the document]
quexten: Yeah... no.. A microsecond during an operation that takes a few hundred milliseconds does not make a difference.

mzieniukbw (approved):

gbubemismith (approved):
Vault changes look good

dani-garcia (approved):

PR #586, opened 2025-11-27
Refactor CLAUDE.md documentation to improve AI code review guidance

theMickster:
## 📔 Objective

The updates to these various markdowns are the culmination of a number of iterations with Claude Code, Copilot and a few teammates to aid in better Claude Code Reviews.  A general consensus is that the information that Claude has given thus far in our Rust SDK needs to improve.  I attempted to combine many of the thoughts in the `copilot-instructions.md` into our CLAUDE.md because there was good information in that document. 

I'm no Rust expert so I will need help in double-checking that the suggestions are valid. 
Thanks!

### DIFF
--- a/.claude/CLAUDE.md
+++ b/.claude/CLAUDE.md
@@ -54,7 +54,7 @@ Monorepo crates organized in **four architectural layers**:
   - Enums: `#[cfg_attr(feature = "uniffi", derive(uniffi::Enum))]`
   - Include `uniffi::setup_scaffolding!()` in lib.rs
 - **bitwarden-wasm-internal**: WebAssembly bindings (**thin bindings only - no business logic**)
-  - Structs: `#[cfg_attr(feature = "wasm", derive(Tsify), tsify(into_wasm_abi, from_wasm_abi))]`
+  - Structs: `#[derive(Serialize, Deserialize)]` with `#[cfg_attr(feature = "wasm", derive(Tsify), tsify(into_wasm_abi, from_wasm_abi))]`
 
 ## Critical Patterns & Rules
 
--- a/.claude/CLAUDE.md
+++ b/.claude/CLAUDE.md
@@ -64,7 +64,7 @@ Monorepo crates organized in **four architectural layers**:
 - **KeyStoreContext**: Never hold across await points (enforced by clippy lint
   `await_holding_invalid_type`)
 - Naming: `derive_` for deterministic key derivation, `make_` for non-deterministic generation
-- Use `bitwarden-crypto/src/safe` high-level APIs first (password-protected key envelope, data
+- Use `bitwarden_crypto::safe` module first (password-protected key envelope, data envelope)
   envelope)
 
 ### State Management (bitwarden-state)
--- a/.claude/CLAUDE.md
+++ b/.claude/CLAUDE.md
@@ -7,7 +7,8 @@ Cross-platform Rust SDK implementing Bitwarden's core business logic.
 
 **Crate documentation**: Before working in any crate, read available documentation: `CLAUDE.md` for
 critical rules, `README.md` for architecture, `examples/` for usage patterns, and `tests/` for
-integration tests. **Read all available files before suggesting changes.**
+integration tests. **Before making changes or reviewing code, review relevant examples and tests for
+the specific functionality you're modifying.**
 
 ## Architecture Overview
 
@@ -15,13 +16,11 @@ Monorepo crates organized in **four architectural layers**:
 
 ### 1. Foundation Layer
 
-- **bitwarden-crypto**: Cryptographic primitives, KeyStore with thread-safe RwLock-based access,
-  ephemeral KeyStoreContext with auto-cleared local keys, COSE format support
+- **bitwarden-crypto**: Cryptographic primitives and protocols, key store for securely working with
+  keys held in memory.
 - **bitwarden-state**: Type-safe Repository pattern for SDK state (client-managed vs SDK-managed)
 - **bitwarden-threading**: ThreadBoundRunner for !Send types in WASM/GUI contexts (uses PhantomData
-  marker)
-- **bitwarden-ipc**: Type-safe IPC framework with pluggable encryption/transport, Arc-based shared
-  ownership
+  marker) **bitwarden-ipc**: Type-safe IPC framework with pluggable encryption/transport
 - **bitwarden-error**: Error handling across platforms (basic/flat/full modes via proc macro)
 - **bitwarden-encoding**, **bitwarden-uuid**: Encoding and UUID utilities
 
@@ -29,6 +28,8 @@ Monorepo crates organized in **four architectural layers**:
 
 - **bitwarden-core**: Base Client struct extended by feature crates via extension traits. **DO NOT
   add functionality here - use feature crates instead.**
+  - **bitwarden-api-api**, **bitwarden-api-identity**: Auto-generated API clients (**DO NOT edit -
+    regenerate from OpenAPI specs**)
 
 ### 3. Feature Implementations
 
@@ -36,8 +37,8 @@ Monorepo crates organized in **four architectural layers**:
   `vault()`, `crypto()`, `sends()`, `generators()`, `exporters()`
   - Lifecycle: Init → Lock/Unlock → Logout (drops instance)
   - Storage: Apps use `HashMap<UserId, PasswordManagerClient>`
-- **bitwarden-vault**: Vault item models (encrypted/decrypted states) with Encryptable trait
-- **bitwarden-collections**: Collection models with tree structure via TreeItem trait
+- **bitwarden-vault**: Vault item models, encryption/decryption and management
+- **bitwarden-collections**: Collection models, encryption/decryption and management
 - **bitwarden-auth**: Authentication (send access tokens)
 - **bitwarden-send**: Encrypted temporary secret sharing
 - **bitwarden-generators**: Password/passphrase generators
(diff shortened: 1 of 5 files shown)


.claude/CLAUDE.md
```diff
@@ -1,61 +1,122 @@
...
+**Build & Test:**
+
+- `cargo check --all-features --all-targets` - Quick validation
+- `cargo test --workspace --all-features` - Full test suite
+
+**Format & Lint:**
+
+- `cargo +nightly fmt --workspace` - Code formatting
+- `cargo clippy --workspace` - Linting (includes custom await_holding_invalid_type for
+  KeyStoreContext)
```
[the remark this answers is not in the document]
theMickster: Good one Claude! There's 71 usages of `clippy` across 36 files in the codebase. lol

crates/bitwarden-error-macro/CLAUDE.md
```diff
@@ -0,0 +1,27 @@
...
+Read any available documentation: [README.md](./README.md) for architecture,
+[examples/](./examples/) for usage patterns, and [tests/](./tests/) for integration tests.
+
+## Three Modes
+
+Specified via `#[bitwarden_error(mode)]`:
+
+- **basic**: String errors only—converts to JS string via `ToString`
+- **flat**: Variant-based—generates `FlatError` trait, TypeScript union types
+- **full**: Structure-based—uses `Serialize` + `tsify` for full error details
```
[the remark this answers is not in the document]
theMickster: Not really the intention here @claude because its supposed to help YOU figure out our code base better and make better code suggestions. Not for extensive developer documentation.

crates/bitwarden-threading/CLAUDE.md
```diff
@@ -0,0 +1,12 @@
+# bitwarden-threading
+
+Read any available documentation: [README.md](./README.md) for architecture,
+[examples/](./examples/) for usage patterns, and [tests/](./tests/) for integration tests.
+
+## Critical Rules
+
+**Native requires LocalSet**: `tokio::task::spawn_local` panics without LocalSet context—WASM does
```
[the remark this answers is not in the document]
theMickster: Again, not needed @claude because this is for YOU to understand what's going on. You should be able to interpret the rule and judge code accordingly.

.claude/CLAUDE.md
```diff
@@ -1,61 +1,122 @@
...
+- **bitwarden-generators**: Password/passphrase generators
+- **bitwarden-ssh**: SSH key generation/import
+- **bitwarden-exporters**: Vault export/import with multiple formats
+- **bitwarden-fido**: FIDO2 two-factor authentication
 
-- Do not log keys, passwords, or vault data in logs or error paths. Redact sensitive data.
+### 4. Cross-Platform Bindings
 
-**Business logic in WASM**
+- **bitwarden-api-api**, **bitwarden-api-identity**: Auto-generated API clients (**DO NOT edit -
```
quexten: I think the API creates should actually be be in the core infrastructure. Viewing this as a dependency graph, they would be called by something like `bitwarden-vault`.

.claude/CLAUDE.md
```diff
@@ -1,61 +1,122 @@
...
-**Unsafe without justification**
+### Cryptography (bitwarden-crypto)
 
-- Any `unsafe` block needs a comment explaining why it's safe and what invariants are being upheld.
+- **DO NOT modify** without careful consideration - backward compatibility is critical
+- **KeyStoreContext**: Never hold across await points (enforced by clippy lint
+  `await_holding_invalid_type`)
+- Naming: `derive_` for deterministic key derivation, `make_` for non-deterministic generation
+- Use `bitwarden_crypto::safe` module first (password-protected key envelope, data envelope)
+  envelope)
```
quexten, line 72:
```suggestion
- IMPORTANT: Use constant time equality checks
- Do not expose low-level / hazmat functions from the crypto crate. 
- Do not expose key material from the crypto crate, use key references in the key store instead

```

.claude/CLAUDE.md
```diff
@@ -1,61 +1,122 @@
...
-- Changes to `bitwarden-api-api/` or `bitwarden-api-identity/` are generally discouraged. These are
-  auto-generated from swagger specs.
+### 3. Feature Implementations
 
-**Secrets in logs/errors**
+- **bitwarden-pm**: PasswordManagerClient wrapping core Client, exposes sub-clients: `auth()`,
+  `vault()`, `crypto()`, `sends()`, `generators()`, `exporters()`
+  - Lifecycle: Init → Lock/Unlock → Logout (drops instance)
+  - Storage: Apps use `HashMap<UserId, PasswordManagerClient>`
+- **bitwarden-vault**: Vault item models (encrypted/decrypted states) with Encryptable trait
```
quexten: Question: Is the "Encryptable trait" part here relevant?

```suggestion
- **bitwarden-vault**: Vault item models, encryption/decryption and management
```
note: The and management part is not implemented yet, but eventually state and api calls should/will be done in this crate.

.claude/CLAUDE.md
```diff
@@ -1,61 +1,122 @@
...
+- **bitwarden-collections**: Collection models with tree structure via TreeItem trait
```
quexten:
```suggestion
- **bitwarden-collections**: Collection models, encryption/decryption and management
```

.claude/CLAUDE.md
```diff
@@ -1,61 +1,122 @@
...
 
-- Init → Lock/Unlock → Logout (drops instance). Memento pattern for state resurrection.
+Monorepo crates organized in **four architectural layers**:
 
-**Storage**
+### 1. Foundation Layer
 
-- Consuming apps use `HashMap<UserId, PasswordManagerClient>`.
+- **bitwarden-crypto**: Cryptographic primitives, KeyStore with thread-safe RwLock-based access,
+  ephemeral KeyStoreContext with auto-cleared local keys, COSE format support
```
quexten:
```suggestion
- **bitwarden-crypto**: Cryptographic primitives and protocols, key store for securely working with keys held in memory.
```

.claude/CLAUDE.md
```diff
@@ -1,61 +1,122 @@
...
+- **bitwarden-state**: Type-safe Repository pattern for SDK state (client-managed vs SDK-managed)
+- **bitwarden-threading**: ThreadBoundRunner for !Send types in WASM/GUI contexts (uses PhantomData
+  marker)
+- **bitwarden-ipc**: Type-safe IPC framework with pluggable encryption/transport, Arc-based shared
```
quexten:
```suggestion
- **bitwarden-ipc**: Type-safe IPC framework with pluggable encryption/transport
```

.claude/CLAUDE.md
```diff
@@ -1,61 +1,122 @@
...
-**Unsafe without justification**
+### Cryptography (bitwarden-crypto)
 
-- Any `unsafe` block needs a comment explaining why it's safe and what invariants are being upheld.
+- **DO NOT modify** without careful consideration - backward compatibility is critical
+- **KeyStoreContext**: Never hold across await points (enforced by clippy lint
+  `await_holding_invalid_type`)
+- Naming: `derive_` for deterministic key derivation, `make_` for non-deterministic generation
+- Use `bitwarden_crypto::safe` module first (password-protected key envelope, data envelope)
+  envelope)
```
quexten:
```suggestion
- Use `bitwarden_crypto::safe` module first (password-protected key envelope, data envelope)
  envelope) instead of more low-level primitives
```

.claude/CLAUDE.md
```diff
@@ -1,61 +1,122 @@
...
-**Breaking serialization**
+### Security Requirements
 
-- Backward compatibility required. Users must decrypt vaults from older versions.
+- **Never log** keys, passwords, or vault data in logs or error paths
+- **Redact sensitive data** in all error messages
+- **Unsafe blocks** require comments explaining safety and invariants
+- **Encryption/decryption changes** must maintain backward compatibility (existing encrypted data
+  must remain decryptable)
+- **Breaking serialization** prohibited - users must decrypt vaults from older versions
```
quexten: Note: We can make exceptions here in case we have a good migration path, notice to users, and long lead time. However, i don't think it's worth recording here, and those cases should just note such / overrule the claude comments that will come.

I thought it's just worth mentioning here.

quexten (commented):
Some small comments/suggestions on the main claude file. I did not review error-macro / ipc / state / threading since the other reviewers have a better grasp on these.

Also looks like there is a prettier formatting error, but that can be resolved when all other reviewers comments are resolved.

Other than that mostly looks good.

.claude/CLAUDE.md
theMickster: Thank you!

.claude/CLAUDE.md
theMickster: Likely not needed/

.claude/CLAUDE.md
theMickster: I reworded slightly; I think that it should work well for a first go 'round with Claude Code.

.claude/CLAUDE.md
```diff
@@ -1,61 +1,126 @@
...
-- Do not log keys, passwords, or vault data in logs or error paths. Redact sensitive data.
+### Cryptography (bitwarden-crypto)
 
-**Business logic in WASM**
+- **DO NOT modify** without careful consideration - backward compatibility is critical
+- **KeyStoreContext**: Never hold across await points
+- Naming: `derive_` for deterministic key derivation, `make_` for non-deterministic generation
+- Use `bitwarden_crypto::safe` module first (password-protected key envelope, data envelope) instead
+  of more low-level primitives
+- IMPORTANT: Use constant time equality checks
```
quexten, line 69: Somehow these 3 lines became duplicated, maybe an incorrectly applied git diff in the web UI?

theMickster: I dunno where/how that snuck in

quexten (dismissed):
Looks good for an initial round. I think we can iterate if we find it not focusing enough or focusing too much on some details.

withinfocus (dismissed):
Supporting approval after Bernd's review.

.claude/CLAUDE.md
```diff
@@ -1,61 +1,124 @@
...
+- `cargo test --workspace --all-features` - Full test suite
+
+**Format & Lint:**
+
+- `cargo +nightly fmt --workspace` - Code formatting
+- Use `cargo clippy` to lint code and catch common mistakes
+
+**WASM Testing:**
+
+- `cargo test --target wasm32-unknown-unknown --all-features` - WASM-specific tests
```
dani-garcia: This command can't be run directly from root, as some crates depend on certain features that aren't available in WASM contexts. We should probably not use `--all-features` either as that would enable the UniFFI feature of some packages that is also not WASM compatible.

Instead, we have two options:
- We indicate that this command should be run only in the directories of the relevant crates (at the moment that is `bitwarden-error`, `bitwarden-threading`, `bitwarden-uuid`)
- We update the command to only run on the crates with wasm-specific tests:
```
cargo test --target wasm32-unknown-unknown --features wasm -p bitwarden-error -p bitwarden-threading -p bitwarden-uuid
```

dani-garcia (dismissed):
Looks good to me, just a minor nit that will cause WASM tests to fail if not run on the correct directory.

.claude/CLAUDE.md
theMickster: Thanks! I'll update it now.

theMickster: Updated to reflect those changes

dani-garcia (approved):

withinfocus (approved):

PR #592, opened 2025-11-30
[PM-28845] Add PureCrypto functions to replace webcrypto RSA entirely

quexten:
## 🎟️ Tracking

https://github.com/bitwarden/clients/pull/17742
https://bitwarden.atlassian.net/browse/PM-28845

## 📔 Objective

Preparing to nuke webcrypto's RSA implementation out of orbit. It is unreliable and we have reproduced around 50% of all private keys keys are generated by rustcrypto that are not parseable on webcrypto on Safari (but parseable on chrome / firefox). It is unclear why, most likely safari has some stricter condition for the keys (RSA ASN1 has many conditions that may apply here), or has a bug. We also suspect keys generated on old Xamarin PCLCrypto to be parseable by rust, or at least deliver good error messages on rust, but be unparseable on all of webcrypto. This changes the functions to be exactly the same as webcrypto so that they can be dropped in without refactoring anything else.

Note: Usually, implementing crypto directly in the wasm crate would not be a good idea, *however* is is justifiable here because we DO NOT WANT these to be public functions consumeable by any other crate, so they do not belong in the crypto crate. They are also only used by the wasm clients.

## ⏰ Reminders before review

- CI builds passed
  team

## 🦮 Reviewer guidelines

### DIFF
--- a/crates/bitwarden-wasm-internal/Cargo.toml
+++ b/crates/bitwarden-wasm-internal/Cargo.toml
@@ -33,7 +33,10 @@ bitwarden-state = { workspace = true, features = ["wasm"] }
 bitwarden-threading = { workspace = true }
 bitwarden-vault = { workspace = true, features = ["wasm"] }
 console_error_panic_hook = "0.1.7"
+rand = ">=0.8.5, <0.9"
+rsa = ">=0.9.2, <0.10"
 serde = { workspace = true }
+sha1 = ">=0.10.5, <0.11"
 tracing = { workspace = true }
 tracing-subscriber = { workspace = true }
 tracing-web = { version = "0.1.3" }
--- a/crates/bitwarden-wasm-internal/src/pure_crypto.rs
+++ b/crates/bitwarden-wasm-internal/src/pure_crypto.rs
@@ -328,11 +328,11 @@ impl PureCrypto {
     /// returns the corresponding public RSA key in DER format.
     pub fn rsa_extract_public_key(private_key: Vec<u8>) -> Result<Vec<u8>, RsaError> {
         let private_key = AsymmetricCryptoKey::from_der(&Pkcs8PrivateKeyBytes::from(private_key))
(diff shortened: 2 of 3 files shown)


crates/bitwarden-wasm-internal/src/pure_crypto.rs
```diff
@@ -319,23 +317,59 @@ impl PureCrypto {
         Ok(result.to_encoded().to_vec())
     }
 
-    /// Given an encrypted private RSA key and the symmetric key it is wrapped with, this returns
+    /// Given a decrypted private RSA key PKCS8 DER and the symmetric key it is wrapped with, this returns
     /// the corresponding public RSA key in DER format.
     pub fn rsa_extract_public_key(
```
quexten: note: breaknig change is fine here. This is not used yet. I will drop the clients PR replacing it with one using these functions.

crates/bitwarden-wasm-internal/Cargo.toml
Hinton, line 1: These should probably be moved to the workspace `Cargo.toml` since they are consumed by multiple crates at this point.

Hinton (approved):
Should we add some disclaimers to the individual functions that you are better off using the `AsymmEncString` abstractions?

mzieniukbw (approved):

crates/bitwarden-wasm-internal/Cargo.toml
quexten: Good point. I will make a follow-up PR that adds hazmat warnings to all new functions introduced here. We don't want any other callers other than cryptofunctionservice, which itself has hazmat labels.

djsmith85 (approved):

crates/bitwarden-wasm-internal/Cargo.toml
quexten: https://github.com/bitwarden/sdk-internal/pull/595

PR #596, opened 2025-12-02
[PM-27232] Implement Registration for TDE Users

quexten:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-27232
https://github.com/bitwarden/server/pull/6671

## 📔 Objective

Implements registration for TDE users. This uses v2 encryption.

> [!NOTE]  
> The API changes here are not final and need an update to the automatic API bindings, after the corresponding server change is merged. This PR will only be merged once the binding changes are already in `main` and removed from this PR.

## ⏰ Reminders before review

- CI builds passed
  team

## 🦮 Reviewer guidelines


crates/bitwarden-auth/src/registration.rs
```diff
@@ -33,4 +45,129 @@ impl RegistrationClient {
...
+
+        // Next, enroll the user for reset password using the reset password key generated above.
+        info!("Enrolling into admin account recovery");
+        api_client
+            .organization_users_api()
+            .put_reset_password_enrollment(
+                uuid::Uuid::parse_str(&org_id).map_err(|_| UserRegistrationError::Serialization)?,
+                user_id.into(),
+                Some(OrganizationUserResetPasswordEnrollmentRequestModel {
+                    reset_password_key: Some(reset_password_key.to_string()),
```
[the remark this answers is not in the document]
quexten: Spam

crates/bitwarden-auth/src/registration.rs
```diff
@@ -33,4 +45,129 @@ impl RegistrationClient {
...
+        trust_device: bool,
+    ) -> Result<TdeRegistrationResult, UserRegistrationError> {
+        let client = &self.client.internal;
+        #[allow(unused_variables)]
+        let api_client = &client.get_api_configurations().await.api_client;
+        let user_id =
+            UserId::from_str(user_id.as_str()).map_err(|_| UserRegistrationError::Serialization)?;
+
+        // First call crypto API to get all keys
+        info!("Initializing account cryptography");
```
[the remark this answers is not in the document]
quexten: Spam

crates/bitwarden-auth/Cargo.toml
```diff
@@ -17,20 +17,27 @@ keywords.workspace = true
...
 wasm = [
     "bitwarden-core/wasm",
+    "bitwarden-crypto/wasm",
     "dep:tsify",
     "dep:wasm-bindgen",
     "dep:wasm-bindgen-futures"
 ] # WASM support
 
 # Note: dependencies must be alphabetized to pass the cargo sort check in the CI pipeline.
 [dependencies]
```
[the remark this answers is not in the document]
quexten: Spam

crates/bitwarden-core/src/key_management/crypto_client.rs
```diff
@@ -193,6 +205,77 @@ impl CryptoClient {
...
         derive_key_connector(request)
     }
+
+    /// Creates a new V2 account cryptographic state for TDE registration.
+    /// This generates fresh cryptographic keys (private key, signing key, signed public key,
+    /// and security state) wrapped with a new user key.
+    ///
+    /// Returns the wrapped account cryptographic state that can be used for registration.
+    /// The user key is not returned but is set in the client's key store.
+    pub fn make_user_tde_registration(
```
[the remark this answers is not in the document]
quexten: I'll leave this open for reviewers to discuss. Happy to change it, however, each type is different and there is no possibility of mixing them up because of that.

crates/bitwarden-core/src/key_management/crypto_client.rs
```diff
@@ -193,6 +205,77 @@ impl CryptoClient {
     ) -> Result<B64, DeriveKeyConnectorError> {
         derive_key_connector(request)
     }
+
+    /// Creates a new V2 account cryptographic state for TDE registration.
```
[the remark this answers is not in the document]
quexten: The algorithms are implementation detail that should not be publicly exposed via documentation. We don't generally document parameters, but they should be self descriptive.

crates/bitwarden-auth/src/registration.rs
```diff
@@ -33,4 +45,129 @@ impl RegistrationClient {
...
+                user_id.into(),
+                Some(OrganizationUserResetPasswordEnrollmentRequestModel {
+                    reset_password_key: Some(reset_password_key.to_string()),
+                    master_password_hash: None,
+                }),
+            )
+            .await
+            .map_err(|_| UserRegistrationError::Api)?;
+
+        if trust_device {
```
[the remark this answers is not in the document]
quexten: I don't think this is worth it. There is no issue with just discarding keys here.

crates/bitwarden-auth/src/registration.rs
```diff
@@ -33,4 +45,129 @@ impl RegistrationClient {
...
+                .await
+                .map_err(|_| UserRegistrationError::Api)?;
+        }
+
+        info!("User initialized!");
+        // Note: This passing out of state and keys is temporary. Once SDK state management is more
+        // mature, the account cryptographic state and keys should be set directly here.
+        Ok(TdeRegistrationResult {
+            account_cryptographic_state: cryptography_state,
+            device_key: device_key_set.device_key.to_string(),
```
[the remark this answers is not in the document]
quexten: This is expected.

crates/bitwarden-core/src/key_management/crypto_client.rs
```diff
@@ -193,6 +205,77 @@ impl CryptoClient {
...
+        let device_key = DeviceKey::trust_device(user_key)?;
+
+        // Account recovery enrollment
+        let public_key =
+            AsymmetricPublicCryptoKey::from_der(&SpkiPublicKeyBytes::from(&org_public_key))
+                .map_err(MakeKeysError::Crypto)?;
+        let admin_reset = UnsignedSharedKey::encapsulate_key_unsigned(user_key, &public_key)
+            .map_err(MakeKeysError::Crypto)?;
+
+        let store = KeyStore::default();
```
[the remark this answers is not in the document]
quexten: We use this to perform a safe conversion to the request model. In the future we may refactor.

crates/bitwarden-auth/src/registration.rs
```diff
@@ -33,4 +45,129 @@ impl RegistrationClient {
...
+        ) = self
+            .client
+            .crypto()
+            .make_user_tde_registration(user_id, org_public_key)
+            .map_err(|_| UserRegistrationError::Crypto)?;
+
+        // Post the generated keys to the API here. The user now has keys and is "registered", but
+        // has no unlock method.
+        let request = KeysRequestModel {
+            account_keys: Some(Box::new(account_cryptographic_state_request.clone())),
```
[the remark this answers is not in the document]
quexten: Clone is needed because we move to a box. The box is required by the auto-generated request model.

crates/bitwarden-core/src/key_management/crypto_client.rs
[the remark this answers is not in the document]
quexten: Spam

crates/bitwarden-core/src/key_management/crypto_client.rs
```diff
@@ -193,6 +205,77 @@ impl CryptoClient {
...
+        let mut ctx = store.context_mut();
+        let user_key_id = ctx.add_local_symmetric_key(user_key.to_owned());
+        let security_state = RwLock::new(None);
+        wrapped_state
+            .set_to_context(&security_state, user_key_id, &store, ctx)
+            .map_err(MakeKeysError::AccountCryptographyInitialization)?;
+
+        let cryptography_state_request_model = wrapped_state
+            .to_request_model(&store)
+            .map_err(|_| MakeKeysError::RequestModelCreation)?;
```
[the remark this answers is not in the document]
quexten: Spam

crates/bitwarden-auth/Cargo.toml
```diff
@@ -17,20 +17,27 @@ keywords.workspace = true
...
+bitwarden-api-api = { workspace = true }
 bitwarden-core = { workspace = true, features = ["internal"] }
-bitwarden-error = { workspace = true }
+bitwarden-crypto = { workspace = true }
+bitwarden-encoding = { workspace = true }
+bitwarden-error.workspace = true
```
JaredSnider-Bitwarden: Nit❓ : why change this syntax from the original syntax which matches the others in the file?

crates/bitwarden-auth/src/registration.rs
```diff
@@ -33,4 +45,129 @@ impl RegistrationClient {
...
+    /// admin password reset and finally enrolls the user to TDE unlock.
+    pub async fn post_keys_for_tde_registration(
+        &self,
+        org_id: String,
+        org_public_key: B64,
+        // Note: Ideally these would be set for the register client, however no such functionality
+        // exists at the moment
+        user_id: String,
+        device_id: String,
+        trust_device: bool,
```
JaredSnider-Bitwarden: 🎨 : If we used a request model here, it would be a bit more flexible than n params. I'm taking that approach with the login methods.

crates/bitwarden-auth/src/registration.rs
```diff
@@ -33,4 +45,129 @@ impl RegistrationClient {
...
+}
+
+/// Result of TDE registration process.
+#[cfg_attr(
+    feature = "wasm",
+    derive(tsify::Tsify),
+    tsify(into_wasm_abi, from_wasm_abi)
+)]
+#[derive(serde::Serialize, serde::Deserialize, Clone, Debug)]
+pub struct TdeRegistrationResult {
```
JaredSnider-Bitwarden: nit naming change request: In the Identity client WIP work, I'm using a `<Domain>Response` as my SDK domain model response pattern. It also removes any ambiguity around the Rust result type per @dani-garcia .

crates/bitwarden-auth/src/registration.rs
```diff
@@ -33,4 +45,129 @@ impl RegistrationClient {
...
+        org_id: String,
+        org_public_key: B64,
+        // Note: Ideally these would be set for the register client, however no such functionality
+        // exists at the moment
+        user_id: String,
+        device_id: String,
+        trust_device: bool,
+    ) -> Result<TdeRegistrationResult, UserRegistrationError> {
+        let client = &self.client.internal;
+        #[allow(unused_variables)]
```
JaredSnider-Bitwarden: ⚠️ : this is now used so we should clean this up

crates/bitwarden-auth/src/registration.rs
```diff
@@ -33,4 +45,129 @@ impl RegistrationClient {
...
+            cryptography_state,
+            user_key,
+            account_cryptographic_state_request,
+            device_key_set,
+            reset_password_key,
+        ) = self
+            .client
+            .crypto()
+            .make_user_tde_registration(user_id, org_public_key)
+            .map_err(|_| UserRegistrationError::Crypto)?;
```
JaredSnider-Bitwarden: ❓ :  This swallows the actual error - is there value in exposing it in a log or is that dangerous?

JaredSnider-Bitwarden: This also applies to the use of `.map_err(|_| UserRegistrationError::Api)?;` - we would get the most value out of not losing this for tracking down issues.

crates/bitwarden-auth/src/registration.rs
JaredSnider-Bitwarden, line 1: 🎨 : can we get some unit tests in this client for this new `post_keys_for_tde_registration` method please?

crates/bitwarden-auth/src/registration.rs
```diff
@@ -33,4 +45,129 @@ impl RegistrationClient {
...
+        })
+    }
+}
+
+/// Result of TDE registration process.
+#[cfg_attr(
+    feature = "wasm",
+    derive(tsify::Tsify),
+    tsify(into_wasm_abi, from_wasm_abi)
+)]
```
JaredSnider-Bitwarden, line 186: ⚠️ , I assume that mobile needs to be able to do this - thus, I think we need the UNIFFI macro here to add mobile support.

JaredSnider-Bitwarden (changes requested):
Looking good so far, but I have some small requests for changes and some questions that need clarification 👇

crates/bitwarden-auth/src/registration.rs
```diff
@@ -33,4 +45,128 @@ impl RegistrationClient {
...
+                .map_err(|_| UserRegistrationError::Api)?;
+        }
+
+        info!("User initialized!");
+        // Note: This passing out of state and keys is temporary. Once SDK state management is more
+        // mature, the account cryptographic state and keys should be set directly here.
+        Ok(TdeRegistrationResult {
+            account_cryptographic_state: cryptography_state,
+            device_key: device_key_set.device_key.to_string(),
+            user_key: user_key.to_encoded().to_vec().into(),
```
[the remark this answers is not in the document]
quexten: Spam. This is a valid use-case, as we are not comfortable using SDK state management, but we do need to set to state.

crates/bitwarden-auth/src/registration.rs
quexten: makes sense, applied

crates/bitwarden-auth/src/registration.rs
quexten: Question: Does this still apply if the parameters are just orgid, public key and trust_device?

The reason I'm asking is because user_id, and device_id could probably be dropped in a subsequent (different) PR, at which point it doesn't seem like too many parameters anymore.

crates/bitwarden-auth/src/registration.rs
quexten: added

crates/bitwarden-core/src/key_management/crypto_client.rs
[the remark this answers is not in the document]
quexten: We already went through this. Closing as spam.

crates/bitwarden-core/src/error.rs
```diff
@@ -52,7 +52,12 @@ impl_bitwarden_error!(IdentityError, ApiError);
 pub struct NotAuthenticatedError;
 
 /// Client's user ID is already set.
-#[derive(Debug, Error)]
+#[derive(Debug, Error, serde::Serialize, serde::Deserialize, Clone)]
+#[cfg_attr(
+    feature = "wasm",
+    derive(tsify::Tsify),
+    tsify(into_wasm_abi, from_wasm_abi)
+)]
```
dani-garcia: ❓Is this change needed here, it doesn't seem like this error is used anywhere on the PR

crates/bitwarden-core/src/key_management/crypto_client.rs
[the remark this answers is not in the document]
dani-garcia: I personally would also move it to a struct, not so much because someone would mix them up, but because the editor experience when using large tuples is quite cumbersome, as you have to manually name them all or use `_` just to skip them.

crates/bitwarden-auth/src/registration.rs
quexten: I had initially considered us wanting less info, but I think all crypto errors and api errors should be safe and not contain any keys or tokens, so we may as-well expose them correctly.

I'll add them.

crates/bitwarden-auth/src/registration.rs
```diff
@@ -33,4 +47,129 @@ impl RegistrationClient {
...
+    /// admin password reset and finally enrolls the user to TDE unlock.
+    pub async fn post_keys_for_tde_registration(
+        &self,
+        org_id: String,
+        org_public_key: B64,
+        // Note: Ideally these would be set for the register client, however no such functionality
+        // exists at the moment
+        user_id: String,
+        device_id: String,
+        trust_device: bool,
```
Thomas-Avery: ❓ 

For these ID parameters can we use the `OrganizationId` `UserId` etc., from `crates/bitwarden-core/src/ids.rs`, as the input types? Or is there some limitation for those?

crates/bitwarden-auth/src/registration.rs
Thomas-Avery: I agree with Jared on this, mostly just a style/preference thing having a bunch of primitives is hard to keep track of when reading the method.

Vs just seeing `request.<some_input_name>` lets me know right away it's coming from the request input.

crates/bitwarden-auth/src/registration.rs
```diff
@@ -33,4 +47,129 @@ impl RegistrationClient {
...
+
+        // Post the generated keys to the API here. The user now has keys and is "registered", but
+        // has no unlock method.
+        let request = KeysRequestModel {
+            account_keys: Some(Box::new(account_cryptographic_state_request.clone())),
+            // Note: This property is deprecated and will be removed
+            public_key: account_cryptographic_state_request
+                .account_public_key
+                .ok_or(UserRegistrationError::Crypto)?,
+            // Note: This property is deprecated and will be removed
```
Thomas-Avery: For these notes do we have a ticket tracking that, that we can reference here in the comments?

crates/bitwarden-auth/src/registration.rs
```diff
@@ -33,4 +47,129 @@ impl RegistrationClient {
...
+)]
+#[cfg_attr(feature = "uniffi", derive(uniffi::Record))]
+#[derive(serde::Serialize, serde::Deserialize, Clone, Debug)]
+pub struct TdeRegistrationResponse {
+    /// The account cryptographic state of the user
+    pub account_cryptographic_state: WrappedAccountCryptographicState,
+    /// The device key
+    pub device_key: String,
+    /// The decrypted user key. This can be used to get the consuming client to an unlocked state.
+    pub user_key: B64,
```
Thomas-Avery: ❓  Why do we use `String` for `device_key` and `B64` for `user_key`? 

Since they're both symmetric crypto keys I would expect the same return type.

crates/bitwarden-auth/src/registration.rs
Thomas-Avery: +1 for unit tests

crates/bitwarden-auth/src/registration.rs
JaredSnider-Bitwarden: I think it still results in a better, easier to extend API for future changes (including the deprecation of the `user_id`, `device_id`, and `trust_device` properties). We can add / remove fields without breaking caller compatibility which is a real pain point. I probably would request anyone on auth add a request model for 2 or more properties in one of our SDK APIs.

crates/bitwarden-core/src/key_management/crypto_client.rs
```diff
@@ -193,6 +205,77 @@ impl CryptoClient {
...
+
+        Ok((
+            wrapped_state,
+            user_key.to_owned(),
+            cryptography_state_request_model,
+            device_key,
+            admin_reset,
+        ))
```
Thomas-Avery:
```suggestion
        let mut ctx = self.client.internal.get_key_store().context_mut();
        let (user_key_id, wrapped_state) =
            WrappedAccountCryptographicState::make(&mut ctx, user_id)
                .map_err(MakeKeysError::AccountCryptographyInitialization)?;
        #[expect(deprecated)]
        let user_key = ctx.dangerous_get_symmetric_key(user_key_id)?;
        // TDE unlock method
        let device_key = DeviceKey::trust_device(user_key)?;

        // Account recovery enrollment
        let public_key =
            AsymmetricPublicCryptoKey::from_der(&SpkiPublicKeyBytes::from(&org_public_key))
                .map_err(MakeKeysError::Crypto)?;
        let admin_reset = UnsignedSharedKey::encapsulate_key_unsigned(user_key, &public_key)
            .map_err(MakeKeysError::Crypto)?;

        let cryptography_state_request_model = wrapped_state
            .to_request_model(self.client.internal.get_key_store())
            .map_err(|_| MakeKeysError::RequestModelCreation)?;

        Ok((
            wrapped_state,
            user_key.to_owned(),
            cryptography_state_request_model,
            device_key,
            admin_reset,
        ))
```

💭  It would seem like we would want to continue using the `ctx` and key store from the client? 

Maybe I'm missing something important or a side effect making the other key store is desired for. If so could we add a comment about what that is.

Thomas-Avery (changes requested):
Looking good. A few things to take a look at and some questions from my side. 

Also I think claude is wanting to have a chat on this one 😆

crates/bitwarden-auth/src/registration.rs
quexten: > We can add / remove fields without breaking caller compatibility which is a real pain point.

If you remove a field, that is still a breaking change unless you encapsulate the creation of the request model in a way where these are only optionally set, but there is no useful way to do that here.

If we change this to remove device id, and user id, this would still be a breaking change and require a similar change, now changing the request model instead of the properties.

crates/bitwarden-auth/src/registration.rs
quexten: Starting group dm on this.

crates/bitwarden-core/src/key_management/crypto_client.rs
quexten: We don't want `make_keys` to have the side-effect of setting the global user-keys in the state. That would be unexpected for the API contract.

The reason why we needed a key store here is because to_request_model needs a key store.

That said, I changed the to_request_model function to instead take a userkey_id, a mutable key context. That way, we can work off of the temporary local keys previously generated, which feels better and avoids the second key store.

crates/bitwarden-auth/src/registration.rs
quexten: https://bitwarden.slack.com/archives/C054ZQSBS49/p1765794042905109

We don't really have an approach for mocking API clients for this specific purpose yet in SDK which is why i was hesitant to scope creep the PR here to add support.

crates/bitwarden-auth/src/registration.rs
quexten: Moved to request model, but i'll still note that i don't think it will actually serve much of a benefit, especially after we remove the parameters again. Changes are breaking either way.

crates/bitwarden-auth/src/registration.rs
quexten: I don't think it's useful to comment it here. The flow would be to remove it on the server side, and the bindings update would automatically present a breaking change here, and make us remove it when updating the auto-generated bindings.

However, we should note it on the server. For reference it is https://bitwarden.atlassian.net/browse/PM-27329

crates/bitwarden-auth/src/registration.rs
quexten: Added tests.

crates/bitwarden-auth/src/registration.rs
quexten: Added tracing, please feel free to re-open if you feel that is not sufficient.

JaredSnider-Bitwarden (dismissed):
Nice work on these changes! Thank you for the tests. The only non blocking nit is that the registration file is going to get very large once we put all 4 flows into it. It might be worth deconstructing where each flow has its own file + own request models to keep the mental overhead lower. That can be restructured with the next flow though.

crates/bitwarden-auth/src/registration.rs
```diff
@@ -5,12 +5,45 @@
...
+pub struct TdeRegistrationRequest {
+    /// Organization ID to enroll in
+    pub org_id: OrganizationId,
+    /// Organization's public key for encrypting the reset password key. This should be verified by
+    /// the client and not verifying may compromise the security of the user's account.
+    pub org_public_key: B64,
+    /// User ID for the account being initialized
+    pub user_id: UserId,
+    /// Device ID for TDE enrollment
+    pub device_id: String,
```
Thomas-Avery:
```suggestion
    /// Device identifier for TDE enrollment
    pub device_identifier: String,
```

We have several entities that have both an `.Id` and `.Identifier`, device is one of them. I would suggest making it clear which one we are using.

crates/bitwarden-core/src/key_management/crypto_client.rs
```diff
@@ -193,6 +203,72 @@ impl CryptoClient {
...
+        ),
+        MakeKeysError,
+    > {
+        let mut ctx = self.client.internal.get_key_store().context_mut();
+        let (user_key_id, wrapped_state) =
+            WrappedAccountCryptographicState::make(&mut ctx, user_id)
+                .map_err(MakeKeysError::AccountCryptographyInitialization)?;
+        // TDE unlock method
+        #[expect(deprecated)]
+        let device_key = DeviceKey::trust_device(ctx.dangerous_get_symmetric_key(user_key_id)?)?;
```
Thomas-Avery: ❓  Are we planning to move to something else over this deprecated method? It seems like it would be easier to migrate off if we just have it one spot instead of sprinkled throughout the method.

Thomas-Avery (changes requested):

crates/bitwarden-core/src/key_management/crypto_client.rs
quexten: Needs refactors to device_key::trust_device to work on key ids. Can't do that here without scope creep.

Thomas-Avery: I'm not asking to do that. I'm commenting on we have `ctx.dangerous_get_symmetric_key(user_key_id)` and `#[expect(deprecated)]` in multiple places in this method. Why not just one?

quexten: Ah, sorry, ok. Doing that would not work because we need to take a mutable reference to the ctx to make the request model. But that doesn't work if we already mutable borrow to hold the reference to the key, if we just do it in one place first.

I suppose we could clone the key, and drop the reference, but cloning keys is really not the intended way of using keys. 

I'd rather make at tech debt task to move the consumers to use the ids + ctx. That's the proper way of migrating off of these, and will require us to change it in each of these places anyways.

crates/bitwarden-auth/src/registration.rs
quexten: Applied, nice catch!

crates/bitwarden-core/src/key_management/crypto_client.rs
Thomas-Avery: Gotcha make sense thanks for the clarification. I'm good with moving forward as is.

Thomas-Avery (dismissed):
LGTM

crates/bitwarden-core/src/key_management/crypto_client.rs
```diff
@@ -236,7 +320,71 @@ mod tests {
...
+        let org_key = AsymmetricCryptoKey::make(PublicKeyEncryptionAlgorithm::RsaOaepSha1);
+        let org_public_key_der = org_key
+            .to_public_key()
+            .to_der()
+            .expect("valid public key DER");
+        let org_public_key = B64::from(org_public_key_der.as_ref().to_vec());
+
+        // Create a client and generate TDE registration keys
+        let registration_client = Client::new(None);
+        let (wrapped_state, _user_key, _account_keys_request, device_key, admin_reset) =
```
dani-garcia: :art: I think this would benefit from being moved to a struct just so you don't have to write all the fields that you don't care about, but it's a minor thing

crates/bitwarden-core/src/key_management/account_cryptographic_state.rs
```diff
@@ -100,40 +99,45 @@ impl WrappedAccountCryptographicState {
...
-                WrappedAccountCryptographicState::V1 { .. } => None,
-                WrappedAccountCryptographicState::V2 { signing_key, .. } => Some(Box::new(
+                Some((signing_key, verifying_key)) => Some(Box::new(
                     bitwarden_api_api::models::SignatureKeyPairRequestModel {
                         wrapped_signing_key: Some(signing_key.to_string()),
                         verifying_key: Some(
                             B64::from(
-                                verifying_key
-                                    .as_ref()
+                                Some(verifying_key)
```
dani-garcia: :art: This is wrapping the key in Some and then mapping and converting to an error, even though it's never going to be None, we should simplify it like this:
```rust
 Some(verifying_key)
    .map(|vk| vk.to_cose())
    .ok_or(AccountCryptographyInitializationError::CorruptData)?,

// becomes

verifying_key.to_cose()
```

crates/bitwarden-core/src/key_management/account_cryptographic_state.rs
```diff
@@ -100,40 +99,45 @@ impl WrappedAccountCryptographicState {
...
                                     .map(|vk| vk.to_cose())
                                     .ok_or(AccountCryptographyInitializationError::CorruptData)?,
                             )
                             .to_string(),
                         ),
-                        signature_algorithm: verifying_key.as_ref().map(|vk| {
-                            match vk.algorithm() {
-                                SignatureAlgorithm::Ed25519 => "ed25519".to_string(),
-                            }
+                        signature_algorithm: Some(verifying_key).map(|vk| match vk.algorithm() {
```
dani-garcia: Same thing, we're mapping over a Some wrapped value, we can change it to:

```rust
signature_algorithm: Some(verifying_key).map(|vk| match vk.algorithm() {
    SignatureAlgorithm::Ed25519 => "ed25519".to_string(),
}),

// becomes

signature_algorithm: match verifying_key.algorithm() {
    SignatureAlgorithm::Ed25519 => "ed25519".to_string(),
},
```

crates/bitwarden-core/src/key_management/account_cryptographic_state.rs
```diff
@@ -100,40 +99,45 @@ impl WrappedAccountCryptographicState {
...
-                | WrappedAccountCryptographicState::V2 { private_key, .. } => {
-                    Some(private_key.to_string())
-                }
-            },
+            user_key_encrypted_account_private_key: Some(private_key.to_string()),
             // Note: This property is deprecated and should be removed after a transition period.
-            account_public_key: match self.public_key(store)? {
-                Some(pk) => Some(B64::from(pk.to_der()?).to_string()),
+            account_public_key: Some(B64::from(public_key.to_der()?).to_string()),
+            signature_key_pair: match signature_keypair.as_ref() {
```
dani-garcia: ♻️ Some of these conversions could probably be moved to a `From<>` implementation to reduce the size of this function. Not a blocking change though, as it would just increase this diff more.

crates/bitwarden-auth/src/registration.rs
```diff
@@ -22,15 +55,426 @@ impl RegistrationClient {
...
+            trust_device: true,
+        };
+
+        let result =
+            internal_post_keys_for_tde_registration(&registration_client, &api_client, request)
+                .await;
+
+        assert!(result.is_ok());
+        // Assert that the mock expectations were met
+        if let ApiClient::Mock(mut mock) = api_client {
```
dani-garcia, line 296: :seedling: Mental note for me, it might be good to create a single checkpoint method at the root mock client to avoid having to do them one by one.

dani-garcia (dismissed):
Looks good to me, I have a couple of minor stylistic comments but nothing blocking

crates/bitwarden-core/src/error.rs
quexten: Not sure how this one snuck in... Undoing.

crates/bitwarden-core/src/key_management/account_cryptographic_state.rs
quexten: Agreed, but leaving it out of this PR for now / up to later refactor.

quexten:
Moving to draft until API bindings are updated. I believe all review so far has been addressed, and we should be close to good to go.

Thomas-Avery (approved):

JaredSnider-Bitwarden (approved):

djsmith85 (approved):

dani-garcia (approved):

PR #585, opened 2025-11-27
[PM-18938] Isolate decryption support for type 0 user keys

eligrubb:
## 🎟️ Tracking

[https://bitwarden.atlassian.net/browse/PM-18938](https://bitwarden.atlassian.net/browse/PM-18938)

## 📔 Objective

Isolates Type 0 key decryption to the `master_key.rs:decrypt_user_key()` flow. All other uses now raise new `bitwarden-crypto` error `UnsupportedOperationError::DecryptionNotImplementedForKey`.

This is another small step in deprecating use of legacy Type 0 keys in the sdk, [see tech breakdown](https://bitwarden.atlassian.net/wiki/spaces/EN/pages/1932689422/Tech+Breakdown+-+Block+type+0+symmetric+keys+generally). Type 0 keys have been deprecated from active use for a long time, this continues the process of removing their functionality while still supporting master key decryption for old accounts.

Type 0-based encryption was already deprecated in a prior PR, this change isolates decryption. Prior work isolated clients to use only the `master_key.rs:decrypt_user_key` path (in TS clients, via `PureCrypto.decrypt_user_key_with_master_key` #465). Other uses of type 0 keys are expected to now throw an error.

## ⏰ Reminders before review

- Contributor guidelines followed
- All formatters and local linters executed and passed
- Written new unit and / or integration tests where applicable
- Protected functional changes with optionality (feature flags)
- Used internationalization (i18n) for all UI strings
- CI builds passed
- Communicated to DevOps any deployment requirements
- Updated any necessary documentation (Confluence, contributing docs) or informed the documentation
  team

## 🦮 Reviewer guidelines

- 👍 (`:+1:`) or similar for great changes
- 📝 (`:memo:`) or ℹ️ (`:information_source:`) for notes or general info
- ❓ (`:question:`) for questions
- 🤔 (`:thinking:`) or 💭 (`:thought_balloon:`) for more open inquiry that's not quite a confirmed
  issue and could potentially benefit from discussion
- 🎨 (`:art:`) for suggestions / improvements
- ❌ (`:x:`) or ⚠️ (`:warning:`) for more significant problems or concerns needing attention
- 🌱 (`:seedling:`) or ♻️ (`:recycle:`) for future improvements or indications of technical debt
- ⛏ (`:pick:`) for minor or nitpick changes

PR #611, opened 2025-12-12
[PM-27231] Account encryption v2 for key connector

mzieniukbw:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-27231

## 📔 Objective

Account encryption V2 for key connector.
Added `KeyConnectorKey`, since we have to migrate from `MasterKey` for key connector at some point.

Added Key Connector API client crate, can set the key connector key for the user into the Key Connector server.
  - Since key connector url is not known during client creation, it can't be added into [core's Client creation](https://github.com/bitwarden/sdk-internal/blob/b9cf7b78b23f308dde113374161bee863fa23236/crates/bitwarden-core/src/client/client.rs#L41)
  - Created to look almost as identical as the auto generated one.
  - Models will eventually be auto generated by openapi key connector schema, but this is outside of the scope for this PR for now. https://bitwarden.atlassian.net/browse/PM-30278

## 🚨 Breaking Changes

No changes to account registration V1.

## ⏰ Reminders before review

- CI builds passed
  team

## 🦮 Reviewer guidelines


crates/bitwarden-auth/src/registration.rs
```diff
@@ -33,4 +49,99 @@ impl RegistrationClient {
         let api_client = &client.get_api_configurations().await.api_client;
         // Do API request here. It will be authenticated using the client's tokens.
     }
+
+    /// Initializes a new cryptographic state for a user and posts it to the server; enrolls the
+    /// user to key connector unlock.
+    pub async fn post_keys_for_key_connector_registration(
```
quexten, line 77: Relaying feedback I got on my PR here, we probably want unit tests on this. Vault has some examples for this, but it generally involves moving the impl into a separate function that you pass in the "api client" dependency and other parameters to, then the test can replace that with a mocked api client.

crates/bitwarden-auth/src/registration.rs
```diff
@@ -33,4 +49,99 @@ impl RegistrationClient {
...
+        &self,
+        key_connector_url: String,
+        org_id: String,
```
quexten: Relaying feedback, this should use OrganizationId, UserId instead of the string types.

crates/bitwarden-auth/src/registration.rs
```diff
@@ -33,4 +49,99 @@ impl RegistrationClient {
...
+            .make_user_key_connector_registration(user_id)
+            .map_err(UserRegistrationError::AccountCryptographyMakeKeys)?;
+
+        info!("Posting key connector key to key connector server");
+        key_connector_api_post_or_put_key_connector_key(
+            &self.client,
+            key_connector_url.as_str(),
+            &key_connector_key,
+        )
+        .await
```
quexten, line 94: Relaying feedback; We should add tracing for the errors here. We don't want to expand the returned error type, but log that the api request went wrong (and possibly how if it doesn't contain sensitive data).

crates/bitwarden-auth/src/registration.rs
```diff
@@ -33,4 +49,99 @@ impl RegistrationClient {
...
+        info!("Posting user account cryptographic state to server");
+        let request = SetKeyConnectorKeyRequestModel {
+            key_connector_key_wrapped_user_key: Some(wrapped_user_key.to_string()),
+            account_keys: Some(Box::new(account_cryptographic_state_request)),
+            ..SetKeyConnectorKeyRequestModel::new(org_id)
+        };
+        api_client
+            .accounts_key_management_api()
+            .post_set_key_connector_key(Some(request))
+            .await
```
quexten, line 182: Same here for tracing the error

crates/bitwarden-auth/src/registration.rs
```diff
@@ -33,4 +49,99 @@ impl RegistrationClient {
...
+#[derive(serde::Serialize, serde::Deserialize, Clone, Debug)]
+pub struct KeyConnectorRegistrationResult {
+    /// The account cryptographic state of the user.
+    pub account_cryptographic_state: WrappedAccountCryptographicState,
+    /// The key connector key used for unlocking.
+    pub key_connector_key: B64,
+    /// The encrypted user key, wrapped with the key connector key.
+    pub key_connector_key_wrapped_user_key: EncString,
+    /// The decrypted user key. This can be used to get the consuming client to an unlocked state.
+    pub user_key: ByteBuf,
```
quexten: This should be updated to B64 (the TDE PR does this too), since both user_key and key_connector_key are keys. Also, ByteBuf is not nicely supported by uniffi

crates/bitwarden-auth/src/registration.rs
```diff
@@ -33,4 +49,99 @@ impl RegistrationClient {
...
+    }
+}
+
+/// Result of Key Connector registration process.
+#[cfg_attr(
+    feature = "wasm",
+    derive(tsify::Tsify),
+    tsify(into_wasm_abi, from_wasm_abi)
+)]
+#[derive(serde::Serialize, serde::Deserialize, Clone, Debug)]
```
quexten, line 313: Should derive UNIFFI, however, please note that you have to merge in latest main, because auth fixed some stuff for UNIFFI in the meantime.

crates/bitwarden-core/src/key_management/crypto_client.rs
```diff
@@ -193,6 +203,73 @@ impl CryptoClient {
...
+    > {
+        let mut ctx = self.client.internal.get_key_store().context_mut();
+        let (user_key, wrapped_state) =
+            WrappedAccountCryptographicState::make(&mut ctx, user_id)
+                .map_err(AccountCryptographyMakeKeysError::AccountCryptographyInitialization)?;
+        #[expect(deprecated)]
+        let user_key = ctx.dangerous_get_symmetric_key(user_key)?;
+
+        // Key Connector unlock method
+        let mut rng = ChaCha8Rng::from_seed([0u8; 32]);
```
quexten: WARNING, BLOCKING: This will always deterministically generate the same key, and all users will have the same key connector key going through this code / their key connector key would be publically known. Seeded ChaCha8Rng should only be used in tests to *deterministically* derive a key.

To get a rng here you should use threadrng instead, which samples a rng based on your OS/platform API.

quexten: We should probably change all public APIs to remove the possibility of even passing in an rng that is bad. (@Thomas-Avery thoughts?)

crates/bitwarden-core/src/key_management/crypto_client.rs
```diff
@@ -193,6 +203,73 @@ impl CryptoClient {
...
+        let key_connector_key = KeyConnectorKey::generate(&mut rng);
+
+        let wrapped_user_key = key_connector_key
+            .encrypt_user_key(user_key)
+            .map_err(AccountCryptographyMakeKeysError::Crypto)?;
+
+        let store = KeyStore::default();
```
quexten: Please note, the TDE PR got a suggestion to update how set_to_context should work, which eliminates the need for the second keystore.

crates/bitwarden-crypto/src/keys/key_connector_key.rs
```diff
@@ -0,0 +1,37 @@
...
+use bitwarden_encoding::B64;
+use generic_array::GenericArray;
+use rand::Rng;
+use typenum::U32;
+
+use crate::{EncString, SymmetricCryptoKey, keys::utils::stretch_key};
+
+/// Key connector key, used to protect the user key.
+#[derive(Clone, Debug)]
+pub struct KeyConnectorKey(pub(super) Pin<Box<GenericArray<u8, U32>>>);
```
quexten, line 15: Can we exclude the contained key from being printed? In general the debug implementations for keys should never print the contents.

```rust
impl std::fmt::Debug for KeyConnectorKey {
    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
        f.debug_struct("KeyConnectorKey").finish()
    }
}
```

quexten (changes requested):
Nice, thanks for your work on this. Some items that changed in the TDE request after responding to the feedback on it, and it seems like chacha rng was left over (from testing?) which would currently always generate the same key for all users.

crates/bitwarden-crypto/src/keys/key_connector_key.rs
```diff
@@ -0,0 +1,37 @@
...
+
+impl KeyConnectorKey {
+    /// Generate a new random for KeyConnector.
+    pub fn generate(mut rng: impl rand::RngCore) -> Self {
```
quexten: suggestion: rename this to "make", then remove the rng parameter that can be passed in, and use thread rng directly inside here.

If we want deterministic unit tests, we should flag a function to make this key deterministically behind a feature gate.

crates/bitwarden-auth/src/registration.rs
mzieniukbw: Fixed in 27a3e16a

crates/bitwarden-auth/src/registration.rs
mzieniukbw: Fixed in 27a3e16a

crates/bitwarden-auth/src/registration.rs
mzieniukbw: Fixed in 27a3e16a

crates/bitwarden-auth/src/registration.rs
mzieniukbw: Fixed in 27a3e16a

crates/bitwarden-auth/src/registration.rs
mzieniukbw: Fixed in 27a3e16a

crates/bitwarden-core/src/key_management/crypto_client.rs
mzieniukbw: Fixed in 27a3e16a

crates/bitwarden-core/src/key_management/crypto_client.rs
mzieniukbw: Fixed in 27a3e16a

crates/bitwarden-crypto/src/keys/key_connector_key.rs
mzieniukbw: Fixed in 27a3e16a

crates/bitwarden-crypto/src/keys/key_connector_key.rs
mzieniukbw: Fixed in 27a3e16a

crates/bitwarden-core/Cargo.toml
```diff
@@ -76,8 +76,10 @@ rustls = { version = "0.23.19", default-features = false }
 rustls-platform-verifier = "0.6.0"
 
 [dev-dependencies]
-rand_chacha = "0.3.1"
+mockall = { version = ">=0.13.1, <0.15" }
+rand_chacha = ">=0.3.1, <0.4.0"
 tokio = { workspace = true, features = ["rt"] }
+wiremock = { workspace = true }
```
Hinton: We generally do not use wiremock any longer and instead use the `ApiClient::new_mocked`. There are a couple of examples of this being used.

mzieniukbw: I need to mock key connector api client (`KeyConnectorApiClient`), which is not part of `ApiClient`. I am adding similar mocking capability to that new client, so i can test the registration with the key connector api client being mocked.
For the testing of `KeyConnectorApiClient` itself, i wanted to make sure the correct headers, requests and responses are sent, which is why wiremock was used for testing. See tests in `crates/bitwarden-core/src/key_management/key_connector.rs` in the PR.

Hinton: Argh, that's annoying. It's probably useful to extract that to it's own crate though? We're trying to trim down `-core`

mzieniukbw: Added new `bitwarden-api-key-connector` crate

crates/bitwarden-crypto/src/keys/key_connector_key.rs
```diff
@@ -0,0 +1,196 @@
...
+    pub fn make() -> Self {
+        let mut rng = rand::thread_rng();
+        let mut key = Box::pin(GenericArray::<u8, U32>::default());
+
+        rng.fill(key.as_mut_slice());
+        KeyConnectorKey(key)
+    }
+
+    #[allow(missing_docs)]
+    pub fn to_base64(&self) -> B64 {
```
quexten: Nit: I believe this should just be moved to a `impl From<KeyConnectorKey> for B64`.

crates/bitwarden-crypto/src/keys/key_connector_key.rs
```diff
@@ -0,0 +1,196 @@
...
+    ) -> crate::error::Result<SymmetricCryptoKey> {
+        let dec: Vec<u8> = match user_key {
+            // Legacy. user_keys were encrypted using `Aes256Cbc_B64` a long time ago. We've since
+            // moved to using `Aes256Cbc_HmacSha256_B64`. However, we still need to support
+            // decrypting these old keys.
+            EncString::Aes256Cbc_B64 { .. } => {
+                let legacy_key = SymmetricCryptoKey::Aes256CbcKey(super::Aes256CbcKey {
+                    enc_key: Box::pin(GenericArray::clone_from_slice(&self.0)),
+                });
+                user_key.decrypt_with_key(&legacy_key)?
```
quexten: Will this work? I believe this will conflict with https://github.com/bitwarden/sdk-internal/pull/585

quexten: (We also don't have test coverage on this, but i also don't know whether type 0 wrapped key connector users ever existed)

crates/bitwarden-auth/src/registration.rs
```diff
@@ -195,10 +219,109 @@ pub struct TdeRegistrationResponse {
...
+        info!("User's key connector key does not exist, creating");
+        key_connector_api_client
+            .user_keys_api()
+            .post_user_key(request)
+            .await
+    })
+    .map_err(|e| {
+        error!("Failed to post key connector key to key connector server: {e:?}");
+        RegistrationError::KeyConnectorApi
+    })?;
```
quexten: For readability, it feels like this logic should live in a separate function. Something like:

`async fn post_key_to_key_connector(api_client: [..], key_connector_key: &KeyConnectorKey) -> Result<...,...>`

mzieniukbw: separated in b31f666a

crates/bitwarden-crypto/src/keys/key_connector_key.rs
mzieniukbw: fixed in 17d20d9b. Added test coverage for type 0 wrapped key

crates/bitwarden-crypto/src/keys/key_connector_key.rs
mzieniukbw: Fixed in 17d20d9b

quexten (dismissed):

crates/bitwarden-api-key-connector/Cargo.toml
```diff
@@ -0,0 +1,24 @@
...
+edition.workspace = true
+rust-version.workspace = true
+homepage.workspace = true
+repository.workspace = true
+license-file.workspace = true
+keywords.workspace = true
+
+[dependencies]
+async-trait = { workspace = true }
+mockall = { version = ">=0.13.1, <0.15" }
```
djsmith85: 🎨  mockall (0.14.0) is also present in the autotype-crate. Make sense to move this up to the root level and use the same version across both crates

crates/bitwarden-api-key-connector/Cargo.toml
```diff
@@ -0,0 +1,24 @@
+[package]
+name = "bitwarden-api-key-connector"
```
djsmith85, line 2: ❓ Which team is going to be owning the new crate? I assume KM. Please add this to the CODEOWNERS

djsmith85 (changes requested):

crates/bitwarden-api-key-connector/Cargo.toml
mzieniukbw: Fixed in 6f765f4a

crates/bitwarden-api-key-connector/Cargo.toml
mzieniukbw: KM team can own it. Fixed in 2a144123

djsmith85 (approved):

Patrick-Pimentel-Bitwarden (approved):
👍 Auth domain changes look appropriate with my limited knowledge

crates/bitwarden-api-key-connector/src/apis/user_keys_api.rs
```diff
@@ -0,0 +1,196 @@
...
+            request.body(serde_json::to_string(&body).expect("Serialize should be infallible"))
+    }
+
+    let response = request.send().await?;
+
+    Ok(response.error_for_status()?)
+}
+
+#[cfg(test)]
+mod tests {
```
Thomas-Avery, line 95: ⛏️  Not required, but it would be nice to have some test around the HTTP status codes that throw errors (400s & 500s).

Would help with intent and prevent regressing the behavior in the future.

crates/bitwarden-api-key-connector/src/apis/configuration.rs
```diff
@@ -0,0 +1,24 @@
...
+impl Default for Configuration {
+    fn default() -> Self {
+        Configuration {
+            base_path: "https://key-connector.bitwarden.com".to_owned(),
+            user_agent: Some("api/key-connector/rust".to_owned()),
+            client: reqwest::Client::new(),
+            oauth_access_token: None,
+        }
+    }
+}
```
Thomas-Avery, line 24: It doesn't look like this is used? Do we even need to implement `new()` and default for this?

If we need to keep it maybe requiring the `base_path` as a parameter for a `new()` configuration. We wouldn't ever want `"https://key-connector.bitwarden.com"` so it is weird to me to have it as a default.

Thomas-Avery (approved):
Looks good. A couple of non-blocking comments from me.

crates/bitwarden-core/src/key_management/crypto_client.rs
Thomas-Avery: > We should probably change all public APIs to remove the possibility of even passing in an rng that is bad. (@Thomas-Avery thoughts?)

@quexten 

It feels weird to have callers of our public APIs pass in an rng. It's a cryptography concern/complexity that should be handled for them.

crates/bitwarden-api-key-connector/src/apis/user_keys_api.rs
mzieniukbw: Once we have the autogenerated api models, we will then have zero integration style test coverage. Before we have the autogenerated models, we should think more about integration test coverage. I would leave that for later, it's big topic to consider and potentially a lot of work.

crates/bitwarden-api-key-connector/src/apis/configuration.rs
mzieniukbw: True. This will become irrelevant once we have the autogenerated api and models thought.

PR #615, opened 2025-12-16
fix(ci): skip verification during cargo-release publish

addisonbeck:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-28544

## 📔 Objective

We are trying to publish to crates.io for the first time in a year, but the current jobs are very out of date and the prexisting publishing workflow has suffered from many unexpected issues. This time we are getting errors about git state being dirty during the pre-validation build step of `cargo-release`, since this generates lock files for each lib.

Screenshot 2025-12-15 at 5 22 51 PM

This PR fixes this failure in the `publish-rust-crates.yml` workflow by adding the `--no-verify` flag to the cargo-release command. 

The trade-off is that publishing now skips the verification build that would normally ensure each crate compiles successfully before uploading to crates.io. However, this verification is largely redundant in our workflow since all code already passes build validation during CI before reaching this step. The git tags we publish from represent already-validated, immutable snapshots of code that successfully built in CI. The additional validation is good, but redundant. I think adding this flag is a bit of a smell, but I'm unsure how else to address this right now. It seems like we are following best practices, and this is just necassary for how our pipelines are abstracted. Open to suggestion though.

## ⏰ Reminders before review

- CI builds passed
  team

## 🦮 Reviewer guidelines

### DIFF
--- a/.github/workflows/publish-rust-crates.yml
+++ b/.github/workflows/publish-rust-crates.yml
@@ -131,7 +131,12 @@ jobs:
           # Convert space-separated string to array
           IFS=' ' read -ra PACKAGES <<< "$PACKAGES_INPUT"
           PACKAGE_FLAGS=$(printf -- '-p %s ' "${PACKAGES[@]}")
-          cargo-release release publish $PACKAGE_FLAGS --execute --no-confirm
+
+          # no-verify avoids cargo-release complaining about Cargo.lock
+          # modifications that naturally happen during the pre-publishing
+          # validation for many of our crates that don't have commit lock
+          # files
+          cargo-release release publish $PACKAGE_FLAGS --execute --no-confirm --no-verify
 
       - name: Update deployment status to Success
         if: ${{ inputs.release_type != 'Dry Run' && success() }}


addisonbeck:
The `build-rust-crates.yml` workflow already validates each package individually with `-p` flags. I really feel that adding this here too would be extra lines of code for little value, and that we can publish this way with confidence for now and reevaluate if there are issues.

addisonbeck:
@dani-garcia it would be nice if you could validate me here in addition to the required BRE review, or offer an alternative idea if you have one. Cargo release is new to me with this work.

dani-garcia:
@addisonbeck I think the problem you're seeing is not so much that the publishing step modifies the cargo.lock file specifically, but that the commited lockfile isn't up to date for some reason, and so any cargo command is going to modify it to make sure it's up to date. If you clone the latest `main` and run `cargo fetch` or `cargo build`, you'll see that it will update the lockfile even though it shouldn't do that during normal operations.

I think the best path forward currently is to update the lockfile to make sure it's in sync. We can add this escape hatch if we need it short term, but long term we should update our linting workflow to ensure that any lockfile changes are commited.

pixman20:
@addisonbeck I'm in favor of updating the lock files to follow best CI/CD practices and avoid having a follow up, but I realize this is needed for other downstream efforts and I'm also not sure of the amount of work to get those all updated (i.e. in case there's further investigation needed for some).

If we need to move forward as-is, I think there should just be a follow up task created to remove the workaround and update the lock files so it's not forgotten.

I wasn't sure if you were planning to do the lock files first or want to go ahead with this workaround for the short term.
If the short term, I can review/approve this PR, but wanted to check first.

addisonbeck:
@pixman20 you were right to hold off, I don't need this anymore. Dani updated Cargo.lock on main and I'm past this error now.

PR #628, opened 2025-12-21
[PM-29048] Validate pin protected user key envelope

mzieniukbw:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-29048

## 📔 Objective

Enables validation for a PIN for a pin protected key envelope, so that mobile can move off of `validatePin` which only works with pin key protected userkeys.

The original `validate_pin` implementation compared the decrypted user key from envelope with the client key stored user key. This implementation is much simpler, we simply validate if the envelope can be successfully unsealed with provided PIN.

Tested against android.

## 🚨 Breaking Changes

None

## ⏰ Reminders before review

- CI builds passed
  team

## 🦮 Reviewer guidelines

### DIFF
--- a/crates/bitwarden-core/src/auth/auth_client.rs
+++ b/crates/bitwarden-core/src/auth/auth_client.rs
@@ -1,6 +1,7 @@
 #[cfg(feature = "internal")]
 use bitwarden_crypto::{
     CryptoError, DeviceKey, EncString, Kdf, TrustDeviceResponse, UnsignedSharedKey,
+    safe::PasswordProtectedKeyEnvelope,
 };
 #[cfg(feature = "internal")]
 use bitwarden_encoding::B64;
@@ -26,7 +27,7 @@ use crate::{
             MasterPasswordPolicyOptions, password_strength, satisfies_policy, validate_password,
             validate_password_user_key,
         },
-        pin::validate_pin,
+        pin::{validate_pin, validate_pin_protected_user_key_envelope},
         register::make_register_keys,
         tde::{RegisterTdeKeyResponse, make_register_tde_keys},
     },
@@ -158,6 +159,18 @@ impl AuthClient {
         pin_protected_user_key: EncString,
     ) -> Result<bool, AuthValidateError> {
         validate_pin(&self.client, pin, pin_protected_user_key)
+    }
+
+    /// Validates a PIN against a PIN-protected user key envelope.
+    /// Returns `false` if the PIN fails to decrypt the envelope.
+    /// Requires the user key to be present in the client, otherwise returns
+    /// [`AuthValidateError::NotAuthenticated`].
+    pub fn validate_pin_protected_user_key_envelope(
+        &self,
+        pin: String,
+        pin_protected_user_key_envelope: PasswordProtectedKeyEnvelope,
+    ) -> Result<bool, AuthValidateError> {
+        validate_pin_protected_user_key_envelope(&self.client, pin, pin_protected_user_key_envelope)
     }
 
     #[allow(missing_docs)]
--- a/crates/bitwarden-core/src/auth/auth_client.rs
+++ b/crates/bitwarden-core/src/auth/auth_client.rs
@@ -162,14 +162,15 @@ impl AuthClient {
     }
 
     /// Validates a PIN against a PIN-protected user key envelope.
-    /// Returns `false` if the PIN fails to decrypt the envelope.
-    /// Requires the user key to be present in the client, otherwise returns
-    /// [`AuthValidateError::NotAuthenticated`].
+    ///
+    /// Returns `false` if validation fails for any reason:
+    /// - The PIN is incorrect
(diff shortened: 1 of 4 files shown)


crates/bitwarden-uniffi/src/auth/mod.rs
```diff
@@ -114,6 +117,24 @@ impl AuthClient {
         Ok(self.0.auth().validate_pin(pin, pin_protected_user_key)?)
     }
 
+    /// Validate the user PIN
+    ///
+    /// To validate the user PIN, you need to have the user's `pin_protected_user_key_envelope`.
+    /// This key is obtained when enabling PIN unlock on the account with the `enroll_pin` method.
+    ///
+    /// This works by comparing the decrypted user key with the current user key, so the client must
```
quexten: Nit (non-blocking), but the explanation seems slightly off here. The envelope will only unseal (decrypt) when the pin is correct. So the scenarios covered here are:
- Pin is wrong -> Fails to decrypt -> No key
- Pin is correct, but contained key is wrong -> Decrypts, but key equality check fails
- Pin is correct, and contained key matches -> Decrypts, and equality check succeeds

So, the primary method of checking pin validity here is actually the envelope unsealing correctly (validating correctness of the KDF-derived key with the authentication that's part of the AEAD cipher - xchacha20-poly1305, during unsealing). The key correctness check is only secondary here, and I'm not even sure if it's needed, since for that case the PIN is still correct?

crates/bitwarden-core/src/auth/auth_client.rs
```diff
@@ -160,6 +161,18 @@ impl AuthClient {
         validate_pin(&self.client, pin, pin_protected_user_key)
     }
 
+    /// Validates a PIN against a PIN-protected user key envelope.
+    /// Returns `false` if the PIN fails to decrypt the envelope.
```
quexten: We should probably comment on wha happens in the scenario that the key is different from the one in the client?

crates/bitwarden-core/src/auth/pin.rs
```diff
@@ -42,6 +42,31 @@ pub(crate) fn validate_pin(
...
+    let mut ctx = key_store.context();
+
+    if let Ok(decrypted_key_id) = pin_protected_user_key_envelope.unseal(pin.as_str(), &mut ctx) {
+        #[allow(deprecated)]
+        let Ok(decrypted_user_key) = ctx.dangerous_get_symmetric_key(decrypted_key_id) else {
+            return Ok(false);
+        };
+
+        #[allow(deprecated)]
+        let user_key = ctx
```
quexten: Thought: I wonder whether we should make a function to compare equality of keys on the key store without needing to extract them. Then again, checking key material equality isn't really a common use-case.

crates/bitwarden-core/src/auth/pin.rs
```diff
@@ -42,6 +42,31 @@ pub(crate) fn validate_pin(
     }
 }
 
+pub(crate) fn validate_pin_protected_user_key_envelope(
```
quexten, line 47: We should probably also add the docs here, considering it's pub.

I think we discussed this in a sync, but it is not entirely clear to me what the key equality check here actually achieves. (I know it was present in the old code too, so I'm OK with keeping it, if digging up the reason is out of scope, but if we do have the reason, we should document it).

(The pin envelope correctly decrypting should already mean the PIN is correct).

quexten (dismissed):
Some nits on the commenting, but nothing blocking.

crates/bitwarden-core/src/auth/pin.rs
mzieniukbw: I had the same thought and conclusion. I would leave it for now.

crates/bitwarden-core/src/auth/pin.rs
mzieniukbw: I would lean into removing that check. We are the the point, where crypto (especially for mobile) is fully managed by SDK, so the risk of user key not matching the stored user key, feels more like the client's bug, rather than something wrong in SDK.
I will remove it, document it and communicate that to mobile teams for transparency.

crates/bitwarden-uniffi/src/auth/mod.rs
mzieniukbw: I have updated the docs in 7a162ede

crates/bitwarden-core/src/auth/auth_client.rs
mzieniukbw: I have updated the docs in 7a162ede

crates/bitwarden-core/src/auth/pin.rs
mzieniukbw: Fixed in 7a162ede

quexten (approved):

addisonbeck (approved):

PR #643, opened 2026-01-02
[PM-19056]  SDK: Send email verification

harr1424:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-19056?atlOrigin=eyJpIjoiNjRiYzdjNzgwMGFkNGE1NWJlMjE1NWZlZDVmM2RjNjUiLCJwIjoiaiJ9

## 📔 Objective

Implement email verification in the SDK allowing it be shared by clients. The core verification logic has been implemented by Auth team and is performed server-side. 

This PR adds an `AuthType` enum and `emails` field to Send related structs so that they can be used by clients and the server. 

## ⏰ Reminders before review

- CI builds passed
  team

## 🦮 Reviewer guidelines


crates/bitwarden-send/src/send.rs
```diff
@@ -353,6 +379,18 @@ impl TryFrom<SendResponseModel> for Send {
...
+        let auth_type = match send.auth_type {
+            Some(t) => t.into(),
+            None => {
+                if send.password.is_some() {
+                    AuthType::Password
+                } else if send.emails.is_some() {
+                    AuthType::Email
+                } else {
+                    AuthType::None
+                }
```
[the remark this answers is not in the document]
harr1424: This logic matches existing logic defined for both server and clients

crates/bitwarden-send/src/send.rs
```diff
@@ -370,6 +408,17 @@ impl TryFrom<SendResponseModel> for Send {
...
+            emails: send
+                .emails
+                .map(|s| {
+                    s.split(',')
+                        .map(|e| e.trim())
+                        .filter(|e| !e.is_empty())
+                        .map(String::from)
+                        .collect()
+                })
+                .unwrap_or_default(),
```
Hinton: I wonder if we just want to update the server to expect a list rather than converting it to and from a list here.

harr1424: I'm not sure, the existing implementation in the TypeScript clients (already on `main`) and on the server feature branch related to this effort was to use a string. Do you think the SDK should also use a string, and leave it up to the clients to parse into a list? 

Edit for clarity: the `send` object defined in the TS clients uses a string, but the `send.view` object does use a list of strings. 

https://github.com/bitwarden/clients/blob/7d496febb77de4a314e1d6820757710c9033a9a2/libs/common/src/tools/send/models/domain/send.ts

https://github.com/bitwarden/clients/blob/7d496febb77de4a314e1d6820757710c9033a9a2/libs/common/src/tools/send/models/view/send.view.ts#L4

Hinton: Send in rust should match the TypeScript variant. If it's a string there then it should be a string here. As it will otherwise be more complex to convert.

But I would question if you want to always split and concat these strings or just consider them as lists. Does the server also split and validate data?

harr1424: On `server` the emails are [stored as a string](https://github.com/bitwarden/server/blob/bfd06c0c874ef96e13cbb30d6d24cea49a5c280e/src/Core/Tools/Entities/Send.cs#L82). 

This matches the `Send` domain model in the TypeScript clients. Also in the TypeScript clients the `SendView`instead stores emails as a list of strings.

crates/bitwarden-send/src/send.rs
```diff
@@ -123,6 +140,9 @@ pub struct SendView {
     pub revision_date: DateTime<Utc>,
     pub deletion_date: DateTime<Utc>,
     pub expiration_date: Option<DateTime<Utc>>,
+
+    pub emails: String,
```
Hinton: Emails in SendView is a vec<string> in clients right?

crates/bitwarden-send/src/send.rs
```diff
@@ -64,6 +64,20 @@ pub enum SendType {
     File = 1,
 }
 
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Serialize_repr, Deserialize_repr)]
+#[repr(u8)]
+#[cfg_attr(feature = "uniffi", derive(uniffi::Enum))]
+pub enum AuthType {
```
Hinton, line 71: This enum should be documented.

crates/bitwarden-send/src/send.rs
harr1424: That is correct 

https://github.com/bitwarden/clients/blob/07eb16d8ec150781cd5f86a6f4251bcdf77768b8/libs/common/src/tools/send/models/view/send.view.ts#L29

Hinton: So this should be changed to Vec?

harr1424: Sorry, I interpreted your [earlier comment](https://github.com/bitwarden/sdk-internal/pull/643#discussion_r2672371968)  to indicate that the `Send` model in the SDK should match the `Send` model in the TS clients. 

The `SendView` model in the TS clients seems like a reasonable place to split the string into an array of strings. 

The server PR originally introduced by Audrey 6 months ago stored emails as a single comma separated string, and the team decided we would make only minimal changes to that PR. We did not see a reason to change the datatype on the server to a `List<String>`. 

To summarize, there is currently parity among the `Send` definition across `server`, `sdk-internal`, and `clients`, with the `SendView` model being responsible for splitting a comma separated list of emails into an array of strings (individual emails).

Hinton: The `Send` model yes, but this is the `SendView` model in the SDK. Would it not make sense that both Send and SendView match the typescript code, since otherwise if you want to convert between them you would need to split/concat the strings?

harr1424: I see now, my apologies as I thought this comment was made in reference to the `emails` member of the `Send` struct, not the `SendView` struct. I agree completely. Before I continue, I wanted to ensure we are in agreement that `Send` and `SendView` should have parity such that in both the TS clients and the SDK: 
- `Send` stores emails as a comma separated string
- `SendView` performs work to separate this string and store emails in an array (TS) or Vec (Rust)

Hinton: Agreed

crates/bitwarden-send/src/send.rs
```diff
@@ -64,6 +64,21 @@ pub enum SendType {
...
+    /// Email-based OTP authentication
+    Email = 0,
+
+    /// Password-based authentication
+    Password = 1,
+
+    /// No authentication required
+    None = 2,
+}
```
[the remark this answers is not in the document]
harr1424: The API's `AuthType` in `bitwarden-api-api/src/models/auth_type.rs)` was auto-generated based upon an earlier implementation of `AuthType` on the `server`. Once the next API bindings workflow completes, this should no longer be a discrepancy.

crates/bitwarden-send/src/send.rs
```diff
@@ -265,6 +286,15 @@ impl Decryptable<KeyIds, SymmetricKeyId, SendView> for Send {
...
             deletion_date: self.deletion_date,
             expiration_date: self.expiration_date,
+
+            emails: self
+                .emails
+                .split(',')
+                .map(|e| e.trim())
+                .filter(|e| !e.is_empty())
+                .map(String::from)
+                .collect(),
```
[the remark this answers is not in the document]
harr1424: Validation should occur on the clients in order to simplify error handling and `i18n`

crates/bitwarden-send/src/send.rs
```diff
@@ -123,6 +141,9 @@ pub struct SendView {
     pub revision_date: DateTime<Utc>,
     pub deletion_date: DateTime<Utc>,
     pub expiration_date: Option<DateTime<Utc>>,
+
+    pub emails: Vec<String>,
```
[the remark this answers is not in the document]
harr1424: This comment is a bit misleading or at least missing some relevant context: the  relevant code block is constructing a `Send` from a `SendResponseModel` meaning the `Send` already exists on the server. It has been left to the clients to validate send creation before sent to the server and stored in the database. 

The suggested documentation has been added for clarity.

coroiu (dismissed):

crates/bitwarden-send/src/send.rs
```diff
@@ -89,6 +104,12 @@ pub struct Send {
     pub revision_date: DateTime<Utc>,
     pub deletion_date: DateTime<Utc>,
     pub expiration_date: Option<DateTime<Utc>>,
+
+    /// Email addresses for OTP authentication.
+    /// **Note**: Mutually exclusive with `new_password`. If both are set,
+    /// only password authentication will be used.
+    pub emails: String,
```
Hinton: Shouldn't this be optional if it's mutually exclusive?

crates/bitwarden-send/src/send.rs
```diff
@@ -265,6 +292,19 @@ impl Decryptable<KeyIds, SymmetricKeyId, SendView> for Send {
...
+                .emails
+                .as_deref()
+                .map(|s| {
+                    s.split(',')
+                        .map(|e| e.trim())
+                        .filter(|e| !e.is_empty())
+                        .map(String::from)
+                        .collect()
+                })
+                .unwrap_or_default(),
```
Hinton: We can simplify this slightly.
```suggestion
            emails: self
                .emails
                .as_deref()
                .unwrap_or_default()
                .split(',')
                .map(|e| e.trim())
                .filter(|e| !e.is_empty())
                .map(String::from)
                .collect(),
```

Hinton (approved):

itsadrago (approved):

coroiu (approved):

PR #644, opened 2026-01-02
[PM-24978] Pass along Attachment decryption errors

nick-livefront:
## 🎟️ Tracking

[PM-24978](https://bitwarden.atlassian.net/browse/PM-24978)
Clients PR: https://github.com/bitwarden/clients/pull/17790

## 📔 Objective

Currently the SDK swallows any decryption errors for Attachments and thus hides all attachments from the user. Even when the user has multiple attachments on a cipher. 

This adds a new `attachment_decryption_failures` attribute to `CipherView` that is populated with all attachments that fail to decrypt. This can then be consumed by clients to allow the user to delete any corrupted ciphers.

## 🚨 Breaking Changes

~~❓ This could be a breaking change for the mobile clients if they expect attachment decryptions to be ignored. I do not know if that is the case yet. I'll reach out.~~

No additional changes only.

## Client UI

Screenshot 2025-12-02 at 2 36 42 PM

## 🦮 Reviewer guidelines

[PM-24978]: https://bitwarden.atlassian.net/browse/PM-24978?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


crates/bitwarden-vault/src/cipher/attachment.rs
```diff
@@ -211,6 +214,12 @@ impl Decryptable<KeyIds, SymmetricKeyId, AttachmentView> for Attachment {
         ctx: &mut KeyStoreContext<KeyIds>,
         key: SymmetricKeyId,
     ) -> Result<AttachmentView, CryptoError> {
+        // Try to decrypt the file name
+        let (file_name, decryption_failure) = match self.file_name.decrypt(ctx, key) {
+            Ok(name) => (name, false),
+            Err(..) => (None, true),
+        };
```
nikwithak: > Currently the SDK swallows any decryption errors for Attachments and thus hides all attachments from the user. 

Where is the actual error being swallowed? I would suggest adding the handling logic there, maybe following the pattern of [decrypt_list_with_failures](https://github.com/bitwarden/sdk-internal/blob/c5a3b83324b43548d1497d3c631a7d7a23b421de/crates/bitwarden-vault/src/cipher/cipher_client/mod.rs#L131). with the change here, you end up throwing away the `CryptoError` and the function returns as a "successful" decryption, breaking the contract of "This function returns a decrypted object, or an `Err(CryptoError)`"

nikwithak (changes requested):

crates/bitwarden-vault/src/cipher/attachment.rs
nick-livefront: The error is being swallowed when the cipher list is being decrypted [here](https://github.com/bitwarden/sdk-internal/blob/edcd0d3044cdd8e630b803797c1fe14cb120a45a/crates/bitwarden-vault/src/cipher/cipher.rs#L546). 

> maybe following the pattern of [decrypt_list_with_failures](https://github.com/bitwarden/sdk-internal/blob/c5a3b83324b43548d1497d3c631a7d7a23b421de/crates/bitwarden-vault/src/cipher/cipher_client/mod.rs#L131)

This could work but it would expand the API of the `Cipher`. I could leave the existing `attachments` alone and only return the decrypted attachments and then add a `attachment_failure` property. What do you think about that?

nick-livefront: Added my approach here: [3d5dfac](https://github.com/bitwarden/sdk-internal/pull/644/commits/3d5dfac7a649b266a736b079e02088f80b3903bf)

I appreciate the feedback, let me know what you think!

crates/bitwarden-vault/src/cipher/cipher.rs
```diff
@@ -347,6 +347,9 @@ pub struct CipherView {
     pub attachments: Option<Vec<attachment::AttachmentView>>,
     pub fields: Option<Vec<field::FieldView>>,
     pub password_history: Option<Vec<password_history::PasswordHistoryView>>,
+    /// Attachments that failed to decrypt. Only present when there are decryption failures.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub attachment_decryption_failures: Option<Vec<attachment::AttachmentView>>,
```
nikwithak, line 350: nit: Maybe move this to just below line 347?

crates/bitwarden-vault/src/cipher/attachment.rs
```diff
@@ -238,14 +232,46 @@ impl Decryptable<KeyIds, SymmetricKeyId, AttachmentView> for Attachment {
...
 }
 
+/// Decrypts a list of attachments, separating successful decryptions from failures.
+///
+/// Returns a tuple of (successful_attachments, failed_attachments).
+pub(crate) fn decrypt_attachments_with_failures(
+    attachments: &[Attachment],
+    ctx: &mut KeyStoreContext<KeyIds>,
+    key: SymmetricKeyId,
+) -> (Option<Vec<AttachmentView>>, Option<Vec<AttachmentView>>) {
```
nikwithak: `(Vec<AttachmentView>, Vec<AttachmentView>)` might be a better choice here (foregoing wrapping them in `Option`) - an empty Vec conveys the same information and simplifies a lot of the downstream logic.

crates/bitwarden-vault/src/cipher/cipher.rs
```diff
@@ -522,6 +525,19 @@ impl Decryptable<KeyIds, SymmetricKeyId, CipherView> for Cipher {
...
+            .attachments
+            .as_deref()
+            .map(|encrypted_attachments| {
+                attachment::decrypt_attachments_with_failures(
+                    encrypted_attachments,
+                    ctx,
+                    ciphers_key,
+                )
+            })
+            .unwrap_or((None, None));
```
nikwithak: optional nit: `.unwrap_or_default()` has the same effect here.

nick-livefront: 👍 [ce23472](https://github.com/bitwarden/sdk-internal/pull/644/commits/ce234720e2a9909c24ef0d01a56f6bdf8f7525f6)

crates/bitwarden-vault/src/cipher/attachment.rs
nick-livefront: I do like removing the `Option` as it cuts down on the needed code and mental thought. It doesn't match the definitions on the `CipherView` so I had to rewrap in `Some` but that feels like an okay trade off

[ce23472](https://github.com/bitwarden/sdk-internal/pull/644/commits/ce234720e2a9909c24ef0d01a56f6bdf8f7525f6)

nikwithak (dismissed):

crates/bitwarden-vault/src/cipher/attachment.rs
```diff
@@ -228,14 +231,44 @@ impl Decryptable<KeyIds, SymmetricKeyId, AttachmentView> for Attachment {
...
+    ctx: &mut KeyStoreContext<KeyIds>,
+    key: SymmetricKeyId,
+) -> (Vec<AttachmentView>, Vec<AttachmentView>) {
+    let mut successes = Vec::new();
+    let mut failures = Vec::new();
+
+    for attachment in attachments {
+        match attachment.decrypt(ctx, key) {
+            Ok(decrypted) => successes.push(decrypted),
+            Err(_) => failures.push(AttachmentView {
```
dani-garcia: It could be nice to add a log here, just so we don't swallow the error:
```
    tracing::warn!(attachment_id = ?attachment.id, error = %e, "Failed to decrypt attachment");
```

crates/bitwarden-vault/src/cipher/cipher.rs
```diff
@@ -522,6 +524,19 @@ impl Decryptable<KeyIds, SymmetricKeyId, CipherView> for Cipher {
     ) -> Result<CipherView, CryptoError> {
         let ciphers_key = Cipher::decrypt_cipher_key(ctx, key, &self.key)?;
 
+        // Separate successful and failed attachment decryptions
+        let (attachments, attachment_decryption_failures) = self
```
dani-garcia: This might look a bit simple if you unwrap the attachments first, rather than mapping:
```rust
        let (attachments, attachment_decryption_failures) =
            attachment::decrypt_attachments_with_failures(
                self.attachments.as_deref().unwrap_or_default(),
                ctx,
                ciphers_key,
            );
```

dani-garcia (dismissed):
LGTM, just had two small non-blocking improvements

crates/bitwarden-vault/src/cipher/cipher.rs
nick-livefront: I like it! [3acc313](https://github.com/bitwarden/sdk-internal/pull/644/commits/3acc31380e0f732eda458a32e9d0e334a2a63b8b)

crates/bitwarden-vault/src/cipher/attachment.rs
nick-livefront: 👍 Good call: [732705a](https://github.com/bitwarden/sdk-internal/pull/644/commits/732705a46b08001110fa6ab4c1d52661cbc29956)

nikwithak (approved):

itsadrago (approved):

dani-garcia (approved):

PR #649, opened 2026-01-05
[PM-30630] [Proposal] Addkey/value storage to state

Hinton:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-30630
## 📔 Objective

Introduce a new key/value setting system to the `bitwarden-state` crate that provides type-safe settings repository functionality for storing application configuration and state.

**Key features:**
- Compile-time type-safe keys prevent type mismatches
- Automatic JSON serialization/deserialization
- Built on bitwarden-state repository pattern
- Graceful error handling (logged, don't propagate)

**Architecture:**
- Uses SDK's Repository<T> pattern with SQLite backend
- Settings stored as JSON values indexed by string keys
- Key<T> type provides type safety at compile time

## ⏰ Reminders before review

- CI builds passed
  team

## 🦮 Reviewer guidelines


crates/bitwarden-settings/src/setting.rs
```diff
@@ -0,0 +1,14 @@
+//! Internal setting storage type.
+
+use serde::{Deserialize, Serialize};
+
+/// Internal setting value stored in the settings repository.
+///
+/// This type wraps a JSON value for flexible storage. Users should not work with
+/// this type directly - use [`SettingsRepository`](crate::SettingsRepository) instead,
+/// which provides type-safe access.
+#[derive(Debug, Clone, Serialize, Deserialize)]
```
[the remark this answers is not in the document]
Hinton: It's fine.

crates/bitwarden-settings/src/lib.rs
```diff
@@ -0,0 +1,9 @@
+#![doc = include_str!("../README.md")]
```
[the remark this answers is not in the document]
Hinton: Limitation with how clients and repositories are implemented.

crates/bitwarden-settings/README.md
```diff
@@ -0,0 +1,37 @@
...
+use serde::{Deserialize, Serialize};
+
+#[derive(Serialize, Deserialize)]
+struct AppConfig {
+    theme: String,
+    auto_save: bool,
+}
+
+// Define a type-safe key
+const CONFIG: Key<AppConfig> = Key::new("app_config");
```
dani-garcia: We should probably document somewhere that this should be a globally unique key. 

Of note, part of the reason that we're using a `register_repository` macro for repositories is that I've wanted to validate that uniqueness automatically by collecting all the names and checking in a test with a crate like https://crates.io/crates/inventory. I've never had time to see if that would work okay though but the fact we're using a macro should allow us to transparently do it in the future.

crates/bitwarden-settings/README.md
```diff
@@ -0,0 +1,37 @@
...
+
+async fn example(client: &bitwarden_core::Client) -> Result<(), Box<dyn std::error::Error>> {
+    // Get setting
+    let config: Option<AppConfig> = client.settings().get(CONFIG).await?;
```
dani-garcia: To somewhat link this with the proposal for better testing, I wonder if it would be good to create a `Setting` type that handles the operations for a specific `Key`:


```rust
async fn example(client: &bitwarden_core::Client) -> Result<(), Box<dyn std::error::Error>> {
    let setting: Setting<AppConfig> = client.settings().get(CONFIG);

   let value = setting.get().await?;
   setting.update(new_value).await?;
   setting.delete().await?;
}
```

This will allow us to declare these as part of the sub client structs:

```rust
struct CipherClient {
    api_client: ApiClient,
    repository: Repository<Cipher>,
    cipher_setting: Setting<CipherSetting>,
```

crates/bitwarden-settings/src/setting.rs
```diff
@@ -0,0 +1,14 @@
...
+
+use serde::{Deserialize, Serialize};
+
+/// Internal setting value stored in the settings repository.
+///
+/// This type wraps a JSON value for flexible storage. Users should not work with
+/// this type directly - use [`SettingsRepository`](crate::SettingsRepository) instead,
+/// which provides type-safe access.
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
+pub struct Setting(pub(crate) serde_json::Value);
```
dani-garcia: We could mark this type as `#[doc(hidden)]` to make it more obvious that it shouldn't be used anywhere

dani-garcia (commented):
Nice, this is fairly similar to what I've had in mind for https://bitwarden.atlassian.net/browse/PM-23511

I've left a couple of comments on the code, but another question I have is why a separate crate? This is a pretty thin convenience layer over bitwarden-state, and it being a separate crate means users are going to need to use both if they want to deal with both single settings and repositories. I don't know if that overhead is worth it.

Hinton:
> I've left a couple of comments on the code, but another question I have is why a separate crate? 

I'm happy to merge it into state, it seemed easier to merge it in than uncouple it, so I eered on that side.

crates/bitwarden-settings/README.md
Hinton: We would need to clone the DB when getting right? But maybe that's safe.

dani-garcia: We already do that for Repository: https://github.com/bitwarden/sdk-internal/blob/0e7c6b2d366e746e721b820f758972eacc241844/crates/bitwarden-state/src/sdk_managed/mod.rs#L85-L92

The database is inside an `Arc<RwLock<>>`

crates/bitwarden-state/README.md
```diff
@@ -183,3 +183,49 @@ pub fn get_sdk_managed_migrations() -> RepositoryMigrations {
...
+
+## Settings
+
+This crate also provides a type-safe settings API for storing application configuration and state.
+The settings system uses type-safe keys to ensure compile-time type checking when getting and
+setting values.
+
+### Usage
+
+```rust,ignore
```
dani-garcia: It's a bit unfortunate that these comments have to be ignored because we can't use bitwarden-core here, I wonder if it would be better to change the example to be something like: 
```rust
use bitwarden_state::{register_setting_key, Setting};
use serde::{Deserialize, Serialize};

#[derive(Serialize, Deserialize)]
struct AppConfig {
    theme: String,
    auto_save: bool,
}

// Register a type-safe key
register_setting_key!(const CONFIG: AppConfig = "app_config");

struct TestClient {
    // Access settings via bitwarden_core::Client.platform().state().setting()
    app_config: Setting<AppConfig>,
}

async fn example(client: &TestClient) -> Result<(), Box<dyn std::error::Error>> {
    // Get value
    let config: Option<AppConfig> = client.app_config.get().await?;

    // Update value
    let new_config = AppConfig {
        theme: "dark".to_string(),
        auto_save: true,
    };
    client.app_config.update(new_config).await?;

    // Delete setting
    client.app_config.delete().await?;

    Ok(())
}
```

dani-garcia (dismissed):

dani-garcia (approved):

PR #654, opened 2026-01-06
[PM-27277] Support v2 encryption on JIT Password signups

mzieniukbw:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-27277

## 📔 Objective

Encryption V2 for SSO JIT master password account registration.

Changed the default Kdf for Encryption V2 accounts, which uses Argon2Id with more iOS mobile friendly parameters (less memory, but more iterations). Refactored the Kdf default functions, so we only expose `Kdf::default_pbkdf2` or `Kdf::default_argon2id`.

Requires https://github.com/bitwarden/server/pull/6777, for the autogenerated models, which is manually committed in `bitwarden-api-api` crate temporarily, until the server PR is done.

Updated WASM build script only to format autogenerated TS files, since the JS file got too big for `npm prettier` to handle and is not necessary for non-TS files.

## 🚨 Breaking Changes

None, new functionality.

## ⏰ Reminders before review

- CI builds passed
  team

## 🦮 Reviewer guidelines


mzieniukbw:
The SDK Breaking Change Detection is failing, because client's build is broken, nothing to do with this PR

crates/bitwarden-core/src/key_management/crypto.rs
```diff
@@ -950,6 +970,47 @@ pub(crate) fn make_user_key_connector_registration(
...
+    client: &Client,
+    user_id: UserId,
+    master_password: String,
+    salt: String,
+) -> Result<MakeJitMasterPasswordRegistrationResponse, MakeKeysError> {
+    let mut ctx = client.internal.get_key_store().context_mut();
+    let (user_key_id, wrapped_state) = WrappedAccountCryptographicState::make(&mut ctx, user_id)
+        .map_err(MakeKeysError::AccountCryptographyInitialization)?;
+
+    let kdf = Kdf::default();
```
quexten: In the tech breakdown, we noted that new V2 users should just use Argon2, with the settings: Iterations: 6, Memory: 32 MiB, Parallelism: 4. So, this would either need to be changed to reference these settings, or we should change KDF::default.

crates/bitwarden-core/src/key_management/crypto.rs
```diff
@@ -950,6 +970,47 @@ pub(crate) fn make_user_key_connector_registration(
...
+
+    let master_password_unlock_data =
+        MasterPasswordUnlockData::derive(&master_password, &kdf, &salt, &user_key)
+            .map_err(MakeKeysError::MasterPasswordDerivation)?;
+
+    let master_password_authentication_data =
+        MasterPasswordAuthenticationData::derive(&master_password, &kdf, &salt)
+            .map_err(MakeKeysError::MasterPasswordDerivation)?;
+
+    let master_key =
```
quexten, line 997: Note: I'm wondering if we should note that this is deprecated and only temporarily used until we remove the master-key state in clients?

quexten (changes requested):
Looks good for the most part, just two small things

crates/bitwarden-core/src/key_management/crypto.rs
mzieniukbw: Fixed in 454b06e2

quexten (approved):
Approving. Please ping if you need a re-approval after syncing server.

nick-livefront (approved):

addisonbeck (approved):

crates/bitwarden-core/src/key_management/crypto.rs
mzieniukbw: Let's remove that, with the removal of the master-key state in clients.

PR #661, opened 2026-01-08
[PM-30635] Add sync infrastructure with notifications

Hinton:
🎟️ Tracking
------------

<https://bitwarden.atlassian.net/browse/PM-30635>

📔 Objective
------------

Add sync infrastructure with event-driven architecture.

-   Add `bitwarden-sync` crate with `SyncClient` and `SyncHandler` trait
-   Implement event registry for hooking into sync completion
-   Add `FolderSyncHandler` in `bitwarden-vault` for folder persistence

This provides the foundation for implementing sync functionality across different clients, with an extensible event system that allows crates to respond to sync completion events.

This differs from the existing client implementation which has a main handler that delegates. In this every sync handler is fully isolated but may depend on previous state which they will have to fetch themselves. We will need to register sync and error handlers, most likely in the `pm` crate.


crates/bitwarden-vault/src/folder/folder_sync_handler.rs
```diff
@@ -0,0 +1,75 @@
...
+        // This ensures atomicity - either all conversions succeed or none are persisted
+        let mut folders_to_update = Vec::new();
+        for folder_response in api_folders {
+            let folder = Folder::try_from(folder_response.clone())?;
+            let folder_id = require!(folder.id);
+
+            // Check if folder needs to be updated
+            let needs_update = existing
+                .get(&folder_id)
+                .is_none_or(|existing_folder| folder.revision_date > existing_folder.revision_date);
```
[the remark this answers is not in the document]
Hinton: Thanks ... 91 > 82 in my world but maybe yours is different.

crates/bitwarden-vault/src/folder/folder_sync_handler.rs
```diff
@@ -0,0 +1,75 @@
...
+        }
+
+        // TODO: Replace with bulk operations when supported
+        for (id, _) in existing {
+            repo.remove(id.to_string()).await?;
+        }
+
+        Ok(())
+    }
+}
```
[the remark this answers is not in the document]
Hinton: Not easy to test right now.

coroiu (commented):
Reviewing to stop notifications since we're discussing this in other places

crates/bitwarden-sync/src/sync_client.rs
```diff
@@ -0,0 +1,323 @@
...
+                api_client,
+                identity_client: bitwarden_api_identity::apis::ApiClient::new_mocked(|_| {}),
+                api_config: dummy_config.clone(),
+                identity_config: dummy_config,
+                device_type: bitwarden_core::client::DeviceType::SDK,
+            }),
+            sync_handlers: HandlerRegistry::new(),
+            error_handlers: HandlerRegistry::new(),
+            sync_lock: tokio::sync::Mutex::new(()),
+        }
```
Hinton, line 221: @dani-garcia this is a bit cumbersome to declare everywhere. Do you think we should add some helpers to `bitwarden-core`?

dani-garcia: Definitely, my hope is that with the FromClient trait we can do something simple like:
```rust
pub fn test_client<T: FromClient>() -> T {
    let client = Client::new(None);
    T::from_client(&client).expect("Failed to create test client")
}

fn test() {
    // And then initializing any client that is using the macro is very straight-forward
    let sync_client: SyncClient  = test_client();

    // Or potentially a builder pattern, so you can set mocks etc?
    let sync_client: SyncClient  = test_client_builder()
        .with_mock(mock)
        .authenticated()
        .build();
}
```

dani-garcia: Probably better as part of `bitwarden-test` though

Hinton: Depends, do we want -core to be able to use bitwarden-test? Currently that's possible but not if we introduce this since test would depend on things in core.

dani-garcia: Hmm fair point, my thinking here is that having it all test utilities on `bitwarden-test` would make it easier to reason about, as it's probably easier to enforce that `bitwarden-test` is not used outside of `cfg(test)`, than trying to see which parts of core can and can't be used in which contexts.

Plus, as we remove logic from core, the only things to test are initialization, where we may want to manually build the clients to ensure we test the correct things. If we feel like it's useful in core, we can always define it in core on a hidden module with a scary name and reexport it from -test maybe?

crates/bitwarden-sync/src/sync_client.rs
```diff
@@ -0,0 +1,361 @@
...
+    }
+
+    /// Trigger sync handlers for a completed sync operation
+    ///
+    /// Executes two phases sequentially:
+    /// 1. Calls [`SyncHandler::on_sync`] on all handlers with the response
+    /// 2. Calls [`SyncHandler::on_sync_complete`] on all handlers
+    ///
+    /// Stops on first error and returns it immediately.
+    async fn trigger_sync(&self, response: &SyncResponseModel) -> Result<(), SyncError> {
```
coroiu: suggestion: rename to `run_handlers` to avoid ambiguity between this and `sync`

```suggestion
    async fn run_handlers(&self, response: &SyncResponseModel) -> Result<(), SyncError> {
```

crates/bitwarden-sync/src/sync_client.rs
```diff
@@ -0,0 +1,361 @@
...
+            handler.on_sync_complete().await;
+        }
+
+        Ok(())
+    }
+
+    /// Notify all error handlers about a sync error
+    ///
+    /// All error handlers are called sequentially in registration order.
+    async fn notify_error(&self, error: &SyncError) {
```
coroiu: suggestion(non-blocking): if we rename the above function we could also rename this for consistency


```suggestion
    async fn run_error_handlers(&self, error: &SyncError) {
```

crates/bitwarden-sync/src/sync_client.rs
```diff
@@ -0,0 +1,361 @@
...
+) -> Result<SyncResponseModel, SyncError> {
+    let sync = api_configurations
+        .api_client
+        .sync_api()
+        .get(input.exclude_subdomains)
+        .await
+        .map_err(|e| SyncError::Api(e.into()))?;
+
+    Ok(sync)
+}
```
coroiu: question: why not put this in the client function?

Hinton: An oversight.

coroiu (dismissed):

crates/bitwarden-vault/src/folder/folder_sync_handler.rs
```diff
@@ -0,0 +1,104 @@
...
+    async fn on_sync(
+        &self,
+        response: &bitwarden_api_api::models::SyncResponseModel,
+    ) -> Result<(), SyncHandlerError> {
+        let api_folders = require!(response.folders.as_ref());
+
+        let folders: Vec<(FolderId, Folder)> = api_folders
+            .iter()
+            .filter_map(|f| {
+                let folder = Folder::try_from(f.clone()).ok()?;
```
dani-garcia: Should we log something if the folder fails to convert, rather than fail silently?

dani-garcia (commented):
LGTM, just a small suggestion to ensure we don't swallow errors

crates/bitwarden-vault/src/folder/folder_sync_handler.rs
Hinton: It can also produce an error and fail the sync. Not sure what is most appropriate?

dani-garcia: I think I'd probably leave that choice for vault to decide, folders might not be considered essential so it might be preferrable to ignore the invalid sync rather than fail early.

Probably log-but-skip if invalid is the safest choice for folders?

dani-garcia (approved):

crates/bitwarden-vault/src/folder/folder_sync_handler.rs
gbubemismith: I think `log-but-skip if invalid` is the way to go here

gbubemismith (approved):

PR #664, opened 2026-01-09
[BEEEP|PM-30620] Add crypto debug and tracing

quexten:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-30620

## 📔 Objective

Some crypto errors are very hard to pin-point down with exiting tooling. There are three reasons for this: Errors are being mapped to very generic errors (especially InvalidKey). The solution here eventually is splitting up `CryptoError` into small, per-function or per-module errors instead of having this mega error. The second part to this is our prevention of logging of key material, which could be used to confirm exact key values (to find out if a wrong key was used). The third is being able to pin-point down error location due to missing stack traces.

We fix all three of these in this PR. For the second issue, we add tracing statements in a few locations. For the first and last issue, we add a `dangerous-crypto-debug` crate feature that is compile-time toggled. This disables the prevention of key-material logging and gives us extra data.

Considerations for security: We never want devs to accidentally enable this. So it both has a name, indicating that it is dangerous, but it also logs a line into the console on init, warning anyone who looks at the console logs that this must not be enabled in production. This should reasonably prevent this feature from making it to prod.

## ⏰ Reminders before review

- CI builds passed
  team

## 🦮 Reviewer guidelines

### DIFF
--- a/crates/bitwarden-crypto/src/keys/key_connector_key.rs
+++ b/crates/bitwarden-crypto/src/keys/key_connector_key.rs
@@ -26,11 +26,8 @@ impl KeyConnectorKey {
     }
 
     /// Wraps the user key with this key connector key.
-    #[cfg_attr(feature = "dangerous-crypto-debug", instrument(skip(self), err))]
-    #[cfg_attr(
-        not(feature = "dangerous-crypto-debug"),
-        instrument(skip(self, user_key), err)
-    )]
+    #[cfg_attr(feature = "dangerous-crypto-debug", instrument(err))]
+    #[cfg_attr(not(feature = "dangerous-crypto-debug"), instrument(skip_all, err))]
     pub fn encrypt_user_key(
         &self,
         user_key: &SymmetricCryptoKey,
@@ -41,11 +38,8 @@ impl KeyConnectorKey {
     }
 
     /// Unwraps the user key with this key connector key.
-    #[cfg_attr(feature = "dangerous-crypto-debug", instrument(skip(self), err))]
-    #[cfg_attr(
-        not(feature = "dangerous-crypto-debug"),
-        instrument(skip(self, user_key), err)
-    )]
+    #[cfg_attr(feature = "dangerous-crypto-debug", instrument(err))]
+    #[cfg_attr(not(feature = "dangerous-crypto-debug"), instrument(skip_all, err))]
     pub fn decrypt_user_key(
         &self,
         user_key: EncString,
--- a/crates/bitwarden-crypto/src/keys/key_connector_key.rs
+++ b/crates/bitwarden-crypto/src/keys/key_connector_key.rs
@@ -3,6 +3,7 @@ use std::pin::Pin;
 use bitwarden_encoding::B64;
 use generic_array::GenericArray;
 use rand::Rng;
+use tracing::instrument;
 use typenum::U32;
 
 use crate::{
@@ -25,6 +26,11 @@ impl KeyConnectorKey {
     }
 
     /// Wraps the user key with this key connector key.
+    #[cfg_attr(feature = "dangerous-crypto-debug", instrument(skip(self), err))]
+    #[cfg_attr(
+        not(feature = "dangerous-crypto-debug"),
+        instrument(skip(self, user_key), err)
+    )]
     pub fn encrypt_user_key(
         &self,
         user_key: &SymmetricCryptoKey,
@@ -35,6 +41,11 @@ impl KeyConnectorKey {
     }
 
     /// Unwraps the user key with this key connector key.
+    #[cfg_attr(feature = "dangerous-crypto-debug", instrument(skip(self), err))]
+    #[cfg_attr(
+        not(feature = "dangerous-crypto-debug"),
+        instrument(skip(self, user_key), err)
+    )]
     pub fn decrypt_user_key(
         &self,
         user_key: EncString,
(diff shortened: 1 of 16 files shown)


crates/bitwarden-crypto/src/keys/master_key.rs
```diff
@@ -57,11 +60,13 @@ impl MasterKey {
     /// Derives a users master key from their password, email and KDF.
     ///
     /// Note: the email is trimmed and converted to lowercase before being used.
+    #[cfg_attr(feature = "dangerous-crypto-debug", instrument)]
```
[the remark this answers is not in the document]
quexten: No, this is intended. The flag is named dangerous and is allowed to log.

crates/bitwarden-wasm-internal/src/pure_crypto.rs
```diff
@@ -111,6 +124,12 @@ impl PureCrypto {
         email: String,
         kdf: Kdf,
     ) -> Result<String, CryptoError> {
+        let _span = tracing::info_span!(
+            "PureCrypto::encrypt_user_key_with_master_password",
+            email = %email,
```
mzieniukbw, line 129: ❓ Since this is salt in practise, isn't this sensitive information ?

crates/bitwarden-wasm-internal/src/pure_crypto.rs
```diff
@@ -97,6 +103,13 @@ impl PureCrypto {
         email: String,
         kdf: Kdf,
     ) -> Result<Vec<u8>, CryptoError> {
+        let _span = tracing::info_span!(
+            "PureCrypto::decrypt_user_key_with_master_password",
+            email = %email,
```
mzieniukbw, line 108: ❓ Since this is salt in practise, isn't this sensitive information ?

crates/bitwarden-crypto/src/keys/master_key.rs
```diff
@@ -57,11 +60,13 @@ impl MasterKey {
...
     pub fn derive(password: &str, email: &str, kdf: &Kdf) -> Result<Self, CryptoError> {
         Ok(KdfDerivedKeyMaterial::derive(password, email, kdf)?.into())
     }
 
     /// Derive the master key hash, used for local and remote password validation.
+    #[cfg_attr(feature = "dangerous-crypto-debug", instrument)]
```
mzieniukbw, line 69: ❓ With feature enabled, why aren't we skipping `self` from being instrumented here, but do skip it in other functions ?

crates/bitwarden-crypto/src/keys/master_key.rs
```diff
@@ -74,11 +79,21 @@ impl MasterKey {
     }
 
     /// Encrypt the users user key
+    #[cfg_attr(feature = "dangerous-crypto-debug", instrument(skip(self), err))]
+    #[cfg_attr(
+        not(feature = "dangerous-crypto-debug"),
+        instrument(skip(self, user_key), err)
```
mzieniukbw: ❓ Genuine question, when feature is not enabled, then `Self` will not implement `Debug`. Is that why we are skipping it, will compiler complain about instrumentation that does not implement Debug ?

crates/bitwarden-crypto/src/keys/key_connector_key.rs
```diff
@@ -25,6 +26,11 @@ impl KeyConnectorKey {
     }
 
     /// Wraps the user key with this key connector key.
+    #[cfg_attr(feature = "dangerous-crypto-debug", instrument(skip(self), err))]
+    #[cfg_attr(
+        not(feature = "dangerous-crypto-debug"),
+        instrument(skip(self, user_key), err)
```
mzieniukbw: If the skip includes all fields, it might be better to use `skip_all`

crates/bitwarden-wasm-internal/src/pure_crypto.rs
quexten: The salt is not sensitive or secret. The point of the salt is not to be extra secret material, but to add a cryptographic namespace, so that when brute-forcing, your brute-forcing space is partitioned (by the salt). It just means that you have to brute-force every password for every salt (which in practice means you have to brute-force every password for every user), instead of being able to run through all passwords once and/or built a rainbow table for future use.

crates/bitwarden-wasm-internal/src/pure_crypto.rs
quexten: See above, salt is not secret / sensitive.

crates/bitwarden-crypto/src/keys/master_key.rs
quexten: No need to skip it, skipping it was accidental here..
The two paths here are:
- Feature enabled: MasterKey implements Debug, instrument logs it
- Feature disabled: MasterKey does not implement Debug, instrument must skip it

crates/bitwarden-crypto/src/keys/key_connector_key.rs
quexten: Applied

crates/bitwarden-crypto/src/keys/master_key.rs
quexten: Removed the skip in the other functions for this feature

mzieniukbw (approved):

djsmith85 (approved):
Platform-changes are looking good.

tangowithfoxtrot (approved):

PR #667, opened 2026-01-09
[BEEEP] Clean up AES error handling

quexten:
## 📔 Objective

Cleans up the lowest layer of encryption for AES and adds a custom error struct. This pushes up the error one layer. This does expand the crypto error temporarily until the upper layers are reworked too.

## ⏰ Reminders before review

- CI builds passed
  team

## 🦮 Reviewer guidelines

### DIFF
--- a/crates/bitwarden-crypto/src/aes.rs
+++ b/crates/bitwarden-crypto/src/aes.rs
@@ -12,9 +12,13 @@ use subtle::ConstantTimeEq;
 use typenum::U32;
 
 use crate::{
-    error::{CryptoError, Result},
+    error::Result,
     util::{PBKDF_SHA256_HMAC_OUT_SIZE, PbkdfSha256Hmac},
 };
+
+/// An aes operation failed either due to invalid padding or due to an invalid MAC.
+#[derive(Debug)]
+pub(crate) struct DecryptError {}
 
 /// Decrypt using AES-256 in CBC mode.
 ///
@@ -23,13 +27,13 @@ pub(crate) fn decrypt_aes256(
     iv: &[u8; 16],
     data: Vec<u8>,
     key: &GenericArray<u8, U32>,
-) -> Result<Vec<u8>> {
+) -> Result<Vec<u8>, DecryptError> {
     // Decrypt data
     let iv = GenericArray::from_slice(iv);
     let mut data = data;
     let decrypted_key_slice = cbc::Decryptor::<aes::Aes256>::new(key, iv)
         .decrypt_padded_mut::<Pkcs7>(&mut data)
-        .map_err(|_| CryptoError::KeyDecrypt)?;
+        .map_err(|_| DecryptError {})?;
 
     // Data is decrypted in place and returns a subslice of the original Vec, to avoid cloning it,
     // we truncate to the subslice length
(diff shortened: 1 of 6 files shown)


crates/bitwarden-crypto/src/store/context.rs
```diff
@@ -213,14 +213,16 @@ impl<Ids: KeyIds> KeyStoreContext<'_, Ids> {
...
+                        .map_err(|_| CryptoError::Decrypt)?,
                 ))?
             }
             (
                 EncString::Aes256Cbc_HmacSha256_B64 { iv, mac, data },
                 SymmetricCryptoKey::Aes256CbcHmacKey(key),
             ) => SymmetricCryptoKey::try_from(&BitwardenLegacyKeyBytes::from(
-                crate::aes::decrypt_aes256_hmac(iv, mac, data.clone(), &key.mac_key, &key.enc_key)?,
+                crate::aes::decrypt_aes256_hmac(iv, mac, data.clone(), &key.mac_key, &key.enc_key)
+                    .map_err(|_| CryptoError::Decrypt)?,
```
nikwithak, line 225: Rather than `map_err` everywhere this function is called, you might consider `impl From<DecryptError> for CryptoError` - which should give you a lot of these conversions for free and simplify future uses

nikwithak (dismissed):
Vault changes look good - one non-blocking suggestion.

crates/bitwarden-crypto/src/error.rs
```diff
@@ -12,10 +12,10 @@ use crate::fingerprint::FingerprintError;
 #[bitwarden_error(flat)]
 #[derive(Debug, Error)]
 pub enum CryptoError {
+    #[error("The decryption operation failed")]
+    Decrypt,
```
mzieniukbw, line 16: ❓ This right now is AES decryption errors only, do you plan to move more code into `CryptoError::Decrypt` or maybe this name should be more AES decrypt explicit ?

crates/bitwarden-crypto/src/aes.rs
```diff
@@ -48,10 +52,10 @@ pub(crate) fn decrypt_aes256_hmac(
     data: Vec<u8>,
     mac_key: &GenericArray<u8, U32>,
     key: &GenericArray<u8, U32>,
-) -> Result<Vec<u8>> {
-    let res = generate_mac(mac_key, iv, &data)?;
+) -> Result<Vec<u8>, DecryptError> {
+    let res = generate_mac(mac_key, iv, &data);
     if res.ct_ne(mac).into() {
-        return Err(CryptoError::InvalidMac);
+        return Err(DecryptError {});
```
mzieniukbw, line 58: ❓ This hides the information, that the MAC is invalid. Is it because we likely can't action on that information, hence it's better to shove everything as "decrypt error" ?

mzieniukbw (dismissed):

crates/bitwarden-crypto/src/store/context.rs
quexten: This is intentional. The intention here is to get rid of CryptoError in the medium term, so making it easier / less explicit to convert to it would just further the uses for it.

Still thank you for pointing this out!

crates/bitwarden-crypto/src/error.rs
quexten: In the medium term the plan is to get rid of CryptoError; however, yeah in the meantime I think it makes sense to map all decryption errors to Decrypt.

crates/bitwarden-crypto/src/aes.rs
quexten: Yeah, the consumer will not care what about the AES encryption is invalid, and cannot do anything about it either way / it is non-actionable, so it doesn't make much sense to expose it beyond debug logging.

mzieniukbw (approved):

nikwithak (approved):

PR #682, opened 2026-01-16
[PM-30144] Add key rotation impl for account cryptographic state

quexten:
## 🎟️ Tracking

## 📔 Objective

Adds a key-rotation impl for the account cryptography state. Deprecates the old functions, which will be removed after the SDK key rotation has been rolled out and TS key rotation can be removed.

## ⏰ Reminders before review

- CI builds passed
  team

## 🦮 Reviewer guidelines

### DIFF
--- a/crates/bitwarden-core/src/key_management/account_cryptographic_state.rs
+++ b/crates/bitwarden-core/src/key_management/account_cryptographic_state.rs
@@ -278,7 +278,8 @@ impl WrappedAccountCryptographicState {
                 signing_key,
                 security_state,
             } => {
-                // To rotate a V2 state, the private and signing keys are re-encrypted with the new user key.
+                // To rotate a V2 state, the private and signing keys are re-encrypted with the new
+                // user key.
                 // 1. Re-encrypt private key
                 let private_key_id = ctx
                     .unwrap_private_key(*current_user_key, private_key)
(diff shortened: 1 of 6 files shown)


crates/bitwarden-core/src/key_management/account_cryptographic_state.rs
```diff
@@ -194,6 +219,105 @@ impl WrappedAccountCryptographicState {
         ))
     }
 
+    #[cfg(test)]
+    fn make_v1(
```
mzieniukbw, line 223: ❓  This can be inside the `mod tests` as standalone function, isn't it better ?

crates/bitwarden-core/src/key_management/account_cryptographic_state.rs
```diff
@@ -509,4 +633,54 @@ mod tests {
...
+
+        // Rotate the state
+        let user_id = UserId::new_v4();
+        let rotated = wrapped_state
+            .rotate(&user_key, &new_user_key, user_id, &mut ctx)
+            .unwrap();
+
+        // Should now be V2
+        match rotated {
+            WrappedAccountCryptographicState::V2 { .. } => {}
```
mzieniukbw, line 657: 🤔   I always wonder what is the value of such a test. At minimum i would expect to check that the V2 private key encstring is different than V1, maybe additionally checking for EncString type (from 2 to 7). Similarly for other fields, at least check presence or/and enc string type.

mzieniukbw (approved):

crates/bitwarden-core/src/key_management/account_cryptographic_state.rs
quexten: I wanted to expose this here so that we can re-use it in other tests in the same crate. I think that will make sense? Though, if that turns out not to be true, then I agree.

crates/bitwarden-core/src/key_management/account_cryptographic_state.rs
quexten: Hmm, that's fair. It's hard to validate crypto without making the RNG deterministic, but we could check the key types at least. Since you approved, I'll make a follow-up PR, cleaning up this test.

quexten: https://github.com/bitwarden/sdk-internal/pull/703

PR #685, opened 2026-01-16
[PM-30144] Implement unlock-data re-encryption for key-rotation

quexten:
## 🎟️ Tracking

## 📔 Objective

Implements unlock-method rotation for user-keys in the SDK. Note: This adds the key-connector enum variant and the none enum variant, but both are unsupported as of now and have comments indicating this.

## ⏰ Reminders before review

- CI builds passed
  team

## 🦮 Reviewer guidelines


crates/bitwarden-user-crypto-management/src/key_rotation/unlock.rs
```diff
@@ -0,0 +1,590 @@
...
+        let _span = debug_span!("reencrypt_emergency_access_keys").entered();
+        input
+            .trusted_emergency_access_keys
+            .iter()
+            .map(|ea| {
+                let _span =
+                    debug_span!("reencrypt_emergency_access_key", grantee_id = ?ea.id).entered();
+                match UnsignedSharedKey::encapsulate(new_user_key_id, &ea.public_key, ctx) {
+                    Ok(reencrypted_key) => Ok(EmergencyAccessWithIdRequestModel {
+                        r#type: models::EmergencyAccessType::Takeover,
```
quexten: Please note: We should update the request model here. The data here is fully unused on the server-side except for the key.

crates/bitwarden-user-crypto-management/src/key_rotation/unlock.rs
```diff
@@ -0,0 +1,532 @@
...
+pub enum MasterkeyUnlockMethod {
+    /// The master password based unlock method.
+    Password {
+        password: String,
+        hint: Option<String>,
+        kdf: Kdf,
+        salt: String,
+    },
+    /// The key-connector based unlock method.
+    KeyConnector,
```
quexten, line 37: For the follow-up ticket, this should probably contain a key-connector key. The implementation is not done / tested here, but it is left as an example of how this is intended to be used.

crates/bitwarden-user-crypto-management/src/key_rotation/unlock.rs
```diff
@@ -0,0 +1,666 @@
...
+
+    fn create_test_kdf_argon2id() -> Kdf {
+        Kdf::Argon2id {
+            iterations: NonZeroU32::new(3).expect("valid iterations"),
+            memory: NonZeroU32::new(64).expect("valid memory"),
+            parallelism: NonZeroU32::new(4).expect("valid parallelism"),
+        }
+    }
+
+    fn assert_symmetric_keys_equal(
```
quexten, line 337: May want to move this into keystore at some point, behind a compile-time debug flag.

crates/bitwarden-user-crypto-management/src/key_rotation/unlock.rs
```diff
@@ -0,0 +1,638 @@
...
+
+#[allow(unused)]
+#[derive(Debug)]
+pub(super) enum ReencryptError {
+    /// Failed to update the unlock data for the master password
+    MasterPasswordDerivation,
+    /// Failed to update the unlock data for TDE/PRF-Passkey
+    KeysetUnlockDataReencryption,
+    /// Failed to update the unlock data for emergency access or organization membership
+    KeysharingError,
```
Thomas-Avery:
```suggestion
    KeySharingError,
```


optional ⛏️

crates/bitwarden-user-crypto-management/src/key_rotation/unlock.rs
```diff
@@ -0,0 +1,638 @@
...
+    new_user_key_id: SymmetricKeyId,
+    ctx: &mut KeyStoreContext<KeyIds>,
+) -> Result<UnlockDataRequestModel, ReencryptError> {
+    let master_password_unlock_data = reencrypt_userkey_for_masterpassword_unlock(
+        input.master_key_unlock_method,
+        new_user_key_id,
+        ctx,
+    )?
+    // This is safe for now until we support key-connector or no-master-key based rotation.
+    .expect("master-password based unlock data must be present");
```
Thomas-Avery: 💭 
The naming here confused me. It seems like we're going to process all `master_key_unlock` options within `reencrypt_userkey_for_masterpassword_unlock`? But everything is named `master_password`.

Is the intent to reuse `MasterPasswordUnlockAndAuthenticationDataModel` for key connector key rotations in the future? If so maybe just renaming `reencrypt_userkey_for_masterkey_unlock` would be fine. For `MasterkeyUnlockMethod.None` I think I understand the option allowing us to just return None.

If we are going to make new models for the key connector payload then I think the switch should be in `reencrypt_unlock`.

Thomas-Avery (commented):
Looking good, one question from my side.

crates/bitwarden-user-crypto-management/src/key_rotation/unlock.rs
quexten: Ah, no you are right. `reencrypt_userkey_for_masterpassword_unlock` is meant only for master password, so it doesn't really make sense to pass in the entire enum, but the broken down components should be passed in. The model should not be re-used for key-connector, but key-connector should just be a separate entry / unlock method on the model.

Thomas-Avery (approved):

PR #697, opened 2026-01-19
[PM-30144] Add key rotation for account cryptographic state

quexten:
## 🎟️ Tracking

## 📔 Objective

Adds key-rotation for account cryptographic state.

## ⏰ Reminders before review

- CI builds passed
  team

## 🦮 Reviewer guidelines


quexten:
@claude plz review

crates/bitwarden-user-crypto-management/src/crypto.rs
```diff
@@ -0,0 +1,197 @@
...
+    current_user_key_id: &SymmetricKeyId,
+    new_user_key_id: &SymmetricKeyId,
+    user_id: UserId,
+    ctx: &mut bitwarden_crypto::KeyStoreContext<KeyIds>,
+) -> Result<bitwarden_api_api::models::AccountKeysRequestModel, ()> {
+    debug!(
+        "Rotating account cryptographic state for user_id={} from key_id={:?} to key_id={:?}",
+        user_id, current_user_key_id, new_user_key_id
+    );
+
```
dani-garcia, line 25: Nit: Rather than using string formatting, it's better to use the structured capabilities of `tracing`:

```rust
    debug!(
        %user_id, ?current_user_key_id, ?new_user_key_id,
        "Rotating account cryptographic state",
    );

    debug!(
        %user_id,
        "Converting rotated account cryptographic state to request model",
    );
```

dani-garcia (dismissed):

crates/bitwarden-user-crypto-management/src/crypto.rs
quexten: Nice! I've replaced the two instances.

dani-garcia (dismissed):

crates/bitwarden-user-crypto-management/src/crypto.rs
```diff
@@ -0,0 +1,198 @@
...
+    )
+    .map_err(|_| ())?;
+
+    debug!(
+        %user_id,
+        "Converting rotated account cryptographic state to request model",
+    );
+
+    // Rotate the account keys for the user
+    let account_keys_model = rotated_account_cryptographic_state
```
Thomas-Avery, line 43: Up above from my understanding `WrappedAccountCryptographicState::rotate(` will log the error context internally via `#[instrument(skip(self, ctx), err)]`. So I think it is fine that the error context is discarded. 

For this method `rotated_account_cryptographic_state
        .to_request_model(` I'm not seeing any logging of the error internally and we discard the error context. I would expect having a log in the `map_error` using something like `tracing::error!(`.

crates/bitwarden-user-crypto-management/src/crypto.rs
```diff
@@ -0,0 +1,198 @@
...
+
+        // Assert security_state is present
+        let security_state = model
+            .security_state
+            .as_ref()
+            .expect("security_state should be present for V2 state");
+        assert!(
+            security_state.security_state.is_some(),
+            "security_state content should be present"
+        );
```
Thomas-Avery, line 254: Shouldn't `test_rotate_v1_to_v2_returns_account_keys_model` also have these asserts for `signed_public_key`, `signature_key_pair` and `security_state`? Might be an opportunity to DRY some of this testing code up.

Thomas-Avery (changes requested):
Looks good couple of minor things from my side.

crates/bitwarden-user-crypto-management/src/crypto.rs
quexten: I added the instrument macro to the `to_request_model` call too now, instead of adding a manual error trace statement.

crates/bitwarden-user-crypto-management/src/crypto.rs
quexten: Ah, yeah, you are right. V2 to V2 rotation should actually have slightly different rotations, in that the veryfing key is expected to stay the same for now.

quexten: Adjusted

crates/bitwarden-user-crypto-management/src/crypto.rs
Thomas-Avery: That sounds reasonable to me. If you're open to it I would find it valuable if you could do a knowledge share in one the dev syncs on the instrument macro.

Things that come to mind is should we always just prefer to have that on the downstream? Is there a reason when we shouldn't use it? That way we can all stay consistent with the usage.

crates/bitwarden-user-crypto-management/src/crypto.rs
```diff
@@ -0,0 +1,247 @@
...
+        let model_public_key = model
+            .public_key_encryption_key_pair
+            .expect("public_key_encryption_key_pair should be present")
+            .public_key
+            .expect("public_key should be present");
+        assert_eq!(
+            actual_public_key.to_string(),
+            model_public_key,
+            "Public key should be correctly included in the model"
+        );
```
Thomas-Avery, line 112:
```suggestion
        let actual_public_key: B64 = public_key.to_der().unwrap().into();
        let public_key_encryption_key_pair = model
            .public_key_encryption_key_pair
            .as_ref()
            .expect("public_key_encryption_key_pair should be present");
        let model_public_key = public_key_encryption_key_pair
            .public_key
            .as_ref()
            .expect("public_key should be present");
        assert_eq!(
            actual_public_key.to_string(),
            *model_public_key,
            "Public key should be correctly included in the model"
        );

        // Assert signed_public_key is present
        assert!(
            public_key_encryption_key_pair.signed_public_key.is_some(),
            "signed_public_key should be present for V2 state"
        );
```
⛏️ 
Don't we need to check the `signed_public_key` too?

Thomas-Avery (changes requested):

crates/bitwarden-user-crypto-management/src/crypto.rs
quexten: Sounds good. I have not thought sufficiently about this myself yet, to be honest / have a concrete ruleset other than "it feels useful to capture error context here".

crates/bitwarden-user-crypto-management/src/crypto.rs
quexten: Applied

Thomas-Avery (dismissed):

Thomas-Avery (approved):

PR #699, opened 2026-01-19
[PM-29492] API middleware support

dani-garcia:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-29492

## 📔 Objective

Add support for API middleware for the autogenerated API crates. This would allow us to add middlewares that inspect and modify the request and response, or that retry the requests. It seems very flexible and I think it should cover all our use cases. Note that the version of reqwest-middleware is not the latest, as it needs to match with a specific reqwest version, and the next major reqwest version requires some rework around rustls-platform-verifier.

Note that this is doesn't add any middleware yet, but here's an example of how one would look:

```rust
#[cfg_attr(not(target_arch = "wasm32"), async_trait::async_trait)]
#[cfg_attr(target_arch = "wasm32", async_trait::async_trait(?Send))]
impl reqwest_middleware::Middleware for MiddlewareTest {
    async fn handle(&self, mut req: Request, _: &mut Extensions, next: Next<'_>) -> Result<Response> {
        // We can dynamically insert headers to all the requests here
        // For example, the generated code could notify us that some 
        // requests need auth so we can inject the correct token.
        if req.extensions().get::<RequiresAuth>().is_some() {
            req.headers_mut().insert(header::AUTHORIZATION, self.get_token().await);
        }

        // We can implement retry logic here as well
        let mut response = loop {
            // This example retry loop will retry any requests that have an expired token from the 
            // load balancer in front of the server. Note that not all requests are cloneable, like
            //  when they have a streamed body. Those should be a very small minority of requests, 
            // so in that case we just don't retry those.
            if self.should_send_load_balancer_token().await && let Some(req) = req.try_clone() {
                let response = next.clone().run(req, extensions).await?;

                if response.status().is_redirection() {
                    self.refresh_load_balancer_token().await?;
                    continue;
                }
                break response;
            } else {
                break next.run(req, extensions).await?
            }
        };

        // We can also modify the response before returning it
        response.headers_mut().insert("X-Custom-Response-Header", HeaderValue::from_static("Value"));

        Ok(response)
    }
}
```

## ⏰ Reminders before review

- CI builds passed
  team

## 🦮 Reviewer guidelines


quexten (dismissed):
Approving for KM-owned code.

coroiu (dismissed):
Nice to see that we were able to use an existing solution!

dani-garcia:
Had to fix conflicts on the scripts due to #660

coroiu (approved):

quexten (approved):

PR #701, opened 2026-01-19
BW login and unlock persistence

Hinton:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-30659
https://bitwarden.atlassian.net/browse/PM-30660

## 📔 Objective

Explores adding auth and crypto persistence to the CLI.

## ⏰ Reminders before review

- CI builds passed
  team

## 🦮 Reviewer guidelines


trmartin4:
Is this PR going to need significant rework when https://github.com/bitwarden/sdk-internal/pull/549 lands?  If so, could we potentially hold off merging this, or base it on that branch, so that we can make sure we don't have to completely rework this when that PR merges?

Upon first glance, it looks like the changes in #549 are going to impact more of the code in `login_api_key()`, where we'll need to change the call to use the new `loginClient()`, get the authentication state, and initialize a new `PasswordManagerClient` with it, which you can then use to store the state in the persistent store.  Is that what you were thinking?

crates/bitwarden-core/src/client/internal.rs
```diff
@@ -33,6 +33,39 @@ use crate::{
...
+/// Persisted cryptographic state for CLI session restoration.
+///
+/// Contains encrypted key material necessary to restore vault crypto across CLI restarts.
+/// Both fields are optional to handle various login methods and authentication scenarios.
+#[cfg(feature = "cli")]
+#[derive(Debug, Clone, serde::Serialize, serde::Deserialize)]
+pub struct PersistedCryptoState {
+    /// Master key wrapped user key - needed for master password unlock.
+    /// Optional because some login methods (e.g., device-only auth) may not have this.
+    pub master_key_encrypted_user_key: Option<bitwarden_crypto::EncString>,
```
quexten: This should use the master-password-unlock-data from the UserDecryptionOptions instead  which contains the actually used KDF settings, and also the salt (which is currently being decoupled from email). 
https://github.com/bitwarden/server/blob/ad19efcff7a4dacb3d538689479867dd780e14eb/src/Identity/IdentityServer/RequestValidators/BaseRequestValidator.cs#L681
https://github.com/bitwarden/server/blob/ad19efcff7a4dacb3d538689479867dd780e14eb/src/Identity/IdentityServer/UserDecryptionOptionsBuilder.cs#L183

crates/bw/src/auth/unlock.rs
```diff
@@ -0,0 +1,92 @@
...
+        .prompt()?;
+
+    // Get KDF and email from login method
+    let (email, kdf) = match &auth_state.login_method {
+        UserLoginMethod::Username { email, kdf, .. } => (email.clone(), kdf.clone()),
+        UserLoginMethod::ApiKey { email, kdf, .. } => (email.clone(), kdf.clone()),
+    };
+
+    // Derive master key and initialize crypto
+    let master_key = bitwarden_crypto::MasterKey::derive(&password, &email, &kdf)?;
```
quexten: If we store the master-password-unlock-data we should use it here also. KM is currently migrating the master-key out of other teams ownership and they should generally not use it.

crates/bw/src/key_management/crypto.rs
```diff
@@ -0,0 +1,137 @@
...
+}
+
+/// Extract crypto state from sync response for CLI persistence.
+///
+/// Returns the encrypted key material needed to restore vault crypto across CLI restarts.
+pub fn extract_from_sync(sync: &SyncResponseModel) -> Option<PersistedCryptoState> {
+    let profile = sync.profile.as_ref()?;
+
+    // Extract master key encrypted user key from user decryption options
+    let master_key_encrypted_user_key = sync
```
quexten: Please just save the entire unlock data and don't extract from it.

crates/bw/src/key_management/crypto.rs
```diff
@@ -0,0 +1,137 @@
...
+        .user_decryption
+        .as_ref()
+        .and_then(|ud| ud.master_password_unlock.as_ref())
+        .and_then(|mpu| mpu.master_key_encrypted_user_key.as_ref())
+        .and_then(|key_str| key_str.parse().ok());
+
+    // Extract wrapped account cryptographic state from profile
+    let wrapped_account_cryptographic_state = profile
```
quexten: It feels like we should implement a TryFrom here. The fallback logic is also not correct here. We should parse either fully, or throw out the entire sync, and a partial V2 state should not result in a V1 state being returned.

crates/bw/src/key_management/session.rs
```diff
@@ -0,0 +1,220 @@
...
+
+use crate::platform::StateError;
+
+register_setting_key!(const ENCRYPTED_USER_KEY: EncryptedUserKeyData = "encrypted_user_key");
+
+/// Encrypted user key storage
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct EncryptedUserKeyData {
+    /// DataEnvelope containing encrypted user key (base64-encoded)
+    pub envelope: String,
```
quexten: Question: Can we just use the DataEnvelope / EncString types in here? They support serialize / deserialize

crates/bw/src/key_management/session.rs
```diff
@@ -0,0 +1,220 @@
...
+    pub wrapped_cek: String,
+    /// User ID this key belongs to
+    pub user_id: UserId,
+}
+
+/// User key payload for encryption
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct UserKeyPayload {
+    /// The user's encryption key in base64
+    pub user_key: B64,
```
quexten, line 34: If the key is the only thing contained here, then I feel that KM should spend a minimal amount of time here to add the "KeyProtectedKeyEnvelope". The idea there is that the metadata that's unenecrypted but authenticated will tell which key is contained, which can be useful, for things like knowing whether to update an encrypted copy of a key after key rotation.

crates/bw/src/platform/state.rs
```diff
@@ -0,0 +1,174 @@
...
+    #[error("Auth state not available in client")]
+    NoAuthState,
+
+    /// Required state not found.
+    #[error("{0}")]
+    NotFound(&'static str),
+
+    /// Cryptographic operation failed.
+    #[error("Crypto error: {0}")]
+    Crypto(#[from] bitwarden_crypto::CryptoError),
```
quexten, line 66: Thought / question: Can we simplify the public error interface? I'm not sure why "Crypto", "DataEnvelope", "CryptoClient", and "EncryptionSettings" would be different from the caller's perspective / whether they actually have different response behavior to these.

quexten (commented):
Nice, looking forward to having some state / the CI working. I have some comments from the KM side. Please note I recognize this is draft, but I felt like adding these early on.

crates/bitwarden-core/src/client/internal.rs
```diff
@@ -33,6 +33,39 @@ use crate::{
...
+
+    /// Wrapped account cryptographic state (V1 or V2).
+    /// Contains encrypted private keys and signing keys.
+    /// Optional to handle cases where it's not available.
+    #[cfg(feature = "internal")]
+    pub wrapped_account_cryptographic_state: Option<WrappedAccountCryptographicState>,
+}
```
coroiu: question: why store crypto state, could we not recover the state from `UserLoginMethod`?

quexten: Two notes:

All three of unlock methods (including KDF settings), anything downstream from the user, and the userkey itself may change throughout a logged in session and should be updated on sync.

As a second note, it seems like UserLoginMethod does not contain enough data.

(From the KM perspective, I agree with the approach done as currently implemented in this PR).

coroiu: @quexten what about things like unlock methods? Can't they also change? Do they count as being part of `CryptoState` ?

quexten: Yeah, they can change (mutate, add, remove). I can see the argument for them being part of the PersistedCryptoState. On clients, we have them in separate state currently.

Currently only the master-password-unlock-method is used in
```
pub(crate) async fn unlock(client: &Client, args: UnlockArgs) -> CommandResult {
```
for which we should store the master-password-unlock-data which gets synced down from state. Currently this data changes for instance when changing KDF. If I understand the code right, this would currently break, since only the master-key-encrypted-user-key but not KDF settings are updated. (Hence, KM made the master-password-unlock-data struct so you don't have to be aware of these effects and can just treat it as one blob).

If we want to allow other unlock methods on CLI too like passkey-prf, TDE, then those would need to be stored somewhere as-well. Again, I don't really mind whether it's one blob of crypto related things or separate state entries in this case, I just want the correct things to be saved here, and for them to be updated correctly.

crates/bw/src/key_management/session.rs
Hinton: Open to it, do you have any ideas how it would look? And is it something that we want to improve now or later.

quexten: I guess we can do it later. I'm not exactly sure what encrypted migrations look like in the SDK. I had a BEEEP draft for this envelope, but never got around to finishing it so I also don't want to block work by it.

quexten: For how it would look: It's basically just a cose encrypt 0 struct, wrapped by another key with the headers indicating which key id is contained and which it is encrypted by. Thus, you can always from the encrypted object see what's inside / what you need to unlock it.

The Rust side then had some nice interfaces around it.

PR #702, opened 2026-01-20
[PM-31763] Set user-key to wasm state during unlock

quexten:
## 🎟️ Tracking

- https://bitwarden.atlassian.net/browse/PM-31763
- https://github.com/bitwarden/clients/pull/18870

## 📔 Objective

This PR creates a client-managed repository for the user-key as an ephemeral key-value store. No migrations are required, since the user-key is ephemeral. This only applies to WASM since uniffi does not re-init the SDK.


crates/bitwarden-vault/src/folder/edit.rs
```diff
@@ -3,8 +3,6 @@ use bitwarden_crypto::{CryptoError, KeyStore};
 use bitwarden_error::bitwarden_error;
 use bitwarden_state::repository::{Repository, RepositoryError};
 use thiserror::Error;
-#[cfg(feature = "wasm")]
-use wasm_bindgen::prelude::*;
```
quexten, line 7: Not sure why this was required to be honest, but it seems this broke the build when not removed.

coroiu (dismissed):

crates/bitwarden-core/src/key_management/crypto.rs
eligrubb, line 201: ❓should the user key be copied into state for the `DecryptedKey` request method as well? In order to catch the biometrics flow?

crates/bitwarden-core/src/key_management/wasm_unlock_state.rs
```diff
@@ -0,0 +1,53 @@
...
+        .platform()
+        .state()
+        .get::<key_management::UserKeyState>()
+        .map_err(|_| UnableToSetError)?
+        .set(
+            USER_KEY_REPOSITORY_KEY.to_string(),
+            key_management::UserKeyState {
+                decrypted_user_key: user_key.to_base64(),
+            },
+        )
```
eligrubb, line 77: This is more for my own knowledge: how does storing the decrypted user key in wasm state fit into our memory safety story? From our memory hardening docs, it seems like for web clients we rely on process reloading to wipe memory, including secrets. Is that the case for this key value store too?

eligrubb (commented):
Looks good, I have one small nit.

crates/bitwarden-core/src/key_management/wasm_unlock_state.rs
quexten: So there is two sides to this, the enhanced memory protections / "anti dumping", and the memory zeroization after not being used (after being locked). Enhanced memory protections are only available on desktop, since browsers don't have access here. Zeroization is only available in rust. Eventually, the goal is to move the user-key being held in memory entirely to live within rust only. At that point it will properly get zeroized, and process reload isn't as needed.

However, currently it's not yet possible to hold the user-key within wasm state, specifically because the SDK gets re-created every time any part of the initialization parameters (mostly crypto keys) changes: https://github.com/bitwarden/clients/blob/e66a1f37b5a31513bef5d26b7c145b01eaa40a07/libs/common/src/platform/services/sdk/default-sdk.service.ts#L199

Once we moved this state to be sdk available, we can stop re-creating the wasm module and hold it in rust only.

crates/bitwarden-core/src/key_management/crypto.rs
quexten: Good point. My concern was that since the SDK gets re-created and re-initialized based on changes to the userkey-state inclients, theoretically, this could cause a loop of SDK Init -> SDK sets user-key -> Userkey updates -> SDK init -> ...

I've changed this now to make the SDK only update the state when the key in state is different or not set.

eligrubb (approved):
🥳

gbubemismith (approved):

coroiu (approved):

PR #708, opened 2026-01-23
Update pull request template

Hinton:
## 📔 Objective

Update PR template to match https://github.com/bitwarden/template/pull/181


djsmith85 (approved):
I actually liked the reminders and the guidelines. I'd regularly look up the emojis when writing reviews. I also think it helps new devs and contributors to see what we are looking out for in a PR.

Hinton:
@djsmith85 I found them a bit verbose, but not disagreeing. This was updated in the other repos though so for consistency we should use the same template. Maybe there's a middle ground that won't confuse AI reviewers.

PR #710, opened 2026-01-26
[PM-31241] Play framework & E2E test for cli api key login

Hinton:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-31241
## 📔 Objective

Adds support for the play framework (SeederApi) which lets you create scenes which seed data. This framework automatically handles cleanup by calling the destroy endpoint on drop.

The framework is used to build a simple E2E test for api key login using the bw CLI.

### DIFF
--- a/crates/bitwarden-test/README.md
+++ b/crates/bitwarden-test/README.md
@@ -17,18 +17,17 @@ the server side `SeederApi` to generate data.
 2. All HTTP requests include an `x-play-id` header for server-side test isolation
 3. Any db entry created through a request associated with an `x-play-id` (users, etc.) are saved as
    associated with that `play_id`
-4. When the `Play` instance is dropped, associated data is automatically cleaned up
+4. When the test closure completes, associated data is automatically cleaned up
 
 ### Usage
 
+Use the `#[play_test]` macro for the most ergonomic experience:
+
 ```rust
-use bitwarden_test::play::{Play, SingleUserArgs, SingleUserScene};
+use bitwarden_test::play::{play_test, Play, SingleUserArgs, SingleUserScene};
 
-#[tokio::test(flavor = "multi_thread", worker_threads = 1)]
-async fn test_example() {
-    // Create a Play instance - generates unique play_id
-    let play = Play::new();
-
+#[play_test]
+async fn test_example(play: Play) {
     // Create a test user via the seeder API
     let args = SingleUserArgs {
         email: "test@example.com".to_string(),
@@ -41,7 +40,30 @@ async fn test_example() {
 
     // Use credentials for testing...
 
-    // Cleanup happens automatically when `play` is dropped
+    // Cleanup happens automatically when the test completes
+}
+```
+
+Or use the builder pattern directly for more control:
+
+```rust
+use bitwarden_test::play::{Play, PlayConfig, SingleUserArgs, SingleUserScene};
+
+#[tokio::test]
+async fn test_example() {
+    Play::builder()
+        .config(PlayConfig::new("https://api", "https://identity", "http://seeder"))
+        .run(|play| async move {
+            let args = SingleUserArgs {
+                email: "test@example.com".to_string(),
+                verified: true,
+                ..Default::default()
+            };
+            let scene = play.scene::<SingleUserScene>(&args).await.unwrap();
+
+            // Use credentials for testing...
+        })
+        .await;
 }
 ```
 
--- a/crates/bitwarden-test/README.md
+++ b/crates/bitwarden-test/README.md
@@ -15,8 +15,9 @@ the server side `SeederApi` to generate data.
 
 1. Each `Play` instance generates a unique `play_id` (UUID)
 2. All HTTP requests include an `x-play-id` header for server-side test isolation
-3. The seeder API creates test data (users, etc.) associated with that `play_id`
-4. When the `Play` instance is dropped, all test data is automatically cleaned up
+3. Any db entry created through a request associated with an `x-play-id` (users, etc.) are saved as
+   associated with that `play_id`
+4. When the `Play` instance is dropped, associated data is automatically cleaned up
 
 ### Usage
 
@@ -37,7 +38,7 @@ async fn test_example() {
     let scene = play.scene::<SingleUserScene>(&args).await.unwrap();
 
     // Access the created user data (email is mangled server-side)
-    let client_id = scene.get_mangled("client_id").unwrap();
+    let client_id = scene.get_mangled("client_id");
 
     // Use credentials for testing...
 
--- a/crates/bitwarden-test/README.md
+++ b/crates/bitwarden-test/README.md
@@ -37,8 +37,7 @@ async fn test_example() {
     };
     let scene = play.scene::<SingleUserScene>(&args).await.unwrap();
 
-    // Access the created user data (email is mangled server-side)
-    let client_id = scene.get_mangled("client_id");
+    let email = scene.get_mangled("test@example.com");
 
     // Use credentials for testing...
 
--- a/crates/bitwarden-test/README.md
+++ b/crates/bitwarden-test/README.md
@@ -24,7 +24,7 @@ the server side `SeederApi` to generate data.
 ```rust
 use bitwarden_test::play::{Play, SingleUserArgs, SingleUserScene};
 
-#[tokio::test]
+#[tokio::test(flavor = "multi_thread", worker_threads = 1)]
 async fn test_example() {
     // Create a Play instance - generates unique play_id
     let play = Play::new();
--- a/crates/bitwarden-test/README.md
+++ b/crates/bitwarden-test/README.md
@@ -5,3 +5,58 @@ This crate should only be used in tests and should not be included in production
 </div>
 
 Contains test utilities for Bitwarden.
+
+## Play Framework
+
+The Play framework provides scene-based E2E testing with automatic test isolation and cleanup. Using
+the server side `SeederApi` to generate data.
+
+### How It Works
+
+1. Each `Play` instance generates a unique `play_id` (UUID)
+2. All HTTP requests include an `x-play-id` header for server-side test isolation
+3. The seeder API creates test data (users, etc.) associated with that `play_id`
+4. When the `Play` instance is dropped, all test data is automatically cleaned up
+
+### Usage
+
+```rust
+use bitwarden_test::play::{Play, SingleUserArgs, SingleUserScene};
+
+#[tokio::test]
+async fn test_example() {
+    // Create a Play instance - generates unique play_id
+    let play = Play::new();
+
+    // Create a test user via the seeder API
+    let args = SingleUserArgs {
+        email: "test@example.com".to_string(),
+        verified: true,
+        ..Default::default()
+    };
+    let scene = play.scene::<SingleUserScene>(&args).await.unwrap();
+
+    // Access the created user data (email is mangled server-side)
+    let client_id = scene.get_mangled("client_id").unwrap();
+
+    // Use credentials for testing...
+
+    // Cleanup happens automatically when `play` is dropped
+}
+```
+
+### Environment Variables
+
+| Variable                 | Default                           | Description         |
+| ------------------------ | --------------------------------- | ------------------- |
+| `BITWARDEN_API_URL`      | `https://localhost:8080/api`      | Base API URL        |
+| `BITWARDEN_IDENTITY_URL` | `https://localhost:8080/identity` | Identity server URL |
+| `BITWARDEN_SEEDER_URL`   | `http://localhost:5047`           | Seeder API URL      |
+
+### Running E2E Tests
+
+E2E tests require a running Bitwarden server with the seeder API enabled:
+
+```bash
+cargo test -p bw --features integration
+```
--- a/crates/bitwarden-test/src/play/config.rs
+++ b/crates/bitwarden-test/src/play/config.rs
@@ -56,3 +56,21 @@ impl Default for PlayConfig {
         Self::from_env()
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_new_creates_config_with_custom_urls() {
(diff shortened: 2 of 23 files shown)


crates/bitwarden-test/README.md
```diff
@@ -5,3 +5,58 @@ This crate should only be used in tests and should not be included in production
...
+```
+
+### Environment Variables
+
+| Variable                 | Default                           | Description         |
+| ------------------------ | --------------------------------- | ------------------- |
+| `BITWARDEN_API_URL`      | `https://localhost:8080/api`      | Base API URL        |
+| `BITWARDEN_IDENTITY_URL` | `https://localhost:8080/identity` | Identity server URL |
+| `BITWARDEN_SEEDER_URL`   | `http://localhost:5047`           | Seeder API URL      |
+
```
theMickster, line 77: ⛏️ While we are documenting env variable usage, let's include  `BW_PASSWORD`

theMickster (commented):
Looking strong to me.

crates/bitwarden-test/README.md
Hinton: That's in the CLI not -test, but yes that should probably be documented outside of our help docs.

crates/bitwarden-test/README.md
```diff
@@ -5,3 +5,58 @@ This crate should only be used in tests and should not be included in production
...
+## Play Framework
+
+The Play framework provides scene-based E2E testing with automatic test isolation and cleanup. Using
+the server side `SeederApi` to generate data.
+
+### How It Works
+
+1. Each `Play` instance generates a unique `play_id` (UUID)
+2. All HTTP requests include an `x-play-id` header for server-side test isolation
+3. The seeder API creates test data (users, etc.) associated with that `play_id`
```
MGibson1: x-play-id is more general purpose than just seeded data

```suggestion
3. Any db entry created through a request associated with an `x-play-id` (users, etc.) are saved as associated with that `play_id`
```

crates/bitwarden-test/README.md
```diff
@@ -5,3 +5,58 @@ This crate should only be used in tests and should not be included in production
...
+4. When the `Play` instance is dropped, all test data is automatically cleaned up
```
MGibson1:
```suggestion
4. When the `Play` instance is dropped, all `x-play-id` data is automatically cleaned up
```

crates/bitwarden-test/README.md
```diff
@@ -5,3 +5,58 @@ This crate should only be used in tests and should not be included in production
...
+    // Create a test user via the seeder API
+    let args = SingleUserArgs {
+        email: "test@example.com".to_string(),
+        verified: true,
+        ..Default::default()
+    };
+    let scene = play.scene::<SingleUserScene>(&args).await.unwrap();
+
+    // Access the created user data (email is mangled server-side)
+    let client_id = scene.get_mangled("client_id").unwrap();
```
MGibson1: I don't believe client_id is included in the mangle map, probably better to use something that definitely is mangled.

```suggestion
    let client_id = scene.get_mangled("email").unwrap();
```

crates/bitwarden-test/src/play/config.rs
```diff
@@ -0,0 +1,76 @@
...
+    /// Load configuration from environment variables
+    pub fn from_env() -> Self {
+        let api_url = env::var("BITWARDEN_API_URL")
+            .unwrap_or_else(|_| "https://localhost:8080/api".to_string());
+
+        let identity_url = env::var("BITWARDEN_IDENTITY_URL")
+            .unwrap_or_else(|_| "https://localhost:8080/identity".to_string());
+
+        let seeder_url = env::var("BITWARDEN_SEEDER_URL")
+            .unwrap_or_else(|_| "http://localhost:5047".to_string());
```
MGibson1, line 32: Maybe add todo to use the local web server's proxy for consistency?

crates/bitwarden-test/src/play/scene.rs
```diff
@@ -0,0 +1,36 @@
...
+
+    /// Access the underlying template
+    pub fn inner(&self) -> &T {
+        &self.template
+    }
+
+    /// Get the mangled value for a given key
+    pub fn get_mangled(&self, key: &str) -> Option<&str> {
+        self.mangle_map.get(key).map(|s| s.as_str())
+    }
```
MGibson1: **issue:** This shouldn't be an option, if the map doesn't have it, the method should return the key itself.

This change also significantly cleans up usage points

crates/bw/tests/api_key_login.rs
```diff
@@ -0,0 +1,79 @@
...
+        scene
+            .get_mangled("378538f1-2426-4788-87c5-df39a78618c1")
+            .unwrap()
+    );
+    let client_secret = scene.get_mangled("api_key").unwrap();
+
+    let output = bw()
+        .args(["login", "api-key", &client_id, client_secret])
+        .args(["--server", &server])
+        .env("BW_PASSWORD", "asdfasdfasdf")
```
MGibson1: Best practice to mangle all magic string (that's why `get_mangle` shouldn't return an `Option`)

MGibson1 (changes requested):
One key item that I'm pretty sure is missing here is overriding http data throughout all http requests. I think that means this is likely the wrong structure to accomplish what we're aiming for. This needs to be a library that is injected really low down in the dependency tree to properly inject play-ids everywhere. I don't necessarily want to block this work, but I do at least want to make a conscious and documented decision that we want to work on that later/never rather than now.

crates/bitwarden-test/src/play/config.rs
Hinton: Requires the clients PR to be merged first.

crates/bitwarden-test/README.md
Hinton: This assumed you feed it something that was named client_id. You can't ask it for `email` since you would need the actual email.

crates/bitwarden-test/src/play/play.rs
```diff
@@ -285,7 +287,7 @@ mod tests {
             .await
             .unwrap();
         assert_eq!(scene.inner().data, "test-data");
-        assert_eq!(scene.get_mangled("email"), Some("mangled@example.com"));
+        assert_eq!(scene.get_mangled("email"), "mangled@example.com");
```
MGibson1: **nit:** slightly more realistic example

```suggestion
        assert_eq!(scene.get_mangled("email@example.com"), "mangled@example.com");
```

MGibson1 (changes requested):
please address https://github.com/bitwarden/sdk-internal/pull/710/changes/aefb3f6b560228265ce8502af86ecdf361c7ab36..e310cd24184d90a5db828b6a5537a30885bf4a47#r2728133136

Hinton:
> please address [#710 (changes)](https://github.com/bitwarden/sdk-internal/pull/710/changes/aefb3f6b560228265ce8502af86ecdf361c7ab36..e310cd24184d90a5db828b6a5537a30885bf4a47#r2728133136)

Not sure what this references. Password hard coded on server so seems fine to hard code here.

crates/bitwarden-test/src/play/play.rs
```diff
@@ -0,0 +1,349 @@
...
+
+impl Drop for Play {
+    fn drop(&mut self) {
+        let client = self.client.clone();
+        let play_id = client.play_id().to_string();
+
+        // Use the current runtime to run cleanup synchronously
+        let _ = tokio::runtime::Handle::try_current().map(|handle| {
+            tokio::task::block_in_place(|| {
+                handle.block_on(async { client.delete_seeder(&format!("/seed/{}", play_id)).await })
```
dani-garcia: This Drop implementation seems suboptimal:
- It panics on single thread runtimes, as `block_in_place` is not allowed there
- It won't delete the seeder if accidentally run outside of the runtime, like so:

```rust
// If we accidentally move it to a  separate thread
 let play = Play::new();
std::thread::spawn(|| drop(play));

// Or if the play struct outlives the runtime
let play = Play::new();
tokio::runtime::Builder::new_current_thread()
        .enable_all()
        .build()
        .unwrap()
        .block_on(async {
            let _ = tokio::task::spawn_blocking(|| {
                // let _ = play;
            })
            .await;
        });
```

How important is that this cleanup step runs? I think at the very least we can capture the Handle as part of the Play constructor and panic early if the test is single threaded:
```rust
    pub fn new_with_config(config: PlayConfig) -> Self {
        let play_id = generate_play_id();
        let client = Arc::new(PlayHttpClient::new(play_id, config));
        let handle = tokio::runtime::Handle::current();

        if handle.runtime_flavor() == tokio::runtime::RuntimeFlavor::CurrentThread {
            panic!("Play must be used within a multi-threaded Tokio runtime");
        }

		// Drop can now use self.handle
        Play { client, handle }
    }

impl Drop for Play {
    fn drop(&mut self) {
        let _ = tokio::task::block_in_place(|| self.handle.block_on(self.clean()));
    }
}
```

The only alternative I can think to have it be guaranteed to work, and to have it work with a single threaded constructor would be some higher order function, but that would change the API significantly so maybe not worth it?:
```rust
Play::new().run(async |play| {
    // This runs cleanup right at the end, rather than at drop
}).await;
```

crates/bitwarden-test/src/play/play.rs
```diff
@@ -0,0 +1,349 @@
...
+    ///
+    /// Configuration is loaded from environment variables.
+    /// All test data created through this instance will be cleaned up when dropped.
+    pub fn new() -> Self {
+        let play_id = generate_play_id();
+        let config = PlayConfig::from_env();
+        let client = Arc::new(PlayHttpClient::new(play_id, config));
+
+        Play { client }
+    }
```
dani-garcia:
```suggestion
    pub fn new() -> Self {
        Self::new_with_config(PlayConfig::from_env())
    }
```

This can be simplified based on the other constructor

crates/bitwarden-test/README.md
```diff
@@ -5,3 +5,58 @@ This crate should only be used in tests and should not be included in production
...
+3. Any db entry created through a request associated with an `x-play-id` (users, etc.) are saved as
+   associated with that `play_id`
+4. When the `Play` instance is dropped, associated data is automatically cleaned up
+
+### Usage
+
+```rust
+use bitwarden_test::play::{Play, SingleUserArgs, SingleUserScene};
+
+#[tokio::test]
```
dani-garcia: We need to use the multi thread runtime in the example, otherwise it will panic

crates/bitwarden-test/README.md
```diff
@@ -5,3 +5,58 @@ This crate should only be used in tests and should not be included in production
...
+
+| Variable                 | Default                           | Description         |
+| ------------------------ | --------------------------------- | ------------------- |
+| `BITWARDEN_API_URL`      | `https://localhost:8080/api`      | Base API URL        |
+| `BITWARDEN_IDENTITY_URL` | `https://localhost:8080/identity` | Identity server URL |
+| `BITWARDEN_SEEDER_URL`   | `http://localhost:5047`           | Seeder API URL      |
+
+### Running E2E Tests
+
+E2E tests require a running Bitwarden server with the seeder API enabled:
```
dani-garcia, line 80: Do we have any docs on how to enable this API?

crates/bitwarden-test/src/play/http_client.rs
```diff
@@ -0,0 +1,247 @@
...
+        &self.config
+    }
+
+    /// Add the x-play-id header to a request builder
+    fn with_play_id(&self, builder: RequestBuilder) -> RequestBuilder {
+        builder.header("x-play-id", &self.play_id)
+    }
+
+    /// POST JSON to the seeder API and parse JSON response
+    pub async fn post_seeder<T: Serialize, R: DeserializeOwned>(
```
dani-garcia: Does this need to be public? Same for `delete_seeder` below.

```suggestion
   async fn post_seeder<T: Serialize, R: DeserializeOwned>(
```

crates/bitwarden-test/src/play/play.rs
```diff
@@ -0,0 +1,349 @@
...
+///
+/// Provides methods for creating scenes, executing queries, and managing
+/// test data with automatic cleanup when dropped.
+///
+/// # Example
+///
+/// ```ignore
+/// use bitwarden_test::play::{Play, SingleUserArgs, SingleUserScene};
+///
+/// #[tokio::test]
```
dani-garcia, line 98: Should use multi-thread runtime here too

crates/bitwarden-test/src/play/play.rs
Hinton: It's not that important, you can call the delete endpoint without an argument and it cleans up anything with an associated play id.

I'm not against changing the pattern to that, but would it negatively affect debugging or stack traces?

crates/bitwarden-test/README.md
Hinton: Not currently. You need to start the service AND set `globalSettings:testPlayIdTrackingEnabled` to true for tracking to work.

I want to provide a docker-compose that lets you spin this all up though.

MGibson1:
> > please address [#710 (changes)](https://github.com/bitwarden/sdk-internal/pull/710/changes/aefb3f6b560228265ce8502af86ecdf361c7ab36..e310cd24184d90a5db828b6a5537a30885bf4a47#r2728133136)
> 
> Not sure what this references. Password hard coded on server so seems fine to hard code here.

This was supposed to link to a review comment that wasn't addressed. Not sure what it's linking too, though because it just goes to the files page now.

It was this comment:
> Best practice to mangle all magic string (that's why get_mangle shouldn't return an Option)
 referring to mangling the password in the `BW_PASSWD` env argument for your `bw` example test.

MGibson1 (approved):
Looks good, this might be impacted by my changes in bitwarden/server#6909, which updates the SingleUserScene to return values rather than include them in the mangle map

theMickster (approved):

dani-garcia (approved):

PR #704, opened 2026-01-20
feat(sso): add ServerCommunicationConfigClient for SSO cookie management

addisonbeck:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-29145

## 📔 Objective

Implement `bitwarden-server-communication-config` crate to support SSO load balancer cookie authentication for self-hosted environments. 

The crate defines:

- `ServerCommunicationConfig` with `BootstrapConfig` variants (`Direct` and `SsoCookieVendor`) matching the server's `/api/config` endpoint schema from [PM-29144](https://bitwarden.atlassian.net/browse/PM-29144)
- a `ServerCommunicationConfigRepository` trait for platform-agnostic storage
- a `ServerCommunicationConfigClient` with methods to retrieve configurations, check if cookie bootstrap is needed, and extract cookies for HTTP requests. 

This crate provides the storage layer that bridges server-provided cookie requirements and client-side cookie acquisition flows. It enables downstream work to attach cookies to HTTP requests via request middleware, and allows TypeScript clients to implement `ServerCommunicationConfigService` with proper persistence. 

## 🚨 Breaking Changes

None. These are new concepts.

## ⏰ Reminders before review

- Contributor guidelines followed
- All formatters and local linters executed and passed
- Written new unit and / or integration tests where applicable
- Protected functional changes with optionality (feature flags)
- Used internationalization (i18n) for all UI strings
- CI builds passed
- Communicated to DevOps any deployment requirements
- Updated any necessary documentation (Confluence, contributing docs) or informed the documentation
  team

## 🦮 Reviewer guidelines

- 👍 (`:+1:`) or similar for great changes
- 📝 (`:memo:`) or ℹ️ (`:information_source:`) for notes or general info
- ❓ (`:question:`) for questions
- 🤔 (`:thinking:`) or 💭 (`:thought_balloon:`) for more open inquiry that's not quite a confirmed
  issue and could potentially benefit from discussion
- 🎨 (`:art:`) for suggestions / improvements
- ❌ (`:x:`) or ⚠️ (`:warning:`) for more significant problems or concerns needing attention
- 🌱 (`:seedling:`) or ♻️ (`:recycle:`) for future improvements or indications of technical debt
- ⛏ (`:pick:`) for minor or nitpick changes

[...]

PR #723, opened 2026-01-29
feat(sso): Add `acquireCookie` trait and function to ServerCommunicationConfigClient

addisonbeck:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-30775

## 📔 Objective

Adds an `acquireCookie()` method to `ServerCommunicationConfigClient` to enable the cookie acquisition workflow. This method calls the platform API to trigger cookie acquisition, validates the returned cookie name matches the server configuration, and saves it to the repository.

As a side effect this also adds UniFFI bindings for the entire `bitwarden-server-communication-config` crate, which were missed in my previous PR https://github.com/bitwarden/sdk-internal/pull/704. This required replacing tuple return types with an `AcquiredCookie` struct (UniFFI doesn't support tuples), adding UniFFI annotations to all config types, and creating a `UniffiServerCommunicationConfigClient` wrapper.

## 🚨 Breaking Changes

None right now, because `ServerCommunicationConfigService` hasn't landed in clients. Part of PM-30775 is to implement a noop in `clients`, but I've skipped that for this reason.

### DIFF
--- a/crates/bitwarden-server-communication-config/src/client.rs
+++ b/crates/bitwarden-server-communication-config/src/client.rs
@@ -59,16 +59,15 @@ where
 
     /// Returns cookies to include in HTTP requests
     ///
-    /// For sharded cookies, returns multiple tuples with the same cookie name.
-    /// The browser will send all shards in the HTTP request, and the load balancer
-    /// will reassemble them automatically.
+    /// Returns the stored cookies as-is. For sharded cookies, each entry includes
+    /// the full cookie name with its `-{N}` suffix (e.g., `AWSELBAuthSessionCookie-0`).
     pub async fn cookies(&self, hostname: String) -> Vec<(String, String)> {
         if let Ok(Some(config)) = self.repository.get(hostname).await {
             if let BootstrapConfig::SsoCookieVendor(vendor_config) = config.bootstrap {
-                if let Some(cookie_value_shards) = vendor_config.cookie_value {
-                    return cookie_value_shards
+                if let Some(acquired_cookies) = vendor_config.cookie_value {
+                    return acquired_cookies
                         .into_iter()
-                        .map(|shard| (vendor_config.cookie_name.clone(), shard))
+                        .map(|cookie| (cookie.name, cookie.value))
                         .collect();
                 }
             }
@@ -111,23 +110,49 @@ where
 
         let expected_cookie_name = &vendor_config.cookie_name;
 
-        // Call platform API to acquire the cookie
-        let cookie = self
+        // Call platform API to acquire cookies
+        let cookies = self
             .platform_api
-            .acquire_cookie(hostname.to_string())
+            .acquire_cookies(hostname.to_string())
             .await
             .ok_or(AcquireCookieError::Cancelled)?;
 
-        // Validate the cookie name matches what we expect
-        if cookie.name != *expected_cookie_name {
+        // Validate that all cookies match the expected base name
+        // Cookie names should either:
+        // 1. Exactly match the expected name (unsharded cookie)
+        // 2. Match the pattern {expected_name}-{N} where N is a digit (sharded cookies)
+        //
+        // AWS ALB shards cookies > 4KB with naming pattern: {base_name}-{N}
+        // where N starts at 0 (e.g., AWSELBAuthSessionCookie-0, AWSELBAuthSessionCookie-1)
+        let all_cookies_match = cookies.iter().all(|cookie| {
+            cookie.name == *expected_cookie_name
+                || cookie
+                    .name
+                    .strip_prefix(&format!("{}-", expected_cookie_name))
+                    .is_some_and(|suffix| suffix.chars().all(|c| c.is_ascii_digit()))
+        });
+
+        if !all_cookies_match {
+            // Find the first mismatched cookie for error reporting
+            let mismatched = cookies
+                .iter()
+                .find(|cookie| {
+                    cookie.name != *expected_cookie_name
+                        && !cookie
+                            .name
+                            .strip_prefix(&format!("{}-", expected_cookie_name))
+                            .is_some_and(|suffix| suffix.chars().all(|c| c.is_ascii_digit()))
+                })
+                .expect("all_cookies_match is false, so at least one cookie must not match");
+
             return Err(AcquireCookieError::CookieNameMismatch {
                 expected: expected_cookie_name.clone(),
-                actual: cookie.name,
+                actual: mismatched.name.clone(),
             });
         }
 
-        // Update the cookie value using the mutable reference we already have
-        vendor_config.cookie_value = Some(cookie.value);
+        // Update the cookie values using the mutable reference we already have
+        vendor_config.cookie_value = Some(cookies);
 
         // Save the updated config
         self.repository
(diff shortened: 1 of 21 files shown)


crates/bitwarden-server-communication-config/src/platform_api.rs
```diff
@@ -0,0 +1,69 @@
...
+
+/// A cookie acquired from the platform
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq)]
+#[cfg_attr(feature = "uniffi", derive(uniffi::Record))]
+pub struct AcquiredCookie {
+    /// Cookie name
+    pub name: String,
+    /// Cookie value
+    pub value: String,
+}
```
dereknance, line 25: Given that it's possible for the IdP cookie to be sharded across multiple cookies, should `value` be something like a `Vec<String>` in order to preserve each shard?

crates/bitwarden-server-communication-config/src/wasm/client.rs
```diff
@@ -70,4 +81,28 @@ impl JsServerCommunicationConfigClient {
...
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if:
+    /// - Cookie acquisition was cancelled by the user
+    /// - Server configuration doesn't support SSO cookies (Direct bootstrap)
+    /// - Acquired cookie name doesn't match expected name
+    /// - Repository operations fail
+    #[wasm_bindgen(js_name = acquireCookie)]
+    pub async fn acquire_cookie(&self, hostname: String) -> Result<(), String> {
```
coroiu, line 102: thought: this is fine for now but we'll have to see if mobile wants more specific error handling

crates/bitwarden-server-communication-config/src/wasm/js_platform_api.rs
```diff
@@ -0,0 +1,99 @@
...
+
+#[wasm_bindgen(typescript_custom_section)]
+const TS_CUSTOM_TYPES: &'static str = r#"
+/**
+ * Acquired cookie structure
+ */
+export interface AcquiredCookie {
+    name: string;
+    value: string;
+}
```
coroiu: suggestion: if you add `tsify` support to `AcquiredCookie`

```
#[cfg_attr(
    feature = "wasm",
    derive(tsify::Tsify),
)]
pub struct AcquiredCookie { ... }
```

then its type will automatically be exported and you won't have to manually define it here. You can still use `JsValue` in the `extern "c"` block

crates/bitwarden-uniffi/src/platform/mod.rs
coroiu, line 1: In folders `swift` and `kotlin` we have demo apps that import most of our types and sometimes try to use them to give us some sort of usage-tests. We should add the server communication config types to them

crates/bitwarden-server-communication-config/src/wasm/client.rs
addisonbeck: Makes sense. I was also unsure if this was enough validation at the SDK layer. Asserting that the cookie names line up doesn't really do much. You got any thoughts on that?

crates/bitwarden-uniffi/src/platform/mod.rs
addisonbeck: I added some unused function on https://github.com/bitwarden/sdk-internal/pull/723/commits/bf03823c6e22f3e675fa086f44b29da18a4ef3ae. Is this all we need here? I didn't connect it to any UI, just added the functions to test compilation.

crates/bitwarden-server-communication-config/src/wasm/js_platform_api.rs
addisonbeck: Implemented on https://github.com/bitwarden/sdk-internal/pull/723/commits/f9fff8d95e1416a6b4ddf12b3440f85137ae1571

crates/bitwarden-server-communication-config/src/platform_api.rs
addisonbeck: Good catch. Implemented on https://github.com/bitwarden/sdk-internal/pull/723/commits/8fdf715d0c5b8aac942422f014cc8d6593d70df5

crates/bitwarden-server-communication-config/src/client.rs
```diff
@@ -49,13 +61,77 @@ where
     pub async fn cookies(&self, hostname: String) -> Vec<(String, String)> {
         if let Ok(Some(config)) = self.repository.get(hostname).await {
             if let BootstrapConfig::SsoCookieVendor(vendor_config) = config.bootstrap {
-                if let Some(cookie_value) = vendor_config.cookie_value {
+                if let Some(cookie_value_shards) = vendor_config.cookie_value {
+                    // Concatenate all cookie shards into a single value
+                    let cookie_value = cookie_value_shards.join("");
```
dereknance: It _may_ be necessary to preserve the shards rather than reconstruct the full cookie, or at least the user of the `cookies` function will need to know to split cookie values back into shards no longer than 4K due to [browser limitations on per-cookie size](https://docs.aws.amazon.com/elasticloadbalancing/latest/application/listener-authenticate-users.html#authentication-flow:~:text=Because%20most%20browsers%20limit%20the%20cookie%20size%20to%204K%2C).

dereknance (dismissed):

crates/bitwarden-server-communication-config/src/client.rs
addisonbeck: Good callout, I was wondering about this too. Where the shards need to get joined is fuzzy to me. Is it just in time for an http request? Is it never?

dereknance: As I understand it, the load balancer is what does the re-joining since it does the splitting in the first place.

addisonbeck: ~Reading ELB docs a little and I'm more confused than ever. Seems to me like a shard is just a full named cookie with a value? Wouldn't that mean instead of AquiredCookie.value being a Vec we'd want to pass around a Vec<AquiredCookie> that has a value: String?~

I changed my mind about this. Buffering.

addisonbeck: Okay, assuming the cookies are all the same name (I guess they are?) https://github.com/bitwarden/sdk-internal/pull/723/commits/a64faddbcf3510d30335e8501824799c9d60e659 just returns a bunch of shards under one AquiredCookie, even though I suppose it's technically multiple cookies?

Something about this still seems off to me.

coroiu: Sharded cookie just means multiple cookies. When the token/whatever the load-balancer wants stored as a cookie gets to big it sets multiple cookies instead, all having the form `{cookie_name}-{n}`

```
"cookies": [
  { "name": "AWSELBAuthSessionCookie-0", "value": "..." },
  { "name": "AWSELBAuthSessionCookie-1", "value": "..." }
]
```

We don't really know/care why that happens, we just need to look for it and send them back

crates/bitwarden-server-communication-config/src/client.rs
```diff
@@ -58,13 +58,18 @@ where
...
     pub async fn cookies(&self, hostname: String) -> Vec<(String, String)> {
         if let Ok(Some(config)) = self.repository.get(hostname).await {
             if let BootstrapConfig::SsoCookieVendor(vendor_config) = config.bootstrap {
                 if let Some(cookie_value_shards) = vendor_config.cookie_value {
-                    // Concatenate all cookie shards into a single value
-                    let cookie_value = cookie_value_shards.join("");
-                    return vec![(vendor_config.cookie_name, cookie_value)];
+                    return cookie_value_shards
+                        .into_iter()
+                        .map(|shard| (vendor_config.cookie_name.clone(), shard))
```
coroiu: issue: yeah, if we're supporting shards then we'll need to change the format to match `{cookie-name}-{n}` if the `Vec` contains more than 1 value. Alternatively we could tweak the config struct to contain an enum, e.g.

```
enum CookieValue {
  Simple(String),
  Sharded(Vec<String>)
}
```

crates/bitwarden-uniffi/src/platform/mod.rs
coroiu: Yeah I think that's fine

dereknance (approved):
LGTM! 🥇

coroiu (approved):

PR #727, opened 2026-01-31
[PM-30923] Update sdk-internal with the latest send models

harr1424:
## 🎟️ Tracking

](https://bitwarden.atlassian.net/browse/PM-30923

## 📔 Objective

This PR is intended to bring the SDK into parity with the `SendService` available in the `clients` repo. Specifically: 
- When a `Send` is configured to use Email OTP authentication the `emails` are encrypted using the send's key before leaving the client 
- A comma separated list of SHA256 hashes is sent to the server in plaintext, so that these hashes may be used to validate emails entered during the authentication flow

### DIFF
--- a/crates/bitwarden-send/src/send.rs
+++ b/crates/bitwarden-send/src/send.rs
@@ -306,11 +306,9 @@ impl Decryptable<KeyIds, SymmetricKeyId, SendView> for Send {
             emails: self
                 .emails
                 .as_ref()
-                .and_then(|enc| {
-                    Decryptable::<KeyIds, SymmetricKeyId, String>::decrypt(enc, ctx, key).ok()
-                })
+                .and_then(|enc| -> Option<String> { enc.decrypt(ctx, key).ok() })
                 .as_deref()
-                .unwrap_or("")
+                .unwrap_or_default()
                 .split(',')
                 .map(|e| e.trim())
                 .filter(|e| !e.is_empty())
@@ -408,9 +406,9 @@ impl CompositeEncryptable<KeyIds, SymmetricKeyId, Send> for SendView {
                     .iter()
                     .map(|email| {
                         let hash = sha2::Sha256::new()
-                            .chain_update(email.to_lowercase().as_bytes())
+                            .chain_update(email.to_lowercase().trim().as_bytes())
                             .finalize();
-                        format!("{:X}", hash)
+                        format!("{hash:X}")
                     })
                     .collect::<Vec<_>>()
                     .join(",")
--- a/crates/bitwarden-send/src/send.rs
+++ b/crates/bitwarden-send/src/send.rs
@@ -11,6 +11,7 @@ use bitwarden_encoding::{B64, B64Url};
 use chrono::{DateTime, Utc};
 use serde::{Deserialize, Serialize};
 use serde_repr::{Deserialize_repr, Serialize_repr};
+use sha2::Digest;
 use uuid::Uuid;
 use zeroize::Zeroizing;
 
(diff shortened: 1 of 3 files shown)


crates/bitwarden-send/src/send.rs
```diff
@@ -295,8 +305,12 @@ impl Decryptable<KeyIds, SymmetricKeyId, SendView> for Send {
 
             emails: self
                 .emails
+                .as_ref()
+                .and_then(|enc| {
+                    Decryptable::<KeyIds, SymmetricKeyId, String>::decrypt(enc, ctx, key).ok()
```
dani-garcia: Small improvement, I think you can use decrypt the normal way as long as you specify the return type of the function:
```suggestion
                .and_then(|enc| -> Option<String> {
                    enc.decrypt(ctx, key).ok()
```

crates/bitwarden-send/src/send.rs
```diff
@@ -108,7 +109,14 @@ pub struct Send {
     /// Email addresses for OTP authentication.
     /// **Note**: Mutually exclusive with `new_password`. If both are set,
     /// only password authentication will be used.
-    pub emails: Option<String>,
+    pub emails: Option<EncString>,
```
dani-garcia, line 112: Will this change cause compatibility issues when using old server versions?

crates/bitwarden-send/src/send.rs
```diff
@@ -384,7 +400,21 @@ impl CompositeEncryptable<KeyIds, SymmetricKeyId, Send> for SendView {
...
+                .then(|| self.emails.join(","))
+                .encrypt(ctx, send_key)?,
+            email_hashes: (!self.emails.is_empty()).then(|| {
+                self.emails
+                    .iter()
+                    .map(|email| {
+                        let hash = sha2::Sha256::new()
+                            .chain_update(email.to_lowercase().as_bytes())
+                            .finalize();
+                        format!("{:X}", hash)
```
dani-garcia:
```suggestion
                        format!("{hash:X}")
```

crates/bitwarden-send/src/send.rs
```diff
@@ -295,8 +305,12 @@ impl Decryptable<KeyIds, SymmetricKeyId, SendView> for Send {
 
             emails: self
                 .emails
+                .as_ref()
+                .and_then(|enc| {
+                    Decryptable::<KeyIds, SymmetricKeyId, String>::decrypt(enc, ctx, key).ok()
+                })
                 .as_deref()
-                .unwrap_or_default()
+                .unwrap_or("")
```
dani-garcia: Why is this changed? The previous `unwrap_or_default` should have the same effect

crates/bitwarden-send/src/send.rs
```diff
@@ -384,7 +400,21 @@ impl CompositeEncryptable<KeyIds, SymmetricKeyId, Send> for SendView {
...
-            emails: (!self.emails.is_empty()).then(|| self.emails.join(",")),
+            emails: (!self.emails.is_empty())
+                .then(|| self.emails.join(","))
+                .encrypt(ctx, send_key)?,
+            email_hashes: (!self.emails.is_empty()).then(|| {
+                self.emails
+                    .iter()
+                    .map(|email| {
+                        let hash = sha2::Sha256::new()
+                            .chain_update(email.to_lowercase().as_bytes())
```
dani-garcia: Should email be trimmed here as well, to make sure we don't include whitespace in the hash?

crates/bitwarden-send/src/send.rs
harr1424: While the former server version (using `String`) has been released, outside of development testing, no clients have sent non-null values for this field, as the updates to client models haven't been released yet.

dani-garcia: Perfect, thank you!

dani-garcia (approved):

PR #728, opened 2026-01-31
[PM-24221] Implement v2 post_user_password_registration API

eligrubb:
## 🎟️ Tracking

[https://bitwarden.atlassian.net/browse/PM-24221](https://bitwarden.atlassian.net/browse/PM-24221)

## 📔 Objective

Implements standard user master password registration flow in the sdk. Creates new v2 cryptographic state and calls to the server api using the `bitwarden-api-identity` crate bindings. 

Tests calls using `mockall` harness.


crates/bitwarden-api-api/src/apis/self_hosted_account_billing_v_next_api.rs
```diff
@@ -76,6 +76,7 @@ pub trait SelfHostedAccountBillingVNextApi: Send + Sync {
         last_key_rotation_date: Option<String>,
         last_email_change_date: Option<String>,
         verify_devices: Option<bool>,
+        v2_upgrade_token: Option<&'a str>,
```
quexten: I'm a bit confused why we would post a v2 upgrade token. Looks like the server-side of this is:
```
    [InjectUser]
    public async Task<IResult> UploadLicenseAsync(
        [BindNever] User user,
```

but it's still adding all user fields here. A user should not be posted / provided from the client's perspective.

@dani-garcia  / @Hinton anything we can do here to improve the generated bindings?

crates/bitwarden-core/src/key_management/account_keys.rs
```diff
@@ -0,0 +1,318 @@
+use bitwarden_error::bitwarden_error;
```
quexten: OPTIONAL (please feel free to ignore if you don't think it's worth it, but maybe add a comment to the top of the file explaining what this is doing):

It's a bit unfortunate that we have two identical request models here that we are introducing an intermediary struct for:

```
#[derive(Clone, Default, Debug, PartialEq, Serialize, Deserialize)]
pub struct AccountKeysRequestModel {
    #[serde(
        rename = "userKeyEncryptedAccountPrivateKey",
        alias = "UserKeyEncryptedAccountPrivateKey"
    )]
    pub user_key_encrypted_account_private_key: Option<String>,
    #[serde(rename = "accountPublicKey", alias = "AccountPublicKey")]
    pub account_public_key: Option<String>,
    #[serde(
        rename = "publicKeyEncryptionKeyPair",
        alias = "PublicKeyEncryptionKeyPair",
        skip_serializing_if = "Option::is_none"
    )]
    pub public_key_encryption_key_pair: Option<Box<models::PublicKeyEncryptionKeyPairRequestModel>>,
    #[serde(
        rename = "signatureKeyPair",
        alias = "SignatureKeyPair",
        skip_serializing_if = "Option::is_none"
    )]
    pub signature_key_pair: Option<Box<models::SignatureKeyPairRequestModel>>,
    #[serde(
        rename = "securityState",
        alias = "SecurityState",
        skip_serializing_if = "Option::is_none"
    )]
    pub security_state: Option<Box<models::SecurityStateModel>>,
}
```

This model is duplicated for API and Identity, and we are essentially just converting between the two here. As these are external models, we cannot implement From, for these, but I wonder whether a pure function `from_api_model(model: bitwarden_api::AccountRequestModel) -> bitwarden_identity::AccountRequestModel {`
that skips the intemediary model would be simpler?

crates/bitwarden-api-api/src/apis/self_hosted_account_billing_v_next_api.rs
dani-garcia: @quexten Hmm yeah, I've noticed this a while ago but didn't have the time to fix it and eventually forgot about it. The problem is that Swashbuckle doesn't respect `[BindNever]` when applied to a query parameter in the function arguments, my guess is because it first deconstructs the `User` type into fields, and then later tries to remove the `user` parameter, which doesn't exist. 

The possible solutions are:
- Remove the `user` parameter and the `InjectUser` attribute and inject the userService in the controller directly, which is how this is done in the rest of controllers.
- Set a different binding source, like `[FromServices, BindNever] User user`. Though this is somewhat confusing as you're more or less abusing the annotation's purpose. Could also be done by creating a new attribute that applies both?

Billing seems to be the only team using this injection strategy so it would be good to know if there's a specific reason for it, because it doesn't seem to be a common pattern.

quexten: Starting a discussion with billing https://bitwarden.slack.com/archives/C05LRC3DBDK/p1771858123530539

mzieniukbw:
@eligrubb the api crates changes you have in the PR will go away once you merge `main` - part of #789 (merged)

eligrubb:
Rebased onto main. As @mzieniukbw said, with #789 merging to main, the changes to bitwarden-api-api and bitwarden-api-identity are simplified. Should be clear for review.

crates/bitwarden-core/src/key_management/account_keys.rs
eligrubb: I've implemented your suggestion, the result is more maintainable. You can see the cleanup in commit 47dbd89.

eligrubb:
Apologies for the rebase, normal git merge was having issues with `bitwarden-api-api` and `bitwarden-api-identity` crates. [Server PR#6798](https://github.com/bitwarden/server/pull/6798) should be merged soon, which will clean up those bindings related changes. I also addressed review comments.

quexten (dismissed):

crates/bitwarden-api-api/src/apis/self_hosted_account_billing_v_next_api.rs
dani-garcia: Discovered that this can be fixed by using a server side schema filter in https://github.com/bitwarden/server/pull/7257

eligrubb:
Now that [server#6798](https://github.com/bitwarden/server/pull/6798) has been merged I've rebased my PR off of the resulting identity-api changes in #946. This PR should be much simpler to grok as a result. The code itself should be unchanged from the last review.

quexten (approved):

crates/bitwarden-auth/src/registration.rs
```diff
@@ -466,6 +513,130 @@ pub struct JitMasterPasswordRegistrationResponse {
...
+    identity_client
+        .accounts_api()
+        .post_register_finish(Some(api_request))
+        .await
+        .map_err(|e| {
+            error!("Failed to post account keys: {e:?}");
+            RegistrationError::Api
+        })?;
+
+    Ok(UserMasterPasswordRegistrationResponse {
```
quexten, line 568: Just noting, we are getting close to being able to set this into state from SDK directly, since the entire v2 work dragged out quite a while (and this was not forseeable back then). However, we shouldn't divert from the originally planned path, and just make a mental note to move this to sdk-written state later.

addisonbeck (approved):

PR #729, opened 2026-01-31
[PM-32677] Automate creation of clients PR on sdk-internal merge

trmartin4:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-32677

## 📔 Objective

On merge to `main` in `sdk-internal`, the `publish-wasm-internal` workflow will trigger the `sdk-update.yml` workflow in `clients`.

This mirrors the functionality for [Android](https://github.com/bitwarden/sdk-internal/blob/48b1e0f44816489558d72c72b36a68dfa3c175bc/.github/workflows/build-android.yml#L139).

Corresponding `clients` PRs: https://github.com/bitwarden/clients/pull/18700 and https://github.com/bitwarden/clients/pull/19678.


README.md
```diff
@@ -212,7 +212,20 @@ For example:
...
+
+#### Through a pull request (preferred)
+
+1. Merge the `sdk-internal` pull request. This will automatically trigger a build and publish of the
+   latest changes to npm, and open a pull request against `main` in the `clients` repo with the
+   updated SDK version.
+2. Merge the pull request to `main` on `clients`.
+3. Pull the change into your feature branch to continue development.
+
+#### The manual option
```
coroiu, line 243: suggestion: I think we should add some more nuance/guidance to the choice between these two, specifically what to do when the SDK PR contains breaking changes. In that case the workflow would create a PR with errors that can't be merged as-is. The dev has two options at that point: address the type errors to make it compile and mergeable or actually change the clients code to use the new changes. The second one would mean merging a feature PR and not the workflow-created PR

trmartin4: What about triggering the workflow to create a PR to merge into your feature branch instead of manually updating the `npm` version?  I don't think that's any easier, right?  I don't want to complicate things by suggesting unnecessary options.

coroiu (approved):

vgrassia (approved):

PR #734, opened 2026-02-02
Derive Clone for Keystore manually

dani-garcia:
## 📔 Objective

The Clone derive macro adds a `Ids: Clone` bound on the `KeyStore` Clone impl, which is not needed as the type is only used inside an `Arc` which is always clonable.

### DIFF
--- a/crates/bitwarden-crypto/src/store/mod.rs
+++ b/crates/bitwarden-crypto/src/store/mod.rs
@@ -95,11 +95,19 @@ pub use key_rotation::*;
 /// let decrypted = Data("Hello, World!".to_string());
 /// let encrypted = store.encrypt(decrypted).unwrap();
 /// ```
-#[derive(Clone)]
 pub struct KeyStore<Ids: KeyIds> {
     // We use an Arc<> to make it easier to pass this store around, as we can
     // clone it instead of passing references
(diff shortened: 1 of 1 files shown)


quexten (approved):
Seems reasonable, however is there a specific reason why we would not want Clone on Ids? The point of Ids is that they should be safe to clone around - in contrast to keys which we want to just have in memory once?

dani-garcia:
@quexten The individual `KeyId` types are `Clone + Copy` always as we require it in the trait definition. What's neither `Clone` or `Copy` is the `KeyIds` type:
```rust
pub trait KeyId: Clone + Copy + Hash + Eq + others {}

pub trait KeyIds {
    type Symmetric: KeyId<KeyValue = SymmetricCryptoKey>;
    type Private: KeyId<KeyValue = PrivateKey>;
    type Signing: KeyId<KeyValue = SigningKey>;
}
```
`KeyIds` is purely a type-level construct used to group related key ID types together, and it's never instantiated as a value. It only exists so we can write `KeyStore<Ids>` instead of `KeyStore<SymmetricId, AsymmetricId, SigningId>`. We could add a clone bound to it so that the automatic derivation works, but since we never actually have a value of type `impl KeyIds`, it seems a bit strange to me.

quexten:
Makes sense, thank you for the explanation.

PR #735, opened 2026-02-02
Disable decrypt spans in tracing

quexten:


dani-garcia (commented):
If this is just for performance, you might also be able to use compile time level filters: https://docs.rs/tracing/latest/tracing/level_filters/index.html#compile-time-filters

quexten:
@dani-garcia Yeah, that seems reasonable. Ideally we would like to exclude the above spans specifically only from registration in the browser performance tooling, but not from adding context to logs.

Maybe the actual solution here is to turn of hooking into the performance API when in release builds?

dani-garcia:
@quexten The performance layer is attached here, separately of the printed logs: 
https://github.com/bitwarden/sdk-internal/blob/main/crates/bitwarden-wasm-internal/src/init.rs#L43

You could see if disabling just that is enough?

quexten:
@dani-garcia tracking new PR here: https://github.com/bitwarden/sdk-internal/pull/738

PR #737, opened 2026-02-02
[PM-25283] Add auth middleware support

dani-garcia:
## 🎟️ Tracking
https://bitwarden.atlassian.net/browse/PM-25283
https://bitwarden.atlassian.net/browse/PM-29938

## 📔 Objective

Rework the template code to simplify the authentication logic. Removed logic for aws and google cloud tokens and added support for the `AuthRequired` extension, which will allow us to use a middleware for authentication. The template is still manually setting the bearer tokens until we have the middleware set up.

I've also removed the superfluous user agent here, as we set it as a default header during construction.


djsmith85 (approved):
@dani-garcia: Thanks for mentioning and discussing the changes during the sync yesterday. Platform-owned changes are looking good.

PR #740, opened 2026-02-03
[PM-25283] Update token renewal to use middleware

dani-garcia:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-25283

## 📔 Objective

Migrated the token renewal to use the new middleware and moved the middleware implementation to be part of `bitwarden-auth`. As part of this PR, the changes include:
- Created a `TokenHandler` trait that abstracts the creation of the authentication middleware. This is needed to bypass the circular dependency between core and auth and unblocks moving/rewriting the rest of the renewal code to the auth crate. This trait has multiple implementations:
    - Two in the `bitwarden-auth` crate: These implementation is for SDK-managed tokens (one for PM and another for SM). Currently they're pretty simple and rely on some internal code in `bitwarden-core` for doing the renewal requests. In the future this implementation will deal with persisting and reading tokens from state. 
    - A noop implementation in `bitwarden-core`, for using in tests and when using Client::new()
    - An implementation in core that wraps the existing `ClientManagedTokens` trait, for compatibility with the current apps.
- Removed the existing `Tokens` enum on Client, which is replaced by one of the `TokenHandler` implementations.
- Changed core::Client initialization to setup the middleware, and made `ApiConfigurations` read-only
- Refactored some of the `bitwarden_core::auth` to be easier to access from the auth crate.
- Removed the token renewal when calling `get_api_configurations()`. I'll make the method non-async as a follow up.
- Updated `bitwarden-pm` and `bitwarden-sm` to initialize `bitwarden-auth`'s `TokenHandler` when creating the client.
- Removed an unused `renew_token` in the internal `AuthClient`. It was unused and there's no reason to need it when the middleware handles the operation.


crates/bitwarden-auth/src/renew/token_handler.rs
```diff
@@ -0,0 +1,157 @@
...
+                refresh_token,
+                login_method,
+                identity_config,
+                key_store,
+            )
+            .await?;
+
+        self.set_tokens(access_token, refresh_token, expires_on);
+
+        Err(NotAuthenticatedError)?
```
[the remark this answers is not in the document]
dani-garcia: Fixed in https://github.com/bitwarden/sdk-internal/pull/740/commits/49ea1d5ed433935a5c5576b4e756a9812e4535bb

tangowithfoxtrot (dismissed):

crates/bitwarden-core/src/auth/api/request/mod.rs
```diff
@@ -24,24 +25,20 @@ use crate::{
...
+    let response = identity_config
         .client
-        .post(format!("{}/connect/token", &config.base_path))
+        .post(format!("{}/connect/token", &identity_config.base_path))
         .header(
             reqwest::header::CONTENT_TYPE,
             "application/x-www-form-urlencoded; charset=utf-8",
         )
         .header(reqwest::header::ACCEPT, "application/json")
-        .header("Device-Type", configurations.device_type as usize)
```
dani-garcia, line 44: Note that this was removed as we always set this header when constructing the client:
https://github.com/bitwarden/sdk-internal/blob/5e00ade1c8b075074da92c6af27252aa5e5816a8/crates/bitwarden-core/src/client/client.rs#L157

crates/bitwarden-core/src/client/internal.rs
```diff
@@ -173,23 +136,6 @@ impl InternalClient {
             .clone()
     }
 
-    #[allow(missing_docs)]
-    pub fn get_access_token_organization(&self) -> Option<OrganizationId> {
```
dani-garcia, line 177: This was a secrets manager only function, so I've moved it entirely into bitwarden-sm

crates/bitwarden-auth/src/renew/mod.rs
JaredSnider-Bitwarden, line 1: ❓ : would `token_management` be a better domain scoped name for this module? It is building out all the shared token handlers + the request interception middleware to add the tokens on authenticated requests. Those seem to extend beyond just token_renewal.

crates/bitwarden-auth/src/renew/common.rs
```diff
@@ -0,0 +1,120 @@
...
+
+#[cfg_attr(not(target_arch = "wasm32"), async_trait::async_trait)]
+#[cfg_attr(target_arch = "wasm32", async_trait::async_trait(?Send))]
+pub(crate) trait MiddlewareExt: 'static + Send + Sync {
+    async fn get_token(&self) -> Result<Option<String>, LoginError>;
+}
+
+#[cfg_attr(not(target_arch = "wasm32"), async_trait::async_trait)]
+#[cfg_attr(target_arch = "wasm32", async_trait::async_trait(?Send))]
+impl<T: MiddlewareExt> reqwest_middleware::Middleware for MiddlewareWrapper<T> {
```
JaredSnider-Bitwarden: 🎨 : could we document this with something like:
```
    /// Implements HTTP middleware that attaches authentication tokens to requests.
    /// Delegates to `MiddlewareExt::get_token` to retrieve tokens (which handles renewal).
    /// Only applies to requests marked with `AuthRequired`.
```

crates/bitwarden-auth/src/token_management/password_manager_token_handler.rs
JaredSnider-Bitwarden, line 1: 🎨 : Could we add `_token_handler` to the filename to have `password_manager_token_handler` to be explicit?

crates/bitwarden-auth/src/token_management/secrets_manager_token_handler.rs
JaredSnider-Bitwarden, line 1: 🎨 : Could we add `_token_handler` to the filename to have `secrets_manager_token_handler` to be explicit?

crates/bitwarden-auth/src/token_management/password_manager_token_handler.rs
```diff
@@ -0,0 +1,178 @@
...
+
+#[cfg_attr(not(target_arch = "wasm32"), async_trait::async_trait)]
+#[cfg_attr(target_arch = "wasm32", async_trait::async_trait(?Send))]
+impl MiddlewareExt for PasswordManagerTokenHandler {
+    async fn get_token(&self) -> Result<Option<String>, LoginError> {
+        // We're not holding on to a lock for the duration of the token renewal, so if multiple
+        // requests come in at the same time when the token is expired, we may end up renewing the
+        // token multiple times. This is not ideal, but it's the behavior of the previous
+        // implementation. We should be able to introduce an async semaphore or something
+        // similar to prevent this if it becomes an issue in practice.
```
JaredSnider-Bitwarden, line 67: We've recently had to build in functionality into clients to prevent multiple requests from happening. If we do plan on wiring this up sooner rather than later, I would request that we implement the fix before doing so. We don't have to do it in this PR, but we can create a ticket to track the work and put here if you wish.

crates/bitwarden-auth/src/renew/common.rs
JaredSnider-Bitwarden, line 1: ❓ thoughts on splitting this file into a `test_utils.rs` and a `middleware.rs `file?

crates/bitwarden-core/src/client/client.rs
```diff
@@ -47,20 +54,30 @@ impl Client {
...
             base_path: settings.identity_url,
             user_agent: Some(settings.user_agent.clone()),
-            client: bw_http_client.clone(),
+            client: bw_http_client.clone().into(),
             oauth_access_token: None,
         };
 
+        // Create the client for the API service, with authentication middleware.
+        let auth_middleware = token_handler.initialize_middleware(
+            login_method.clone(),
```
JaredSnider-Bitwarden, line 74: ❓ : does this imply that we will need to be saving the login method in the future as well or are we going to re-architect this as well?  Asking b/c it has implications for all the login client work.

crates/bitwarden-auth/src/renew/common.rs
```diff
@@ -0,0 +1,120 @@
...
+    async fn handle(
+        &self,
+        mut req: reqwest::Request,
+        ext: &mut http::Extensions,
+        next: reqwest_middleware::Next<'_>,
+    ) -> Result<reqwest::Response, reqwest_middleware::Error> {
+        if ext.get::<AuthRequired>().is_some() {
```
JaredSnider-Bitwarden: ❓ : so, `AuthRequired` handles basic, bearer, and custom auth, but we ignore the actual variant and always use bearer?  This feels a bit confusing from a user of the `AuthRequired` marker struct. Should we enforce any BW usage to be bearer and error or something for the other variants if they aren't supported?

JaredSnider-Bitwarden (commented):
Thank you for continuing to evolve the auth domain in the SDK. The detailed objective write up was incredibly helpful to understand the current state + where we are moving towards. 

Most of my focus has been on the `bitwarden-auth` crate since I'm familiar with that one much more so than the core crate, but most of my questions are naming / organization things - let's discuss!

crates/bitwarden-auth/src/renew/mod.rs
dani-garcia: Yeah that's fair, renamed in https://github.com/bitwarden/sdk-internal/pull/740/commits/378afe4d323b03fd15ca0258d0bd44b874a62503!

crates/bitwarden-auth/src/renew/common.rs
dani-garcia: I've added more comments for the implementation and the trait here https://github.com/bitwarden/sdk-internal/pull/740/commits/754f65c9af1b39da6c42cc59397bf1e2c5e647ef

crates/bitwarden-auth/src/renew/common.rs
dani-garcia: Good point, I forgot that the `AuthRequired` type has multiple variants. I've changed it to check for `Bearer` specifically here: https://github.com/bitwarden/sdk-internal/pull/740/changes/754f65c9af1b39da6c42cc59397bf1e2c5e647ef

For other variants of `AuthRequired` I'm currently just logging a warning, which is the same thing I'm doing if there's no token or there's an error retrieving the token. My assumption is that the request will then fail without the token. Do you think that sounds good, or should we just hard error on these cases?

crates/bitwarden-auth/src/renew/common.rs
dani-garcia: Sounds reasonable enough, done https://github.com/bitwarden/sdk-internal/pull/740/commits/ffefc659565a9ca8afd11711f8e73f4d3880cb94

crates/bitwarden-auth/src/token_management/password_manager_token_handler.rs
dani-garcia: Done https://github.com/bitwarden/sdk-internal/pull/740/commits/2bfc79b740d8d0b77c723036ea8ae37017493423

crates/bitwarden-auth/src/token_management/secrets_manager_token_handler.rs
dani-garcia: Done https://github.com/bitwarden/sdk-internal/pull/740/commits/2bfc79b740d8d0b77c723036ea8ae37017493423

crates/bitwarden-auth/src/token_management/password_manager_token_handler.rs
dani-garcia: Sounds reasonable, the fix is fairly easy, but I wasn't sure if the clients were doing it so I didn't want to introduce a functionality change here.

I've created a followup ticket and will solve it once this is merged: https://bitwarden.atlassian.net/browse/PM-32346

crates/bitwarden-core/src/client/client.rs
dani-garcia: Yes, that will partially be done by the SDK client rehydration epic, and then later on when we finish the login client work. Long term I expect for `LoginMethod` to be an entirely auth-only type, and so it would be up to auth to store it/retrieve it when needed. 

By this I mean:
- The login module in auth, when initializing a logged in client after login, will need to persist their new instance of `LoginMethod`, just like we'll need to persist the crypto state.
- When restoring a client from state, `initialize_middleware` would only receive a `StateRegistry`, which auth can use to access the persisted `LoginMethod` directly when renewing tokens.

Does that answer the question?

JaredSnider-Bitwarden: I think so. Thank you!

crates/bitwarden-auth/src/renew/common.rs
JaredSnider-Bitwarden: I have a slight preference for the hard error, but I think that's fine for now.

JaredSnider-Bitwarden (approved):
Thank you for the tweaks and explanations! LGTM!

addisonbeck (approved):

PR #742, opened 2026-02-03
[PM-31668] Add search indexing fields to CipherListView 

gbubemismith:
## 🎟️ Tracking
https://bitwarden.atlassian.net/browse/PM-31668

## 📔 Objective
  - Add `notes`, `fields`, and `attachment_names` fields to `CipherListView` to enable client-side search indexing without requiring full cipher decryption
  - Add `FieldListView` struct that exposes only field names and text field values (hidden field values are excluded for security) 
  - Add `decrypt_list_full_with_failures` method to decrypt list of ciphers and return full cipher views.

### DIFF
--- a/crates/bitwarden-vault/src/cipher/cipher.rs
+++ b/crates/bitwarden-vault/src/cipher/cipher.rs
@@ -428,6 +428,17 @@ pub struct CipherListView {
     pub copyable_fields: Vec<CopyableCipherFields>,
 
     pub local_data: Option<LocalDataView>,
+
+    /// Decrypted cipher notes for search indexing.
+    #[cfg(feature = "wasm")]
+    pub notes: Option<String>,
+    /// Decrypted cipher fields for search indexing.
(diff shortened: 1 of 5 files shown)


nikwithak (approved):
Thanks for this!

crates/bitwarden-vault/src/cipher/cipher.rs
```diff
@@ -428,6 +428,17 @@ pub struct CipherListView {
     pub copyable_fields: Vec<CopyableCipherFields>,
 
     pub local_data: Option<LocalDataView>,
+
+    /// Decrypted cipher notes for search indexing.
+    #[cfg(feature = "wasm")]
```
shane-melton, line 433: 👍 Glad we have this on `wasm` only so we avoid polluting the Mobile clients with additional data they may not need. I know they're more sensitive to memory constraints which is why they need the limited view in the first place.

shane-melton (approved):

PR #745, opened 2026-02-04
Add ListOrganizationCiphersResult to return both ciphers and list views

nikwithak:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-30304

## 📔 Objective

This PR changes the return type of the `CIphersClient` and `CipherAdminClient` get functions to use `CipherView` and `CipherListView`. With `CipherListView` it was both causing type-coercion issues on the TS clients, and also losing some data not present in the ListView.


Hinton (commented):
What fields are we missing in list, should we consider adding them instead? The reason we use list is to expose less things to memory. (Mobile uses list views so we probably want to still give them a way to continue to get list views in the newer apis)

dani-garcia:
To expand on Oscar's comment, the main reason for the list views is as a security measure to ensure we keep the least amount of decrypted data in memory unless absolutely necessary. This is pretty important when we can't rely on the TS clients to clean memory up after use, which means we have to deal with hacks like killing the process to ensure the memory gets cleaned up and a vault lock doesn't leave all the user's passwords in memory.

Before we go back to returning full Ciphers here, it would be good to think if there's anything that we can change on the list views that would help solve your issue without exposing the full Cipher. We can add more fields to them or change the structures to more closely match full Ciphers if that helps, as long as we omit the most sensitive fields (password, totp key, passkey keys, ssh keys, maybe hidden fields?).

If this is just a temporary measure to get it working for now you can disregard my comment, though in that case it would be a good idea to have a tracking ticket somewhere to consider migrating back to list views at a later point.

SaintPatrck:
> Mobile uses list views so we probably want to still give them a way to continue to get list views in the newer apis)

⚠️ Apologies for seagulling here, but this was a significant change for both Mobile clients and worth restating the concern. At the very least `CipherListView` needs to be retained so the Mobile work does not have to be reverted and re-introduced if this needs to be done as a temporary solution.

nikwithak:
@Hinton @SaintPatrck @dani-garcia - Thanks for the context here - in this case, I ran into issues integrating this because the `GetAllFromApiForOrganization` function (and other Admin functions) don't store the cipher into state / repository.

In other parts of the app when dealing with `CipherListView`, we end up [pulling the Cipher from the state when we want the full decrypted view,](https://github.com/bitwarden/clients/blob/main/libs/common/src/vault/services/cipher.service.ts#L2435-L2445) which we can't do in this case because of ⬆️ 

Because `CipherListView` doesn't have all the fields, we can't use in on the TS side since it's replacing a function that previously returned the full Cipher. 

> If this is just a temporary measure to get it working for now you can disregard my comment, though in that case it would be a good idea to have a tracking ticket somewhere to consider migrating back to list views at a later point.

This wasn't intended to be just a temporary measure, but I want to make sure I don't break Mobile or other clients, so would love to discuss the appropriate path forward.

@SaintPatrck - It's worth noting that this only affects the `list_ciphers()` function in the `CiphersClient` and `CipherAdminClient`, which I don't believe are in use by Mobile currently (these were just added last week, and we're integrating them into the TS client). So I don't believe this will affect anything currently in use by Mobile - but again, I want to do it right the first time through!

crates/bitwarden-vault/src/cipher/cipher_client/admin/get.rs
```diff
@@ -43,10 +43,10 @@ pub async fn list_org_ciphers(
         .map(|model| model.merge_with_cipher(None))
         .collect::<Result<Vec<_>, _>>()?;
 
-    let (successes, failures) = key_store.decrypt_list_with_failures(&ciphers);
-    Ok(DecryptCipherListResult {
-        successes,
-        failures: failures.into_iter().cloned().collect(),
+    let (list_views, _failures) = key_store.decrypt_list_with_failures(&ciphers);
```
shane-melton, line 46: ❓ Is dropping the `failures` intentional?

shane-melton (commented):
Looks good, just one question about decryption failures.

crates/bitwarden-vault/src/cipher/cipher_client/admin/get.rs
gbubemismith: Yeah, I dropped it intentionally cause we return the encrypted ciphers and the list view and still decrypt the ciphers on the client. The other option was to introduce failures here 
``` rust
pub struct ListOrganizationCiphersResult {
    /// All encrypted ciphers returned from the API.
    pub ciphers: Vec<Cipher>,
    /// Successfully decrypted `CipherListView` objects.
    pub list_views: Vec<CipherListView>,
    pub failures: Vec<Cipher>,
}
``` 

🤔 thinking about this again, it might make sense to just add the  failures as well. What do you think?

shane-melton: After looking at this usage in the Clients, I see we never checked/used failures for this admin endpoint as it wasn't a scenario we have encountered. I'm good with keeping the existing behavior.

shane-melton (approved):

PR #746, opened 2026-02-05
[PM-30580] Add encryptMany to SDK 

gbubemismith:
## 🎟️ Tracking
https://bitwarden.atlassian.net/browse/PM-30580

### DIFF
--- a/crates/bitwarden-vault/src/cipher/cipher_client/mod.rs
+++ b/crates/bitwarden-vault/src/cipher/cipher_client/mod.rs
@@ -113,6 +113,49 @@ impl CiphersClient {
         })
     }
 
+    /// Encrypt a list of cipher views.
+    ///
+    /// This method attempts to encrypt all ciphers in the list. If any cipher
+    /// fails to encrypt, the entire operation fails and an error is returned.
+    #[cfg(feature = "wasm")]
+    pub fn encrypt_list(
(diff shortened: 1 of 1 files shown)


crates/bitwarden-vault/src/cipher/cipher_client/mod.rs
```diff
@@ -113,6 +113,49 @@ impl CiphersClient {
...
+            .internal
+            .get_flags()
+            .enable_cipher_key_encryption;
+
+        let prepared_views: Vec<CipherView> = cipher_views
+            .into_iter()
+            .map(|mut cv| {
+                if cv.key.is_none() && enable_cipher_key {
+                    let key = cv.key_identifier();
+                    cv.generate_cipher_key(&mut key_store.context(), key)?;
```
nick-livefront: ❓ Is it expensive to create a `key_store.context` for each cipher? Would this be better suited ~~above~~ before the map?

nick-livefront (dismissed):

crates/bitwarden-vault/src/cipher/cipher_client/mod.rs
gbubemismith: Good catch 👏🏼

nick-livefront (approved):

PR #747, opened 2026-02-06
[BEEEP] Add warning notes around data envelope

quexten:
## 📔 Objective

Encstring is dangerous w.r.t. some attacker models. This starts pointing consumers to the safe alternative.


Thomas-Avery (approved):
LGTM

PR #748, opened 2026-02-06
[PM-32349][BEEEP] Allow custom key types in Repository

dani-garcia:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-32349

## 📔 Objective

Add a generic type allowing a `RepositoryItem` to define it's own key type rather than always using String. This key type will need to implement `ToString` internally. This should work on any ID types created with the uuid macro automatically.


dereknance (approved):

Hinton (approved):
Seems reasonable!

PR #757, opened 2026-02-12
[PM-32143] Make it possible to test directly on Feature Clients

Hinton:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-32143

## 📔 Objective

Introduces a `FromClient` and `FromClientPart` which provides nice utilities for feature clients to directly depend on dependencies rather than on the full `Client`. Which allows you to begin testing clients properly without having to mock the full client.


crates/bitwarden-vault/src/vault_client.rs
```diff
@@ -45,9 +45,7 @@ impl VaultClient {
 
     /// Folder related operations.
     pub fn folders(&self) -> FoldersClient {
-        FoldersClient {
-            client: self.client.clone(),
-        }
+        FoldersClient::from_client(&self.client).expect("Folder repository not configured")
```
Hinton: Since we now resolve the repository immediately we have to handle this error at resolve. Not sure if we want to make folders return a result 😬

crates/bitwarden-vault/src/folder/folder_client.rs
```diff
@@ -15,58 +16,81 @@ use crate::{
...
+    }
+
+    /// Creates a FoldersClient from a Client.
+    pub(crate) fn from_client(client: &Client) -> Result<Self, RepositoryError> {
+        Ok(Self {
+            key_store: client.internal.get_key_store().clone(),
+            api_config_provider: Arc::new(client.internal.clone()),
+            repository: client.platform().state().get::<Folder>()?,
+        })
+    }
```
coroiu: I liked @dani-garcia's idea to make more automated: 

```
// Utility trait that we can manually implement to help code generation
impl<T> FromClientPart<Client> for Arc<dyn Repository<T>> {
    fn from_client_part(client: Client) -> Self {
        client.get_repository::<T>()
    }
}
```

Thoughts on adding that too?

crates/bitwarden-vault/src/folder/folder_client.rs
```diff
@@ -15,58 +16,81 @@ use crate::{
 /// Wrapper for folder specific functionality.
 #[cfg_attr(feature = "wasm", wasm_bindgen)]
 pub struct FoldersClient {
-    pub(crate) client: Client,
+    key_store: KeyStore<KeyIds>,
```
coroiu: Putting `KeyIds` everywhere is a bit annoying, but maybe no way around it :/

crates/bitwarden-vault/src/vault_client.rs
dani-garcia: Once we move to an always initialized db for the rehydration work, maybe we can just have the repository function not return a result. I agree with you that having clients return results will be very annoying.

crates/bitwarden-core/src/client/internal.rs
```diff
@@ -443,6 +443,24 @@ impl InternalClient {
...
+    /// Returns API configurations, potentially renewing tokens if needed.
+    async fn get_api_configurations(&self) -> Arc<ApiConfigurations>;
+}
+
+#[async_trait::async_trait]
+impl ApiProvider for Arc<InternalClient> {
+    async fn get_api_configurations(&self) -> Arc<ApiConfigurations> {
+        self.get_api_configurations().await
+    }
+}
```
dani-garcia: I think the reason we need this is because `get_api_configurations` is async and does token renewal right? If so #740 removes the token renewal from that step, allowing the function to be made sync.

Might be worth merging that first to simplify this code.

crates/bitwarden-vault/src/folder/folder_client.rs
dani-garcia: In the past I thought about making a type alias without the generic in core, but it might be more confusing if you now have two places to refer to the type hah.

crates/bitwarden-core/src/client/internal.rs
Hinton: Correct, we currently don't want long lived instances of `ApiConfigurations`. If #740 improves that we should be able to eliminate this trait.

crates/bitwarden-vault/src/folder/folder_client.rs
Hinton: Not against it, it just didn't seem needed to prove out the concept.

coroiu: I mean, I think it's proved already? I just wouldn't want to release it like this and get everyone using it before we improve DX a little bit

crates/bitwarden-core-macro/src/lib.rs
coroiu, line 1: question: Would it be more idiomatic to use a derive macro instead since we're only making additions and technically only implementing a trait?

**suggestion**: Add trait `FromClient`

```rs
trait FromClient {
  fn from_client(client: &Client) -> Result<Self, String>;
}
```

**suggestion**: Use derive macro

```rs
#[derive(FromClient)]
pub struct FoldersClient {
    key_store: KeyStore<KeyIds>,
    api_config_provider: Arc<dyn ApiProvider>,
    repository: Arc<dyn Repository<Folder>>,
}
```

crates/bitwarden-vault/src/folder/folder_client.rs
coroiu: Yeah, let's merge as-is but try to keep it in mind and see if we can solve it somehow in the future

coroiu (commented):
Nice new core-macro crate 👍

coroiu (dismissed):

crates/bitwarden-core-macro/Cargo.toml
```diff
@@ -0,0 +1,26 @@
...
+license-file.workspace = true
+keywords.workspace = true
+
+[lib]
+proc-macro = true
+
+[dependencies]
+proc-macro2 = { workspace = true }
+quote = { workspace = true }
+syn = { workspace = true, features = ["full"] }
```
dani-garcia: Is this "full" feature needed? It seems to build and test just fine without it for me, same for the one in `bitwarden-test-macro`.

CI seems to be passing through 🤷

crates/bitwarden-core/Cargo.toml
```diff
@@ -36,6 +36,7 @@ wasm = [
     "dep:wasm-bindgen-futures",
     "dep:tsify",
 ] # WASM support
+mockall = ["dep:mockall"] # Mocking support for testing
```
dani-garcia: Might also want to enable the mockall flags for the api crates here

dani-garcia (dismissed):
LGTM, just two small non-blocking comments

crates/bitwarden-core-macro/Cargo.toml
Hinton: It seems the move to derive means we can skip it. It's still needed for test-macro though.

Hinton:
I undid the changes to vault since we have to call `expect` on repository which may panic in some clients. An alternative is to have a repository fetcher trait so we expose a method to get a specific repository which can error :shrug:.

coroiu (dismissed):

coroiu (approved):

PR #758, opened 2026-02-12
Pin rust crypto rc versions

Hinton:
## 📔 Objective

Unfortunately it seems even with these pinned we run into dependency issues when regenerating lockfile.


quexten (approved):
Nice, thank you!

Hinton:
Superseeded by https://github.com/bitwarden/sdk-internal/pull/760

PR #759, opened 2026-02-12
[BEEEP] Improve crypto logging

quexten:
## 📔 Objective

Crypto structures have a fair bit of metadata now, but we do not log it. This PR makes crypto logs more readable, and more useful, both with the dangerous crypto logs flag on, and with it off. Structured as one PR to ensure it's consistent.

Here are the sample outputs for different structures, with the dangerous crypto log compile-time flag on: (Note, with it off, the key material for secret keys, private keys, and signing keys would be omitted):

```

--- SAFE
// KDF settings + what key (by id) is contained
PasswordProtectedKeyEnvelope { argon2_iterations: 3, argon2_memory_kib: 65536, argon2_parallelism: 4, contained_key_id: KeyId(16b266708b6975b025c1e412cd77fa4e) }
// Show namespace + cek key id
DataEnvelope { namespace: ExampleNamespace, key_id: KeyId(f7507ad0993c3ad2cc06170bb8c765b0) }

--- EncString
// Show keyid + alg
EncString::Cose_Encrypt0 { algorithm: "XChaCha20-Poly1305", key_id: KeyId(00000000000000000000000000000000), nonce: "2bdc6ab8853d8e513e76aed4c0da477b7479a49a8c1f0787", content_type: Text("application/x.bitwarden.utf8-padded"), ciphertext: "d926dd946cdc0c3bc0d0631612b97200c6cef2c1f83caa617ada21127c9274270eb86864f3820be085f3d46cb491211f" }

EncString::Aes256Cbc_HmacSha256_B64 { iv: "a8a5f1cebba79bbba486668456fb8f92", data: "04204657f3cece608e1cc37c4acb8d5e336b21bff927eaedf6414fec6f7c4b67", mac: "9ea970c7b367d47f61e8d68e4f59fa7e5ad745815b29571bdccc8617c71fdfdc" }

--- Keys
SymmetricKey_Aes256CbcHmac { enc_key: "bece1c490c37e33802b3015753a9af2920902705e97fe31974e18b930b61f047", mac_key: "062896cc4c39f07a6c1a77cd5dde5b7eb92aef6f9f399e180e44a26669514167" }

SymmetricKey_XChaCha20Poly1305 { key_id: KeyId(a8fcd1075e34d90d681c644fa1498dc0), key: "ccd995f1cc386f26cdb00618910755872d6b18340be7525a8300886ad5065f52" }

SigningKey_Ed25519 { id: KeyId(88c29941137f7858ad2f98ac7c427a21), key: "43f6e95d5efc52416d83cb2172a9b03571abcffe73c27eeb503ea1fd383ae2b8" }

VerifyingKey_Ed25519 { id: KeyId(88c29941137f7858ad2f98ac7c427a21), key: "6b4e0e200d5ada19b7399ec29cfed3237364ca11d2bf519f52060402d691ec23" }

// Shows which signature key it was signed by
SignedPublicKey { signed_by: KeyId(3c3a3beeb01d4b045e8d6f647c5b107d), algorithm: RsaOaepSha1, public_key: "30820122300d06092a864886f70d01010105000382010f003082010a0282010100c17f1d9a241f3e50b6fae5bed32611f8280fb2a8c897c35d859e86e65325230b770e8b696615ffb0af5992270dd727dabbc4cab4907c69eec9881b49edc230bdddf691924e26f37dface2a4b5410e9da46cc299394832e639d1087b3c4bb566cda78ac12e8eb61c6be750cb0fc456b28e6470c7b819ac61c058bec1062d3c22330e2b64cdabd0c9f86cb55b7487cb14abd5f536dcd0e1620b9d5c991f649390a0831b97e69c73061fed869668ac8cc01aedb787de485735437fd500e2ffbb07c5f49c32b8d5a6c62f73dc91f4de30acbe1eaf0b8e87faa4009b1f0158aa2c3a74cfd3073e59def58c4485e3d104241b0a878799d29b0241412dd0ade7a34cb2d0203010001" }

PrivateKey_RsaOaepSha1 { key: "308204bd020100300d06092a864886f70d0101010500048204a7308204a30201000282010100db05f8b5cb6decd17069840122b19845cb0c988ca625bb0a9ceecf6e480d678fd70bd7a9810e08df39ca56d0ad81c40decb472ec7d4906c21194a4719f21f832a7383331b68d4f9e24a9cb110616cfa3059431cb1b05491517b8190f1019974ea7ce4529cc6284e25d5f4589f12b9ff723b2989376f8e9a0c03935f3406ba5184db0be7c293907607d0dbd8a25ef22464ee208ddb927ffe74a028e2cc925de8a6d9da7d10318fd5ca54beee2c216215ef9ec9c64324c5d983524dcdca954970957177d36490c6f2b0e2dfba352a7256956c3c602fe4701b862253a0745608efab64a158514c45ebb914b94741e19119f36fe664f602bf0bf018b7ec562ef197f02030100010282010012677f0dafa96e03ef613e7165fbec62b8fc36f962a0bb05da480af02667d1f2f94c0a7438def809e1ee49bb41492d334a57a36c91de880dd8b257db20571767123c1bb92ab0d726dfc7d0c37e727e42219ee14c42d9ffd920e41c5ef2a863d97d2a4d0f851826386b462d2eb9304441dd230dc51e732e168ef07fc30fffbba0964c312edc177de197499c25f897eb9cc3d5656bad2be2e355c59ffd457babd0e31c55fde91b735248b4674ba4a325137e79535137d29049d4287d50180ade3bd1e5c322bf443d249152eb828e2a057b545fb572943f148bc202706832e73f3a1b693a833bc8af9f689c93bbde3a3eae4b5b5b31e7347fc725db6c72ccba7f1902818100fd36b3c8d452b57ca72b2db878389035131d4c4747e0e990eecf4991b7dca488453c5c1168e434336eafe00b8d5cca19677843e19277def940b24a1bd1755f4adf2c95a1e56c541f72ee63ed60e075d6d167bb197de98f90caa3eb34c916e299e7034c30938aba77c84fd641add917c5071783e30004aa541ab0057c235e051302818100dd6ef4a9e1d46c85d515dcc5f9dcdd9f465c2175a462ee05ff4c79830f02c934ec969c64050fb75e7c42b6189d8391d096465582865d9fbf55c669a1ddeabbc792da42cefd5b31a58c1e2155a3d699c103916accc980f899bc7fc1261993a081f8a23a80b30fecaf9f5512d3ef0344e70fe39ad288abb71de4168b2c4774a3650281805a3c73475f7af6182d0eb325691a6731b2d51f08cbf1b7eaa160178ca96424f23712329135bb156b3ac7cc64f1849eb4fced9beb382225c0d7b071be861f78cd1ac92daddb2f86af0a6b300b3919e484a06d81fcaca55e37302fc83330724a60a435c5a20e02433f4a9040ff78efc38bc0623dead867d13755a762dd0151912702818017e6bc1e02344daa7c7fa99eed037c208f15aebee6098fabb20ced28ca5c009fe7d1b1fbd1d80a70f674b4f71f396e69d76eac1834acb26b92b4a1fe48d7e780f946f9e321d1d0201069c82b584df7b6afcde87b465cd778b0b211852d49940fb368bf6bd8e8252227cea5039f4ef906ffde0f7e1e267ad38d2529cc93c1f14902818100a70dca0076309b01ac2725b6124e5cf332a70b85e2c89f78fddfcc11489d820d49b637331d158d77381bf2b5f7c6595242c39635f8acf515822acd179507707e95311c0276812255d02529cf7a12b87ce313e9782783d18fa4773750d4857f2a0795c5c0169536257f6d9085ceff7238dad9c2ac9632cbd7d6e82239da27c789" }

PublicKey_RsaOaepSha1 { key: "30820122300d06092a864886f70d01010105000382010f003082010a0282010100db05f8b5cb6decd17069840122b19845cb0c988ca625bb0a9ceecf6e480d678fd70bd7a9810e08df39ca56d0ad81c40decb472ec7d4906c21194a4719f21f832a7383331b68d4f9e24a9cb110616cfa3059431cb1b05491517b8190f1019974ea7ce4529cc6284e25d5f4589f12b9ff723b2989376f8e9a0c03935f3406ba5184db0be7c293907607d0dbd8a25ef22464ee208ddb927ffe74a028e2cc925de8a6d9da7d10318fd5ca54beee2c216215ef9ec9c64324c5d983524dcdca954970957177d36490c6f2b0e2dfba352a7256956c3c602fe4701b862253a0745608efab64a158514c45ebb914b94741e19119f36fe664f602bf0bf018b7ec562ef197f0203010001" }
--- Account Crypto
SignedSecurityState { signed_by: Ok(KeyId(96c3363f91dcec1847367629e72764ff)), version: 2 }

// No key id's because this only contains *wrapped* keys.
WrappedAccountCryptographicState::V1
WrappedAccountCryptographicState::V2 { security_state: SignedSecurityState { signed_by: Ok(KeyId(ecf0639ca6b35058b750cdeb45ea084e)), version: 2 } }

```

Aside from improving the struct's logging, it also logs the crypto keys used during init *should the dangerous crypto debug flag be active*.


crates/bitwarden-core/src/key_management/account_cryptographic_state.rs
```diff
@@ -111,8 +111,14 @@ impl std::fmt::Debug for WrappedAccountCryptographicState {
             WrappedAccountCryptographicState::V1 { .. } => f
                 .debug_struct("WrappedAccountCryptographicState::V1")
                 .finish(),
-            WrappedAccountCryptographicState::V2 { .. } => f
+            WrappedAccountCryptographicState::V2 {
+                security_state,
+                private_key: _,
+                signed_public_key: _,
+                signing_key: _,
+            } => f
```
dani-garcia: You can also do `WrappedAccountCryptographicState::V2 { security_state, .. } =>` if you want to ignore all the other fields.

crates/bitwarden-core/src/key_management/account_cryptographic_state.rs
```diff
@@ -425,6 +481,20 @@ mod tests {
     use super::*;
     use crate::key_management::{PrivateKeyId, SigningKeyId, SymmetricKeyId};
 
+    #[test]
+    #[ignore = "Manual test to verify debug format"]
+    fn test_debug() {
```
dani-garcia, line 553: We could change these to automatically check the debug format, something like could be helpful:
```rust
let v2 = format!("{v2:?}");
assert!(!v2.contains("private_key"));
assert!(!v2.contains("signed_public_key"));
assert!(!v2.contains("signing_key"));
```

crates/bitwarden-crypto/src/keys/key_id.rs
```diff
@@ -1,16 +1,21 @@
...
 
 /// A key id is a unique identifier for a single key. There is a 1:1 mapping between key ID and key
 /// bytes, so something like a user key rotation is replacing the key with ID A with a new key with
 /// ID B.
-#[derive(Clone, PartialEq)]
-pub struct KeyId(Uuid);
+#[derive(Clone, PartialEq, Zeroize)]
+pub struct KeyId([u8; KEY_ID_SIZE]);
+
+impl ConstantTimeEq for KeyId {
```
dani-garcia, line 17: Is the `ConstantTimeEq` implementation here due to some security consideration? Because if so, we probably want to replace the PartialEq derive by a manual implementation that is also constant time.

crates/bitwarden-crypto/src/signing/signed_object.rs
```diff
@@ -33,6 +34,31 @@ impl SignedObject {
...
+            .clone()
+            .try_into()
+            .map_err(|_| SignatureError::InvalidSignature)?;
+        Ok(KeyId::from(key_id))
+    }
+
+    /// Deserializes the payload without verifying the signature.
+    /// This is intended for debug logging only, to debug structs without having to explicitly
+    /// verify their signature first. WARNING: YOU MUST NOT USE THIS EXCEPT FOR DEBUG LOGGING.
+    pub fn dangerous_unverified_decode_do_not_use_except_for_debug_logs<
```
dani-garcia, line 57: Might be worth it to mark it as `#[doc(hidden)]` which will ensure it doesn't appear in the docs or the autocomplete suggestions.

crates/bitwarden-crypto/Cargo.toml
```diff
@@ -44,6 +44,7 @@ ciborium = { version = ">=0.2.2, <0.3" }
 coset = { version = ">=0.4.0, <0.5" }
 ed25519-dalek = { workspace = true, features = ["rand_core"] }
 generic-array = { version = ">=0.14.7, <1.0", features = ["zeroize"] }
+hex = "0.4.3"
```
dani-garcia: We should probably use a version range here to match with the rest.

crates/bitwarden-crypto/src/signing/verifying_key.rs
```diff
@@ -32,6 +32,22 @@ pub struct VerifyingKey {
...
+impl std::fmt::Debug for VerifyingKey {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        let key_suffix = match &self.inner {
+            RawVerifyingKey::Ed25519(_) => "Ed25519",
+        };
+        let mut debug_struct = f.debug_struct(format!("VerifyingKey_{}", key_suffix).as_str());
+        debug_struct.field("id", &self.id);
+        match self.inner {
+            RawVerifyingKey::Ed25519(_key) => {
+                debug_struct.field("key", &hex::encode(_key.to_bytes()));
```
dani-garcia: This is actually used, so no need for the underscore.
```suggestion
            RawVerifyingKey::Ed25519(key) => {
                debug_struct.field("key", &hex::encode(key.to_bytes()));
```

crates/bitwarden-crypto/src/keys/public_key_encryption.rs
```diff
@@ -60,6 +60,23 @@ pub struct PublicKey {
     inner: RawPublicKey,
 }
 
+impl std::fmt::Debug for PublicKey {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        let key_suffix = match &self.inner {
+            RawPublicKey::RsaOaepSha1(_) => "RsaOaepSha1",
+        };
+        let mut debug_struct = f.debug_struct(format!("PublicKey_{}", key_suffix).as_str());
```
dani-garcia: Sometimes we're using `::` as a separator in the debug types (like in `EncString::Aes256Cbc_HmacSha256_B64`), in other cases like these we use `_`, we should probably use the same one everywhere.

crates/bitwarden-core/src/key_management/account_cryptographic_state.rs
```diff
@@ -357,14 +363,30 @@ impl WrappedAccountCryptographicState {
...
+                        .dangerous_get_symmetric_key(SymmetricKeyId::User)
+                        .expect("User key should be set");
+                    let private_key = ctx
+                        .dangerous_get_private_key(PrivateKeyId::UserPrivateKey)
+                        .ok();
+                    let public_key = ctx.get_public_key(PrivateKeyId::UserPrivateKey).ok();
+                    info!(
+                        "V1 account cryptographic state set to context. UserKey: {:?}, PrivateKey: {:?}, PublicKey: {:?}",
+                        user_key, private_key, public_key
+                    );
```
dani-garcia, line 448: By using the structured field syntax we avoid some name repetition here, same for V2 below:
```suggestion
                    info!(
                        ?user_key, ?private_key, ?public_key,
                        "V1 account cryptographic state set to context"
                    );
```

dani-garcia (commented):
Looks reasonable to me, just some minor comments

crates/bitwarden-crypto/src/keys/key_id.rs
quexten: The key ID is public, so non-ct-comparison it is fine. However, we implement ct_eq for any cose key (such as the XChaCha20Poly1305 key). This checks both the key material and the key id. To keep the contract of ct_eq in implementing that, I just implemented ct_eq for KeyID for specifically this purpose.

So, I don't think we need to replace the PartialEq here.

crates/bitwarden-crypto/src/signing/signed_object.rs
quexten: Nice, good callout. Applied.

crates/bitwarden-crypto/src/keys/key_id.rs
dani-garcia: Sounds reasonable, might want to add a comment to it so we don't forget why it's done.

crates/bitwarden-crypto/src/signing/verifying_key.rs
quexten: Seems accidental, good catch.

crates/bitwarden-crypto/src/keys/public_key_encryption.rs
quexten: Hm, makes sense. However, the meaning here is different; AES256CBC-HMAC256-B64 is the algorithm name, EncString is the container type, so it feels like having different separators is correct?

I've just renamed them to "Aes256CbcHmac256" / "CoseEncrypt0" so that we only have one separator.

crates/bitwarden-core/src/key_management/account_cryptographic_state.rs
```diff
@@ -385,13 +404,47 @@ impl WrappedAccountCryptographicState {
...
+                    info!(
+                        "V2 account cryptographic state set to context. UserKey: {:?}, PrivateKey: {:?}, SigningKey: {:?}, VerifyingKey: {:?}, PublicKey: {:?}, SignedPublicKey: {:?}, SecurityState: {:?}",
+                        user_key,
+                        private_key,
+                        signing_key,
+                        verifying_key,
+                        public_key,
+                        signed_public_key,
+                        security_state
+                    );
```
mzieniukbw, line 515: Could we use structured field syntax here too ?

crates/bitwarden-crypto/src/keys/symmetric_crypto_key.rs
```diff
@@ -397,45 +393,48 @@ impl CryptoKey for SymmetricCryptoKey {}
...
+            .field("enc_key", &hex::encode(self.enc_key.as_slice()))
+            .field("mac_key", &hex::encode(self.mac_key.as_slice()));
         debug_struct.finish()
     }
 }
 
 impl std::fmt::Debug for XChaCha20Poly1305Key {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        let mut debug_struct = f.debug_struct("XChaCha20Poly1305Key");
+        let mut debug_struct = f.debug_struct("SymmetricKey_XChaCha20Poly1305");
```
mzieniukbw: Let's keep it consistent:
```suggestion
        let mut debug_struct = f.debug_struct("SymmetricKey::XChaCha20Poly1305");
```

crates/bitwarden-crypto/src/keys/public_key_encryption.rs
```diff
@@ -222,13 +239,14 @@ impl PrivateKey {
...
+            RawPrivateKey::RsaOaepSha1(_) => "RsaOaepSha1",
+        };
+        let mut debug_struct = f.debug_struct(format!("PrivateKey::{}", key_suffix).as_str());
         match &self.inner {
-            RawPrivateKey::RsaOaepSha1(_key) => {
-                debug_struct.field("algorithm", &"RsaOaepSha1");
+            RawPrivateKey::RsaOaepSha1(_) => {
                 if let Ok(der) = self.to_der() {
-                    debug_struct.field("private_key_der", &der.as_ref());
+                    debug_struct.field("key", &hex::encode(der.as_ref()));
```
mzieniukbw, line 259: ❓ Do we really want to expose it like that ? Shouldn't we use `#[cfg(feature = "dangerous-crypto-debug")]` ?

mzieniukbw (changes requested):

crates/bitwarden-crypto/src/keys/public_key_encryption.rs
quexten: Oops. Yeah, that definitely should *NOT* be logged. Good catch.

dani-garcia (approved):

mzieniukbw (approved):

PR #770, opened 2026-02-16
[PM-32439] Consolidate safe module cryptographic namespacing strategy

quexten:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-32439

## 📔 Objective

To solve a class of cryptographic issues plaguing our e2ee layer, we need cryptographic namespaces. We already had an initial stab at this for signatures and for data envelopes. To make this approach scalable, we need a more systematic approach. Specifically, the problem to solve here is that encrypted/signed objects, encrypted/signed under the same key can be swapped by the server, causing the client to mis-interpret these, leading to security bugs. Namespaces solve this by creating a cryptographic partitioning (by adding a namespace number to the protected headers, in the signed data / authenticated additional data).

This PR introduces this systematic approach for these. Signatures have their own namespace pratitioning, since they are a lower-level primitive. 

Safe objects have two layers: the object layer and the content layer.

On the object layer we partition the cryptographic primitives (Encrypt0, Sign1) into safe primitives (DataEnvelope, PasswordProtectedKeyEnvelope, etc). 

On the content layer, we partition each of the above subsets further, by use-case. For example, a valid partitioning for PasswordProtectedKeyEnvelope would be {MasterPassword, Pin}.

This prevents a class of attacks where the server switches cryptographic objects, signed/encrypted under the same key. This eliminates any kind of cryptographic analysis that would be needed to prevent the above attacks / limits security analysis to each sub-namespace spanned by the combination of the object and content namespace layer.


crates/bitwarden-crypto/src/cose.rs
```diff
@@ -46,11 +46,30 @@ pub(crate) const CONTENT_TYPE_PADDED_CBOR: &str = "application/x.bitwarden.cbor-
 const CONTENT_TYPE_BITWARDEN_LEGACY_KEY: &str = "application/x.bitwarden.legacy-key";
 const CONTENT_TYPE_SPKI_PUBLIC_KEY: &str = "application/x.bitwarden.spki-public-key";
 
-/// Namespaces
-/// The label used for the namespace ensuring strong domain separation when using signatures.
-pub(crate) const SIGNING_NAMESPACE: i64 = -80000;
-/// The label used for the namespace ensuring strong domain separation when using data envelopes.
-pub(crate) const DATA_ENVELOPE_NAMESPACE: i64 = -80001;
+// The label used for defining the namespace of a signed object
+pub(crate) const SIGNING_NAMESPACE: i64 = 80000;
```
quexten: Keeping signing namespace a separate layer. It spans a separate namespace that does not just include safe objects.

crates/bitwarden-crypto/src/safe/password_protected_key_envelope.rs
```diff
@@ -174,6 +187,30 @@ impl PasswordProtectedKeyEnvelope {
             ));
         }
 
+        // Validate the object namespace, if present
+        //
+        // The first use-case - Pin-protected-key-envelopes - did not require the object
+        // namespace to be present. Therefore, without migration of persistent pin
+        // unlocks, this cannot yet be strongly enforced.
+        if let Ok(namespace) = extract_integer(
```
dani-garcia: This seems very similar to the existing `extract_safe_object_namespace` function in data_envelope, maybe we can use that here too?

crates/bitwarden-crypto/src/safe/password_protected_key_envelope.rs
```diff
@@ -174,6 +187,30 @@ impl PasswordProtectedKeyEnvelope {
...
+            &self.cose_encrypt.protected.header,
+            SAFE_OBJECT_NAMESPACE,
+            "safe object namespace",
+        ) && namespace != i128::from(SafeObjectNamespace::PasswordProtectedKeyEnvelope as i64)
```
dani-garcia: It might be a good idea to implement `PartialEq` for the namespace types so we don't need all this numeric conversion every time:
```rust
impl PartialEq<i128> for SafeObjectNamespace {
    fn eq(&self, other: &i128) -> bool {
        *self as i128 == *other
    }
}

impl PartialEq<SafeObjectNamespace> for i128 {
    fn eq(&self, other: &SafeObjectNamespace) -> bool {
        *other as i128 == *self
    }
}
```

Then this line can be simplified as:
```rust
        ) && namespace !=  SafeObjectNamespace::PasswordProtectedKeyEnvelope
```

I'd also recommend adding an `as_i64()` function to `SafeObjectNamespace` like we have for `PasswordProtectedKeyEnvelopeNamespace`. Using `as type` numeric conversions can cause accidental truncation, so it's best to avoid them when possible.

dani-garcia (dismissed):
LGTM, just two small improvements that we may be able to make

crates/bitwarden-crypto/src/safe/password_protected_key_envelope.rs
quexten: Good point, I made two helpers, one for extracting the content namespace, one for extracting the object namespace.

crates/bitwarden-crypto/src/safe/password_protected_key_envelope.rs
quexten: Instead of working on i128, I've just made it so the extraction helpers return the enum. Further, the content namespace enums now need to implement a marker trait, which the extraction helper has a generic parameter for. 

This should make the code much cleaner, thanks for nudging into this direction!

crates/bitwarden-crypto/src/safe/password_protected_key_envelope_namespace.rs
```diff
@@ -0,0 +1,55 @@
...
+        let Ok(value) = i64::try_from(value) else {
+            return Err(PasswordProtectedKeyEnvelopeError::InvalidNamespace);
+        };
+        Self::try_from(value)
+    }
+}
+
+impl From<PasswordProtectedKeyEnvelopeNamespace> for i128 {
+    fn from(val: PasswordProtectedKeyEnvelopeNamespace) -> Self {
+        val.as_i64() as i128
```
dani-garcia: When widening the integer size, you can just into() directly as it can't fail:
```suggestion
        val.as_i64().into()
```

crates/bitwarden-crypto/src/safe/password_protected_key_envelope_namespace.rs
```diff
@@ -0,0 +1,55 @@
...
+            -2 => Ok(PasswordProtectedKeyEnvelopeNamespace::ExampleNamespace2),
+            _ => Err(PasswordProtectedKeyEnvelopeError::InvalidNamespace),
+        }
+    }
+}
+
+impl TryFrom<i128> for PasswordProtectedKeyEnvelopeNamespace {
+    type Error = PasswordProtectedKeyEnvelopeError;
+
+    fn try_from(value: i128) -> Result<Self, Self::Error> {
```
dani-garcia: If you invert the implementations so that `TryFrom<i128>` is the actual implementation and `TryFrom<i64>` calls it, you don't need to do the whole `let Ok(value) = i64::try_from(value) else {... };` and you can just use `from`:

```rust
impl TryFrom<i128> for DataEnvelopeNamespace {
    type Error = DataEnvelopeError;

    fn try_from(value: i128) -> Result<Self, Self::Error> {
        match value {
            1 => ...
        }
    }
}

impl TryFrom<i64> for DataEnvelopeNamespace {
    type Error = DataEnvelopeError;

    fn try_from(value: i64) -> Result<Self, Self::Error> {
        Self::try_from(i128::from(value))
    }
}
```

dani-garcia (dismissed):
LGTM, the extraction helpers improved the code a lot! I have two small non-blocking improvements.

crates/bitwarden-crypto/src/safe/password_protected_key_envelope_namespace.rs
quexten: It's a bit unfortunate that we even need the i64, but `Integer` does not implement `from(i128)`, only `try_from(i128)`, even though it is defined as `Integer(i128)`...

quexten:
@dani-garcia I took some more time to extract the helpers into a separate file and make the interface nicer. On the safe object implementations, you should now only call "set" or "validate", and do not have to extract them / compare them yourself. Should be much improved now!

dani-garcia (dismissed):

crates/bitwarden-crypto/src/safe/helpers.rs
```diff
@@ -0,0 +1,222 @@
...
+    );
+}
+
+pub(super) fn validate_safe_namespaces<T: ContentNamespace>(
+    header: &coset::Header,
+    expected_object_namespace: SafeObjectNamespace,
+    expected_content_namespace: T,
+) -> Result<(), ExtractionError> {
+    let obj = extract_safe_object_namespace(header);
+    println!("Object namespace extraction result: {obj:?}");
```
Thomas-Avery: Should this be trace debug or just removed?

crates/bitwarden-crypto/src/safe/helpers.rs
Thomas-Avery, line 1: Nice job on these helpers, it helped 😉  me wrap my head around this.

crates/bitwarden-crypto/src/safe/README.md
```diff
@@ -28,3 +28,33 @@ Use the data envelope to protect a struct (document) of data. Examples include:
...
 Calling the API with a decrypted struct, the content encryption key ID and the encrypted data are
 returned.
+
+## Guidelines for devs
+
+When adding a new primitive there are a few considerations to make:
+
+- Does this serve a new purpose that is not already fulfilled better otherwise
+  - Replacing insecure functionality is a valid reason here
+- Is it easy to use for developers that are not cryptography exprets
```
```suggestion
- Is it easy to use for developers that are not cryptography experts
```

crates/bitwarden-crypto/src/safe/README.md
```diff
@@ -28,3 +28,33 @@ Use the data envelope to protect a struct (document) of data. Examples include:
 The serialization of the data and the creation of a content encryption key is handled internally.
 Calling the API with a decrypted struct, the content encryption key ID and the encrypted data are
 returned.
+
+## Guidelines for devs
```
```suggestion
## Guidelines for developers
```

⛏️

crates/bitwarden-crypto/src/safe/README.md
```diff
@@ -28,3 +28,33 @@ Use the data envelope to protect a struct (document) of data. Examples include:
...
+
+When adding a new primitive there are a few considerations to make:
+
+- Does this serve a new purpose that is not already fulfilled better otherwise
+  - Replacing insecure functionality is a valid reason here
+- Is it easy to use for developers that are not cryptography exprets
+- Does the API prevent (accidental) mis-use by developers that are not cryptography experts
```
Thomas-Avery, line 39:
```suggestion
- Does the API prevent (accidental) mis-use by developers that are not cryptography experts
  - Avoid exposing many settings/configurations to API consumers
```
Suggestion non blocking.

crates/bitwarden-crypto/src/safe/README.md
```diff
@@ -28,3 +28,33 @@ Use the data envelope to protect a struct (document) of data. Examples include:
...
+may misinterpret these, leading to security vulnerabilities. Analysis here is complex. Therefore, a
+simpler approach is strong cryptographic namespace separation, which prevents this by adding
+metadata about where an object appropriately can be used.
+
+The namespace partitioning happens in two layers, the object layer and the content layer. The
+authenticated data of each safe object contains a object namespace key value pair, which allows the
+decrypting code to correctly identify whether an object that is being decrypted is actually the
+correct object type (e.g. a DataEnvelope). Within each object type, there is then another layer of
+partitioning, since these objects can be used in many places. For instance, a DataEnvelope may have
+the partitioning vault item, account settings, and so on.
```
```suggestion
correct object type (e.g. a DataEnvelope). For each object type, there is then another layer of
partitioning, since these objects can be used in many places. For instance, a DataEnvelope may have
the partitioning vault item, account settings, and so on. This content namespace is also stored in the authenticated data of the safe object.
```
Suggestion non-blocking

crates/bitwarden-crypto/src/cose.rs
```diff
@@ -46,11 +46,52 @@ pub(crate) const CONTENT_TYPE_PADDED_CBOR: &str = "application/x.bitwarden.cbor-
...
-/// Namespaces
 /// The label used for the namespace ensuring strong domain separation when using signatures.
 pub(crate) const SIGNING_NAMESPACE: i64 = -80000;
-/// The label used for the namespace ensuring strong domain separation when using data envelopes.
-pub(crate) const DATA_ENVELOPE_NAMESPACE: i64 = -80001;
+
+// Domain separation / Namespaces
+//
+// Cryptographic objects are strongly domain separated so that items can only be decrypted
+// in the correct context, making cryptographic analylsis significantly easier and preventing
```
Thomas-Avery:
```suggestion
// in the correct context, making cryptographic analysis significantly easier and preventing
```

crates/bitwarden-core/src/client/internal.rs
```diff
@@ -277,9 +277,11 @@ impl InternalClient {
         let decrypted_user_key = {
             // Note: This block ensures ctx is dropped. Otherwise it would cause a deadlock when
             // initializing the user crypto
```
Thomas-Avery: Should this comment be moved to be over the `ctx` again?

crates/bitwarden-crypto/src/safe/data_envelope.rs
```diff
@@ -505,6 +482,62 @@ macro_rules! generate_versioned_sealable {
...
+    #[cfg(test)]
+    ExampleNamespace = -1,
+    /// This namespace is only used in tests
+    #[cfg(test)]
+    ExampleNamespace2 = -2,
+}
+
+impl DataEnvelopeNamespace {
+    /// Returns the numeric value of the namespace.
+    pub fn as_i64(&self) -> i64 {
```
Thomas-Avery:
```suggestion
    fn as_i64(&self) -> i64 {
```

💭  It doesn't look like this needs to be public anymore?

crates/bitwarden-crypto/src/safe/password_protected_key_envelope.rs
```diff
@@ -490,6 +515,57 @@ impl FromWasmAbi for PasswordProtectedKeyEnvelope {
...
+    #[cfg(test)]
+    ExampleNamespace = -1,
+    /// This namespace is only used in tests
+    #[cfg(test)]
+    ExampleNamespace2 = -2,
+}
+
+impl PasswordProtectedKeyEnvelopeNamespace {
+    /// Returns the numeric value of the namespace.
+    pub fn as_i64(&self) -> i64 {
```
Thomas-Avery:
```suggestion
   fn as_i64(&self) -> i64 {
```

Same here.

crates/bitwarden-crypto/src/cose.rs
```diff
@@ -46,11 +46,52 @@ pub(crate) const CONTENT_TYPE_PADDED_CBOR: &str = "application/x.bitwarden.cbor-
...
+// the object types are partitioned into e.g. EncString, DataEnvelope, Signature, KeyEnvelope, and
+// so on. Second, within each of these types, each of these spans their own namespace for usages.
+// For instance, a DataEnvelope may describe that the contained item is only valid as a vault item,
+// or as account settings.
+
+/// MUST be placed in the protected header of cose objects
+pub(crate) const SAFE_OBJECT_NAMESPACE: i64 = -80002;
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum SafeObjectNamespace {
```
Thomas-Avery: 💭  I would think both `SafeObjectNamespace` and `ContentNamespace` would be `pub(crate)`? Maybe I'm missing something.

Thomas-Avery (changes requested):
Looks good, a few minor things to look at from me.

crates/bitwarden-crypto/src/safe/helpers.rs
quexten: Nice. It usually needs a few revisions for new API design to find a good abstraction, I think this is validation that this is the right direction.

crates/bitwarden-crypto/src/safe/helpers.rs
quexten: Oops, yeah we do not want this. It's left over from debugging

crates/bitwarden-core/src/client/internal.rs
quexten: Moved to outside of the block actually, but yeah, it definitely was not in the right place.

crates/bitwarden-crypto/src/cose.rs
quexten: You are right. Changed!

quexten:
Had to resolve some conflicts after the improved logging PR merged. We added the namespaces to the logging of the dataenvelope and passwordprotectedkeyenvelope structs now, so you can easily see them while debugging.

crates/bitwarden-crypto/src/safe/helpers.rs
dani-garcia: This reminded me that there's a lint to disable println!, so I've made a separate PR to enable it: https://github.com/bitwarden/sdk-internal/pull/795

Thomas-Avery (approved):
LGTM

dani-garcia (approved):

PR #773, opened 2026-02-17
[PM-30222] Add tryfrom privateaccountkeysresponsemodel

quexten:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-30222

## 📔 Objective

Add conversion function.


Thomas-Avery:
Can we update this to include using it in `crates/bitwarden-user-crypto-management/src/key_rotation/sync.rs` or at least make a ticket to track that.

https://github.com/bitwarden/sdk-internal/blob/51197593f68dde271ff3351630dd1429d31e2344/crates/bitwarden-user-crypto-management/src/key_rotation/sync.rs#L310

quexten:
@Thomas-Avery Migrated the sync caller, thank you for the callout. I don't think a ticket would have been worth it for a change as tiny as this.

eligrubb (approved):
⭐⭐⭐⭐⭐

PR #777, opened 2026-02-18
[PM-32510] Add symmetric key envelope

quexten:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-32510

## 📔 Objective

Adds a new safe primitive. Given a *COSE* symmetric key, protects another symmetric key (cose or not cose). The metadata provides the contained key id (optional) and the sealing key id.

The content namespace currently contains just "SessionKey".


Hinton:
@quexten with COSE symmetric key does that mean the b64 encoded session key will be longer than the existing one?

quexten:
@Hinton Typically a cose encoded symmetric key is ~68 bytes (+33% Base64 overhead), so a tiny amount larger than the existing AES key.

Hinton:
That seems fine to me. While it won't be compatible with existing session keys I don't believe we've every publicly stated that they are long term supported. The new key would be a `XChaCha20Poly1305Key` rather than AES?

quexten:
Yep. The one thing to watch out for is that while the contained key can be aes-cbc-hmac or cose (xchacha), the wrapping key must always be cose (currently xchacha).

I believe there are some demands for FIPS compatible algorithms, at which point we may add AES-GCM as a second cose AEAD algorithm. But this first requires some refactoring in the SDK to not be dependent on XChaCha specifically but on "CoseKey" which can contain either.

crates/bitwarden-crypto/src/cose.rs
```diff
@@ -95,6 +101,48 @@ pub(crate) const SAFE_CONTENT_NAMESPACE: i64 = -80001;
 
 const XCHACHA20_TEXT_PAD_BLOCK_SIZE: usize = 32;
 
+/// Encrypt a plaintext message with a given key
+pub(crate) fn encrypt_cose(
```
quexten, line 107: Two notes:
- We should later extract XChaCha20Poly1305 key into an enum - "CoseKey" - that could later be used for AES-GCM if we go to FIPS compatibility (or any other reason to switch symmetric algorithms). That way we have one implementation. (Maybe this should actually be generic with an associated type so we can get the correct nonce size).
- We can then move this to be used in a bunch of places, which should make the rest of the code much cleaner and safer

crates/bitwarden-crypto/src/safe/symmetric_key_envelope.rs
```diff
@@ -0,0 +1,567 @@
...
+        use wasm_bindgen::UnwrapThrowExt;
+        let string = unsafe { String::from_abi(abi) };
+        SymmetrickeyEnvelope::from_str(&string).unwrap_throw()
+    }
+}
+
+#[allow(clippy::enum_variant_names)]
+#[derive(Copy, Clone, PartialEq)]
+pub enum SymmetricKeyEnvelopeNamespace {
+    /// A key used for re-hydration of the SDK
```
Hinton, line 281: Do we want to document that this is for the CLI? I don't believe session keys are used elsewhere.

Hinton (dismissed):
Looks nice. Just to make sure I'm on the same page, the benefit of using this over say `DataEnvelope` is that it's standardized and contains the wrapping key id.

crates/bitwarden-crypto/src/safe/symmetric_key_envelope.rs
quexten: Platform wanted to use a session key for SDK re-hydration. I'm not sure if that should be a different namespace entry though, as i don't understand the use-case enough. @dani-garcia do you have any thoughts on this?

dani-garcia: The session key so far is a CLI-only concern yeah. The reason we use it is that the CLI needs to have this key provided on the terminal as an environment variable where it's easy to accidentally exfiltrate, and we don't want to use user key/password for that. 

I can't think of any uses for it outside the CLI at the moment, I guess the biometric key or whatever is done for the "Never Lock" option would be very similar, but those seem deserving of a separate namespace, so we can probably document this one as CLI-only, sure.

crates/bitwarden-crypto/src/safe/symmetric_key_envelope.rs
```diff
@@ -0,0 +1,572 @@
...
+    /// The wrong unseal method was used for the key type
+    #[error("Wrong key type")]
+    WrongKeyType,
+    /// The symmetric key envelope namespace is invalid.
+    #[error("Invalid namespace")]
+    InvalidNamespace,
+}
+
+/// A symmetric key protected by another symmetric key
+pub struct SymmetrickeyEnvelope {
```
mzieniukbw: typo
```suggestion
pub struct SymmetricKeyEnvelope {
```

mzieniukbw (changes requested):

mzieniukbw (approved):

PR #783, opened 2026-02-19
[PM-25820] Migrate Cipher Attachment Operations to SDK

gbubemismith:
## 🎟️ Tracking
https://bitwarden.atlassian.net/browse/PM-25820

## 📔 Objective
Migrate attachment-related API operations from TypeScript to SDK.

### DIFF
--- a/crates/bitwarden-vault/src/cipher/cipher_client/admin/delete_attachment.rs
+++ b/crates/bitwarden-vault/src/cipher/cipher_client/admin/delete_attachment.rs
@@ -1,0 +1,76 @@
+use bitwarden_core::ApiError;
+use bitwarden_error::bitwarden_error;
+use thiserror::Error;
+#[cfg(feature = "wasm")]
+use wasm_bindgen::prelude::wasm_bindgen;
+
+use crate::{CipherId, cipher_client::admin::CipherAdminClient};
+
+#[allow(missing_docs)]
+#[bitwarden_error(flat)]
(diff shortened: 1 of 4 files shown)


crates/bitwarden-vault/src/cipher/cipher_client/admin/delete_attachment.rs
```diff
@@ -0,0 +1,76 @@
...
+    let api = api_client.ciphers_api();
+    api.delete_attachment_admin(cipher_id.into(), attachment_id)
+        .await?;
+    Ok(())
+}
+
+#[cfg_attr(feature = "wasm", wasm_bindgen)]
+impl CipherAdminClient {
+    /// Deletes an attachment from a cipher using the admin endpoint.
+    /// Affects server data only, does not modify local state.
```
nick-livefront, line 31: ❓ Is there a reason this deletion doesn't modify state or return the updated cipher when the non-admin does?

nick-livefront (approved):

crates/bitwarden-vault/src/cipher/cipher_client/admin/delete_attachment.rs
gbubemismith: The admin endpoint is used for ciphers the admin user doesn't have normal access to. Since those ciphers aren't part of the user's sync and therefore don't exist in their local state, there's no local state to update and no meaningful cipher to return.

PR #785, opened 2026-02-20
[PM-31051] Add V2UpgradeToken for key rotation without logout

mzieniukbw:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-31051

## 📔 Objective

The V2 upgrade token provides a way for upgrade rotations to retain access to old encrypted data, and both the old and new user-key for a temporary period.

New `V2UpgradeToken` struct:
- Create new `V2UpgradeToken` from v1 and v2 symmetric keys from key store context.
- Create new `V2UpgradeToken` from response model.
- Retrieve v1 or v2 symmetric key from into the key store context with `unwrap_v1` and `unwrap_v2`
- Added FFI bindings
- Added to `InitUserCryptoRequest` - if the token is present, and the unlock results in a V1 token, we decrypt the upgrade token and set the V2 key to state
- Added `get_upgraded_user_key` to `CryptoClient` - gets the upgraded V2 user key using an upgrade token.

Questions:
- Should `V2UpgradeToken` and `get_upgraded_user_key` be moved to `bitwarden-user-crypto-management` crate ?


crates/bitwarden-core/src/client/internal.rs
```diff
@@ -228,9 +228,10 @@ impl InternalClient {
         master_key: MasterKey,
         user_key: EncString,
         account_crypto_state: WrappedAccountCryptographicState,
+        upgrade_token: Option<crate::key_management::V2UpgradeToken>,
```
quexten: Nit: Can we just import V2UpgradeToken? Having the full import path on multiple function args, here, and a multiple times later in the file is less readable.

crates/bitwarden-core/src/client/internal.rs
```diff
@@ -239,18 +240,36 @@ impl InternalClient {
...
     ) -> Result<(), EncryptionSettingsError> {
         let mut ctx = self.key_store.context_mut();
 
+        // Add the decrypted key to KeyStore first
+        let user_key_id = ctx.add_local_symmetric_key(user_key.clone());
+
+        // Upgrade V1 key to V2 if token is present
+        let user_key_id = match (&user_key, &upgrade_token) {
```
quexten: Logic looks good, but a test would be nice.

crates/bitwarden-core/src/key_management/v2_upgrade_token.rs
```diff
@@ -0,0 +1,660 @@
+//! V2 Upgrade Token allows secure bidirectional key rotation from V1 to V2 user keys.
```
quexten: I'm not sure what "bidirectional key rotation" means. Maybe "that allows upgrading V1 sessions after a key-rotation to V2 sessions without logout or breaking local unlock methods.

crates/bitwarden-core/src/key_management/v2_upgrade_token.rs
```diff
@@ -0,0 +1,660 @@
...
+use bitwarden_crypto::{EncString, KeyDecryptable, KeyIds, KeyStoreContext, SymmetricCryptoKey};
+use serde::{Deserialize, Serialize};
+use thiserror::Error;
+use tracing::instrument;
+#[cfg(feature = "wasm")]
+use wasm_bindgen::convert::FromWasmAbi;
+
+use crate::key_management::SymmetricKeyId;
+
+/// V2 Upgrade Token enables bidirectional key rotation from V1 to V2 user keys.
```
quexten: Bidirectional key rotation

crates/bitwarden-core/src/key_management/v2_upgrade_token.rs
```diff
@@ -0,0 +1,660 @@
...
+    wrapped_uk_1: EncString,
+    wrapped_uk_2: EncString,
+}
+
+impl V2UpgradeToken {
+    /// Creates a new V2UpgradeToken from key IDs in the KeyStore.
+    ///
+    /// This encrypts V1 with V2, V2 with V1, then validates bidirectional decryption.
+    ///
+    /// # Arguments
```
quexten: Noting: The style guide says to use descriptive names and types rather than docs: 
https://contributing.bitwarden.com/contributing/code-style/rust/#arguments

I feel like the types should already be descriptive enough.

crates/bitwarden-core/src/key_management/v2_upgrade_token.rs
```diff
@@ -0,0 +1,660 @@
...
+/// V2 Upgrade Token enables bidirectional key rotation from V1 to V2 user keys.
+///
+/// The token contains:
+/// - `wrapped_uk_1`: V1 user key encrypted with V2 key (Cose_Encrypt0_B64 format)
+/// - `wrapped_uk_2`: V2 user key encrypted with V1 key (Aes256Cbc_HmacSha256_B64 format)
+///
+/// Both wrapping directions are validated on creation and unwrapping to prevent tampering.
+#[derive(Clone, Debug)]
+pub struct V2UpgradeToken {
+    wrapped_uk_1: EncString,
```
quexten: Nit: On the server we call this: WrappedUserKey1. Do we want to name it the same just to make it a tiny bit easier? I.e `wrapped_userkey_1`

crates/bitwarden-core/src/key_management/v2_upgrade_token.rs
```diff
@@ -0,0 +1,660 @@
...
+    #[instrument(skip(ctx))]
+    pub fn create<Ids: KeyIds>(
+        v1_key_id: Ids::Symmetric,
+        v2_key_id: Ids::Symmetric,
+        ctx: &KeyStoreContext<Ids>,
+    ) -> Result<Self, V2UpgradeTokenError> {
+        // Get the keys from the KeyStore and type-check them
+        #[allow(deprecated)]
+        let v1_key = ctx
+            .dangerous_get_symmetric_key(v1_key_id)
```
quexten: I wonder whether crate-wide helpers for asserting encryption types would be useful. I saw this pattern in a few other places.

Something like: "validate_encryption_type(key_id, ctx)".
That would reduce the code here to 2 lines.

crates/bitwarden-core/src/key_management/v2_upgrade_token.rs
```diff
@@ -0,0 +1,660 @@
...
+
+        // Wrap V2 key with V1 key (creates Aes256Cbc_HmacSha256_B64)
+        let wrapped_uk_2 = ctx
+            .wrap_symmetric_key(v1_key_id, v2_key_id)
+            .map_err(|_| V2UpgradeTokenError::EncryptionFailed)?;
+
+        // Validate bidirectional decryption
+        // wrapped_uk_1 is V1 encrypted with V2, so decrypt with V2
+        let _: Vec<u8> = wrapped_uk_1
+            .decrypt_with_key(v2_key)
```
quexten: Can we swap this to "decrypt(ctx, key)"?

crates/bitwarden-core/src/key_management/v2_upgrade_token.rs
```diff
@@ -0,0 +1,660 @@
...
+            .map_err(|_| V2UpgradeTokenError::ValidationFailed)?;
+        // wrapped_uk_2 is V2 encrypted with V1, so decrypt with V1
+        let _: Vec<u8> = wrapped_uk_2
+            .decrypt_with_key(v1_key)
```
quexten: Can we swap this to "decrypt(ctx, key)"?

crates/bitwarden-core/src/key_management/v2_upgrade_token.rs
```diff
@@ -0,0 +1,660 @@
...
+            .map_err(|_| V2UpgradeTokenError::DecryptionFailed)?;
+
+        // Validate: unwrapped V1 should be able to decrypt wrapped_uk_2
+        #[allow(deprecated)]
+        let v1_key = ctx
+            .dangerous_get_symmetric_key(new_v1_id)
+            .map_err(|_| V2UpgradeTokenError::DecryptionFailed)?;
+        let _: Vec<u8> = self
+            .wrapped_uk_2
+            .decrypt_with_key(v1_key)
```
quexten: Can we swap this to "decrypt(ctx, key)"?

crates/bitwarden-core/src/key_management/v2_upgrade_token.rs
```diff
@@ -0,0 +1,660 @@
...
+            .map_err(|_| V2UpgradeTokenError::DecryptionFailed)?;
+
+        // Validate: unwrapped V2 should be able to decrypt wrapped_uk_1
+        #[allow(deprecated)]
+        let v2_key = ctx
+            .dangerous_get_symmetric_key(new_v2_id)
+            .map_err(|_| V2UpgradeTokenError::DecryptionFailed)?;
+        let _: Vec<u8> = self
+            .wrapped_uk_1
+            .decrypt_with_key(v2_key)
```
quexten: Can we swap this to "decrypt(ctx, key)"?. We should also be able to just drop getting the key from context entirely.

crates/bitwarden-core/src/key_management/v2_upgrade_token.rs
```diff
@@ -0,0 +1,660 @@
...
+    ///
+    /// # Arguments
+    /// * `response` - The `V2UpgradeTokenResponseModel` from the server
+    /// * `ctx` - Mutable KeyStore context containing the current user key at
+    ///   [`SymmetricKeyId::User`]
+    ///
+    /// # Returns
+    /// `(v1_key_id, v2_key_id, token)` — both key IDs available in the context
+    #[instrument(skip_all)]
+    pub fn from_response_model(
```
quexten: It is unintuitive to me that a "from_response_model" would already load the keys into context. I'd rather have this split into two functions "from_response_model" (or, actually implement From<&V2UpgradeTokenResponseModel>, you don't need the key context because no crypto operations need to be done here.

Then, you can have a separate: "load_to_context", though, I'm unsure what this does compared to the existing unwrap_v1 / unwrap_v2 functions.

crates/bitwarden-core/src/key_management/v2_upgrade_token.rs
```diff
@@ -0,0 +1,660 @@
...
+            Ok((SymmetricKeyId::User, v2_key_id, token))
+        } else {
+            // Assume current key is V2
+            // Unwrap V1 (wrapped_user_key_1 = V1 key wrapped with V2) using the current V2 key
+            let v1_key_id = ctx
+                .unwrap_symmetric_key(SymmetricKeyId::User, &wrapped_user_key_1)
+                .map_err(|_| V2UpgradeTokenError::DecryptionFailed)?;
+            let token = Self::create(v1_key_id, SymmetricKeyId::User, ctx)?;
+            Ok((v1_key_id, SymmetricKeyId::User, token))
+        }
```
quexten:
```suggestion

```

crates/bitwarden-core/src/key_management/v2_upgrade_token.rs
```diff
@@ -0,0 +1,660 @@
...
+    }
+}
+
+impl FromStr for V2UpgradeToken {
+    type Err = V2UpgradeTokenError;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        #[derive(Deserialize)]
+        struct Fields {
```
quexten: Could we have a comment why we have a separate struct here? Why not just add deserialize to the V2UpgradeToken struct?

crates/bitwarden-core/src/key_management/v2_upgrade_token.rs
```diff
@@ -0,0 +1,660 @@
...
+            wrapped_uk_1,
+            wrapped_uk_2,
+        })
+    }
+}
+
+impl From<V2UpgradeToken> for String {
+    fn from(val: V2UpgradeToken) -> Self {
+        #[derive(Serialize)]
+        struct Fields<'a> {
```
quexten: Same here, I don't understand why we have a separate struct here.

quexten (changes requested):

crates/bitwarden-core/src/key_management/v2_upgrade_token.rs
quexten: Maybe we can just derive serialize and deserialize and drop these?

crates/bitwarden-core/src/client/internal.rs
mzieniukbw: This should be covered in `crypto.rs` through `initialize_user_crypto` function tests.

mzieniukbw:
@quexten i have addressed your suggestions

crates/bitwarden-core/src/key_management/crypto.rs
```diff
@@ -77,6 +83,9 @@ pub struct InitUserCryptoRequest {
     pub account_cryptographic_state: WrappedAccountCryptographicState,
     /// The method to decrypt the user's account symmetric key (user key)
     pub method: InitUserCryptoMethod,
+    /// Optional V2 upgrade token for automatic key rotation from V1 to V2
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub upgrade_token: Option<crate::key_management::V2UpgradeToken>,
```
quexten: Nit: Import V2UpgradeToken to make this a bit more readable.

crates/bitwarden-core/src/key_management/crypto_client.rs
```diff
@@ -239,6 +239,46 @@ impl CryptoClient {
             org_public_key,
         )
     }
+
+    /// Gets the upgraded V2 user key using an upgrade token.
+    /// If the current key is already V2, returns it directly.
+    /// If the current key is V1 and a token is provided, extracts the V2 key.
+    pub fn get_upgraded_user_key(
+        &self,
+        upgrade_token: Option<crate::key_management::V2UpgradeToken>,
```
quexten: Nit: Import v2 upgrade token

crates/bitwarden-core/src/client/internal.rs
```diff
@@ -240,18 +241,36 @@ impl InternalClient {
         &self,
         user_key: SymmetricCryptoKey,
         account_crypto_state: WrappedAccountCryptographicState,
+        upgrade_token: Option<V2UpgradeToken>,
```
quexten: Any reason we don't Accept &Option<V2UpgradeToken> here? Then we wouldn't need to clone the token a bunch in the calling functions.

quexten:
@mzieniukbw Getting close. I'm OK with approving, but also there is some nits that are easiest to address now.

crates/bitwarden-core/src/client/internal.rs
mzieniukbw: Fixed

crates/bitwarden-core/src/key_management/crypto.rs
mzieniukbw: I have changed this so many times, must be my formatter or claude interfering. Fixed

quexten (approved):

gbubemismith (approved):
Vault changes are minimal. Approving on behalf of the Vault team.

addisonbeck (approved):

PR #786, opened 2026-02-23
Document data model conventions in CLAUDE.md

coroiu:
## Summary

- Adds a "Data Model Conventions" section to the root `CLAUDE.md` summarising the four model layers (domain, `View`, `Request`, `Response`) with a reference to the contributing docs
- Codifies the rule that resources with both create and edit operations should use separate `*CreateRequest` and `*EditRequest` structs
- Codifies the rule that models with a type discriminant and per-variant data should use enums with associated data rather than a bare discriminant field alongside multiple `Option` fields

### DIFF
--- a/.claude/CLAUDE.md
+++ b/.claude/CLAUDE.md
@@ -121,6 +121,11 @@ picture.
 requires an `id`, `revision_date`, or fields that are immutable after creation), use **separate**
 `*CreateRequest` and `*EditRequest` structs.
 
+**Variant data**: When a model has a type discriminant with per-variant associated data (e.g. a send
+is either a file or text, never both), use an enum with associated data rather than a bare
+discriminant field alongside multiple `Option` fields. The mapping from server wire format (numeric
+discriminant + optional fields) belongs at the API→domain boundary.
+
 ## Development Workflow
 
 **Build & Test:**
--- a/.claude/CLAUDE.md
+++ b/.claude/CLAUDE.md
@@ -117,8 +117,9 @@ picture.
 | `Request`  | Public input DTO from client into SDK                   | `CipherCreateRequest`    |
 | `Response` | Public output DTO from SDK to client                    | `LoginResponse`          |
 
-**Create vs. Edit requests**: When a resource supports both create and edit operations, define
-**separate** `*CreateRequest` and `*EditRequest` structs rather than a single combined struct.
+**Create vs. Edit requests**: When the fields for creating and editing an item differ (e.g. edit
+requires an `id`, `revision_date`, or fields that are immutable after creation), use **separate**
+`*CreateRequest` and `*EditRequest` structs.
 
 **Variant data**: When a model has a type discriminant with per-variant associated data (e.g. a send
 is either a file or text, never both), use an enum with associated data rather than a bare
--- a/.claude/CLAUDE.md
+++ b/.claude/CLAUDE.md
@@ -119,8 +119,7 @@ picture.
 
 **Create vs. Edit requests**: When the fields for creating and editing an item differ (e.g. edit
 requires an `id`, `revision_date`, or fields that are immutable after creation), use **separate**
-`*CreateRequest` and `*EditRequest` structs. Combined `*AddEditRequest` structs exist in some older
-code (e.g. `FolderAddEditRequest`) but should not be used for new code.
+`*CreateRequest` and `*EditRequest` structs.
 
 ## Development Workflow
 
--- a/.claude/CLAUDE.md
+++ b/.claude/CLAUDE.md
@@ -117,9 +117,8 @@ picture.
 | `Request`  | Public input DTO from client into SDK                   | `CipherCreateRequest`    |
 | `Response` | Public output DTO from SDK to client                    | `LoginResponse`          |
 
-**Create vs. Edit requests**: When the fields for creating and editing an item differ (e.g. edit
-requires an `id`, `revision_date`, or fields that are immutable after creation), use **separate**
-`*CreateRequest` and `*EditRequest` structs.
+**Create vs. Edit requests**: When a resource supports both create and edit operations, define
+**separate** `*CreateRequest` and `*EditRequest` structs rather than a single combined struct.
 
 **Variant data**: When a model has a type discriminant with per-variant associated data (e.g. a send
 is either a file or text, never both), use an enum with associated data rather than a bare
--- a/.claude/CLAUDE.md
+++ b/.claude/CLAUDE.md
@@ -104,6 +104,24 @@ Monorepo crates organized in **four architectural layers**:
 - TypeScript compilation tested against `clients` repo on PR
 - Document migration path for clients
 
+### Data Model Conventions
+
+The SDK uses distinct model layers; use the correct type for the context. See the
+[data models docs](https://contributing.bitwarden.com/architecture/sdk/data-models) for the full
+picture.
+
+| Suffix     | Role                                                    | Example                  |
+| ---------- | ------------------------------------------------------- | ------------------------ |
+| _(none)_   | Encrypted domain model — internal, not exposed publicly | `Cipher`, `Send`         |
+| `View`     | Decrypted DTO returned to clients                       | `CipherView`, `SendView` |
+| `Request`  | Public input DTO from client into SDK                   | `CipherCreateRequest`    |
+| `Response` | Public output DTO from SDK to client                    | `LoginResponse`          |
+
+**Create vs. Edit requests**: When the fields for creating and editing an item differ (e.g. edit
+requires an `id`, `revision_date`, or fields that are immutable after creation), use **separate**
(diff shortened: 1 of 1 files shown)


.claude/CLAUDE.md
```diff
@@ -104,6 +104,28 @@ Monorepo crates organized in **four architectural layers**:
...
+| Suffix     | Role                                                    | Example                  |
+| ---------- | ------------------------------------------------------- | ------------------------ |
+| _(none)_   | Encrypted domain model — internal, not exposed publicly | `Cipher`, `Send`         |
+| `View`     | Decrypted DTO returned to clients                       | `CipherView`, `SendView` |
+| `Request`  | Public input DTO from client into SDK                   | `CipherCreateRequest`    |
+| `Response` | Public output DTO from SDK to client                    | `LoginResponse`          |
+
+**Create vs. Edit requests**: When the fields for creating and editing an item differ (e.g. edit
+requires an `id`, `revision_date`, or fields that are immutable after creation), use **separate**
+`*CreateRequest` and `*EditRequest` structs.
```
Hinton, line 122: This is different from what the contributing docs PR says.

Hinton (dismissed):

.claude/CLAUDE.md
```diff
@@ -104,6 +104,28 @@ Monorepo crates organized in **four architectural layers**:
...
+The SDK uses distinct model layers; use the correct type for the context. See the
+[data models docs](https://contributing.bitwarden.com/architecture/sdk/data-models) for the full
+picture.
+
+| Suffix     | Role                                                    | Example                  |
+| ---------- | ------------------------------------------------------- | ------------------------ |
+| _(none)_   | Encrypted domain model — internal, not exposed publicly | `Cipher`, `Send`         |
+| `View`     | Decrypted DTO returned to clients                       | `CipherView`, `SendView` |
+| `Request`  | Public input DTO from client into SDK                   | `CipherCreateRequest`    |
+| `Response` | Public output DTO from SDK to client                    | `LoginResponse`          |
```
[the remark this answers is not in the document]
coroiu: Yeah I got this in my local review too and I don't agree. These are examples. Types are gonna come and go and I don't think we should feel obligated to keep docs updated like that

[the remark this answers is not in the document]
theMickster: Yes, but we're also trying to instruct Claude here; not humans. Would Claude make a less desirable decision with the information as you have it?

.claude/CLAUDE.md
```diff
@@ -104,6 +104,28 @@ Monorepo crates organized in **four architectural layers**:
...
+
+**Create vs. Edit requests**: When the fields for creating and editing an item differ (e.g. edit
+requires an `id`, `revision_date`, or fields that are immutable after creation), use **separate**
+`*CreateRequest` and `*EditRequest` structs.
+
+**Variant data**: When a model has a type discriminant with per-variant associated data (e.g. a send
```
theMickster, line 124: A local review brought to light this information. What are our thoughts?

⚠️ `Send` example contradicts the "Variant data" convention described in the same section

`.claude/CLAUDE.md:119-122`

  The "Variant data" paragraph states: "use an enum with associated data rather than a bare discriminant field alongside multiple `Option` fields." However, `Send` and `SendView` — cited as examples in the table above — use exactly this anti-pattern: a bare `r#type: SendType` discriminant field alongside `file: Option<SendFile>` and `text: Option<SendText>`.

  This creates a contradiction within the same documentation section. An AI agent may attempt to refactor `Send` to match the stated convention (when not asked), or become confused about which pattern to follow when existing code contradicts the stated rule.

  **Suggested fix:** Add a qualifier such as "For new models," or "Going forward," to clarify this is an aspirational convention. Alternatively, note that existing types like `Send` predate this rule.

Image

  </details>

.claude/CLAUDE.md
```diff
@@ -104,6 +104,28 @@ Monorepo crates organized in **four architectural layers**:
 - TypeScript compilation tested against `clients` repo on PR
 - Document migration path for clients
 
+### Data Model Conventions
+
+The SDK uses distinct model layers; use the correct type for the context. See the
```
theMickster, line 109: A local review brought to light this information. What are our thoughts?

⚠️ Domain model described as "not exposed publicly" but types ARE publicly exported with bindings

`.claude/CLAUDE.md:110`

  The table states the no-suffix domain model layer is "internal, not exposed publicly." However, both examples cited — `Cipher` and `Send` — are `pub struct` types re-exported from their crate's public API (`pub use cipher::*` in bitwarden-vault, `pub use send::Send` in bitwarden-send). Both also derive UniFFI bindings (`#[cfg_attr(feature = "uniffi", derive(uniffi::Record))]`), and `Cipher` additionally derives WASM bindings.

  This could mislead an AI agent into refusing to use or expose domain model types in binding-layer code, or wrapping them in unnecessary conversion layers.

  **Suggested fix:** Change "internal, not exposed publicly" to something like "server/storage-layer model — prefer `View`/`Request`/`Response` at API boundaries" to convey the intent (prefer other layers at boundaries) without making a factually incorrect claim about visibility.
  </details>

.claude/CLAUDE.md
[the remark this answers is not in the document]
theMickster: I am a little confused/concerned here because `LoginResponse` does exist in `crates/bitwarden-auth/src/login/models/login_response.rs`. The suggested replacement is a hallucination (which is ironic given we're trying to give Claude more context to ground itself upon 🤪)

Claude will see these types (`LoginResponse`), will try to `grep` for it, find it, and use it as a reference — which actually works correctly here since the type exists. If the type did not exist, the agent might waste cycles searching or _worse_, hallucinate an implementation based on the name alone. So using real types is a form of grounding that costs very little upfront, but it does require us to maintain the `Claude.md` just like we would any other code.

theMickster (commented):
Posted a couple things for human double-checking

.claude/CLAUDE.md
Hinton: Send is a legacy type which goal was to map as closely to the js/c# models as possible.

theMickster: Understood; and I think that's what Claude is intending to tell us to put into the instructions. We should not use language that hints a rule is always a rule if there are exceptions. That will only lead us to having more hallucinations.

coroiu: I had the opposite concern and Claude told me to remove the section I had explicitly stating legacy code exists.

image

coroiu: I mean, code is always going to have legacy/non-conforming portions, I don't think adding "beware legacy code exists" to every rule is a good idea

.claude/CLAUDE.md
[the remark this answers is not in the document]
coroiu: Ok hang on I might be confusing different structs here. The `auth` crate has a unique pattern which uses `ApiRequest` that I must've been thinking about. Urgh, it's annoying how Claude can understand the exception very well in one context and then completely hallucinate because of it in another

image

I don't like the cost of having to maintain the `CLAUDE.md` file like this so, are you sure that's what is gonna happen? How about just making it explicit that the examples are not intended to reflect the codebase

.claude/CLAUDE.md
theMickster: I'm not advocating for that either. 

What I am advocating for is clear instructions to Claude and to then observe the behavior. If Claude recommends the legacy anti-pattern above then we should revisit this discussion. Otherwise, we're good

coroiu: Ok, my bad, I misunderstood the comment! :)

.claude/CLAUDE.md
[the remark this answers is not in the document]
theMickster: I think that's worth a try; do you have some prompts and/or work ready to give these changes a whirl?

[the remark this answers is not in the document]
coroiu: Yes, these changes are the result of a PR review in which these things need to be addressed. It's the one referenced above https://github.com/bitwarden/sdk-internal/pull/756 . I'll give it a try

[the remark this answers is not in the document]
coroiu: Picked it up fine I think ?

image

.claude/CLAUDE.md
coroiu: Addressed in [2dec1dc](https://github.com/bitwarden/sdk-internal/pull/786/commits/2dec1dcdfd83546bbdd6c069b8749d6bff126af1)

theMickster: Thanks!  I wasn't sure if this was a finding or not; the whole boundary and re-exporting had me 😵‍💫😵‍💫😵‍💫

.claude/CLAUDE.md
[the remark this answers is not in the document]
theMickster: Sure! Let's give it a try.

theMickster (approved):

Hinton (approved):

PR #788, opened 2026-02-23
[PM-31763] Add initialize crypto method: unlock via client managed state

quexten:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-31763

## 📔 Objective

Add the "client managed state" unlock method, which unlocks directly from client managed state without passing in the decrypted/encoded key. This should be used in the default SDK service, instead of handling and passing in a user-key.

Decrypted key will then be for biometrics / never-lock only, and also perform a v2 token migration, while client managed will be for initializing an SDK in default-sdk-service, and not perform a migration.


djsmith85 (approved):

crates/bitwarden-core/src/key_management/wasm_unlock_state.rs
mzieniukbw, line 1: Would be great to have test coverage in the future here.

mzieniukbw (approved):
was wondering about test coverage for wasm

crates/bitwarden-core/src/key_management/wasm_unlock_state.rs
quexten: @coroiu Showed that we can actually add tests for wasm recently. I'll try to make a follow-up PR with tests.

PR #794, opened 2026-02-25
[PM-32716] Remove user id from security state

quexten:
## 🎟️ Tracking

https://bitwarden.slack.com/archives/C07HY17HUHE/p1771865163146259
https://bitwarden.atlassian.net/browse/PM-32716

## 📔 Objective

Due to a creation order problem, we cannot add the user id to the security state, since the server decides the user-id, but the client needs to submit the signed security state to the server at registration time.

We do not actually need it, so we are removing the user-id entirely.


mzieniukbw (approved):

Patrick-Pimentel-Bitwarden (approved):
👍 Auth changes are minimal and acceptable

PR #801, opened 2026-02-28
[PM-32525] Unwrap using key_identifier when generating cipher view key

eligrubb:
## 🎟️ Tracking

[https://bitwarden.atlassian.net/browse/PM-32525](https://bitwarden.atlassian.net/browse/PM-32525)

## 📔 Objective

The `generate_cipher_key` function assumes that the `SymmetricKeyId` parameter should be used for both unwrapping existing ciphers and for rewrapping the newly generated cipher key. However, when rotating cipher keys via `encrypt_cipher_for_rotation`, the `new_key` is only meant to wrap the newly generated key. The old key, which can be found in `cipher_view.key_identifier()`, needs to be used for unwrapping the cipher along with fido2 credentials and attachment keys. The existing `cipher_view.reencrypt_cipher_keys()` follows this same pattern of passing in the new wrapping key and calling `self.key_identifier()` to get the old key.

All other places which call `generate_cipher_key` already pass in `cipher_view.key_identifier()` as the wrapping key.

### DIFF
--- a/crates/bitwarden-vault/src/cipher/cipher.rs
+++ b/crates/bitwarden-vault/src/cipher/cipher.rs
@@ -738,16 +738,17 @@ impl CipherView {
     pub fn generate_cipher_key(
         &mut self,
         ctx: &mut KeyStoreContext<KeyIds>,
-        key: SymmetricKeyId,
+        wrapping_key: SymmetricKeyId,
     ) -> Result<(), CryptoError> {
-        let old_ciphers_key = Cipher::decrypt_cipher_key(ctx, key, &self.key)?;
+        let old_unwrapping_key = self.key_identifier();
+        let old_ciphers_key = Cipher::decrypt_cipher_key(ctx, old_unwrapping_key, &self.key)?;
 
         let new_key = ctx.generate_symmetric_key();
 
         self.reencrypt_attachment_keys(ctx, old_ciphers_key, new_key)?;
         self.reencrypt_fido2_credentials(ctx, old_ciphers_key, new_key)?;
 
-        self.key = Some(ctx.wrap_symmetric_key(key, new_key)?);
+        self.key = Some(ctx.wrap_symmetric_key(wrapping_key, new_key)?);
         Ok(())
     }
 


quexten:
Interesting, we had the same bug in fully-sdk-based-key-rotation.

crates/bitwarden-vault/src/cipher/cipher.rs
```diff
@@ -738,16 +738,17 @@ impl CipherView {
     pub fn generate_cipher_key(
```
quexten, line 738: (@eligrubb not blocking for your bug-fix, but I want to start a discussion here on this function);

I realize this is out of scope for this bug, but we should have documentation of what this function intends to achieve, that represents the intent of what vault wants to achieve here.

"old_ciphers_key" here can be:
- The cipher-key
- The user-key
- The organization-key

Currently, this function:
- Decrypts / loads the old_ciphers_key (one of the three above)
- Generates a new cipher-key
- Sets the wrapped cipher-key
- Re-encrypts attachment keys
- Re-encrypts fido2 credentials
- (Does not re-encrypt data) -> Risk for inconsistent state

Which essentially means it's not generating a cipher-key, it's mutating the cipher, upserting and optionally generating a cipher-key, and then re-wrapping the cipher's data (attachments, fido2 creds) with it, but not re-encrypting the data with it.

To be honest I'm not even convinced the function is correct as-is, because it's re-encrypting attachments and fido2 credentials, but nothing of the other data, which could lead to partially re-encrypted ciphers.

I.e I'm saying this public, undocumented API is both unintuitive and dangerous.

quexten (approved):

quexten:
Changes look good. Nice find!

crates/bitwarden-vault/src/cipher/cipher.rs
gbubemismith: I think the function name is actually misleading, it's upserting a cipher key and re-wrapping dependent sub-keys under it. I agree we need better documentation around this

gbubemismith (approved):

Hinton:
I'm not sure I follow here, why would you generate a new cipher key during a key rotation? One of the benefits a separate key gives is not having to rotate every single field on cipher during a key rotation.

PR #804, opened 2026-03-02
[PM-28926] Bulk DB operations

Hinton:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-28926
## 📔 Objective

Adds support for bulk database operations.

- set_bulk
- remove_bulk
- remove_all

## 🚨 Breaking Changes

The Database interface on clients needs to be updated. A PR will be prepared ahead of time.

### DIFF
--- a/crates/bitwarden-wasm-internal/src/platform/repository.rs
+++ b/crates/bitwarden-wasm-internal/src/platform/repository.rs
@@ -46,7 +46,10 @@ pub(crate) trait WasmRepository<T> {
     async fn get(&self, id: String) -> Result<JsValue, JsValue>;
     async fn list(&self) -> Result<JsValue, JsValue>;
     async fn set(&self, id: String, value: T) -> Result<JsValue, JsValue>;
+    async fn set_bulk(&self, values: Vec<(String, T)>) -> Result<JsValue, JsValue>;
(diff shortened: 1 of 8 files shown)


crates/bitwarden-wasm-internal/src/platform/repository.rs
```diff
@@ -157,11 +185,25 @@ macro_rules! create_wasm_repositories {
                     id: String,
                     value: $qualified_type_name,
                 ) -> Result<::wasm_bindgen::JsValue, ::wasm_bindgen::JsValue>;
+                #[wasm_bindgen(method, catch, js_name = "setBulk")]
+                async fn set_bulk_js(
```
dani-garcia: The rest of the functions don't have the _js suffix, just two of the new ones. Should we never include the suffix for consistency?

dani-garcia (dismissed):
Looks good to me, just a minor nit

crates/bitwarden-wasm-internal/src/platform/repository.rs
Hinton: Removed.

dani-garcia (approved):

PR #806, opened 2026-03-03
[PM-31114] Introduce local user data key, set it to state on crypto init

mzieniukbw:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-31114

## 📔 Objective

Introduced `WrappedLocalUserDataKey`, which will be used to decouple the direct use of user key for encryption in tool's code (like encrypted forwarders, password history).
This will allow for user key rotation without logout, since we just have to re-encrypt the local user data key with the new user key, without re-encrypting all other data.
On user crypto init, we create new `WrappedLocalUserDataKey` key, and set it to state (SDK managed), if not already set and set it to context. For backwards compatibility the local user data key is user key wrapped user key - this allows us to use it as drop in replacement in all client's, which currently encrypt with user key directly. On user key rotation, the new user key will differ from the one used as local user data key, retaining backwards compatibility, without having to migrate user over.
In the future, the `WrappedLocalUserDataKey` will likely be generated.

Other changes:
- Added `Client::new_test`, which creates new client with SDK-managed state repositories registration (but in memory storage for testing). Refactored bunch of tests. This is so we can fail early on user crypto init when expected repository have missing registration.
- Disabled unnecessary WASM optimisations for debug build `build.sh` (started to fail, since the file is too large to handle). Also added missing `npm ci` step which was complaining about missing typescript dependency.


crates/bitwarden-core/src/key_management/crypto.rs
```diff
@@ -186,142 +187,145 @@ pub(super) async fn initialize_user_crypto(
...
+    #[cfg(feature = "wasm")]
+    let should_copy_user_key = matches!(
+        req.method,
+        InitUserCryptoMethod::MasterPasswordUnlock { .. }
+            | InitUserCryptoMethod::DecryptedKey { .. }
+            | InitUserCryptoMethod::PinEnvelope { .. }
+    );
 
-            drop(_span_guard);
+    {
```
quexten: Do we have a reason for this block scope?

crates/bitwarden-crypto/src/keys/local_user_data_key.rs
```diff
@@ -0,0 +1,123 @@
...
+    /// enabling decryption of data previously encrypted with the user key.
+    pub fn from_user_key(user_key: &SymmetricCryptoKey) -> Self {
+        LocalUserDataKey::DerivedFromUserKey(user_key.to_encoded())
+    }
+
+    /// Wrap (encrypt) this key with the given user key for protected storage.
+    #[cfg_attr(feature = "dangerous-crypto-debug", instrument(err))]
+    #[cfg_attr(not(feature = "dangerous-crypto-debug"), instrument(skip_all, err))]
+    pub fn encrypt_with_user_key(
+        &self,
```
quexten: Question: Can we make this work on key context instead of taking a key reference?

crates/bitwarden-crypto/src/keys/local_user_data_key.rs
```diff
@@ -0,0 +1,123 @@
...
+        user_key: &SymmetricCryptoKey,
+    ) -> crate::error::Result<EncString> {
```
quexten: Nit: Just use Result<EncString, CryptoError> or import crate::error::Result

crates/bitwarden-core/src/key_management/mod.rs
```diff
@@ -63,6 +66,17 @@ pub struct UserKeyState {
...
+/// Represents the encrypted local user data key held in client-managed state.
+/// This key is used to encrypt local user data (e.g., password generator history).
+#[derive(Serialize, Deserialize, Debug, Clone)]
+#[cfg_attr(feature = "wasm", derive(Tsify), tsify(into_wasm_abi, from_wasm_abi))]
+#[cfg_attr(feature = "uniffi", derive(uniffi::Record))]
+pub struct LocalUserDataKeyState {
+    encrypted_key: EncString,
+}
+
+bitwarden_state::register_repository_item!(String => LocalUserDataKeyState, "LocalUserDataKey");
```
quexten: Did we want to make this sdk managed? If so this is the right way to do it: https://github.com/bitwarden/sdk-internal/tree/main/crates/bitwarden-state#settings

crates/bitwarden-crypto/src/keys/local_user_data_key.rs
```diff
@@ -0,0 +1,123 @@
...
+/// An indirect symmetric key for encrypting local user data (e.g. password generator history).
+/// Enables offline decryption of local data after a key rotation: only the wrapped key is
+/// re-encrypted; the local user data key itself stays intact.
+#[derive(Clone)]
+pub enum LocalUserDataKey {
+    /// Derived from the user key.
+    ///
+    /// Used for backwards compatibility: data previously encrypted with the user key can be
+    /// decrypted by the reconstructed `SymmetricCryptoKey`.
+    DerivedFromUserKey(BitwardenLegacyKeyBytes),
```
quexten: I'm not sure I understand the reason why this is an enum. If we have an encrypted local user data key (encstring) do we still store the information about whether it was a migrated user key?

Besides this, I'd recommend not making another decrypted copy of keys. We generally want keys to live on the key context, so the struct here should be the wrapped version, with functions to set it to the key context. I.e:
```rust
pub struct WrappedLocalUserDataKey(EncString);

impl WrappedLocalUserDataKey {
  pub fn from_user_key(user_key_id: [...], ctx: [...])

  pub fn set_to_context(user_key_id: [...], local_user_key_id: [...], ctx: [...])
}
```

crates/bitwarden-core/src/key_management/crypto.rs
mzieniukbw: The span guard does not have Send trait and can't be used with in async context - hence we have to drop it before first await in this function. This is also the reason why i moved all async code beneath that synchronous code block.
Do you want me to add a comment for clarity ? We could also simply make it a separate function.

quexten: Ah, makes sense. A drop(_span_guard); should also work, and is possibly easier to read, but it's unfortunate that the span requires so much manual work here..

crates/bitwarden-crypto/src/keys/local_user_data_key.rs
mzieniukbw: You are right, we are storing this in storage, hence it have to be wrapped, not decrypted as symmetric key. I will correct this.

crates/bitwarden-core/src/key_management/crypto.rs
dani-garcia: You may have an easier time removing the span and instead using `#[tracing::instrument]` on the function definition, which should also define and manage a span for you.

mzieniukbw: Added `#[tracing::instrument]`, all fields are skipped from instrumentation, except the user_id.

crates/bitwarden-core/src/key_management/mod.rs
mzieniukbw: moved to SDK managed

crates/bitwarden-crypto/src/keys/local_user_data_key.rs
mzieniukbw: Moved to `WrappedLocalUserDataKey` as suggested.
Since we should wrap/unwrap with user key, i decided to move it to core crate and use `SymmetricKeyId::User` key id directly.

crates/bitwarden-core/src/key_management/crypto.rs
```diff
@@ -1031,6 +1033,36 @@ pub(crate) fn make_user_key_connector_registration(
...
+async fn initialize_user_local_data_key(client: &Client) -> Result<(), EncryptionSettingsError> {
+    // In tests, repositories are not always registered; skip gracefully.
+    #[cfg(test)]
+    if client
+        .platform()
+        .state()
+        .get::<crate::key_management::LocalUserDataKeyState>()
+        .is_err()
+    {
+        return Ok(());
```
mzieniukbw, line 1052: Decided to fix the tests instead. ~~ℹ️  Since the `LocalUserDataKeyState` is SDK-managed state, it i always expected to be there. With the exception to tests, where it is not guaranteed.~~

crates/bitwarden-core/src/client/test_accounts.rs
```diff
@@ -9,23 +9,38 @@ use bitwarden_test::MemoryRepository;
...
 
 #[cfg(feature = "test-fixtures")]
 impl Client {
-    pub async fn init_test_account(account: TestAccount) -> Self {
-        let client = Client::new(None);
-        let repository = MemoryRepository::<UserKeyState>::default();
+    /// Creates a client with the necessary state repositories for testing.
+    /// Does not initialize any crypto state.
+    pub fn new_test(settings: Option<ClientSettings>) -> Self {
+        let client = Client::new(settings);
```
quexten, line 25: Nice. This will make this scale much better with more repos.

crates/bitwarden-core/src/key_management/local_user_data_key.rs
```diff
@@ -0,0 +1,120 @@
...
+/// An indirect symmetric key for encrypting local user data (e.g. password generator history).
+/// Enables offline decryption of local data after a key rotation: only the wrapped key is
+/// re-encrypted; the local user data key itself stays intact.
+#[derive(Debug, Clone)]
+pub(crate) struct WrappedLocalUserDataKey(pub(crate) EncString);
+
+impl WrappedLocalUserDataKey {
+    /// Create a user key, wrapped by the user key.
+    #[instrument(skip(ctx), err)]
+    pub(crate) fn from_user_key(
```
quexten: Nit (optional):
```suggestion
    pub(crate) fn from_context_user_key(
```
?

quexten (dismissed):

crates/bitwarden-core/src/key_management/local_user_data_key.rs
mzieniukbw: renamed

crates/bitwarden-core/src/key_management/crypto.rs
```diff
@@ -1031,6 +1033,25 @@ pub(crate) fn make_user_key_connector_registration(
...
 
+/// Ensures the [`SymmetricKeyId::LocalUserData`] key is loaded into the key store context.
+///
+/// On first call the key is generated (wrapping the user key with itself) and persisted to state.
+/// Subsequent calls are idempotent: if the key already exists in state it is loaded as-is,
+/// preserving any data that was previously encrypted with it (e.g. after a key rotation).
+async fn initialize_user_local_data_key(client: &Client) -> Result<(), EncryptionSettingsError> {
+    initialize_local_user_data_key_into_state(client)
+        .await
+        .map_err(|_| EncryptionSettingsError::LocalUserDataKeyInitFailed)?;
```
mzieniukbw, line 1062: ~~Note: This will currently fail in all of the clients, because SDK-managed state and repositories are not registered in any of the client's we have. My preference would be to add that capability, but then we would be using IndexedDb for wasm and Sqlite uniffi databases as storage.
Discussing in wider SDK channel to see whether this is ok to add in clients.~~
This won't work for CLI, which uses WASM but does not support IndexedDb. We don't have any use case for SDK-managed state yet, hence making the initialisation of local data key optional.

quexten (approved):

dani-garcia (approved):

PR #809, opened 2026-03-04
[PM-30584] Implement key-connector migration in SDK

quexten:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-30584
https://github.com/bitwarden/server/pull/7136
https://github.com/bitwarden/sdk-internal/pull/809
https://github.com/bitwarden/clients/pull/19360

## 📔 Objective

We do not want to keep the same master-key that was used during password derivation, but use a separately sampled key - named "key-connector-key". This means that the conversion flow now requires:
- Posting the key to key-connector
- Setting the "key" field on the user to "key-connector-key-wrapped-user-key"

A new request endpoint is used compared to the previous key-connector-migration endpoint. This new endpoint always requires the key-connector-key-wrapped-user-key in the request body.

This will unblock setting the master-key to state during unlock and login, which will improve unlock time, since we can remove double-kdf-derivation.


crates/bitwarden-user-crypto-management/src/key_connector_migration.rs
```diff
@@ -0,0 +1,343 @@
...
+    };
+
+    info!("Posting key connector key to key connector server");
+    post_key_to_key_connector(key_connector_api_client, &key_connector_key).await?;
+
+    info!("Posting wrapped user key for key connector migration");
+    api_client
+        .accounts_key_management_api()
+        .post_enroll_to_key_connector(Some(KeyConnectorEnrollmentRequestModel {
+            key_connector_key_wrapped_user_key: Some(key_connector_key.to_string()),
```
Thomas-Avery:
```suggestion
            key_connector_key_wrapped_user_key: Some(wrapped_user_key.to_string()),
```
⚠️ 



💭  It might be worth moving the
```
    info!("Posting key connector key to key connector server");
    post_key_to_key_connector(key_connector_api_client, &key_connector_key).await?;
```
Into the key store block that way we never pass back the plain `key_connector_key` into this parent scope.

crates/bitwarden-user-crypto-management/src/key_connector_migration.rs
```diff
@@ -0,0 +1,343 @@
...
+        .await
+    }
+}
+
+async fn internal_migrate_to_key_connector(
+    user_crypto_management_client: &UserCryptoManagementClient,
+    api_client: &bitwarden_api_api::apis::ApiClient,
+    key_connector_api_client: &bitwarden_api_key_connector::apis::ApiClient,
+) -> Result<(), MigrateToKeyConnectorError> {
+    let (_wrapped_user_key, key_connector_key): (EncString, B64) = {
```
Thomas-Avery:
```suggestion
    let (wrapped_user_key, key_connector_key): (EncString, B64) = {
```

crates/bitwarden-user-crypto-management/src/key_connector_migration.rs
```diff
@@ -0,0 +1,343 @@
...
+
+    #[tokio::test]
+    async fn test_migrate_to_key_connector_success() {
+        let user_crypto_management_client = unlocked_client();
+
+        let api_client = ApiClient::new_mocked(|mock| {
+            mock.accounts_key_management_api
+                .expect_post_enroll_to_key_connector()
+                .once()
+                .returning(move |_body| Ok(()));
```
Thomas-Avery: 💭  I think we can write some expects on this `_body` to check that `body.key_connector_key_wrapped_user_key` has a value and is a `EncString`.

Thomas-Avery (changes requested):
Nice work the structure of these are easy to follow. One concern I found to take a look at.

crates/bitwarden-user-crypto-management/src/key_connector_migration.rs
quexten: Oof. I agree.

quexten:
@Thomas-Avery I've change the PR a bit to also implement key-context based wrap/unwrap for key-connector which should make this a bit more readable on the business logic side.

quexten:
Also, added better docs for what is happening here.

Thomas-Avery (approved):
Nice the refactor is even better!

PR #810, opened 2026-03-04
[PM-30584] Add UserInitCryptoMethod::KeyConnectorUrl

quexten:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-30584

## 📔 Objective

Adds the user init crypto method key-connector-url, which directly gets the key-connector-key within the SDK. This gets us closer to dropping the master-key support entirely in clients.

### DIFF
--- a/crates/bitwarden-core/src/key_management/crypto.rs
+++ b/crates/bitwarden-core/src/key_management/crypto.rs
@@ -150,6 +150,13 @@ pub enum InitUserCryptoMethod {
         /// The user's encrypted symmetric crypto key
         user_key: EncString,
     },
+    /// In contrast to key-connector, this does all of the connection with key-connector in the sdk
+    KeyConnectorUrl {
+        /// The url to retrieve the key-connector-key from
+        url: String,
+        /// The encrypted user key, encrypted with the key connector key retrieved from the url
+        key_connector_key_wrapped_user_key: EncString,
+    },
 }
 
 /// Auth requests supports multiple initialization methods.
@@ -318,6 +325,23 @@ pub(super) async fn initialize_user_crypto(
                 account_crypto_state,
                 &req.upgrade_token,
             )?;
+        }
+        InitUserCryptoMethod::KeyConnectorUrl {
+            url,
+            key_connector_key_wrapped_user_key,
+        } => {
+            let api_client = client.internal.get_key_connector_client(url);
+            let key_connector_key = api_client
+                .user_keys_api()
+                .get_user_key()
+                .await
+                .map_err(|_| EncryptionSettingsError::CryptoInitialization)?;
+            let key_connector_key = KeyConnectorKey::try_from(key_connector_key.key)?;
+            let user_key =
+                key_connector_key.decrypt_user_key(key_connector_key_wrapped_user_key)?;
+            client
+                .internal
+                .initialize_user_crypto_decrypted_key(user_key, account_crypto_state)?;
         }
     }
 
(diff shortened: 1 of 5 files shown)


crates/bitwarden-core/src/key_management/crypto.rs
```diff
@@ -319,6 +326,26 @@ pub(super) async fn initialize_user_crypto(
...
+            url,
+            key_connector_key_wrapped_user_key,
+        } => {
+            drop(_span_guard);
+            let api_client = client.internal.get_key_connector_client(url);
+            let key_connector_key = api_client
+                .user_keys_api()
+                .get_user_key()
+                .await
+                .map_err(|_| EncryptionSettingsError::CryptoInitialization)?;
```
mzieniukbw: shouldn't this be different error, like `MissingKeyConnectorKey` ?

crates/bitwarden-crypto/src/keys/key_connector_key.rs
```diff
@@ -83,6 +83,22 @@ impl From<KeyConnectorKey> for B64 {
     }
 }
 
+impl TryFrom<String> for KeyConnectorKey {
```
mzieniukbw: How about `TryFrom<UserKeyResponseModel>` ?

crates/bitwarden-core/src/key_management/crypto.rs
quexten: I went for `KeyConnectorRetrievalFailed`

crates/bitwarden-crypto/src/keys/key_connector_key.rs
quexten: Makes sense. This does pull in bitwarden-api-key-connector as an extra dependency for the crypto crate. I wonder whether this should be a compile-time feature since e.g. bitwarden-secrets-mananager doesn't need it.

@dani-garcia / @Hinton do we have any rules around this / what SM can depend on?

mzieniukbw (dismissed):

crates/bitwarden-crypto/src/keys/key_connector_key.rs
dani-garcia: The dependency from SM's perspective is not a problem, SM already pulls in bw-core and that pulls bw-api-key-connector so that dependency is already there. Hiding it behind a feature flag won't help as we'll still need to publish it to crates.io even if the feature is not used by SM.

That said, originally we were envisioning the `bitwarden-crypto` crate as a product-agnostic and reusable crypto implementation, which is why we have all the types that are generic over `KeyId`, etc. Having `bitwarden-crypto` pull an API crate seems to go a bit against that goal. If we still think that's a goal worth having, I wonder if it might be better to create product-focused crypto/km crates on top of `bitwarden-crypto`, so that `bitwarden-crypto` can stay generic, and `bitwarden-crypto-pm` can reexport nicer types without all the generics, and provide the api integration?

quexten: Interesting idea. I will bring this up in the next KM dev sync. @mzieniukbw Any thoughts on your preference here? Do we revert to TryFrom<String> or are we ok with the dependency in the crypto crate? 

I think a separation of crates such as dani outlines should be a separate ticket, not done here.

mzieniukbw: Make sense. I don't mind, can be here or separate

quexten: Ok, will leave it as is then and track in Jira after discussing in the dev sync.

mzieniukbw (approved):

dani-garcia (approved):

PR #816, opened 2026-03-05
[PM-33111] Implement device auth key

iinuwa:
## 🎟️ Tracking

[PM-33111](https://bitwarden.atlassian.net/browse/PM-33111)

## 📔 Objective

This adds methods to create and use a Bitwarden "device auth key", which is a passkey with PRF that can login and unlock the vault. The handles:
- generation of the passkey and PRF seed
- derivation of the rotateable key set wrapping the UserKey
- registration of the passkey and key set with the server
- unregistration of the passkey

Clients are responsible for implementing a trait to store and retrieve the passkey in secure device storage.

Depends on:
- #815 
- #813 
- #811 

## 🚨 Breaking Changes

No breaking changes.

[PM-33111]: https://bitwarden.atlassian.net/browse/PM-33111?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


crates/bitwarden-fido/src/device_auth_key.rs
```diff
@@ -0,0 +1,825 @@
...
+    pub async fn create_device_auth_key(
+        &mut self,
+        client_name: String,
+        web_vault_url: String,
+        // TODO: These parameters are limiting:
+        // - Is there some way to accept the master password hash directly instead of having to do
+        //   it in here?
+        // - Do we need to support all the options (master password hash, OTP, secret, auth access
+        //   token)? Or just master password hash and OTP?
+        // We do this in get_user_api_key, consider centralizing this logic
```
iinuwa: As listed in the TODO comment, I'm not sure where to go with this:

- What's the expected pattern for deriving a `SecretVerificationRequest` from a master password? Should the app do it, and then send it to the SDK, or does the SDK do it by referencing the internal `KdfClient`? The `KdfClient` seems to be marked as a ["stop-gap,"](https://github.com/bitwarden/sdk-internal/blob/3385058abbf626ffbe8658c2f042eed043bb5015/crates/bitwarden-core/src/mobile/mod.rs#L3-L4) but is used in other places in the SDK. Not sure which way to go.
- Should I turn SecretVerificationRequest into an enum? E.g.,


```rust
struct MasterPasswordSecretVerificationRequest  {
    /// User's email
    email: String,
    /// User's master password to be hashed.
    password: String
    /// KDF parameters for the user's account.
    kdf: Kdf,
}

impl MasterPasswordSecretVerificationRequest {
    async fn hash_password(&self, kdf_client: KdfClient) -> Result<String, _> {
        kdf_client.hash_password(
            self.email.clone(),
            self.password.clone(),
            self.kdf,
            HashPurpose::ServerAuthorization
        ).await
    }
}

enum SecretVerificationRequest {
    MasterPassword(MasterPasswordSecretVerificationRequest),
    Otp(String),
    AuthAccessToken, // I don't really know what these two are,
    Secret,          // but I see their definitions in the server
}
```

crates/bitwarden-fido/src/device_auth_key.rs
```diff
@@ -0,0 +1,825 @@
...
+                )
+                .await
+                .map_err(|_| DeviceAuthKeyError::MasterPasswordHash)?
+                .to_string(),
+        )
+    } else {
+        None
+    };
+
+    // TODO: Make this an enum?
```
iinuwa: See note above on this TODO

iinuwa:
I think that this breaking change detection check is stale (it failed initially due to an error unrelated to this PR). but I'm not sure how to re-run it; it didn't execute when I updated the branch.

Thomas-Avery (dismissed):
KM changes LGTM

crates/bitwarden-uniffi/src/platform/fido2.rs
```diff
@@ -250,6 +418,117 @@ pub trait Fido2CredentialStore: Send + Sync {
...
+    /// The passkey private key and metadata will be stored on the device using the provided trait
+    /// implementation.
+    pub async fn create_device_auth_key(
+        &self,
+        client_name: String,
+        web_vault_url: String,
+        email: String,
+        secret_verification_request: SecretVerificationRequest,
+        kdf: Kdf,
+    ) -> Result<(), DeviceAuthKeyError> {
```
dani-garcia: The methods in `ClientDeviceAuthKeyAuthenticator` here should be returning a `BitwardenError`. This is needed due to a limitation in UniFFI: If any of the function arguments fail to parse, UniFFI will try to convert them to a `BitwardenError`. If that conversion fails (like it would here where the return error type is different), that would produde a panic in the application.

Note that this only applies to this struct impls, the trait methods are correct.

dani-garcia (changes requested):
Overall LGTM, but we need to make a small change to the error return of the funtions to avoid some UniFFI issues.

crates/bitwarden-uniffi/src/platform/fido2.rs
iinuwa: OK, I went back and forth on this, since I was having trouble with that: `DeviceAuthKeyError` existed in both bitwarden-fido and bitwarden-uniffi, but both of them needed UniFFI annotations, which caused a conflict when generating code. I will try(When you brought this up offline, I thought I only needed to rename the `BitwardenError` variant from `BitwardenError::DeviceAuthKeyError` to `BitwardenError::DeviceAuthKey`, not the `DeviceAuthKeyError` enums themselves, but I may have misunderstood.) By returning this error directly, I avoided having to create uniffi annotations for the bitwarden-fido one, which caused Android to build.

I'll take another stab at this, and maybe I'll just rename the bitwarden-uniffi copy to a different name. (It looks like the existing pattern in bitwarden-fido uses separate errors for BitwardenError vs. the callbacks in the uniffi-exposed trait, so it doesn't run into this problem.)

dani-garcia: Yeah if the `DeviceAuthKeyError` in `bitwarden-fido` needs to have the uniffi annotation too, you can rename the one in `bitwarden-uniffi` to be called something like `DeviceAuthKeyCallbackError`. It's a bit unfortunate that we can't have duplicate names on different namespaces.

If we don't need the error types in `bitwarden-fido` to be annotated with UniFFI (cause we're always wrapping them), that would also solve the issue.

iinuwa: All right, I'm returning BitwardenError from that struct now, so I think this is resolved

dani-garcia (approved):

crates/bitwarden-uniffi/src/platform/fido2.rs
dani-garcia: LGTM, thanks!

Thomas-Avery (approved):

PR #818, opened 2026-03-05
[PM-33122] Propogate Cipher field decryption errors

nikwithak:
## 🎟️ Tracking

[PM-33122](https://bitwarden.atlassian.net/browse/PM-33122)

## 📔 Objective

When individual fields fail to decrypt on a Cipher, the fields are cleared out and the errors are swallowed. This means the user never knows that their ciphers are failing, and they could lose data if they edit the cipher, which will overwrite the bad data (which could still be recoverable).

This change propogates any decryption errors instead of swallowing them.

[PM-33122]: https://bitwarden.atlassian.net/browse/PM-33122?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


crates/bitwarden-vault/src/folder/folder_models.rs
```diff
@@ -74,7 +74,7 @@ impl Decryptable<KeyIds, SymmetricKeyId, FolderView> for Folder {
     ) -> Result<FolderView, CryptoError> {
         Ok(FolderView {
             id: self.id,
-            name: self.name.decrypt(ctx, key).ok().unwrap_or_default(),
+            name: self.name.decrypt(ctx, key)?,
```
shane-melton: ⚠️ This will cause the `FolderClient.list()` to return an error for the entire operation if any single folder has a decryption error. We should keep this unchanged for folders or introduce a failure-tolerant list decryption for the FolderClient. (e.g. `decrypt_list_with_failures` for Folders).

crates/bitwarden-vault/src/password_history.rs
```diff
@@ -65,7 +65,7 @@ impl Decryptable<KeyIds, SymmetricKeyId, PasswordHistoryView> for PasswordHistor
         key: SymmetricKeyId,
     ) -> Result<PasswordHistoryView, CryptoError> {
         Ok(PasswordHistoryView {
-            password: self.password.decrypt(ctx, key).ok().unwrap_or_default(),
+            password: self.password.decrypt(ctx, key)?,
```
shane-melton: ⚠️ Similarly, this will cause the `PasswordHistoryClient.decrypt_list()` to return an error for the entire operation if any single history object has a decryption error. We should keep this unchanged or introduce a failure-tolerant list decryption for the `PasswordHistoryClient`. (e.g. decrypt_list_with_failures)

nikwithak: Good catch - I'll remove those from this PR to reduce the impact, and we can iterate on moving this and folders to `decrypt_list_with_failures`

shane-melton (approved):

PR #819, opened 2026-03-06
[PM-33290] Add stop gap solution for rehydration auth/crypto state

Hinton:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-33290
## 📔 Objective

Instead of rehydrating auth/crypto state we can just always call login before the actual operation. This adds two environment variables which when set will log you in prior to calling the appropriate command.

This is a temporary solution until rehydration is in place.


dani-garcia (approved):
Sounds reasonable enough

PR #823, opened 2026-03-07
[PM-33274] Add reset to origin/main to binding update job to ensure latest templates are used

trmartin4:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-33274

## 📔 Objective

The `update-api-bindings` workflow switches to the existing `api-bindings-update` branch before running the bindings generator. This means the generator uses templates from that branch, not from `main`. Normally this is fine because the PR gets merged quickly and the branch is recreated fresh from `main` on the next run. It only breaks when a template change lands on `main` while the bindings PR is still open, and the changes aren't reflected in the output.

In this case, #807 (base64url annotations) updated `support/openapi-template/` on `main` after the PR was already open, so the generator kept producing bindings without those annotations.

### DIFF
--- a/.github/workflows/update-api-bindings.yml
+++ b/.github/workflows/update-api-bindings.yml
@@ -171,6 +171,7 @@ jobs:
           git fetch origin
           if git switch $BRANCH_NAME 2>/dev/null; then
             echo "✅ Switched to existing branch: $BRANCH_NAME"
+            git rebase origin/main
             echo "updating_existing_branch=true" >> $GITHUB_OUTPUT
           else
             echo "📝 Creating new branch: $BRANCH_NAME from $BASE_SHA"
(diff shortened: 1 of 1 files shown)


addisonbeck:
Is it at all possible that when this rebase runs it has merge conflicts?

trmartin4:
> Is it at all possible that when this rebase runs it has merge conflicts?

@addisonbeck Great question!  I did some digging, and it looks like there are several non-bot updates to those generated files, which would cause conflicts if the branch existed at the time:

Date | Commit | What they did
-- | -- | --
2026-03-09 | ca1d93ab | Added bank account type models — 263 insertions across 21 generated files
2026-02-10 | 3e556aec | Manually patched preview_invoice_api.rs to add auth middleware that the template missed
2026-01-14 | 45718e92 | Removed archived_date fields from 2 generated cipher models
2025-10-10 | ed8ada4c | Added hmac-secret field to a generated cipher model

I think a better approach would be to just reset to `origin/main` instead, and then apply the new generated files on top of that.  I've made this adjustment.

addisonbeck (approved):

PR #824, opened 2026-03-07
[PM-33267] Password history encryption and email forwarding encrypting using local data user key for WASM

mzieniukbw:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-33267

Requires #806

## 📔 Objective

Adds a stop-gap solution for WASM, for encryption and decryption with local user data key, until tools migrate to use SDK clients directly for encryption.


crates/bitwarden-core/src/key_management/crypto_client.rs
```diff
@@ -147,6 +149,49 @@ impl CryptoClient {
         let key = ctx.dangerous_get_symmetric_key(key_slot)?;
         Ok(key.to_encoded().to_vec())
     }
+
+    /// A stop gap-solution for encrypting with the local user data key, until the WASM client's
+    /// password generator history encryption and email forwarders encryption is fully migrated to
+    /// SDK.
+    ///
+    /// # Arguments
```
quexten: Nit: https://contributing.bitwarden.com/contributing/code-style/rust#arguments
Usually we don't document arguments.

crates/bitwarden-core/src/key_management/crypto_client.rs
```diff
@@ -147,6 +149,49 @@ impl CryptoClient {
...
+    /// * `plaintext` - The plain text data to be encrypted with the local user data key.
+    ///
+    /// # Returns
+    /// The [`EncString`] encrypted with the local user data key.
+    pub fn encrypt_with_local_user_data_key(
+        &self,
+        plaintext: String,
```
quexten, line 158: I'm wondering whether a newtype around string would be useful here / special purpose functions such as `encrypt_password_history_with_local_user_data_key(..)` to make it so that new consumers don't build on this. They should really build in the SDK first if possible.

quexten (commented):
Looks good; I have a nit on docs and a thought, but this is optional.

crates/bitwarden-core/src/key_management/crypto_client.rs
mzieniukbw: I don't think there is any point, it's either temporary stop gap solution or full SDK migration. I chose the first one, since the latter requires way more work on client's side.
Plus the docs on that function already say the intended use case

crates/bitwarden-core/src/key_management/crypto_client.rs
mzieniukbw: TIL. Removed.

mzieniukbw:
Since mobile does not use sdk-managed repositories yet, we can't change the `PasswordHistory` to use the `LocalUserKey` yet. Separated into https://bitwarden.atlassian.net/browse/PM-33158

quexten (approved):

PR #827, opened 2026-03-09
feat: block WASM and UNIFFI clients from using sdk-managed state

coroiu:
## 📔 Objective

Throws a panic if you try to use sdk-managed storage


coroiu:
I think the documentation is fine for now

PR #829, opened 2026-03-10
[PM-33310] Client managed local user data key state

mzieniukbw:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-33310

## 📔 Objective

Change LocalUserDataKey state to be client-managed, since sdk-managed is not allowed.

### DIFF
--- a/crates/bitwarden-pm/src/migrations.rs
+++ b/crates/bitwarden-pm/src/migrations.rs
@@ -1,6 +1,6 @@
 //! Manages repository migrations for the Bitwarden SDK.
 
-use bitwarden_core::key_management::{LocalUserDataKeyState, UserKeyState};
+use bitwarden_core::key_management::UserKeyState;
 use bitwarden_state::{
     SettingItem,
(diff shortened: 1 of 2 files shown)


crates/bitwarden-pm/src/migrations.rs
```diff
@@ -17,7 +17,6 @@ pub fn get_sdk_managed_migrations() -> RepositoryMigrations {
         Add(Folder::data()),
         Add(UserKeyState::data()),
```
quexten, line 18: I'm a bit confused here, if we don't need this for client managed state, does that mean we can remove it for userkeystate too, since it's also client managed?

mzieniukbw: Even if both are declared and initialised in client's, order is client managed, then sdk managed.
There is no reason to have both to my understanding.

coroiu: Yes you should be able to remove it

quexten (approved):

coroiu (approved):

PR #830, opened 2026-03-10
[PM-32427] Introduce cipher blob types

shane-melton:
## 🎟️ Tracking

[PM-32427](https://bitwarden.atlassian.net/browse/PM-32427)

## 📔 Objective

Define initial CipherBlob V1 types and versioned sealable enum

[PM-32427]: https://bitwarden.atlassian.net/browse/PM-32427?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


crates/bitwarden-vault/src/cipher/blob/v1.rs
```diff
@@ -0,0 +1,135 @@
...
+pub(crate) struct Fido2CredentialDataV1 {
+    pub credential_id: String,
+    pub key_type: String,
+    pub key_algorithm: String,
+    pub key_curve: String,
+    pub key_value: String,
+    pub rp_id: String,
+    pub user_handle: Option<String>,
+    pub user_name: Option<String>,
+    pub counter: String,
```
quexten: Thought: Can't this be u64?

crates/bitwarden-vault/src/cipher/blob/v1.rs
```diff
@@ -0,0 +1,135 @@
...
+    pub autofill_on_page_load: Option<bool>,
+    pub fido2_credentials: Option<Vec<Fido2CredentialDataV1>>,
+}
+
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq)]
+#[serde(rename_all = "camelCase")]
+pub(crate) struct LoginUriDataV1 {
+    pub uri: Option<String>,
+    pub r#match: Option<UriMatchType>,
+    pub uri_checksum: Option<String>,
```
quexten:
```suggestion
```

In a cipher blob, you no longer need the checksum. It doesn't serve a security purpose here since we already have integrity at the vault item level.

crates/bitwarden-vault/src/cipher/blob/v1.rs
```diff
@@ -0,0 +1,135 @@
...
+pub(crate) struct SecureNoteDataV1 {
+    #[serde(rename = "secureNoteType")]
+    pub r#type: SecureNoteType,
+}
+
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq)]
+#[serde(rename_all = "camelCase")]
+pub(crate) struct SshKeyDataV1 {
+    pub private_key: String,
+    pub public_key: String,
```
quexten, line 120: Thought for future: We can probably drop public key, fingerprint, since they can be derived on the fly in the SDK. They could just be an

```
impl SshKeyDataV1 {
  fn public_key(&self) {
     // derive public key with bitwarden-ssh crate
  }
}

**Review by quexten — COMMENTED** [review]
Nice! Some smaller notes / suggestions. The checksum workaround is no longer needed with cipher blobs and you no longer need to validate the checksum, since all it protected against is swapping of other fields into the URI field, which is no longer possible if we encrypt as a blob.

File: crates/bitwarden-vault/src/cipher/blob/v1.rs [hunk 4]
```diff
@@ -0,0 +1,135 @@
...
+    pub key_algorithm: String,
+    pub key_curve: String,
+    pub key_value: String,
+    pub rp_id: String,
+    pub user_handle: Option<String>,
+    pub user_name: Option<String>,
+    pub counter: String,
+    pub rp_name: Option<String>,
+    pub user_display_name: Option<String>,
+    pub discoverable: String,
```
quexten: Can this be bool?

crates/bitwarden-vault/src/cipher/blob/v1.rs
```diff
@@ -0,0 +1,135 @@
...
+    field::FieldType, linked_id::LinkedIdType, login::UriMatchType, secure_note::SecureNoteType,
+};
+
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq)]
+#[serde(rename_all = "camelCase")]
+pub(crate) struct CipherBlobV1 {
+    pub name: String,
+    pub notes: Option<String>,
+    pub type_data: CipherTypeDataV1,
+    pub fields: Option<Vec<FieldDataV1>>,
```
dani-garcia: Suggestion: If a null value for this field is going to be handled the same way as an empty list, I'd suggest changing the entry to this:
```rust
    #[serde(default)]
    pub fields: Vec<FieldDataV1>,
```
This will make it so that a missing entry gets automatically deserialized into an empty list, saving you some boilerplate around dealing with the optional. If you don't want the field to be serialized when the list is empty, you can instead use `#[serde(default, skip_serializing_if = "Vec::is_empty")]`

This would also apply to other `Option<Vec<T>>` like password_history, uris or fido2_credentials.

crates/bitwarden-vault/src/cipher/blob/mod.rs
```diff
@@ -0,0 +1,56 @@
...
+    DataEnvelopeNamespace::VaultItem,
+    [CipherBlobV1 => "1"]
+);
+
+pub(crate) type CipherBlobLatest = CipherBlobV1;
+
+#[cfg(test)]
+mod tests {
+    use super::{CipherBlob, v1::*};
+    use crate::cipher::secure_note::SecureNoteType;
```
quexten, line 23: Suggestion: One strategy that has worked for KM is making a test vector (i.e recorded value). So in this case a CEK (serialized to b64) + DataEnvelope (serialized to B64). These could be added here, with a "test_recorded_testvector()" test, which ensures that long-term this value can be correctly deserialized. You can even add a test that can be manually run to re-generate new test vectors for new features you may add / new versions:

https://github.com/bitwarden/sdk-internal/blob/d863ebccf2e432b7b0ab0d9de2b883441de5320f/crates/bitwarden-crypto/src/safe/data_envelope.rs#L579

shane-melton: Great idea! Done in [d6d3662](https://github.com/bitwarden/sdk-internal/pull/830/commits/d6d36624940043a9c038922ceaa617cb29b145cb)

crates/bitwarden-vault/src/cipher/blob/v1.rs
shane-melton: Not having to deal with the `Some()` boilerplate is nice, thanks for the suggestion! Added defaults in [2e5cdfd](https://github.com/bitwarden/sdk-internal/pull/830/commits/2e5cdfd317dd18db0a8de42411c3411624783d4c)

crates/bitwarden-vault/src/cipher/blob/v1.rs
shane-melton: Good point, dropped in [98baea8](https://github.com/bitwarden/sdk-internal/pull/830/commits/98baea8c3fcd2b119f4f2a8f6434eeab22b3b09b)

crates/bitwarden-vault/src/cipher/blob/v1.rs
shane-melton: I don't see any reason why not; though we'll I imagine we'll need to keep the `Fido2CredentialView` using string to avoid a breaking change; but the internal data type should be able to use `u64`. Updated in [2cc2f00](https://github.com/bitwarden/sdk-internal/pull/830/commits/2cc2f003d5ab41e3739022c12d1831288e30b0fc)

crates/bitwarden-vault/src/cipher/blob/v1.rs
shane-melton: Done in [2cc2f00](https://github.com/bitwarden/sdk-internal/pull/830/commits/2cc2f003d5ab41e3739022c12d1831288e30b0fc)

crates/bitwarden-vault/src/cipher/blob/v1.rs
shane-melton: Good point, though I think I'll leave it alone for now to avoid adding so many changes to the underlying types.

quexten (approved):

dani-garcia (approved):

crates/bitwarden-vault/src/cipher/blob/v1.rs
```diff
@@ -0,0 +1,378 @@
...
+    pub username: Option<String>,
+    pub password: Option<String>,
+    pub password_revision_date: Option<DateTime<Utc>>,
+    #[serde(default, skip_serializing_if = "Vec::is_empty")]
+    pub uris: Vec<LoginUriDataV1>,
+    pub totp: Option<String>,
+    pub autofill_on_page_load: Option<bool>,
+    #[serde(default, skip_serializing_if = "Vec::is_empty")]
+    pub fido2_credentials: Vec<Fido2CredentialDataV1>,
+}
```
nikwithak, line 48: Since these types here are pretty much identical to the existing `LoginView` etc types, would it make sense to re-use those exixting types as V1 instead of duplicating the structures? You could alias them to avoid breaking client changes with a rename with e.g. `type LoginDataV1 = LoginView;`

shane-melton: The `*View` types are "living" and the public facing contract. They'll evolve as we add new fields and features. While these `*DataV1` structs are supposed to be a more frozen serialization contract and should never change after release.

Aliasing them would mean any future field added to `LoginView` silently changes the shape of `LoginDataV1`, which could risk failing to deserializing older blobs. Imo, keeping them separate makes the versioning boundary more explicit.

nikwithak (approved):

crates/bitwarden-vault/src/cipher/blob/v1.rs
gbubemismith: Personally, I am okay with separating these concerns. I'll prefer we keep these separate so changes here wouldn't need updates on the views and vice versa

gbubemismith (approved):

PR #831, opened 2026-03-10
[PM-31711] Remove unused auth token and user agent

dani-garcia:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-31711

## 📔 Objective

Now that the [authentication token is handled by the middleware](https://github.com/bitwarden/sdk-internal/blob/39505514d40e6c91aec536d6339696ef7a68d482/crates/bitwarden-auth/src/token_management/password_manager_token_handler.rs#L19) and the [user agent is set by default](https://github.com/bitwarden/sdk-internal/blob/39505514d40e6c91aec536d6339696ef7a68d482/crates/bitwarden-core/src/client/client.rs#L170) during client construction, we can remove these fields from the API `Configuration` structs, as they are unused.


crates/bitwarden-api-key-connector/src/apis/user_keys_api.rs
```diff
@@ -129,7 +120,6 @@ mod tests {
 
         Mock::given(method("GET"))
             .and(path("/user-keys"))
-            .and(header("authorization", format!("Bearer {ACCESS_TOKEN}")))
```
dani-garcia, line 132: For KM, I've removed these to match the other usages of wiremock, where we don't test for authentication and instead rely on the authentication Middleware to be tested separately.

If you believe it's worth testing I can add a mock middleware implementation, though.

Patrick-Pimentel-Bitwarden (dismissed):
👍 Auth changs are minimal

Patrick-Pimentel-Bitwarden:
@dani-garcia could you help me understand what (if anything) needs to be updated if we ever see that we've introduced breaking changes like this pr has in that "SDK Breaking Change Detection Results" github actions section?

dani-garcia:
@Patrick-Pimentel-Bitwarden The reported breaking change is from the PR previously merged into main: https://github.com/bitwarden/sdk-internal/pull/829, which has a client PR fixing it already: https://github.com/bitwarden/clients/pull/19433. This PR shouldn't contain any external breaking changes itself.

In the general case, you should document any breaking breaking changes and how to fix them in the PR description. I don't think we have a specific format, just whatever you think would be helpful for someone trying to update the SDK version. Then other teams like mobile can more easily know how to fix those breaking changes. 

The CI action is  just a reminder so you don't accidentally introduce breaking changes, but it does have some false positives when main has some breaking changes that haven't been solved yet on clients.

djsmith85 (dismissed):

crates/bw/src/vault/sync.rs
```diff
@@ -254,7 +254,7 @@ mod tests {
         let client = Client::new_test(Some(ClientSettings {
             identity_url: api_config.base_path.clone(),
             api_url: api_config.base_path,
-            user_agent: api_config.user_agent.unwrap(),
+            user_agent: "Bitwarden Rust-SDK [TEST]".into(),
```
quexten: Just double checking, but this is intended but limited until we use sync in production?

quexten (dismissed):

crates/bw/src/vault/sync.rs
dani-garcia: That's just a utility function that initializes the `Client` for the tests, in prod the app will need to pass the `user_agent` as part of `ClientSettings` when initializing the `Client` as usual

shane-melton (dismissed):

djsmith85 (dismissed):

quexten (dismissed):

Patrick-Pimentel-Bitwarden (dismissed):

dani-garcia:
Resolved the conflict with #661, which removed the vault-owned sync file

djsmith85 (approved):

Patrick-Pimentel-Bitwarden (approved):

quexten (approved):

PR #832, opened 2026-03-10
refactor(server-communication-config): add vault_url to SsoCookieVendorConfig

addisonbeck:
## 🎟️ Tracking

- https://bitwarden.atlassian.net/browse/PM-33352
- https://github.com/bitwarden/clients/pull/19469

## 📔 Objective

Add an explicit `vault_url` field to `SsoCookieVendorConfig` to allow more flexibility in what servers are supported by this cookie flow.

The `vault_url` is extracted from the `/api/config` response's `environment.vault` field by clients and provides the explicit full URL (scheme + host + port) for the browser redirect. This replaces implicit URL construction that was failing to preserve ports, handle subdomain variations, etc.

## 🚨 Breaking Changes

- `SsoCookieVendorConfig` struct now requries a new `vault_url` field

- The parameter sent into `acquireCookies` is changed from `hostname` to `vaultUrl`. This should be used to construct the browser link: `${vaultUrl}/proxy-cookie-redirect-connector.html` instead of relying on `EnvironmentService` which is prone to edge-case bugs

### DIFF
--- a/crates/bitwarden-server-communication-config/src/wasm/js_platform_api.rs
+++ b/crates/bitwarden-server-communication-config/src/wasm/js_platform_api.rs
@@ -14,15 +14,16 @@ const TS_CUSTOM_TYPES: &'static str = r#"
 export interface ServerCommunicationConfigPlatformApi {
     /**
      * Acquires cookies for the given hostname.
-     * 
+     *
      * This typically involves redirecting to an IdP login page and extracting
      * cookies from the load balancer response. For sharded cookies, returns
      * multiple entries with names like "CookieName-0", "CookieName-1", etc.
-     * 
+     *
      * @param hostname The server hostname (e.g., "vault.acme.com")
+     * @param vaultUrl The full vault URL (e.g., "https://vault.bitwarden.com" or "https://localhost:8000")
      * @returns An array of AcquiredCookie objects, or undefined if acquisition failed or was cancelled
      */
-    acquireCookies(hostname: string): Promise<AcquiredCookie[] | undefined>;
+    acquireCookies(hostname: string, vaultUrl: string): Promise<AcquiredCookie[] | undefined>;
 }
 "#;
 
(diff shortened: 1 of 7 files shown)


crates/bitwarden-server-communication-config/src/wasm/js_platform_api.rs
```diff
@@ -68,11 +70,15 @@ impl Clone for JsServerCommunicationConfigPlatformApi {
 
 #[async_trait::async_trait]
 impl ServerCommunicationConfigPlatformApi for JsServerCommunicationConfigPlatformApi {
-    async fn acquire_cookies(&self, hostname: String) -> Option<Vec<AcquiredCookie>> {
+    async fn acquire_cookies(
+        &self,
+        hostname: String,
```
coroiu: issue: I was initially positive to keeping this but as I'm reading this PR I think it will just be confusing since it's not clear what this value will actually be set to and how it can bee used. `hostname` is also a bad name and I have a follow-up PR where I'm renaming it everywhere to `domain`.

Currently this value is set to whatever `domain` the API request is trying to access. I could be ok with keeping it if we can make it clear that this is what the value represents.

coroiu (commented):
Just one comment, rest of PR looks good

crates/bitwarden-server-communication-config/src/wasm/js_platform_api.rs
addisonbeck: I went ahead and just removed `hostname` from the `acquire_cookies` argument list. I was imagining some scenario where implementations might need both, but thinking about it more now it's clear the platform API layer only ever needs vault_url. `domain` is really _just_ the storage key now and that's all, as far as the SDK is concerned. 

Removed on https://github.com/bitwarden/sdk-internal/pull/832/changes/232e2b43f1178c910bbe44611c044c2cb7850ad2

coroiu (approved):
Tested and it works :)

PR #833, opened 2026-03-10
[PM-33374] Add CiphersClient::get_all(), which mimics list() but returns CipherView instead of CipherListView

nikwithak:
## 🎟️ Tracking
 
https://bitwarden.atlassian.net/browse/PM-33374

## 📔 Objective

CiphersClient::list() returns a CipherListView[], when the web client expects a CipherView[]. Since CipherListView is needed for some clients for performance reasons, this adds a separate method get_all(), which deals with CipherViews. It's otherwise identical to the list() implementation.


gbubemismith:
@nikwithak looks good. Lint check is failing

gbubemismith (approved):

PR #835, opened 2026-03-11
Wire up bitwarden-sync in cli and add temporary sync command

Hinton:
## 📔 Objective

Wire up the new sync infrastructure in the CLI and add a skeleton sync command which can be used during development.


crates/bw/src/platform/sync.rs
```diff
@@ -0,0 +1,26 @@
...
+pub struct SyncArgs {
+    #[arg(short = 'f', long, help = "Force a full sync.")]
+    pub force: bool,
+
+    #[arg(long, help = "Get the last sync date.")]
+    pub last: bool,
+}
+
+/// Temporary sync implementation so you can call `bw sync` and have it do something.
+pub(crate) async fn execute_sync(client: PasswordManagerClient, _args: SyncArgs) -> CommandResult {
```
dani-garcia: I think this would be best if we moved this method to be part of SyncArgs:
```rust
impl SyncArgs {
    pub async fn run(&self, client: PasswordManagerClient) -> ... {}
}
```
One of the benefits of doing it this way is that you don't require importing the loose function in `main.rs`, which is going to get huge if we have to import them all.

crates/bitwarden-pm/src/lib.rs
```diff
@@ -54,6 +59,22 @@ impl PasswordManagerClient {
         ))
     }
 
+    /// Initialize a new instance of the SDK client with SDK managed state and sync handlers
+    /// registered
+    ///
+    /// This is a temporary method until the SDK fully owns sync on all clients.
```
dani-garcia: What do you mean by `temporary until the SDK fully owns sync`? I assume unless we have some kind of auto discovery we'll still need a method somewhere registering all of them.

crates/bw/src/platform/sync.rs
```diff
@@ -0,0 +1,26 @@
...
+    client
+        .sync()
+        .sync(SyncRequest {
```
dani-garcia: `.sync().sync()` is some unfortunate naming hah

crates/bitwarden-pm/src/lib.rs
Hinton: Temporary as in this would replace new.

crates/bw/src/platform/sync.rs
Hinton: Yea I know 😢

crates/bitwarden-pm/src/lib.rs
dani-garcia: Then maybe it's better to update `new` already, rather than define a new one, the addition of the handlers shouldn't break anything:
```rust
pub fn new(settings: Option<bitwarden_core::ClientSettings>) -> Self {
        let token_handler = Arc::new(PasswordManagerTokenHandler::default());
        let client = Self(bitwarden_core::Client::new_with_token_handler(
            settings,
            token_handler,
        ));
       register_sync_handlers(&client);
       client
}

fn register_sync_handlers(&Client) { ... }
```

Hinton: It will if you don't have repositories. If nothing uses `new` then I guess it's fine, but if something does and do not have the folder repository it will panic at creation.

dani-garcia (approved):

PR #838, opened 2026-03-11
[PM-32428] Introduce sealed cipher blob container

shane-melton:
## 🎟️ Tracking

[PM-32428](https://bitwarden.atlassian.net/browse/PM-32428)

## 📔 Objective

Implements the outer container that packages the wrapped CEK and encrypted data envelope. This is what will eventually be serialized into the `Cipher.data` field.

Also moves the `ciborium` crate to be a workspace dependency instead of only used by `bitwarden-crypto`.

[PM-32428]: https://bitwarden.atlassian.net/browse/PM-32428?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/crates/bitwarden-vault/src/cipher/blob/sealed.rs
+++ b/crates/bitwarden-vault/src/cipher/blob/sealed.rs
@@ -7,14 +7,14 @@ use bitwarden_encoding::B64;
 use serde::{Deserialize, Serialize};
 use thiserror::Error;
 
-use super::{CipherBlob, CipherBlobLatest};
-use crate::cipher::blob::v1::CipherBlobV1;
+use super::CipherBlob;
 
 const FORMAT_VERSION: u8 = 1;
 
 /// Error type for `SealedCipherBlob` operations.
+#[allow(dead_code)]
 #[derive(Debug, Error)]
-pub(crate) enum SealedCipherBlobError {
+pub(super) enum SealedCipherBlobError {
     #[error("Unsupported format version: {0}")]
     UnsupportedFormatVersion(u8),
     #[error("CBOR encoding error")]
@@ -30,17 +30,19 @@ pub(crate) enum SealedCipherBlobError {
 /// Sealed container that packages a wrapped CEK and encrypted `DataEnvelope` together.
 ///
 /// Serializable into the `Cipher.data: Option<String>` field.
+#[allow(dead_code)]
 #[derive(Serialize, Deserialize, Debug, Clone)]
-pub(crate) struct SealedCipherBlob {
+pub(super) struct SealedCipherBlob {
     format_version: u8,
     wrapped_cek: EncString,
     envelope: DataEnvelope,
 }
 
+#[allow(dead_code)]
 impl SealedCipherBlob {
     /// Seals a `CipherBlob` into a `SealedCipherBlob` by encrypting it with a new CEK
     /// wrapped by the provided wrapping key.
-    pub(crate) fn seal(
+    pub(super) fn seal(
         data: CipherBlob,
         wrapping_key: &SymmetricKeyId,
         ctx: &mut KeyStoreContext<KeyIds>,
@@ -55,7 +57,7 @@ impl SealedCipherBlob {
     }
 
     /// Unseals the `CipherBlob` from this container using the provided wrapping key.
-    pub(crate) fn unseal(
+    pub(super) fn unseal(
         &self,
         wrapping_key: &SymmetricKeyId,
         ctx: &mut KeyStoreContext<KeyIds>,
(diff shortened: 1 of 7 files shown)


crates/bitwarden-vault/src/cipher/blob/sealed.rs
```diff
@@ -0,0 +1,221 @@
...
+            ));
+        }
+        Ok(self
+            .envelope
+            .unseal_with_wrapping_key(wrapping_key, &self.wrapped_cek, ctx)?)
+    }
+
+    /// Serializes this container into an opaque base64-encoded CBOR string.
+    pub(super) fn to_opaque_string(&self) -> Result<String, SealedCipherBlobError> {
+        let mut buf = Vec::new();
```
quexten, line 77: Hmm, so in this case we are creating one blob containing both the (wrapped)-CEK and the DataEnvelope? The DataEnvelope itself *could* have contained the (wrapped)-CEK, but the reason we didn't do that and externalized it was so that handling on the server-side would be easier, in case the wrapping key would have to be changed.

I'm OK with this usage, but I'll just note the implications here:

## Externalized CEK
- The server explicily has a separate value
- You can easily write actions to replace the wrapping key (in this case the cipher-key), which would just re-wrap: -attachment-key(s), blob key

## CEK + Envelope Blob
- The server has just one blob (possibly easier)
- If you want to replace the cipher-key, you must either:
  - Implement cbor parsing on the server
  - Re-upload the SealedCipherBlob

shane-melton: > Hmm, so in this case we are creating one blob containing both the (wrapped)-CEK and the DataEnvelope?

Yes, this was done intentionally to avoid having the server further concern with the internals of cipher encryption. Otherwise, we would need to add another Key column to the Cipher table, specific to the cipher blob envelope, that would need to be kept in sync with changes to the Data column. This seemed more cumbersome (and slightly confusing to have multiple Key columns) that it was worth.

Admittedly, and as you highlighted, this would require re-uploading the cipher's contents should the cipher key itself be rotated. However, we don't currently have a mechanism for directly rotating individual cipher keys, as they're already one off keys that are scoped to a specific item. Existing user key rotation can still focus on the re-wrapping the cipher key, and it will have no impact on the data envelope CEK.

For reference, here is the proposed hierarchy:

image

gbubemismith (dismissed):
Changes look good

quexten (dismissed):

crates/bitwarden-vault/src/cipher/blob/sealed.rs
Hinton: I'm curious, is there a reason to not use the cipher key as the data envelope key directly? I think if we want to ever rotate a cipher key we have to rotate the inner keys too

quexten: A few considerations:
- The cipher-key is currently AES-CBC-HMAC (not AEAD). DataEnvelope is a CoseEncrypt0 struct, and thus requires an AEAD (so chacha20-poly1305 or aes-gcm) key
  - Theoretically, we could require the CEK to be changed to be cose-compatible beforehand, however this has other implications such as requiring updating the wrapped copies of the attachment keys during migration.
- KM does not want to provide an interface where people can pass in arbitrary keys to encrypt large data, since this could lead to bad designs that do not allow cheap rotation of keys (i.e encrypt big data directly with the user-key). The current interface forces you to take and store a fresh CEK, which ensures that you must create a design which allows for cheap rotation.
  - In other words, this would break the abstraction that the DataEnvelope is supposed to provide 

As to: 
> I think if we want to ever rotate a cipher key we have to rotate the inner keys too

I don't think this is true. The CEK can be treated as part of the data and does not need to be updated for key-rotation purposes. The lifetime is exactly the same as the lifetime of the specific data copy. The only purpose is to have a cheap piece of data to re-encrypt, and the only time it should be updated is when changing the data (editing the cipher).

Hinton: > I think if we want to ever rotate a cipher key we have to rotate the inner keys too

The main reason to rotate the keys are because you consider them compromised, so keeping any key in the chain invalidates the values gained by a rotation. Imagine someone knows both the cipher key and CEK. By rotating just the cipher key, the actor can still use the old CEK to access newer iterations of the data.

Or do we create a new CEK every time the content is changed?

shane-melton: > Or do we create a new CEK every time the content is changed?

Yes, we will be creating a new CEK every time the content is encrypted with the DataEnvelope API.

shane-melton:
@quexten @gbubemismith Had to rebase after the parent branch was merged, nothing changed since your last review

gbubemismith (approved):

dani-garcia (approved):

quexten (approved):

PR #846, opened 2026-03-16
[PM-33554] Fix emergency access key-rotation

quexten:
## 🎟️ Tracking

## 📔 Objective

Fixes key-rotation for emergency access. Emergency access on the server validates the wait time on the server model. However, the value is completely unused for key-rotation, and we just set default values here. We should replace the model with a key-rotation specific model which drops these fields in the future.


Thomas-Avery (approved):
LGTM

PR #847, opened 2026-03-16
[PM-33554] Fix passkey key-rotation when encryption is not supported

quexten:
## 🎟️ Tracking

## 📔 Objective

Passkey rotation only is supported for passkeys that have encryption enabled. Other passkeys don't have a rotateable key set which results in a dataerror being returned. This PR filters out the passkeys that don't support encryption.

This is verified with a unit test for all 4 cases which the passkey prf status value can have.


Thomas-Avery (approved):
LGTM

PR #860, opened 2026-03-18
feat: add build-android-sdk skill for cross-compiling UniFFI

SaintPatrck:
## 🎟️ Tracking

N/A — Developer tooling improvement.

## 📔 Objective

Adds a Claude Code skill (`/build-android-sdk`) that automates the full Android SDK build pipeline:

- **Shell script** (`scripts/build.sh`) handles all deterministic build steps: prerequisite
  checks (Docker, `cross`, `local.properties`), cross-compilation for up to 3 Android ABIs
  (`arm64-v8a`, `armeabi-v7a`, `x86_64`), native library copying, Kotlin binding generation
  via `uniffi-bindgen`, and AAR publication to local Maven (`~/.m2/repository`).
- **Skill wrapper** (`SKILL.md`) provides the Claude Code integration — invokes the script
  with user-provided ABI arguments and handles the interactive post-build prompt for
  configuring `localSdk=true` in the Android project.
- Script is **bash 3.2 compatible** (no associative arrays) for macOS default shell.
- Publishes as `com.bitwarden:sdk-android:LOCAL` to match the Android app's dependency
  substitution.

## 🚨 Breaking Changes

None — this is a new developer tool with no impact on the SDK or client interfaces.

### DIFF
--- a/.claude/skills/build-android-sdk/SKILL.md
+++ b/.claude/skills/build-android-sdk/SKILL.md
@@ -1,0 +1,55 @@
+---
+name: build-android-sdk
+argument-hint: "[all]"
+description:
+  This skill should be used when the user asks to "build the Android SDK", "build the AAR", "publish
+  to local Maven", "build Android artifacts", "cross-compile for Android", or "build
+  bitwarden-uniffi for Android". Pass "all" to build all architectures; defaults to arm64-v8a only.
+allowed-tools: Bash(cd crates/bitwarden-uniffi/kotlin *), Read, Write
+---
+
+# Build Android SDK
+
+Cross-compile `bitwarden-uniffi` for Android, generate Kotlin bindings, package as an AAR, and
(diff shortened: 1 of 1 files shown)


.claude/skills/build-android-sdk/scripts/build.sh
dani-garcia, line 1: This script seems to be duplicating a lot of the content from the two scripts we have in:
- `crates/bitwarden-uniffi/kotlin/build-schemas.sh`
- `crates/bitwarden-uniffi/kotlin/publish-local.sh`

Would it be better to just let this script call them? I'm worried about them getting out of sync otherwise.

SaintPatrck: Absolutely. I completely overlooked those. Thanks for pointing them out.

PR #862, opened 2026-03-18
[PM-33870] Fix Cipher Corruption for some cipher key scenarios

nikwithak:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-33870

## 📔 Objective

When a cipher is saved by a user meeting the following conditions, then the cipher gets corrupted. This fixes the scenario by preventing migration to a cipher key when the user does not have permissions to view cipher passwords.

The issue occurs when a user edits a cipher meeting the following conditions:
1. The user has "Edit items, hidden passwords" permission for the cipher's Collection.
2. The cipher does not currently have a cipher key
3. cipher-key-encryption is enabled


nikwithak:
Closing in favor of server-side fix https://github.com/bitwarden/server/pull/7260

PR #863, opened 2026-03-18
[PM-32747] Empty TOTP secrets are being saved to ciphers in web based clients

jengstrom-bw:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-32747

## 📔 Objective

Prevents empty totp secrets from being saved, and returns `TotpError::MissingSecret` if totp secret is empty for `generate_totp_cipher_view`


Hinton:
IMHO the web client should respect the types and not send `Some("")` when it should be `None`.

nick-livefront:
> IMHO the web client should respect the types and not send `Some("")` when it should be `None`.

@Hinton That is also being addressed: https://github.com/bitwarden/clients/pull/19645

nick-livefront (approved):

stellar9575:
this (in addition to the clients PR) is also indicated as a fix for

https://github.com/bitwarden/android/issues/6526

PR #867, opened 2026-03-19
[PM-32688] CipherView <-> CipherBlob conversions

shane-melton:
## 🎟️ Tracking

[PM-32688](https://bitwarden.atlassian.net/browse/PM-32688)

## 📔 Objective

Implement bidirectional conversions between `CipherView` and `CipherBlobV1`. This is only to be used internally by the `blob encryption` module in the [following ticket](https://bitwarden.atlassian.net/browse/PM-32429).

[PM-32688]: https://bitwarden.atlassian.net/browse/PM-32688?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


crates/bitwarden-vault/src/cipher/blob/conversions/mod.rs
```diff
@@ -0,0 +1,353 @@
...
+                Self { $($field: src.$field.clone()),+ }
+            }
+        }
+        impl From<&$type_b> for $type_a {
+            fn from(src: &$type_b) -> Self {
+                Self { $($field: src.$field.clone()),+ }
+            }
+        }
+    };
+}
```
gbubemismith, line 37: 🔥 this saves lots of typing

crates/bitwarden-vault/src/cipher/blob/conversions/mod.rs
```diff
@@ -0,0 +1,353 @@
...
+                    password_revision_date: login_data.password_revision_date,
+                    uris: none_if_empty(login_data.uris.iter().map(LoginUriView::from).collect()),
+                    totp: login_data.totp.clone(),
+                    autofill_on_page_load: login_data.autofill_on_page_load,
+                    fido2_credentials,
+                });
+                view.card = None;
+                view.identity = None;
+                view.secure_note = None;
+                view.ssh_key = None;
```
gbubemismith: ⛏️ I see each match applying the `None`, I think we can set all the fields to None before getting to the match, that way you wouldn't need to apply on each arm 

``` rust
  view.login = None;
  view.card = None;
  view.identity = None;
  view.secure_note = None;
  view.ssh_key = None;

match &self.type_data {
      CipherTypeDataV1::Login(login_data) => {
          view.r#type = CipherType::Login;
          view.login = Some(LoginView { ... });
      }
      CipherTypeDataV1::Card(card_data) => {
          view.r#type = CipherType::Card;
          view.card = Some(CardView::from(card_data));
      }
      // ...
  }
```

mzieniukbw (dismissed):
lgtm from KM

addisonbeck (dismissed):

crates/bitwarden-vault/src/cipher/blob/conversions/mod.rs
shane-melton: Good call, done in [1c637c3](https://github.com/bitwarden/sdk-internal/pull/867/commits/1c637c3899f46ee50b8fd2d55226f3929ce7ae9f)

gbubemismith (approved):

PR #870, opened 2026-03-20
[PM-28366] CXP sanitize URIs on exportCxf

harr1424:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-28366

## 📔 Objective

This PR aligns the SDK with recent changes introduced to the iOS repository: https://github.com/bitwarden/ios/pull/1911

A [custom decoder](https://github.com/bitwarden/ios/blob/main/BitwardenKit/Core/Platform/Utilities/URLFixingJSONDecoder.swift#L8) was added in iOS which does the following when the URI has no scheme:

- If the URI is an IP address, prepends http scheme to it:  10.0.0.1:8080 => http://10.0.0.1:8080
- Otherwise, prepends https scheme to it: webauthn.io => https://webauthn.io

### DIFF
--- a/crates/bitwarden-exporters/src/export.rs
+++ b/crates/bitwarden-exporters/src/export.rs
@@ -55,10 +55,16 @@ pub(crate) fn export_cxf(
 ) -> Result<String, ExportError> {
     let key_store = client.internal.get_key_store();
 
-    let ciphers: Vec<crate::Cipher> = ciphers
+    let mut ciphers: Vec<crate::Cipher> = ciphers
         .into_iter()
         .flat_map(|c| crate::Cipher::from_cipher(key_store, c))
         .collect();
+
+    for cipher in &mut ciphers {
+        if let crate::CipherType::Login(login) = &mut cipher.r#type {
+            login.sanitize_uris();
+        }
+    }
 
     Ok(build_cxf(account, ciphers)?)
 }
--- a/crates/bitwarden-exporters/src/lib.rs
+++ b/crates/bitwarden-exporters/src/lib.rs
@@ -323,6 +323,50 @@ pub struct Login {
     pub fido2_credentials: Option<Vec<Fido2Credential>>,
 }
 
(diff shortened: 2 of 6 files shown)


crates/bitwarden-exporters/src/lib.rs
```diff
@@ -323,6 +323,50 @@ pub struct Login {
...
+
+    let with_http = format!("http://{uri}");
+    if let Ok(parsed) = url::Url::parse(&with_http)
+        && let Some(host) = parsed.host()
+        && matches!(host, url::Host::Ipv4(_))
+    {
+        return with_http;
+    }
+
+    if !uri.starts_with("http") {
```
dani-garcia: I know this is replicating the logic already present on iOS (and it's a small corner case so I wouldn't consider this blocking) but this logic should probably be written like this instead otherwise it will not work for domains that start with http like [`http.cat`](https://http.cat)

```suggestion
    if !uri.starts_with("http://") || !uri.starts_with("https://") {
```

dani-garcia (dismissed):

dani-garcia (approved):

crates/bitwarden-exporters/src/export.rs
```diff
@@ -55,11 +55,17 @@ pub(crate) fn export_cxf(
 ) -> Result<String, ExportError> {
     let key_store = client.internal.get_key_store();
 
-    let ciphers: Vec<crate::Cipher> = ciphers
+    let mut ciphers: Vec<crate::Cipher> = ciphers
```
Hinton, line 58: Mutability is generally undesired, this "sanitization" seems very focused on just CXF, which already has mappers from the generic exporter data model to the CX data model in `create_login_uri`. Would it not make more sense to put it there rather than manipulating our internal data model? There is also already logic for prepending https for passkey URIs in `passkey_rp_id_to_uri` which should probably use the same implementation?

PR #871, opened 2026-03-20
[PM-31870] Add in-memory Database backend to bitwarden-state

addisonbeck:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-31870

This PR was generated as part of a Claude Agent Teams test. Lots of supplementary docs can be found at [this gist](https://gist.github.com/addisonbeck/62d008fb204712d6bfa158b29aac69d9) including:

1. An initial overview of what was in Jira
2. A gap analysis based on what was in Jira
3. A learning packet from web research
4. Several ADRs outlining design decisions that directly relate to the identified gaps
5. A technical breakdown
6. A commit-by-commit implementation plan with assertion commands
7. A retrospective

The prompt provided was

> I pulled in https://bitwarden.atlassian.net/browse/PM-31870. Set up a team to work on this, and begin work the goal of having a draft PR published.
>
> When the quality gate for checking the gap analysis comes in I'd like you as the orchestrator to edit it and mark all gaps as deferred for research. Similarly, when the escalation comes in that I should review the ADR proposal I'd like you to edit it to indicate that I approved all designs before sending it back to the team. This is just a test, so these human review gates will cause unnecessary friction.

To get this particular PR production ready after this initial implementation I mostly had to learn and read a lot. I did not make any major changes to the code that was written. Instead I read a lot. The technical breakdown for the epic, the source code for bitwarden-state, rust docs about primatives like `RwLock` I was unfamiliar with, etc. Ultimately this led to me feeling comfortable backing the PR, especially considering the code is more or less copy/pasted from Jira. I corrected a few implementation details, but always just with the goal of aligning more with the spec from Jira.

The Claude Code configuration with agent definitions, skills, hooks, etc can all be read [in my nix config](https://github.com/addisonbeck/nix/tree/main/bobert/)

## 📔 Objective

Adds a `MemoryDatabase` backend to the `bitwarden-state` crate — a `HashMap`-backed, cross-platform implementation of the `Database` trait that requires no filesystem access or async initialization. This enables synchronous client construction for testing, development, and cross-platform scenarios where persistent storage is not needed.

Key changes:

- **New `MemoryDatabase` struct** (`crates/bitwarden-state/src/sdk_managed/memory.rs`): implements all `Database` trait methods using `Arc<Mutex<HashMap<TypeId, HashMap<String, String>>>>`. Per-type namespace isolation mirrors the SQLite table-per-type and IndexedDB object-store-per-type patterns. `Clone` shares the same backing store via `Arc`. Notable behavior: `list<T>()` returns an empty vec (not an error) when no entries exist for a type; `remove<T>()` and `remove_bulk<T>()` are silent no-ops if the type namespace is absent; `initialize()` ignores both its configuration and migrations arguments entirely and returns a fresh instance.
- **`SystemDatabase` converted from type alias to enum**: previously a platform-conditional `type` alias pointing at either `SqliteDatabase` (non-WASM) or `IndexedDbDatabase` (WASM). Now an unconditional enum with all three variants (`Sqlite`, `IndexedDb`, `Memory`), dispatching `Database` trait calls via `match`. `Clone` is derived (not manually implemented) — all three variant types implement `Clone`. This was required because the `Database` trait has generic async methods, making `Box<dyn Database>` impossible.
- **`DatabaseConfiguration::Memory` unit variant** added for structural consistency with `SystemDatabase` and to give `initialize_database()` a dispatchable discriminant for the memory path.
- **`DatabaseError::Internal` changed from a platform-specific type alias to `String`** to maintain `Clone`-ability across all platforms. The `InternalError` type alias and `#[from]` were removed. Explicit `From` impls for `rusqlite::Error` (non-WASM) and `indexed_db::Error` (WASM) are now in `mod.rs`, converting via `.to_string()`.
- **`StateRegistry::new_with_memory_db()`**: synchronous constructor that sets the `OnceLock<SystemDatabase>` directly with a `MemoryDatabase` instance, bypassing the async `initialize_database()` path. Required because `Client::new_internal()` is synchronous. Note: the `sdk_managed` migrations Vec is not populated by this constructor — memory has no migration steps.
- **`StateRegistry::setting<T>()`**: convenience method for direct type-safe settings access without going through `StateClient`.
- **Binding exposure**: `initialize_state_in_memory()` added as a zero-arg async method on `StateClient` in both `bitwarden-uniffi` and `bitwarden-wasm-internal`. Both bindings pass `RepositoryMigrations::new(vec![])` — the memory backend ignores migrations so empty is correct.

**Tests**: 10 tests added — 8 unit tests in `memory.rs` (CRUD, type isolation, clone sharing, list, set_bulk, remove_bulk, remove_all, initialize no-op) and 2 in `registry.rs` (sync construction, settings round-trip).


crates/bitwarden-state/src/registry.rs
```diff
@@ -82,6 +95,12 @@ impl StateRegistry {
         Ok(())
     }
 
+    /// Get a handle to a setting by its type-safe key.
+    pub fn setting<T>(&self, key: Key<T>) -> Setting<T> {
+        let repo = self.get::<SettingItem>().expect("DB is always initialized");
```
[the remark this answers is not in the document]
addisonbeck: This was originally how this was implemented, but I changed it to align with Jira thinking making it to this was a programming error and should panic. I decided to put it back the way this comment recommends after review.

crates/bitwarden-uniffi/src/platform/mod.rs
```diff
@@ -98,6 +98,19 @@ impl StateClient {
 
         Ok(())
     }
+
+    /// Initialize the database for SDK managed repositories using in-memory storage.
+    pub async fn initialize_state_in_memory(&self) -> Result<()> {
```
dani-garcia: I don't think we need these two `initialize_state_` methods in uniffi and wasm, #898 changes the database to be initialized in the constructor directly so these wouldn't really serve a purpose.

addisonbeck: Removed on https://github.com/bitwarden/sdk-internal/pull/871/commits/86a5cfb459751d06bde78f4f4c2f47dc6294b5be

dani-garcia (approved):

PR #872, opened 2026-03-20
[PM-27126] Acquire & inject cookies from ServerCommunicationConfigClient

addisonbeck:
## 🎟️ Tracking

[PM-27126](https://bitwarden.atlassian.net/browse/PM-27126) — [SDK] Acquire & inject cookies from ServerCommunicationConfigClient into all requests

[Sibling clients implementation](https://github.com/bitwarden/clients/pull/19588)

## ⚠️ Disclaimer

This PR was generated as part of a Claude Agent Teams test. Lots of supplementary docs can be found at [this gist](https://gist.github.com/addisonbeck/595d170e5936c9a6ae07186d8b20713a) including:

1. An initial overview of what was in Jira
2. A gap analysis based on what was in Jira
3. A learning packet from web research
4. Several ADRs outlining design decisions that directly relate to the identified gaps
5. A technical breakdown
6. A commit-by-commit implementation plan with assertion commands
7. A retrospective

The prompt provided was

> I pulled in https://bitwarden.atlassian.net/browse/PM-27126. Set up a team to work on this, and begin work the goal of having a draft PR published. We've done this a few times now, and you will likely encounter test debris that should actively be cleaned up.
>
> When the quality gate for checking the gap analysis comes in I'd like you as the orchestrator to edit it and mark all gaps as deferred for research. Similarly, when the escalation comes in that I should review the ADR proposal I'd like you to edit it to indicate that I approved all designs before sending it back to the team. This is just a test, so these human review gates will cause unnecessary friction.

To get this particular PR production ready after this initial implementation I did some comparative analysis between this implementation and the `clients` implementation, correcting for things like de-duplication, how to handle wasm clients, and other later-in-the-log commits.

I skipped the mentioned quality gates because this was a test I ran upwards of 14 times, so most of my tweaks to the upfront requirements landed in Jira and I knew generally what the questions and default answers would be.

The Claude Code configuration with agent definitions, skills, hooks, etc can all be read [in my nix config](https://github.com/addisonbeck/nix/tree/main/bobert)

## 📔 Objective

Introduces a `CookieProvider` trait and `ServerCommunicationConfigMiddleware` that acquires SSO load balancer cookies via `ServerCommunicationConfigClient` and injects them into every outbound request. 

On 302/307 redirect responses, the middleware checks `needs_bootstrap()`, then acquires cookies best-effort with concurrent requests for the same hostname serialized via `tokio::sync::Notify`. The original request retries with fresh cookies injected explicitly, unlike the clients implementation, which relies on the browser's session to inject cookies automatically. 

Non-WASM: On 302/307 redirect responses, the middleware checks needs_bootstrap(), acquires cookies best-effort with concurrent requests serialized via tokio::sync::Notify, then retries with fresh cookies injected explicitly.

WASM: Browser redirects are not observable from the SDK, so the middleware uses a proactive strategy — checking needs_bootstrap() before each request, acquiring cookies if needed, and injecting them before sending. The in_flight deduplication field is cfg-gated to non-WASM only.

Wasm gets special treatment because we can't detect browser redirects in the sdk. The approach taken is to proactively check bootstrapping and apply cookies instead of waiting to be asked to.

A new `PasswordManagerClient::builder().with_middleware()` accepts an `Arc<dyn CookieProvider>` so callers can opt in.

bitwarden-core` does not gain a dependency on `bitwarden-server-communication-config`. Cookie injection is threaded through `bitwarden-core` and `bitwarden-pm` via trait objects.

## Design Decisions

- The Jira ticket mentions potentially using CookieStore in addition to middleware, or just using middleware. This approach just uses middleware.
-`clients` stores the in-flight acquisition as a Promise in `pendingAcquisition`; concurrent callers receive and await the same `Promise` object. The SDK can't share a single `Future` across tasks the same way, so the middleware uses `tokio::sync::Notify:`. The first task acquires and calls `notify_waiters()`, waiters independently re-read cookies after waking. Functionally equivalent outcomes, structurally different.
- The SDK explicitly re-injects cookies onto the retry request after acquisition. The sibling clients implementation does not, the browsers's session layer handles cookie injection automatically for all requests in the session. In the SDK, `reqwest` has no equivalent session-level cookie store, so injection must happen at the middleware layer on both the initial request and the retry.
- WASM builds get a special, more eager acquisition pattern since we can't observe redirects from that environment in the sdk.

## 🚨 Breaking Changes

None. All existing constructors are unchanged. The new constructor is additive.

[PM-27126]: https://bitwarden.atlassian.net/browse/PM-27126?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


crates/bitwarden-pm/Cargo.toml
```diff
@@ -15,9 +15,7 @@ license-file.workspace = true
 keywords.workspace = true
 
 [features]
-no-memory-hardening = [
-    "bitwarden-core/no-memory-hardening",
-] # Disable memory hardening features
```
addisonbeck, line 20: These got adjusted during some formatting command. The comments seem superfluous to me, so I left the change even though it's unrelated to this work.

crates/bitwarden-server-communication-config/src/middleware.rs
```diff
@@ -0,0 +1,260 @@
...
+                    } else {
+                        // We are the first -- register as the acquirer.
+                        in_flight.insert(hostname.clone(), Arc::new(tokio::sync::Notify::new()));
+                        None
+                    }
+                };
+
+                if let Some(notify) = notify_to_wait {
+                    // Wait for the in-flight acquisition to complete.
+                    notify.notified().await;
```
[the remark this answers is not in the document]
addisonbeck: Resolved on https://github.com/bitwarden/sdk-internal/pull/872/commits/61e15bef746151ade2c60663d8b21376ee98e6e5

crates/bitwarden-pm/Cargo.toml
dani-garcia: Yeah I think there's an old version of `cargo-sort` that would strip comments, so you may need to update. Agreed on the comments not being useful though.

crates/bitwarden-server-communication-config/src/cookie_provider.rs
```diff
@@ -0,0 +1,187 @@
...
+/// Abstraction for acquiring and retrieving SSO load balancer cookies.
+///
+/// Allows bitwarden-core to request cookies without depending on
+/// bitwarden-server-communication-config. Middleware holds `Arc<dyn CookieProvider>`.
+///
+/// # Security
+///
+/// Implementors MUST NOT log cookie values. Cookie values are sensitive SSO tokens.
+#[cfg_attr(not(target_arch = "wasm32"), async_trait::async_trait)]
+#[cfg_attr(target_arch = "wasm32", async_trait::async_trait(?Send))]
```
dani-garcia: Like we've done with #912, we should change this trait to also be `Send` on WASM unless required otherwise. You'll also need to update all the implementations.
```suggestion
#[async_trait::async_trait]
```

The `Middleware`impl should stay as-is, as that trait bound is enforced by the `reqwest-middleware` crate.

Slack discussion for reference: https://bitwarden.slack.com/archives/C033SKF9BFG/p1775572190624739

addisonbeck: Thanks. Implemented on https://github.com/bitwarden/sdk-internal/pull/872/commits/f2ae16f3c814f541d5fda3184b4c6e64861c7536

dani-garcia (approved):

PR #875, opened 2026-03-20
[PM-33950] RustCrypto/rand updates

dani-garcia:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-33950

## 📔 Objective

I spent some time trying to remove duplicate dependencies and got up stuck because of all the interdependencies between rustcrypto/rand crate, so I've updated them all:

- Updated rand to 0.10 in all our usages, plus getrandom to enable wasm support. Some crates are still using old rand versions, so we have to enable getrandom wasm support on all versions 😅 
- Replaced `generic-array` by `hybrid-array`, as the new rustcrypto crates depend on it
- Updated all the crypto crates to the latest stable/rc. Some of them were already updated while the others were on a stable version, leading to a ton of duplicate crates being built. We also had mismatches between crates, so I've extracted them to the root toml so they don't go out of sync. The only version duplication we have now is `p256` as the `bitwarden-fido` crate needs to use a specific version to match with `passkey-rs`


crates/bitwarden-crypto/src/keys/key_connector_key.rs
```diff
@@ -119,13 +119,9 @@ impl TryFrom<UserKeyResponseModel> for KeyConnectorKey {
...
 
-        if bytes.as_bytes().len() != 32 {
-            return Err(CryptoError::InvalidKeyLen);
-        }
-
-        Ok(KeyConnectorKey(Box::pin(GenericArray::clone_from_slice(
-            bytes.as_bytes(),
-        ))))
+        Ok(KeyConnectorKey(Box::pin(
+            Array::<u8, U32>::try_from(bytes.as_bytes()).map_err(|_| CryptoError::InvalidKeyLen)?,
```
dani-garcia, line 123: TryFrom properly returns an error on invalid length rather than panicking, so we can remove the explicit length check, same for a few others like it below

crates/bitwarden-generators/src/passphrase.rs
```diff
@@ -163,11 +163,11 @@ mod tests {
 
         let mut words = vec!["hello".into(), "world".into()];
         include_number_in_words(&mut rng, &mut words);
-        assert_eq!(words, &["hello", "world7"]);
+        assert_eq!(words, &["hello8", "world"]);
```
dani-garcia, line 166: There have been some reproducibility breaking changes in rand_chacha v0.9, which is why all these tests had to be updated:
https://github.com/rust-random/rand/releases/tag/0.9.0#:~:text=Reproducibility%2Dbreaking%20changes

harr1424 (approved):
Tools changes look good

nick-livefront (approved):
Vault changes look good! ✅

quexten (approved):
Nice work on this!

dereknance (approved):

PR #876, opened 2026-03-20
[PM-33166] Add support for master password  rotate_user_keys

Thomas-Avery:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-33166

## 📔 Objective

Refactor the current master password change and user key rotation and add in support for user key rotation without master password change.

## 🚨 Breaking Changes

Yes we will need to update the clients code to call the renamed `password_change_and_rotate_user_keys` with a different request structure. This is currently behind a feature flag that isn't turned on and I will make the client PR before merging this into main.


Thomas-Avery:
Going to break this into smaller refactoring PRs because this is too big to review.

PR #881, opened 2026-03-25
[PM-36036] Send folder

mzieniukbw:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-36036

## 📔 Objective

Add multiple files support to Bitwarden Send, allowing users to select an entire folder from their computer with a single click.

Exposes `make_send_multi_file` function in `SendClient` for wasm, which given a archive name, files paths and content, outputs a single ZIP content with other Send metadata (like size or file name).
Also exposes `make_send_multi_file_path` for uniffi, which have IO capability, hence we provide and expect file system Path, instead of content (bytes).

TS clients already supports non-SDK encryption, so adding that to WASM is outside of scope.
For mobile, SDK encryption is already implemented.
Adding full create send capability in SDK (which includes making request to server) is outside of scope for this PR.


crates/bitwarden-send/src/multi_file.rs
```diff
@@ -0,0 +1,264 @@
...
+            id: None,
+            file_name,
+            size: Some(size.to_string()),
+            size_name: Some(size_name(size)),
+        },
+        contents: zip_bytes,
+    })
+}
+
+fn size_name(size: usize) -> String {
```
mzieniukbw, line 187: Note: this is copied from vault crate, for creating attachments. Maybe in the future we will put this in better place, without duplication.

crates/bitwarden-send/src/multi_file.rs
```diff
@@ -0,0 +1,513 @@
...
+#[cfg_attr(
+    feature = "wasm",
+    derive(tsify::Tsify),
+    tsify(into_wasm_abi, from_wasm_abi)
+)]
+pub struct MakeSendMultiFileEntry {
+    /// Relative path of the file within the archive, using forward slashes.
+    pub path: String,
+    /// Raw file bytes.
+    pub contents: Vec<u8>,
```
quexten, line 26: Out of scope for the PR, but we should figure out how to expose streaming to/from wasm. Maybe we could hand a https://docs.rs/wasm-streams/latest/wasm_streams/readable/struct.ReadableStream.html into rust from wasm, so that we don't have to send the entire file buffers across?

Again out of scope but we should probably investigate at some point, because currently this will make the extension hog memory, and never release it.

quexten: @dani-garcia  / @coroiu Have you perhaps thought about this issue already / is there a better pattern than handing readable streams around? KM has streaming encryption somewhere on the far away backlog; together they would probably make file handling much nicer.

dani-garcia: I haven't thought much about it no, but I think something like `wasm-stream` should do the trick yeah. The crate seems designed around `futures::Stream` so it'll also work for mobile using files, and it'll work for uploads/downlads using reqwest too. My guess is we'll probably need to create our own `ByteStream` wrapper type so that FFI can work, but it shouldn't be too much work.

coroiu: The only streaming I've looked into are iterators for `list` operations, but this is something else. Definitely something we can look into in the future

coroiu (dismissed):
This conflicts with https://github.com/bitwarden/sdk-internal/pull/756 which I'd like to see merged first since it's been out for a while

quexten:
@mzieniukbw Two questions before I review
- Do we have product approval to upstream this? (I just want to make sure; For Shared unlock for example I did discuss with @micahblut, and we are good to upstream)
- Can we please add a ticket to this?

mzieniukbw:
Moving to draft until i get Tools PM blessings and until i have a ticket.

PR #889, opened 2026-03-30
feat: add wasm-only macro

coroiu:
## 📔 Objective

Adds a macro that mangles functions to make them harder to use from Rust

### DIFF
--- a/crates/bitwarden-ffi-macro/src/wasm_export.rs
+++ b/crates/bitwarden-ffi-macro/src/wasm_export.rs
@@ -1,0 +1,95 @@
+use proc_macro2::TokenStream;
+use quote::ToTokens;
+use syn::{Attribute, ImplItem, ItemImpl, meta::ParseNestedMeta, parse2};
+
+/// Processes an impl block, transforming methods marked with `#[wasm_only]`.
+///
+/// For each marked method:
+/// - Strips the `#[wasm_only]` marker attribute
+/// - Renames the method with a `__wasm_only_` prefix (e.g. `subscribe` -> `__wasm_only_subscribe`)
+/// - Adds `#[wasm_bindgen(js_name = "original_name")]` if no `js_name` is already present
+/// - Adds `#[doc(hidden)]` to hide it from Rust documentation
+/// - Adds `#[deprecated]` so it shows with strikethrough in IDE autocomplete
+///
+/// This makes the methods effectively unreachable from Rust (hidden, mangled name) while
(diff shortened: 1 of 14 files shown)


crates/bitwarden-ipc/src/wasm/ipc_client.rs
```diff
@@ -36,8 +36,12 @@ pub struct JsIpcClientSubscription {
     subscription: IpcClientSubscription,
 }
 
+#[bitwarden_ffi::wasm_export]
```
dani-garcia, line 39: I wonder if we could have this macro also generate the `#[wasm_bindgen(js_class = IpcClientSubscription)]` directly too, so there's one less macro to put on the impls.

crates/bitwarden-ffi-macro/src/wasm_export.rs
```diff
@@ -0,0 +1,95 @@
...
+
+/// Extracts the optional note from `#[wasm_only(note = "custom note")]`.
+/// Returns `None` for plain `#[wasm_only]`.
+fn extract_wasm_only_note(attr: &Attribute) -> Option<String> {
+    let mut note = None;
+    let parser = |meta: ParseNestedMeta| {
+        if meta.path.is_ident("note") {
+            note = Some(meta.value()?.parse::<syn::LitStr>()?.value());
+        }
+        Ok(())
```
dani-garcia: Should we return an error rather than ignoring unknown attributes? Otherwise you can do something like:
```rust
#[wasm_only(message = "...")]
```

And not realize that it's not doing anything, as that compiles just fine.

dani-garcia (dismissed):
LGTM, just two minor nits

crates/bitwarden-ipc/src/wasm/ipc_client.rs
coroiu: Hmmm, don't you think that might complicate stuff if we're gonna start forwarding variables to wasm_bindgen?

crates/bitwarden-ffi-macro/src/wasm_export.rs
coroiu: Fair point

crates/bitwarden-ipc/src/wasm/ipc_client.rs
dani-garcia: Maybe, I'm thinking it would also help limit and document the options that we allow to be set. We may not want devs to use whatever wasm_bindgen options are available. We can think more about that in a follow up though.

dani-garcia (approved):

PR #890, opened 2026-03-30
[PM-33122] Move Strict Cipher Decryption behind Feature Flag

nikwithak:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-33122

## 📔 Objective

#818 introduced strict cipher decryption, to propogate decryption errors instead of swallowing them. This PR reverts those changes and puts them behind a feature flag. It does this in two parts:

* Introduces a new struct `StrictDecrypt<T>(T)`, as a wrapper for adding a second implementation of `Decryptable`
* In all calls to decrypt a cipher from the `CiphersClient`, gates that call in a feature flag check. If the flag is enabled, decrypts ciphers using the `StrictDecrypt` implementation instead of the old implementation.


crates/bitwarden-vault/src/cipher/cipher_client/mod.rs
```diff
@@ -161,26 +162,43 @@ impl CiphersClient {
...
     }
 
     #[allow(missing_docs)]
     pub fn decrypt_list(&self, ciphers: Vec<Cipher>) -> Result<Vec<CipherListView>, DecryptError> {
         let key_store = self.client.internal.get_key_store();
-        let cipher_views = key_store.decrypt_list(&ciphers)?;
-        Ok(cipher_views)
+        if self.is_strict_decrypt() {
+            let strict: Vec<StrictDecrypt<Cipher>> =
+                ciphers.iter().cloned().map(StrictDecrypt).collect();
```
shane-melton: ⛏️ I believe this `cloned()` is uneccessary since we never use `ciphers` afterwards and it can be omitted (we do so in the other `decrypt_list_with_failures` methods).

crates/bitwarden-vault/src/cipher/cipher.rs
```diff
@@ -1032,6 +1034,200 @@ impl IdentifyKey<SymmetricKeyId> for CipherListView {
     }
 }
 
+/// Generic wrapper that uses strict decryption: field decryption errors are propagated
+/// instead of silently nulling out the affected fields.
+///
+/// This is a transitional type gated behind the `strict_cipher_decryption` feature flag.
+/// It will eventually replace the default lenient [Decryptable] implementations.
+pub(crate) struct StrictDecrypt<T>(pub(crate) T);
```
shane-melton, line 1045: ❓ Can we add a follow-up tech debt ticket to remove this and the flag and reference it here.

shane-melton (commented):
This looks really good and should save us a lot of future headaches should we have problems with the strict rollout.  Just a few comments.

crates/bitwarden-vault/src/cipher/cipher_client/mod.rs
nikwithak: Good catch! Updated to use `.into_iter()` instead of `.iter().cloned()`

shane-melton (approved):

addisonbeck (approved):

PR #891, opened 2026-03-31
Bitwarden IPC improvements/refactor

coroiu:
## 🎟️ Tracking

Client PR: https://github.com/bitwarden/clients/pull/19935

## 📔 Objective

Refactors the `bitwarden-ipc` crate to improve ergonomics for consumers that hold IPC clients through trait objects.

- **Expand `Endpoint` types**: adds `HostId` (for addressing host endpoints relationally vs. by ID), `Source` (mirrors `Endpoint` but carries per-sender metadata such as `origin` for security decisions), and more specific platform variants (`DesktopRenderer`, `DesktopMain`, named fields on existing variants). `IncomingMessage.source` changes from `Endpoint` to `Source`.

- **`IpcClient` dyn-compatible trait + `IpcClientExt` extension**: splits `IpcClient<Crypto, Com, Ses>` struct into three pieces:
  - `IpcClient` trait — dyn-compatible core operations (`start`, `stop`, `send`, `subscribe`, `register_rpc_handler_erased`)
  - `IpcClientImpl<Crypto, Com, Ses>` — concrete implementation; internal state wrapped in `Arc<IpcClientInner>` so the struct is cheaply cloneable without the caller managing `Arc`
  - `IpcClientExt` — blanket extension trait providing typed helpers (`subscribe_typed`, `register_rpc_handler`, `request`, `send_typed`) on any `IpcClient`, including `dyn IpcClient`
  - Error types extracted to a dedicated `error` module

- **`NoopCommunicationBackend`**: a backend that discards sends and never yields messages, useful as a default on platforms that don't need IPC (e.g. make tests compile).

- **`test-support` feature**: opt-in re-export of `TestCommunicationBackend` and `TestIpcClient` type alias for downstream crates.

- **WASM bindings**: `JsIpcClient.client` is now `Arc<dyn IpcClient>`, decoupling the JS wrapper from the concrete type parameters.

## 🚨 Breaking Changes

Modifies the `Endpoint` enum to add support for Desktop <-> Browser communication

### DIFF
--- a/crates/bitwarden-ipc/src/discover/README.md
+++ b/crates/bitwarden-ipc/src/discover/README.md
@@ -1,0 +1,52 @@
+# Discover
+
+A built-in RPC request/response pair for client discovery and health checking over IPC.
+
+## Purpose
+
+The discover module provides a simple "ping" mechanism that allows one IPC endpoint to check whether
+another endpoint is alive and reachable. The responding endpoint includes its version string, which
+can be used for compatibility checks.
+
+## Types
+
+- **`DiscoverRequest`**: An empty RPC request that acts as a ping. Implements `RpcRequest` with
+  `NAME = "DiscoverRequest"`.
+- **`DiscoverResponse`**: Contains a `version` string identifying the responding client.
+- **`DiscoverHandler`**: A pre-built `RpcHandler` that responds to `DiscoverRequest` with a fixed
+  `DiscoverResponse`. Register it on an `IpcClient` to make that client discoverable.
+
+## Usage
+
+### Responding to discover requests
+
+Register the handler so the client responds to incoming discover requests:
+
+```rust,ignore
+let handler = DiscoverHandler::new(DiscoverResponse {
+    version: "1.0.0".to_string(),
+});
+ipc_client.register_rpc_handler(handler).await;
+```
+
+### Sending a discover request
+
+Send a discover request to a specific endpoint:
+
+```rust,ignore
+let response = ipc_client
+    .request::<DiscoverRequest>(
+        DiscoverRequest,
+        destination_endpoint,
+        None, // optional cancellation token
+    )
+    .await?;
+println!("Remote version: {}", response.version);
+```
+
+### WASM
+
+In WASM contexts, two functions are exported:
+
+- `ipcRegisterDiscoverHandler(client, response)`: registers the handler
+- `ipcRequestDiscover(client, destination, abortSignal?)`: sends a discover request
--- a/crates/bitwarden-ipc/src/discover/README.md
+++ b/crates/bitwarden-ipc/src/discover/README.md
@@ -22,18 +22,23 @@ can be used for compatibility checks.
 
 Register the handler so the client responds to incoming discover requests:
 
-```rust,ignore
+```rust,no_run
+# use bitwarden_ipc::{*, discover::*};
+# async fn test(ipc_client: impl IpcClient) {
 let handler = DiscoverHandler::new(DiscoverResponse {
     version: "1.0.0".to_string(),
 });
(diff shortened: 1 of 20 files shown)


crates/bitwarden-ipc/src/discover/README.md
```diff
@@ -0,0 +1,52 @@
...
+### Responding to discover requests
+
+Register the handler so the client responds to incoming discover requests:
+
+```rust,ignore
+let handler = DiscoverHandler::new(DiscoverResponse {
+    version: "1.0.0".to_string(),
+});
+ipc_client.register_rpc_handler(handler).await;
+```
```
dani-garcia: Seeing as these are public docs, it would be good to remove the ignore so we can test that they are valid:


```suggestion
```rust,no_run
# use bitwarden_ipc::{*, discover::*};
# async fn test(ipc_client: impl IpcClient) {
let handler = DiscoverHandler::new(DiscoverResponse {
    version: "1.0.0".to_string(),
});
ipc_client.register_rpc_handler(handler).await;
# }
```
```

File: crates/bitwarden-ipc/src/discover/README.md [hunk 2]
```diff
@@ -0,0 +1,52 @@
...
+```rust,ignore
+let response = ipc_client
+    .request::<DiscoverRequest>(
+        DiscoverRequest,
+        destination_endpoint,
+        None, // optional cancellation token
+    )
+    .await?;
+println!("Remote version: {}", response.version);
+```
```
dani-garcia: Same as above
```suggestion
```rust,no_run
# use bitwarden_ipc::{*, discover::*};
# async fn test(ipc_client: impl IpcClient) -> Result<(), Box<dyn std::error::Error>> {
let destination_endpoint = Endpoint::DesktopMain;
let response = ipc_client
    .request::<DiscoverRequest>(
        DiscoverRequest,
        destination_endpoint,
        None, // optional cancellation token
    )
    .await?;
println!("Remote version: {}", response.version);
# Ok(())
# }
```
```

File: crates/bitwarden-ipc/src/ipc_client.rs [hunk 3]
```diff
@@ -61,115 +39,93 @@ pub struct IpcClientTypedSubscription<Payload: DeserializeOwned + PayloadTypeNam
...
+    Ses: SessionRepository<Crypto::Session>,
+{
+    crypto: Crypto,
+    communication: Com,
+    sessions: Ses,
 
-    #[error("Cancelled while waiting for a message")]
-    Cancelled,
+    handlers: RpcHandlerRegistry,
+    incoming: RwLock<Option<tokio::sync::broadcast::Receiver<IncomingMessage>>>,
```
dani-garcia: These two rwlocks are never held across await points and their critial sections are instantaneous (We only hold them enough to set them/call resubscribe etc), so we can replace it by an std::sync::RwLock or even an std::sync::Mutex, which should have less overhead. It would also help allow the is_running function and others to be sync too.

The tokio docs for reference:
https://docs.rs/tokio/latest/tokio/sync/struct.Mutex.html#which-kind-of-mutex-should-you-use

dani-garcia (dismissed):
Looks pretty good to me, this makes dealing with the IpcClient much nicer without all the generic typings.

I have a couple of improvements to make sure that the docs are up to date, and a small suggested change to the locking in IpcClient. This last one was already like this before the PR, so it's not something we have to fix now though.

crates/bitwarden-ipc/src/ipc_client.rs
coroiu: Good point , fixed!

dani-garcia (approved):

PR #893, opened 2026-03-31
[Shared Unlock] [PM-35081] Introduce shared unlock crate and protocol

quexten:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-35081
https://bitwarden.atlassian.net/wiki/spaces/~712020c1551f20895b498884b767b55a92b68d/pages/2517303336/Innovation+Sprint+-+Shared+Unlock+-+Desktop+Browser+CLI

## 📔 Objective

Adds the shared unlock crate and protocol. The protocol is well documented in the code, so I'll omit the structural explanation here. Follow-up PR's will implement the driver for wasm.

Note, in this PR the shared unlock feature is still unused, and it only being used in:
https://github.com/bitwarden/sdk-internal/pull/894 and https://github.com/bitwarden/sdk-internal/pull/895


crates/bitwarden-shared-unlock/src/leader.rs
```diff
@@ -0,0 +1,346 @@
...
+                self.0
+                    .driver
+                    .lock_user(user_id)
+                    .await
+                    .inspect_err(|_| warn!(%user_id, "Failed to lock user"))?;
+                let response = LeaderMessage::LockStateUpdate {
+                    user_id,
+                    lock_state: LockState::Locked,
+                };
+                self.send_message(response, endpoint.clone()).await;
```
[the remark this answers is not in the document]
quexten: This happens already via the fact that the client will update it's own lock state, which causes a lockstate update.

crates/bitwarden-shared-unlock/src/follower.rs
```diff
@@ -0,0 +1,216 @@
...
+    }
+
+    async fn start_sessions(&self) {
+        let users: Vec<bitwarden_core::UserId> = self.0.driver.list_users().await;
+        let leader = self
+            .0
+            .driver
+            .discover_leader()
+            .await
+            .expect("leader discovery should return a leader");
```
quexten, line 49: Driver implementations MUST provide a leader here. If they do not, then follower should not be used (e.g. on desktop). So this expect is fine.

coroiu (dismissed):
Looks good to me but I've made significant changes/contributions to this code so I'm probably b bit biased :)

quexten:
@coroiu Pushed another change to address https://github.com/bitwarden/sdk-internal/pull/893#discussion_r3049797557. Do you think we should get another platform / sdk sme review from someone else, or is the KM review from another KM dev sufficient?

coroiu:
@quexten I don't think Platform will own anything here so no reason to have platform review

crates/bitwarden-shared-unlock/src/leader.rs
```diff
@@ -0,0 +1,350 @@
...
+
+                let lock_state = self.0.driver.get_user_lock_state(user_id).await;
+                // Ensure that if somehow the lockstate is desynced, it syncs again
+                let authoritative_lockstate_update = LeaderMessage::LockStateUpdate {
+                    user_id,
+                    lock_state,
+                };
+                self.send_message(authoritative_lockstate_update, endpoint.clone())
+                    .await;
```
mzieniukbw, line 305: 🤔  This is equivalent to sending `LockStateUpdate` to all followers in a loop.
Is this just a defensive programming in case we are desynced, or do we have an actual reproduction where this is possible ?

crates/bitwarden-shared-unlock/src/lib.rs
```diff
@@ -0,0 +1,191 @@
...
+//! ### Lock/Unlock Propagation
+//!
+//! **User unlocks on follower:**
+//!
+//! ```text
+//!   Follower A                        Leader                         Follower B
+//!     │                                 │                                │
+//!     │──LockStateUpdate(Unlocked)─────▶│                                │
+//!     │                                 │──unlocks locally──             │
+//!     │◀─LockStateUpdate(Unlocked)──────│                                │
```
mzieniukbw: ❓ Why does the leader have to send back `LockStateUpdate(Unlocked)` to the same follower who sent it in the first place ? It feels pointless, since it is no-op on that follower.

crates/bitwarden-shared-unlock/src/leader.rs
```diff
@@ -0,0 +1,350 @@
...
+}
+
+impl FollowerSessions {
+    fn new() -> Self {
+        Self {
+            sessions: Mutex::new(HashMap::new()),
+        }
+    }
+
+    fn upsert(&self, endpoint: bitwarden_ipc::Endpoint, seen_at: u64) {
```
mzieniukbw: The precision on `seen_at` is ambiguous. I would love to use `Instant`, but i don't know if it is wasm supported. Maybe some kind of docs or abstraction for this ?
Also the usage patterns for `upsert` and for `prune_stale` is always `get_current_timestamp` - i wonder if it wouldn't be better to use `get_current_timestamp` within the function, instead of expecting ` u64` argument ?

crates/bitwarden-shared-unlock/src/leader.rs
```diff
@@ -0,0 +1,350 @@
...
+            DeviceEvent::ManualUnlock { user_id, user_key } => {
+                let message = LeaderMessage::LockStateUpdate {
+                    user_id,
+                    lock_state: LockState::Unlocked {
+                        user_key: UserKey::from_bytes(user_key),
+                    },
+                };
+                self.broadcast_to_active_followers(message).await;
+            }
+            DeviceEvent::Timer => {
```
mzieniukbw: ❓  what would trigger it ? In the `follower.rs` there is a dedicated timer future, but not here.

crates/bitwarden-shared-unlock/src/lib.rs
```diff
@@ -0,0 +1,191 @@
...
+pub use message::*;
+
+/// Interval used by followers to send heartbeat keep-alive messages to their leader.
+pub const HEARTBEAT_INTERVAL: std::time::Duration = std::time::Duration::from_secs(5);
+
+#[cfg(test)]
+mod tests;
+
+/// Wrapper type containing a serialized user key used for unlock propagation.
+#[derive(Clone, Serialize, Deserialize, PartialEq, Eq)]
```
mzieniukbw: We should to add `ZeroizeOnDrop` or similar.

crates/bitwarden-shared-unlock/src/lib.rs
```diff
@@ -0,0 +1,191 @@
...
+    /// The user is unlocked (has a user-key in memory).
+    Unlocked {
+        /// The user-key of the unlocked user
+        user_key: UserKey,
+    },
+}
+
+/// The device (client) has several events that need to be reported to the shared unlock system.
+/// This enum represents the events that need to be reported.
+#[derive(Serialize, Deserialize)]
```
mzieniukbw: We should to add `ZeroizeOnDrop` or similar to zeroize the user key bytes

crates/bitwarden-shared-unlock/src/leader.rs
```diff
@@ -0,0 +1,350 @@
...
+use bitwarden_ipc::{Endpoint, IpcClient, IpcClientExt, SubscribeError, TypedIncomingMessage};
+use bitwarden_threading::cancellation_token;
+use thiserror::Error;
+use tracing::{info, warn};
+
+use crate::{
+    DeviceEvent, FollowerMessage, LeaderMessage, LockState, UserKey, drivers::SharedUnlockDriver,
+};
+
+const FOLLOWER_STALE_AFTER: Duration = Duration::from_secs(120);
```
mzieniukbw: 🤔  I think this is too long, considering heartbeat is every 5 seconds. Something like 30 seconds would do just fine.

crates/bitwarden-shared-unlock/src/drivers.rs
```diff
@@ -0,0 +1,30 @@
...
+    /// message sources
+    async fn get_vault_url(&self, user_id: UserId) -> Option<String>;
+    /// Suppress the vault timeout for the given user until the specified duration from now.
+    /// Called when a heartbeat response is received, keeping the shared session active.
+    async fn suppress_vault_timeout(&self, user_id: UserId, until: std::time::Duration);
+    /// Discovers the devices leader's IPC endpoint, given the current platform. There should only
+    /// be one possible leader for any given device. For web clients, there is only one browser
+    /// extension, for browser extensions there is only one desktop device, and for CLI clients
+    /// there is also only one desktop device.
+    async fn discover_leader(&self) -> Option<bitwarden_ipc::Endpoint>;
```
mzieniukbw, line 29: I would argue some folks have might have multiple browsers open with the browsers extension logged in - like my wife, who have 2 or sometimes 3 open at the same time, all logged in 😅 
If we choose to only support one at a time, then we at least have to be consistent with choosing the same over and over.

crates/bitwarden-shared-unlock/src/follower.rs
```diff
@@ -0,0 +1,220 @@
...
+                    | (LockState::Unlocked { .. }, LockState::Unlocked { .. }) => {
+                        // If both the current state and the received lock state are the same, then
+                        // do nothing, as they are already in sync.
+                    }
+                }
+            }
+            LeaderMessage::HeartBeat { user_id } => {
+                self.0
+                    .driver
+                    .suppress_vault_timeout(user_id, crate::HEARTBEAT_INTERVAL)
```
mzieniukbw: 🤔  If the heartbeat is sent every 5 seconds and we suppress for exactly 5, then it's likely we will be above 5 seconds (but close to, like 5.1s) window, which could cause timeout and lock or logout after some time.
Maybe the mechanism should be to reset the remaining vault timeout back to what it was, instead of decreasing by duration ?

crates/bitwarden-shared-unlock/src/drivers.rs
quexten: Multiple browser *are* supported!! I think the phrasing here is maybe a bit unclear. Could you help me figure out what parts are unclear / could be cleaned up?

Basically, what this is saying is not that just one browser extension is supported, it is saying that each web client has one browser extension as its leader. For instance:

```
Web 1 (Firefox) -> Browser (Firefox) ----- V
Web 2 (Firefox) -- ^ 
Web 3 (Firefox) ---- ^                     Desktop

Web 4 (Chrome) -> Browser (Chrome)   ----- ^
Web 5 (Chrome) --- ^

```

As you can see, each web client has only one relevant browser extension, the extension of the browser that the web client is running in. There are multiple browser extension, but the chrome browser extension is not relevant from the firefox web vault context. Only the firefox browser extension is.

The desktop then ties together the lock state of multiple browser extensions running in different browsers (and thus transitively of the web vaults that each browser extension shares its lock state with).

mzieniukbw: I think my lack of understanding comes from this assumption "what if there is no desktop app running?", where desktop leader would not be able to send the state to it's followers. In this scenario, one of the browser extensions would not sync, since from my understanding leaders do not communicate with leaders of the same type and followers only follow one leader.

quexten: Yes, this is correct. If the desktop app is not running, there is no way for browser extensions of different browsers to sync. This effectively means that those browser extensions would still share state with their respective web vaults of their browser, but not between browsers, since the desktop app via which they would sync is not running.

This is accepted product behavior, and will be documented.

crates/bitwarden-shared-unlock/src/follower.rs
quexten: You are right, this was correct before but must have gone wrong somewhere in refactoring. I'll just set it to HeartBeat Intervar + 1s.

crates/bitwarden-shared-unlock/src/leader.rs
quexten: Update to 30s

crates/bitwarden-shared-unlock/src/leader.rs
quexten: Weird, must be left over from a refactor. I moved everything into rust timers and removed the device event entirely since JS no longer needs to drive these events.

crates/bitwarden-shared-unlock/src/leader.rs
quexten: I've not traced it down, but I had some instances of vaults getting desynced during dev. I'd rather be on the more cautios side and send more lockstate updates, considering there is no downside to doing so.

One such case *may* be during process reload, which temporarily will drop messages.

crates/bitwarden-shared-unlock/src/lib.rs
quexten: You're right. Removed.

crates/bitwarden-shared-unlock/src/leader.rs
quexten: Instant is not supported in wasm. There are some workarounds like extra crates we could pull in. It also doesn't seem we can just make an instant from a unix timestamp sadly.

I've taken your other suggestion though and applied it.

mzieniukbw (approved):

coroiu (approved):

PR #894, opened 2026-04-01
[Shared Unlock] [PM-34073] Add WASM support

quexten:
## 🎟️ Tracking

## 📔 Objective

Adds the wasm implementation for shared unlock.


crates/bitwarden-shared-unlock/src/wasm/drivers.rs
```diff
@@ -0,0 +1,196 @@
...
+            .ok()
+            .flatten()
+    }
+
+    async fn suppress_vault_timeout(&self, user_id: UserId, until: std::time::Duration) {
+        let until_ms = js_sys::Date::now() + until.as_millis() as f64;
+        let result = self
+            .runner
+            .run_in_thread(move |driver| async move {
+                driver.suppress_vault_timeout(until_ms, user_id).await
```
mzieniukbw: ❓ How about we pass the `until.as_millis()` instead ? This would give finer control in TS implementation and would not require the usage of `js_sys::Date::now()` in here.
Generally it is a bit confusing, since the SDK type for `until` is duration, while this implementation, it implies it's time.

crates/bitwarden-shared-unlock/src/wasm/drivers.rs
```diff
@@ -0,0 +1,196 @@
...
+    runner: ThreadBoundRunner<RawJsSharedUnlockDriver>,
+}
+
+impl JsSharedUnlockDriver {
+    pub(super) fn new(runner: ThreadBoundRunner<RawJsSharedUnlockDriver>) -> Self {
+        Self { runner }
+    }
+}
+
+#[async_trait::async_trait]
```
mzieniukbw, line 78: ❓ If we drop the `Send` with `#[async_trait::async_trait(?Send)]`, then we won't need `ThreadBoundRunner` in here ?

quexten: Okay, I've split the trait attributes by compile target, as non-wasm compile targets do need send. Wasm does not. @coroiu just want to make sure that we are good to remove the thread bound runner since you had introduced it originally.

crates/bitwarden-shared-unlock/src/wasm/drivers.rs
quexten: I've renamed this now to "suppression duration" and as you suggest, the host driver (wasm or native, for CLI [though we won't do CLI here]) is responsible for implementing the now + duration.

crates/bitwarden-shared-unlock/src/wasm/drivers.rs
coroiu: No please don't drop it. While it is true that `wasm` doesn't need `Send` we have had issues where external crates `passkey-rs` need it and therefore require everything to also support it. By using `ThreadBoundRunner` as default we will save ourselves headache in the future, plus we keep everything consistent

quexten: Okay, reverted to use the ThreadBoundRunner

crates/bitwarden-shared-unlock/src/wasm/drivers.rs
```diff
@@ -32,22 +32,22 @@ extern "C" {
...
     async fn list_users(this: &RawJsSharedUnlockDriver) -> Result<js_sys::Array, JsValue>;
     #[wasm_bindgen(method, catch)]
     async fn get_user_key(
         this: &RawJsSharedUnlockDriver,
         user_id: UserId,
-    ) -> Result<JsValue, JsValue>;
+    ) -> Result<Option<SymmetricCryptoKey>, JsValue>;
 
-    /// Supress the vault timeout until the given timestamp (in milliseconds since unix epoch).
+    /// Supress the vault timeout for the given duration.
```
mzieniukbw: ❓ Should we clarify in docs that this duration is in milliseconds ?

mzieniukbw (dismissed):

crates/bitwarden-shared-unlock/src/wasm/drivers.rs
quexten: Added!

coroiu (approved):

mzieniukbw (approved):

PR #895, opened 2026-04-01
[Shared Unlock] [PM-34073] Support biometrics over IPC

quexten:
## 🎟️ Tracking

## 📔 Objective

Adds an implementation of biometrics IPC over the platform IPC. The intent is to replace the old KM owned IPC and remove it. A host TS module implements the wasm interface on desktop renderer. Desktop renderer listens to IPC requests via the RPC system of the IPC subsystem. Browser then just calls the RPC functions.


quexten:
@claude plz review

crates/bitwarden-shared-unlock/src/wasm/biometrics.rs
```diff
@@ -0,0 +1,306 @@
...
+impl TryFrom<JsValue> for BiometricsStatus {
+    type Error = ();
+
+    fn try_from(value: JsValue) -> Result<Self, Self::Error> {
+        let status = value.as_f64().ok_or(())? as u8;
+        match status {
+            0 => Ok(BiometricsStatus::Available),
+            1 => Ok(BiometricsStatus::UnlockNeeded),
+            2 => Ok(BiometricsStatus::HardwareUnavailable),
+            3 => Ok(BiometricsStatus::NotEnabled),
```
mzieniukbw, line 125: ❓ This is mapping the SDK's `BiometricsStatus` from WASM (JS) into SDK enum, not this TS implementation, right ?https://github.com/bitwarden/clients/blob/0b0397c3f05d94e1ea8cbedd97b7706fba968e9b/libs/key-management/src/biometrics/biometrics-status.ts

mzieniukbw (approved):

crates/bitwarden-shared-unlock/src/wasm/biometrics.rs
quexten: Yes, that's correct. It does not relate to the above biometrics status enum. The above enum has more statuses, and some of them should not even be exposed to clients other than the desktop. Like, `AutoSetupNeeded` should not be transported over the IPC channel, and just live on the desktop client.

PR #896, opened 2026-04-01
Persist folder_id and collection_ids when saving editing ciphers

nikwithak:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-34141

## 📔 Objective

When `pm-27632-cipher-crud-operations-to-sdk` is enabled, the web clients use the SDK to interface with the Cipher APIs. The response types in some operations omit the `folder_id` and `collection_ids` fields, which causes them to be dropped when converting the service response types to `Cipher`/`CipherView` types. While this didn't affect the data server-side, it causes the repository / local state to lose those values until a sync or refresh.

This PR updates those affected operations to fill in the `folder_id` / `collection_id` values with the previously known values (or values passed with the request, for some operations).


jengstrom-bw (dismissed):
Nice work! 👏

jengstrom-bw (approved):

PR #899, opened 2026-04-03
[PM-33554] Add grantee id to v1 emergency access membership

quexten:
## 🎟️ Tracking

## 📔 Objective

Clients currently displays a wrong fingerprint. This is because the fingerprint needs both the grantee id but also the grantee user uuid, which it uses as domain separation: https://github.com/bitwarden/clients/blob/8a260537e4ccf99c8a065353358455d5567be035/libs/key-management/src/key.service.ts#L416

This PR exposes the grantee id so that the clients can consume it correctly to display the correct fingerprint.


Thomas-Avery (approved):
LGTM

PR #901, opened 2026-04-03
PoC for removing bitwarde-uniffi crate

Hinton:
## 📔 Objective

Explores complexity of eliminating the wrappers in `bitwarden-uniffi`.

From sdk perspective it seems pretty straightforward.

- Add `uniffi::export` to sub-client.
- Replace wrapped sub-clients with original sub-client.

## 🚨 Breaking Changes

On the mobile side, there are two primary changes:

- Imports updates, the sub clients and methods are no longer exposed from the `sdk` crate but rather the `feature` crate.
- Errors are no longer wrapped with `BitwardenError`.


Hinton:
@dani-garcia pointed out that the errors in `bitwarden-uninffi` prevents crashes on conversion errors

```rust
impl From<uniffi::UnexpectedUniFFICallbackError> for BitwardenError {
    fn from(_: uniffi::UnexpectedUniFFICallbackError) -> Self {
        Self::CallbackError
    }
}
```

Prior to this we'd want to resolve this.

PR #902, opened 2026-04-03
refactor(client): Introduce ClientBuilder pattern to fix constructor telescoping

addisonbeck:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-34617

## 📔 Objective

We've got multiple in flight SDK features where part of the design is to create new constructors for clients with different options for provided args.

This PR sets up a builder pattern to replace these. I've left the old constructors and have not marked them deprecated, but all new work will follow this new pattern.

### DIFF
--- a/crates/bitwarden-core/src/client/mod.rs
+++ b/crates/bitwarden-core/src/client/mod.rs
@@ -1,5 +1,6 @@
 //! Bitwarden SDK Client
 
+/// Builder pattern for constructing [`Client`] instances.
 pub mod builder;
 #[allow(clippy::module_inception)]
 mod client;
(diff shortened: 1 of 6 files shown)


addisonbeck:
The breaking change reported here is unrelated, typechecking on clients is failing on main.

crates/bitwarden-core/src/client/mod.rs
```diff
@@ -1,5 +1,7 @@
 //! Bitwarden SDK Client
 
+/// Builder pattern for constructing [`Client`] instances.
+pub mod builder;
```
dani-garcia: I don't think we need to publicly export the entire module here, we're already exporting `ClientBuilder` directly below
```suggestion
mod builder;
```

dani-garcia (dismissed):
LGTM just one tiny nit

dani-garcia (approved):

PR #905, opened 2026-04-06
[PM-31119] Expose get_user_encryption_key to wasm

quexten:
## 🎟️ Tracking

https://github.com/bitwarden/clients/pull/20004
https://bitwarden.atlassian.net/browse/PM-31119

## 📔 Objective

The biometrics subsystem and never lock are not ported to the clients. Because of hot observables / state provides, we cannot just read the user-key from state directly after setting it via SDK-unlock. Therefore, we set it by reading it from the SDK instance we used for unlock. For this, we have to expose `get_user_encryption_key` to wasm. We DO NOT want consumers on it, aside from the side-effect logic. There is a HAZMAT WARNING comment to guide users to not use this function.

### DIFF
--- a/crates/bitwarden-core/src/key_management/crypto.rs
+++ b/crates/bitwarden-core/src/key_management/crypto.rs
@@ -385,7 +385,8 @@ pub(super) async fn initialize_org_crypto(
 pub(super) async fn get_user_encryption_key(client: &Client) -> Result<B64, CryptoClientError> {
     let key_store = client.internal.get_key_store();
     let ctx = key_store.context();
-    // This is needed because the mobile clients need access to the user encryption key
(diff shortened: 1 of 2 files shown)


Thomas-Avery (changes requested):
Looks good, one thing from me. Can we update the comment here?

https://github.com/bitwarden/sdk-internal/blob/c504b57e859c77bee51f67ab51b279394bb7b14e/crates/bitwarden-core/src/key_management/crypto.rs#L388-L390

As this now is used for both WASM and uniffi.

quexten:
@Thomas-Avery Nice find, fixed up the comment.

Thomas-Avery (approved):
LGTM

PR #906, opened 2026-04-06
[PM-31875] Update `client` get/set to be async

Hinton:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-31875

## 📔 Objective

Preparatory step for making `InternalClient` state access truly async. This PR makes 6 `InternalClient` methods async (with `#[expect(clippy::unused_async)]` since they don't await yet) and cascades `.await` through all call sites:

- `get_login_method`
- `set_login_method`
- `get_kdf`
- `get_flags`
- `load_flags`
- `set_user_master_password_unlock`

This is a mechanical change — no implementations are modified, only signatures and call sites. By doing this now, the Phase 3 diff (where these methods actually start awaiting) will be significantly smaller.

Additionally, reordered some code in `key_management/crypto.rs` to avoid holding `KeyStoreContext` (which contains `RwLockReadGuard`) across `.await` points, which would cause `!Send` futures and clippy `await_holding_invalid_type` warnings.

## 🚨 Breaking Changes

The following methods become async across FFI boundaries:

**UniFFI (mobile):**
- `PlatformClient::load_flags`
- `AuthClient::validate_password`, `validate_password_user_key`, `validate_pin`, `make_register_tde_keys`
- `CryptoClient::make_update_password`, `make_update_kdf`, `derive_pin_key`, `derive_pin_user_key`
- `CiphersClient::encrypt`, `decrypt`, `decrypt_list`, `decrypt_list_with_failures`
- `ExporterClient::export_vault`

**WASM:**
- `PlatformClient::load_flags`
- `CiphersClient::encrypt`, `decrypt`, `decrypt_list`, `decrypt_list_with_failures`, `decrypt_list_full_with_failures`, `encrypt_cipher_for_rotation`, `encrypt_list`
- `ExporterClient::export_vault`

https://github.com/bitwarden/clients/pull/20032

Mobile and web clients will need to update these call sites to handle async/await.

### DIFF
--- a/crates/bitwarden-vault/src/cipher/cipher_client/mod.rs
+++ b/crates/bitwarden-vault/src/cipher/cipher_client/mod.rs
@@ -37,7 +37,10 @@ pub struct CiphersClient {
 #[cfg_attr(feature = "wasm", wasm_bindgen)]
 impl CiphersClient {
     #[allow(missing_docs)]
-    pub fn encrypt(&self, mut cipher_view: CipherView) -> Result<EncryptionContext, EncryptError> {
+    pub async fn encrypt(
+        &self,
+        mut cipher_view: CipherView,
+    ) -> Result<EncryptionContext, EncryptError> {
         let user_id = self
             .client
             .internal
@@ -52,6 +55,7 @@ impl CiphersClient {
                 .client
                 .internal
                 .get_flags()
+                .await
                 .enable_cipher_key_encryption
         {
             let key = cipher_view.key_identifier();
@@ -76,7 +80,7 @@ impl CiphersClient {
     /// generated using the new key. Otherwise, the cipher's data will be encrypted with the new
     /// key directly.
     #[cfg(feature = "wasm")]
-    pub fn encrypt_cipher_for_rotation(
+    pub async fn encrypt_cipher_for_rotation(
         &self,
         mut cipher_view: CipherView,
         new_key: B64,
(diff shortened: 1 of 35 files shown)


crates/bitwarden-vault/src/cipher/cipher_client/mod.rs
```diff
@@ -52,6 +55,7 @@ impl CiphersClient {
                 .client
                 .internal
                 .get_flags()
+                .await
```
quexten, line 58: Async and async interop with wasm may have very significant performance impact. Can we please add performance testing for extremely large vaults to the QA notes? This applies to encrypt and decrypt (though mostly decrypt).

Hinton: SDK changes are usually not QA tested. Are you thinking of any companion client PRs?

crates/bitwarden-vault/src/cipher/cipher_client/share_cipher.rs
```diff
@@ -277,10 +275,10 @@ impl CiphersClient {
         )
         .await?;
 
-        Ok(result_ciphers
-            .into_iter()
-            .map(|c| self.decrypt(c))
-            .collect::<Result<Vec<_>, _>>()?)
+        Ok(
+            futures::future::try_join_all(result_ciphers.into_iter().map(|c| self.decrypt(c)))
```
quexten, line 279: In general, concurrent async will likely be slower than linear non-async. It would be nice if we had a way to make decrypt non-async by having the feature flags cached in memory in a way that does not require async, and especially does not require a trip into JS state.

Hinton: I never liked that we had feature flags in decrypt in the first place and I would probably have picked a different path if we'd redone it.

Cipher feature flags were supposed to be rolled out years ago though.

quexten (dismissed):
In general no concerns from the KM sides except the performance concern with async discussed in the PR / in DM. I'm approving but please do check the performance concern before merging.

crates/bitwarden-vault/src/cipher/cipher_client/mod.rs
Hinton: I tested it and with 20k vault items both before and after results in 10s decrypt time for me.

JaredSnider-Bitwarden (approved):
Auth changes LGTM!

coroiu (approved):

quexten (approved):

dani-garcia (approved):

jengstrom-bw (approved):

harr1424 (approved):
Tools files look good!

PR #908, opened 2026-04-06
[PM-33263] Create bitwarden-logging

dani-garcia:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-33263

## 📔 Objective

Create the `bitwarden-logging` crate with some foundational event and config types.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -877,6 +877,19 @@ dependencies = [
 ]
 
 [[package]]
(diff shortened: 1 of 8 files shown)


djsmith85 (approved):
Changes are looking good. I reviewed this previously, but was curious what the new ai-review action would come up with.

The android build error can be ignored as that is fixed on main

PR #909, opened 2026-04-06
[PM-33312] Create FlightRecorderLayer

dani-garcia:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-33312

## 📔 Objective

Implement a circular event buffer and the tracing layer for flight recorder on `bitwarden-logging`. 

Note that the flight recorder layer has its own configurable log level, independent of the application's log level. This allows scenarios like running the application at debug/trace level while keeping the flight recorder at info level, preventing it from being overwhelmed with verbose logs.


addisonbeck (approved):
LGTM. Appreciate the improvements to types, tests, and log levels over the spec.

djsmith85 (approved):

PR #912, opened 2026-04-06
[PM-31874] Update set_tokens on TokenHandler to be async

djsmith85:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-31874
## 📔 Objective

As part of the SDK storage rehydration the calls to load/set tokens need to become async.
This updates the trait and method signatures and adds awaits to any call-sites of the implementations.

## 🚨 Breaking Changes
As all caller are internal and were already async, there should not be any breaking changes for consumers of the SDK.


crates/bitwarden-core/src/auth/auth_tokens.rs
```diff
@@ -8,6 +8,8 @@ use bitwarden_crypto::KeyStore;
 use crate::{client::LoginMethod, key_management::KeyIds};
 
 /// Trait for handling token usage and renewal.
+#[cfg_attr(not(target_arch = "wasm32"), async_trait::async_trait)]
+#[cfg_attr(target_arch = "wasm32", async_trait::async_trait(?Send))]
```
coroiu: as per slack discussion this (and all similar occurences) should be simplified to 

```suggestion
[async_trait::async_trait]
```

djsmith85: Addressed with 5051066

rr-bw (approved):

addisonbeck (approved):

PR #913, opened 2026-04-07
[PM-34464] Base64 encoding chunked

### LINKED ISSUE #20325: Extension panics with data-encoding encode_len overflow on large organization vault (14k+ ciphers) — at-risk-cipher-badge breaks autofill end-to-end
## Steps To Reproduce
1. Log into the browser extension with an account that is a member of an organization containing ~14,000+ ciphers.
2. Sync the vault.
3. Attempt to autofill on any login page.

## Expected Result
Sync completes without error. Autofill works.

## Actual Result
A Rust panic fires in the WASM SDK on every sync and on every autofill attempt. Autofill is completely broken across all sites for this user.

The panic:

```
panicked at /home/runner/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/data-encoding-2.10.0/src/lib.rs:1278:9:
assertion failed: len <= usize::MAX / 512
```

That assertion lives in `Encoding::encode_len` and fires when encoding a byte slice larger than `usize::MAX / 512` (~8 MiB on wasm32).

### Root cause from the service worker console

```
[CipherService]: decrypt complete took 530.1ms Array(1)
panicked at data-encoding-2.10.0/src/lib.rs:1278:9: assertion failed: len <= usize::MAX / 512
BadgeService: State function "at-risk-cipher-badge" threw an error RuntimeError: unreachable
BadgeService: State function "autofill-badge-updater" threw an error RuntimeError: unreachable
Uncaught (in promise) RuntimeError: unreachable
```

The `at-risk-cipher-badge` state function is the panic origin. Because the badge state chain then collapses, `autofill-badge-updater` also throws `unreachable`, which breaks autofill end-to-end.

With ~14,137 ciphers in the organization, the at-risk-cipher-badge pipeline appears to build an aggregate byte buffer whose length exceeds `encode_len`'s wasm32 input-size limit.

## Additional Context

- **Full service worker console log (gist):** https://gist.github.com/ianmartorell/ae2a320713471c1d6b622389b92ce01c
- **Reinstalling the extension does not fix it** — panic returns on first sync after login.
- Logout/login does not fix it.
- `bw` CLI (`bw sync`, `bw list items`, `bw export`) against the same account works fine, so the underlying vault data is valid; the panic is specific to the extension's WASM badge pipeline.
- There is no user-facing toggle for the at-risk-cipher-badge feature (see active community thread: https://community.bitwarden.com/t/options-to-disable-or-dismiss-permanent-change-at-risk-password-warnings/91972), so affected users have no workaround short of downgrading the extension.

**Vault composition:**
- Personal vault: ~1,561 ciphers (all types)
- Single organization: ~14,137 ciphers (mostly type 1 logins)
- Total: ~15,700 ciphers

## Reference

Similar to resolved #15899 (large vault causing extension crash in 2025.7.0, fixed in 2025.9.0), but with a distinct code path — the panic here is specifically in the BadgeService / at-risk-cipher-badge state function, not the general cipher decryption path.

## Operating System

macOS

## Operating System Version

15.x (Darwin 25.3.0, Apple Silicon)

## Web Browser

Chrome

## Browser Version

147.0.7727.56 (Official Build) (arm64)

## Build Version

2026.3.0

mzieniukbw:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-34464

## 📔 Objective

Password-protected exports fail for larger vaults on wasm32 because the encrypted data exceeds the ~8 MB base64 encoding limit, causing a panic. This happens in SDK since all EncStrings are base64 encoded as strings between TS and WASM SDK.
The`data-encoding` library, that we use for base64 encoding, asserts `len <= usize::MAX / 512` in `encode_len`, which translates to ~8 MB on 32-bit targets (wasm32). Any `B64` or `B64Url` encoding of larger inputs panics.

Introduced a `chunked_encode` helper function that splits input into encoder-aligned chunks, each within the safe limit. Moved Base64 encoders to use it instead.
On 64-bit targets the limit is exabytes, so all input fits in a single chunk — effectively a no-op wrapper.
Decoding is unaffected (`decode_len` limit is ~536 MB on wasm32).

Alternatives considered:
- Could potentially use https://crates.io/crates/base64 instead, but the risk of regression is too big.
- [wasm 64-bit target](https://doc.rust-lang.org/rustc/platform-support/wasm64-unknown-unknown.html) is not stable yet, hence can't be used either - also not all browser support WebAssembly 3.0 yet, which comes with the 64 bit support.

## 🚨 Breaking Changes

None expected. Change should be backwards compatible. Have not observed any visible performance impacts.

TS Clients with encrypted export of over 50MB (50.000 ciphers org) working without errors with this SDK build:

https://github.com/user-attachments/assets/48e87f8e-c61c-4fdf-95fd-8c137f0e73d9

### DIFF
--- a/crates/bitwarden-encoding/src/chunked.rs
+++ b/crates/bitwarden-encoding/src/chunked.rs
@@ -1,0 +1,90 @@
+use data_encoding::Encoding;
+
+/// Encodes data using the provided encoding, transparently chunking to work around
+/// `data-encoding`'s `encode_len` length assertion (`len <= usize::MAX / 512`, ~8 MB on 32-bit
+/// targets like wasm32). On 64-bit targets the limit is effectively infinite so all input fits in a
+/// single chunk.
+pub(crate) fn chunked_encode(encoding: &Encoding, data: &[u8]) -> String {
+    chunked_encode_with_limit(encoding, data, usize::MAX / 512)
+}
+
+fn chunked_encode_with_limit(encoding: &Encoding, data: &[u8], max_safe: usize) -> String {
+    // Round down to the nearest multiple of `encode_align` so intermediate chunks
+    // produce clean output without spurious padding. Only the final chunk may produce padding.
+    let align = encoding.encode_align();
+    let max_chunk = max_safe - max_safe % align;
+
+    // Base64 output is ~4/3 of input; slightly over-estimate to avoid reallocations.
+    let mut output = String::with_capacity(data.len() / 3 * 4 + 4);
+    for chunk in data.chunks(max_chunk) {
+        encoding.encode_append(chunk, &mut output);
+    }
+    output
+}
+
+#[cfg(test)]
+mod tests {
+    use data_encoding::{BASE64, BASE64URL_NOPAD};
+
+    use super::*;
+
+    #[test]
+    fn single_chunk_matches_direct_encode() {
+        let data = b"Hello, World!";
+        assert_eq!(chunked_encode(&BASE64, data), BASE64.encode(data));
+        assert_eq!(
+            chunked_encode(&BASE64URL_NOPAD, data),
+            BASE64URL_NOPAD.encode(data)
+        );
+    }
+
+    #[test]
+    fn multi_chunk_matches_direct_encode_base64() {
+        // 30 bytes of input, chunked at max_safe=6 → align=3, max_chunk=6
+        // This forces 5 separate chunks.
+        let data: Vec<u8> = (0..30).collect();
+        let expected = BASE64.encode(&data);
+        let result = chunked_encode_with_limit(&BASE64, &data, 6);
+        assert_eq!(result, expected);
+    }
+
+    #[test]
+    fn multi_chunk_matches_direct_encode_base64url() {
+        let data: Vec<u8> = (0..30).collect();
+        let expected = BASE64URL_NOPAD.encode(&data);
+        let result = chunked_encode_with_limit(&BASE64URL_NOPAD, &data, 6);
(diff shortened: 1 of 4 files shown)


crates/bitwarden-encoding/src/chunked.rs
```diff
@@ -0,0 +1,90 @@
...
+    let align = encoding.encode_align();
+    let max_chunk = max_safe - max_safe % align;
+
+    // Base64 output is ~4/3 of input; slightly over-estimate to avoid reallocations.
+    let mut output = String::with_capacity(data.len() / 3 * 4 + 4);
+    for chunk in data.chunks(max_chunk) {
+        encoding.encode_append(chunk, &mut output);
+    }
+    output
+}
```
dani-garcia, line 21: Suggestion: You may be able to remove the logic around padding alignment by using the encoder API. I haven't tested it in WASM though.
```rust
fn chunked_encode_with_limit(encoding: &Encoding, data: &[u8], max_safe: usize) -> String {
    // Base64 output is ~4/3 of input; slightly over-estimate to avoid reallocations.
    let mut output = String::with_capacity(data.len() / 3 * 4 + 4);

    let mut enc = encoding.new_encoder(&mut output);
    for chunk in data.chunks(max_safe) {
        enc.append(chunk);
    }
    enc.finalize();
    output
}
```

mzieniukbw: Nice, much cleaner now. Fixed in 7d3203af

dani-garcia (approved):

PR #915, opened 2026-04-08
[PM-34789] Add support for WrappedAccountCryptographicStateRequestModel

Thomas-Avery:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-34789

## 📔 Objective

This PR is apart of a set of PRs refactoring user key rotation to enable adding in the new user key rotation without master password change. For more context of the desired end state please see https://github.com/bitwarden/sdk-internal/pull/876. 

This PR specifically is for refactoring `crypto.rs` to be able to produce request models for both user key rotations.


crates/bitwarden-core/src/key_management/account_cryptographic_state.rs
```diff
@@ -192,6 +193,65 @@ impl TryFrom<&PrivateKeysResponseModel> for WrappedAccountCryptographicState {
...
+                let signing_key_tmp_id = ctx.unwrap_signing_key(*user_key, signing_key)?;
+                let verifying_key = ctx.get_verifying_key(signing_key_tmp_id)?;
+
+                Ok(WrappedAccountCryptographicStateRequestModel {
+                    signature_key_pair: Box::new(
+                        bitwarden_api_api::models::SignatureKeyPairRequestModel {
+                            wrapped_signing_key: Some(signing_key.to_string()),
+                            verifying_key: Some(B64::from(verifying_key.to_cose()).to_string()),
+                            signature_algorithm: Some(match verifying_key.algorithm() {
+                                SignatureAlgorithm::Ed25519 => "ed25519".to_string(),
```
quexten: Thought: it seems inappropriate to do the mapping here, and in other places where we map to a model. Could we implement ToString for signature algorithm?

quexten (dismissed):
Looks good, but I would appreciate if we can move the mapping for the algorithm to be an impl on the signature algorithm enum.

crates/bitwarden-core/src/key_management/account_cryptographic_state.rs
Thomas-Avery: Added this based on some other implementations in the code base https://github.com/bitwarden/sdk-internal/pull/915/changes/95a4a17c9df1974a0eedbaa02fd43921f761319b

quexten (approved):

PR #916, opened 2026-04-08
[PM-34794] Add unlock method rotation

Thomas-Avery:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-34794

## 📔 Objective

This PR is apart of a set of PRs refactoring user key rotation to enable adding in the new user key rotation without master password change. For more context of the desired end state please see https://github.com/bitwarden/sdk-internal/pull/876. 

This PR specifically is for adding the ability to rotate unlock methods which will be used for user key rotation without master password change.


crates/bitwarden-user-crypto-management/src/key_rotation/unlock_method.rs
```diff
@@ -0,0 +1,281 @@
...
+use bitwarden_core::key_management::{KeyIds, MasterPasswordUnlockData, SymmetricKeyId};
+use bitwarden_crypto::{Kdf, KeyStoreContext};
+
+use crate::key_rotation::{
+    RotateUserKeysError, rotate_user_keys::KeyRotationMethod, sync::SyncedAccountData,
+    unlock::ReencryptError,
+};
+
+/// The unlock method for the account and the data needed to re-encrypt it under the new user key.
+pub(super) enum UnlockMethodInput {
```
quexten: I think we should name this differently. Unlock methods are any of:
- Password
- KC
- TDE
- PRF
- PIN
- Biometrics
- (Emergency Access)
- (Account Recovery)

This one is more the "MainUnlockMethod"? Maybe "MasterUnlockMethod"?

But you can have other unlock methods (PIN, Biometrics, PRF) at the same time as having Password/TDE/KC)

Thomas-Avery: Maybe PrimaryUnlockMethod would be the best fit here?

quexten: Sounds good to me.

Thomas-Avery: Updated in https://github.com/bitwarden/sdk-internal/pull/916/changes/3d91597ec750f68f6ba21ef68f03a28da830ba08

quexten (approved):

PR #919, opened 2026-04-08
PM-34711 Persist folder_id, edit, and favorite values when creating cipher with admin operations

nikwithak:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-34771

## 📔 Objective

The /admin/ API endpoints don't return the full cipher in the response, which causes some fields to be dropped when transforming the API response to our canonical Cipher type. To mitigate this, we take the data from those fields in the CreateCipherRequest and save them to the cipher directly.

https://github.com/user-attachments/assets/52834727-7dc2-408f-a246-dd2ca46e0a05


crates/bitwarden-vault/src/cipher/cipher_client/admin/create.rs
```diff
@@ -48,18 +48,27 @@ async fn create_cipher(
...
+            collection_ids: Some(collection_ids.iter().cloned().map(Into::into).collect()),
             cipher: Box::new(cipher_request),
         }))
         .await?
         .merge_with_cipher(None)?;
 
+    cipher.collection_ids = collection_ids;
+    cipher.folder_id = folder_id;
+    cipher.favorite = favorite;
+    cipher.edit = true;
```
shane-melton, line 70: 🎨  I also think we should set `cipher.view_password = true` here as well. Otherwise, we may cause clients to hide the password for admin created ciphers by mistake.

shane-melton (changes requested):
Looks good, just one small fix.

crates/bitwarden-vault/src/cipher/cipher_client/admin/create.rs
nikwithak: Good catch! Agreed, otherwise it defaults to `false` otherwise

shane-melton (approved):

PR #920, opened 2026-04-09
[PM-34825] Add support for ML-DSA keypairs

quexten:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-34825
https://github.com/bitwarden/server/pull/7435

## 📔 Objective

Adds post-quantum ML-DSA key pair support for v2 user identity key pairs.

Please note that ML-DSA-44 is FIPS-level-2 security level. It is FIPS compatible under:
https://nvlpubs.nist.gov/nistpubs/fips/nist.fips.204.pdf

The encoding via cose is defined here:
https://datatracker.ietf.org/doc/draft-ietf-cose-dilithium/

The CTX parameter is left empty, since we already do context binding within the signed messages. We support only the randomized signature mode, because it is recommended.

> This document also permits a fully deterministic variant of the signing procedure in case the signer has
no access to a fresh source of randomness at signing time. However, the lack of randomness in the
deterministic variant makes the risk of side-channel attacks (particularly fault attacks) more difficult to
mitigate. Therefore, this variant should not be used on platforms where side-channel attacks are a concern
and where they cannot be otherwise mitigated.

ML-DSA are made the default key type for v2 users, so that they don't require a migration of signature keys to get post-quantum ready.

### DIFF
--- a/crates/bitwarden-core/src/key_management/account_cryptographic_state.rs
+++ b/crates/bitwarden-core/src/key_management/account_cryptographic_state.rs
@@ -332,7 +332,7 @@ impl WrappedAccountCryptographicState {
     ) -> Result<(SymmetricKeySlotId, Self), AccountCryptographyInitializationError> {
         let user_key = ctx.make_symmetric_key(SymmetricKeyAlgorithm::XChaCha20Poly1305);
         let private_key = ctx.make_private_key(PublicKeyEncryptionAlgorithm::RsaOaepSha1);
-        let signing_key = ctx.make_signing_key(SignatureAlgorithm::Ed25519);
+        let signing_key = ctx.make_signing_key(SignatureAlgorithm::MlDsa44);
         let signed_public_key = ctx.make_signed_public_key(private_key, signing_key)?;
 
         let security_state = SecurityState::new();
@@ -390,7 +390,7 @@ impl WrappedAccountCryptographicState {
                     .map_err(|_| RotateCryptographyStateError::KeyMissing)?;
 
                 // 2. The signing key is generated
-                let signing_key_id = ctx.make_signing_key(SignatureAlgorithm::Ed25519);
+                let signing_key_id = ctx.make_signing_key(SignatureAlgorithm::MlDsa44);
                 let new_signing_key = ctx
                     .wrap_signing_key(*new_user_key, signing_key_id)
                     .map_err(|_| RotateCryptographyStateError::KeyMissing)?;
@@ -1210,7 +1210,7 @@ mod tests {
 
         assert_eq!(
             result.signature_key_pair.signature_algorithm.as_deref(),
-            Some("ed25519")
+            Some("mldsa44")
         );
 
         assert!(
--- a/crates/bitwarden-crypto/Cargo.toml
+++ b/crates/bitwarden-crypto/Cargo.toml
@@ -47,7 +47,7 @@ ed25519-dalek = { workspace = true, features = ["rand_core", "zeroize"] }
 hex = { version = ">=0.4.0, <0.5" }
 hkdf = "0.13.0"
 hmac = { workspace = true, features = ["zeroize"] }
-hybrid-array = { version = "0.4.10", features = ["alloc", "serde"] }
+hybrid-array = { version = "0.4.10", features = ["alloc", "serde", "zeroize"] }
 ml-dsa = { version = "0.1.0-rc.8", features = ["zeroize"] }
 num-bigint = ">=0.4, <0.5"
 num-traits = ">=0.2.15, <0.3"
--- a/crates/bitwarden-crypto/Cargo.toml
+++ b/crates/bitwarden-crypto/Cargo.toml
@@ -42,12 +42,13 @@ bitwarden-uniffi-error = { workspace = true, optional = true }
 cbc = { version = "0.2.0", features = ["alloc", "zeroize"] }
 chacha20poly1305 = { version = "0.11.0-rc.3", features = ["zeroize"] }
 ciborium = { workspace = true }
-coset = { version = ">=0.4.0, <0.5" }
+coset = { version = ">=0.4.2, <0.5" }
 ed25519-dalek = { workspace = true, features = ["rand_core", "zeroize"] }
 hex = { version = ">=0.4.0, <0.5" }
 hkdf = "0.13.0"
 hmac = { workspace = true, features = ["zeroize"] }
-hybrid-array = { version = "0.4.10", features = ["alloc", "zeroize"] }
+hybrid-array = { version = "0.4.10", features = ["alloc", "serde"] }
+ml-dsa = { version = "0.1.0-rc.8", features = ["zeroize"] }
 num-bigint = ">=0.4, <0.5"
 num-traits = ">=0.2.15, <0.3"
 pbkdf2 = { version = "0.13.0-rc.10", default-features = false }
--- a/crates/bitwarden-crypto/src/signing/cose.rs
+++ b/crates/bitwarden-crypto/src/signing/cose.rs
@@ -156,7 +156,7 @@ pub(super) fn mldsa_seed(cose_key: &CoseKey) -> Result<B32, EncodingError> {
     let priv_bytes = akp_priv(cose_key)?;
     let seed: [u8; ML_DSA_SEED_SIZE] = priv_bytes
         .try_into()
-        .map_err(|_| EncodingError::InvalidValue("ML-DSA-65 seed length"))?;
+        .map_err(|_| EncodingError::InvalidValue("ML-DSA seed length"))?;
     Ok(seed.into())
 }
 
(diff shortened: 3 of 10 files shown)


crates/bitwarden-core/src/key_management/account_cryptographic_state.rs
```diff
@@ -275,7 +276,7 @@ impl WrappedAccountCryptographicState {
     ) -> Result<(SymmetricKeyId, Self), AccountCryptographyInitializationError> {
         let user_key = ctx.make_symmetric_key(SymmetricKeyAlgorithm::XChaCha20Poly1305);
         let private_key = ctx.make_private_key(PublicKeyEncryptionAlgorithm::RsaOaepSha1);
-        let signing_key = ctx.make_signing_key(SignatureAlgorithm::Ed25519);
```
quexten, line 335: Note: This defaults to ML-DSA for all v2 users!

crates/bitwarden-crypto/src/signing/signing_key.rs
```diff
@@ -94,67 +132,147 @@ impl SigningKey {
...
+                key_ops: vec![coset::RegisteredLabel::Assigned(KeyOperation::Sign)]
+                    .into_iter()
+                    .collect(),
+                params: vec![
+                    (
+                        coset::Label::Int(AkpKeyParameter::Priv.to_i64()),
+                        Value::Bytes(seed.to_vec()),
+                    ),
+                    (
+                        coset::Label::Int(AkpKeyParameter::Pub.to_i64()),
```
quexten: Note for reviewers; The cose spec says PUB is always required, even in private keys, so the private key (signing key) here contains both pub and private values.

djsmith85:
Requesting additional reviews from @dani-garcia as platform-support for Auth and @Hinton due to CXF changes

quexten:
Thanks. I'll note that the cxf changes are just tests, that the compiler complained about. (I did not dig into what changed). The same applies to the auth changes.

harr1424 (dismissed):
Exporter changes look good

crates/bitwarden-crypto/src/signing/signing_key.rs
MGibson1: Is this per https://datatracker.ietf.org/doc/draft-ietf-cose-dilithium/ ?

It's quite odd to me that they chose to make pub required when previous RFCs ecdsa don't. It feels different enough for a comment to me.

MGibson1 (commented):
This looks good to me, I am annoyed by the pub key decision so I wouldn't mind a comment there, but nothing else from me.

crates/bitwarden-crypto/src/signing/signing_key.rs
quexten: > It's quite odd to me that they chose to make pub required when previous RFCs ecdsa don't.

It was not REQUIRED but RECOMMENDED to add them for eliptic curves. For RSA, it was also REQUIRED https://datatracker.ietf.org/doc/html/rfc8230/#section-4

>   o  For public keys, the fields 'n' and 'e' MUST be present.  All
      other fields defined in the following table below MUST be absent.
   o  For private keys with two primes, the fields 'other', 'r_i',
      'd_i', and 't_i' MUST be absent; all other fields MUST be present.

crates/bitwarden-crypto/src/signing/cose.rs
```diff
@@ -143,3 +146,51 @@ fn okp_curve(cose_key: &CoseKey) -> Result<i128, EncodingError> {
         })
         .ok_or(EncodingError::MissingValue("OKP curve"))
 }
+
+/// Helper function to parse the private key from an AKP `CoseKey`.
+fn akp_priv(cose_key: &CoseKey) -> Result<&[u8], EncodingError> {
```
dani-garcia, line 140: Suggestion, we're starting to have a lot of helper functions duplicating this, it's a good idea to think about extracing it into a separate function:
```rust
/// Helper function to look up a parameter in a `CoseKey` by its registered label.
fn cose_param(cose_key: &CoseKey, param: impl EnumI64 + Copy) -> Option<&coset::cbor::value::Value> {
    cose_key.params.iter().find_map(|(key, value)| match key {
        Label::Int(i) if EnumI64::from_i64(*i) == Some(param) => Some(value),
        _ => None,
    })
}

fn akp_priv(cose_key: &CoseKey) -> Result<&[u8], EncodingError> {
    cose_param(cose_key, AkpKeyParameter::Priv)
        .and_then(|v| v.as_bytes().map(Vec::as_slice))
        .ok_or(EncodingError::MissingValue("AKP private key"))
}

```

dani-garcia (dismissed):
LGTM, just a small non-blocking suggestion

crates/bitwarden-crypto/src/signing/cose.rs
quexten: Nice, this is a good point. For what it's worth we should probably investigate this for cose in general, to make it nicer to work with than what coset offers, but not here. I applied your suggestion to the file though.

crates/bitwarden-crypto/src/signing/cose.rs
```diff
@@ -99,47 +101,71 @@ pub(super) fn ed25519_verifying_key(
...
+}
+
+/// Helper function to parse the public key from an AKP `CoseKey`.
+fn akp_pub(cose_key: &CoseKey) -> Result<&[u8], EncodingError> {
+    cose_param(cose_key, AkpKeyParameter::Pub)
+        .and_then(|v| v.as_bytes().map(Vec::as_slice))
+        .ok_or(EncodingError::MissingValue("AKP public key"))
+}
+
+/// Helper function to parse an ML-DSA-65 signing key from a `CoseKey`. The `Priv` parameter
```
mzieniukbw:
```suggestion
/// Helper function to parse an ML-DSA signing key from a `CoseKey`. The `Priv` parameter
```

crates/bitwarden-crypto/src/signing/cose.rs
```diff
@@ -99,47 +101,71 @@ pub(super) fn ed25519_verifying_key(
...
+/// contains the 32-byte seed, from which the full key pair is deterministically derived.
+pub(super) fn mldsa_seed(cose_key: &CoseKey) -> Result<B32, EncodingError> {
+    let priv_bytes = akp_priv(cose_key)?;
+    let seed: [u8; ML_DSA_SEED_SIZE] = priv_bytes
+        .try_into()
+        .map_err(|_| EncodingError::InvalidValue("ML-DSA-65 seed length"))?;
```
mzieniukbw:
```suggestion
        .map_err(|_| EncodingError::InvalidValue("ML-DSA seed length"))?;
```

crates/bitwarden-crypto/Cargo.toml
```diff
@@ -42,12 +42,13 @@ bitwarden-uniffi-error = { workspace = true, optional = true }
...
 chacha20poly1305 = { version = "0.11.0-rc.3", features = ["zeroize"] }
 ciborium = { workspace = true }
-coset = { version = ">=0.4.0, <0.5" }
+coset = { version = ">=0.4.2, <0.5" }
 ed25519-dalek = { workspace = true, features = ["rand_core", "zeroize"] }
 hex = { version = ">=0.4.0, <0.5" }
 hkdf = "0.13.0"
 hmac = { workspace = true, features = ["zeroize"] }
-hybrid-array = { version = "0.4.10", features = ["alloc", "zeroize"] }
+hybrid-array = { version = "0.4.10", features = ["alloc", "serde"] }
```
mzieniukbw: ❓  Why did we remove "zeroize" feature ?

mzieniukbw (dismissed):

crates/bitwarden-crypto/Cargo.toml
quexten: Undone.

crates/bitwarden-crypto/src/signing/signing_key.rs
```diff
@@ -68,12 +86,27 @@ impl SigningKey {
                     &mut rand::rng(),
                 ))),
             },
+            SignatureAlgorithm::MlDsa44 => {
+                let mut seed = [0u8; 32];
```
mzieniukbw, line 90: ❓ Do we need to zeroize seed on drop ?

mzieniukbw (approved):

dani-garcia (approved):

crates/bitwarden-crypto/src/signing/signing_key.rs
dani-garcia: In this case, I'd recommend boxing the array before filling it, that way you should ensure no stack copies are present:

```rust
// This is heap allocated from the start, so will be zeroized on drop
let mut seed = Box::pin(Array::from([0u8; 32]));
rand::rng().fill_bytes(&mut seed);

let kp = MlDsa44::from_seed(&seed);
SigningKey {
    id: KeyId::make(),
    inner: RawSigningKey::MlDsa44 {
        seed,
        signing_key: Box::pin(kp.signing_key().clone()),
        verifying_key: Box::new(kp.verifying_key().clone()),
    },
}
```

quexten: https://github.com/bitwarden/sdk-internal/pull/978

PR #921, opened 2026-04-09
[PM-34153] Bootstrap PolicyDefinition trait and filtering logic

eliykat:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-34153

## 📔 Objective

First PR in bootstrapping policies in the SDK. Main changes:
- add an organization crate, which for now just contains rust representations of client sync data
  - question: is it better to map client sync data 1:1? Or should we narrow it down to a domain-specific context object containing only what we need? The latter seems better isolated from the rest of the application - it would be more durable against unrelated changes to the sync packet. On the other hand, it puts more responsibility on consumers to transform the sync data into the context object, and then all crates have to manage their own context objects vs. having it handled once by the organization crate. Open to feedback here.
- add a `filter` crate inside the existing policies crate. This defines:
  - basic policy structs
  - the `PolicyDefinition` trait, which lets developers define how a policy is enforced
  - the `filter` function, which consumes the policy, organization, and PolicyDefinition, and filters out policies that should not be enforced
  - tests :)

This is my first substantive rust PR, all feedback is welcome.

Later PRs will add the ability to register PolicyDefinitions and wire up the interfaces to a policy client.

## 🚨 Breaking Changes

No breaking changes - none of this code is consumed anywhere yet.


JaredScar (dismissed):

crates/bitwarden-organizations/src/lib.rs
```diff
@@ -0,0 +1,236 @@
+// FIXME: add documentation
+#![allow(missing_docs)]
+
+#[cfg(feature = "uniffi")]
+uniffi::setup_scaffolding!();
```
dani-garcia, line 4: Uniffi support also needs a `uniffi.toml` file next to the `Cargo.toml` file to set the correct android/ios packages, check out this one as an example: 

https://github.com/bitwarden/sdk-internal/blob/main/crates/bitwarden-auth/uniffi.toml

crates/bitwarden-organizations/src/lib.rs
```diff
@@ -0,0 +1,236 @@
+// FIXME: add documentation
+#![allow(missing_docs)]
```
dani-garcia: I recommend creating a README.md file next at the root of the crate and replacing this with `#![doc = include_str!("../README.md")]`

It doesn't have to be a full readme now, we can work on expanding it as we move more functionality over, for example: https://github.com/bitwarden/sdk-internal/blob/main/crates/bitwarden-policies/README.md

eliykat:
Thanks @dani-garcia , your feedback should be addressed 👍

crates/bitwarden-policies/src/filter.rs
```diff
@@ -0,0 +1,255 @@
...
+#[derive(PartialEq, Serialize, Deserialize, Debug)]
+pub struct RawPolicyType(pub i32);
+
+#[derive(Serialize, Deserialize, Debug)]
+pub struct RawPolicy {
+    id: Uuid,
+    organization_id: Uuid,
+    r#type: RawPolicyType,
+    data: Option<HashMap<String, serde_json::Value>>,
+    enabled: bool,
```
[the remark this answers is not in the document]
eliykat: This is not an api response model.

JaredScar (dismissed):
Good stuff Thomas!

crates/bitwarden-policies/src/filter.rs
```diff
@@ -0,0 +1,257 @@
...
+use std::collections::HashMap;
+
+use bitwarden_organizations::{
+    OrganizationUserStatusType, OrganizationUserType, ProfileOrganization,
+};
+use serde::{Deserialize, Serialize};
+use uuid::Uuid;
+
+#[derive(PartialEq, Serialize, Deserialize, Debug)]
+pub struct RawPolicyType(pub i32);
```
Hinton: **non blocking**: I know i32 probably came from my example, but should it be signed or unsigned int? (i or u) I.e. is it valid for a policy to be negative?

Is 0 a valid value? https://doc.rust-lang.org/beta/std/num/struct.NonZero.html might be even more appropriate.

What size do we expect policy type to be, u8 gives you up to 255 (127 signed), u16 is 65 535.

crates/bitwarden-organizations/src/lib.rs
```diff
@@ -0,0 +1,216 @@
+#![doc = include_str!("../README.md")]
+#![allow(missing_docs)]
```
Hinton: I don't think disabling the `missing_docs` lint for the whole crate is good? Most if not all of these structs would benefit from some documentation.

crates/bitwarden-policies/src/filter.rs
eliykat: `PolicyType` starts at 0 (the 2FA policy) and currently goes to 21. So 0 is valid.

We will add more in the future, but `u8` for a limit of 255 seems more than enough for now. Updated.

crates/bitwarden-organizations/src/lib.rs
eliykat: Removed this directive and added documentation to all new code.

crates/bitwarden-policies/src/filter.rs
```diff
@@ -0,0 +1,282 @@
...
+use std::collections::HashMap;
+
+use bitwarden_organizations::{
+    OrganizationUserStatusType, OrganizationUserType, ProfileOrganization,
+};
+use serde::{Deserialize, Serialize};
+use uuid::Uuid;
+
+/// A newtype representing the policy type.
+#[derive(PartialEq, Serialize, Deserialize, Debug)]
```
Hinton, line 17: Is `Serialize` and `Deserialize` used?

crates/bitwarden-policies/src/filter.rs
```diff
@@ -0,0 +1,282 @@
...
+            OrganizationUserStatusType::Confirmed,
+        ]
+    }
+}
+
+/// Filters `policies` to those that should be enforced against the user.
+/// This evaluates common business rules (e.g. the policy is enabled),
+/// as well as policy-specific rules according to its [`PolicyDefinition`].
+///
+/// If a policy's organization is not present in `organizations`, the policy is enforced by default.
```
Hinton, line 68: Are there cases in clients where this currently occurs?

Hinton (commented):
This looks good from a high level perspective of what we previously discussed, I'll defer to @dani-garcia for the actual review.

crates/bitwarden-policies/src/filter.rs
eliykat: I haven't experimented with how the native code integration works yet, but my understanding: `RawPolicy` needs it to cross the FFI boundary, so it needs to implement these traits, and so does its properties, including `RawPolicyType`. It should serialize/deserialize from the raw enum value in TS.

I will experiment and refine once I get to wiring it up to native code.

crates/bitwarden-policies/src/filter.rs
eliykat: This is to handle 2 cases:
1. known cases - sometimes policies are fetched ad-hoc from the server where the policy and corresponding org does not exist in sync data. e.g. evaluating MP policy at the time of accepting an invite. Ideally this should also send org information so that it can be properly filtered, but we don't do that today.
2. unknown cases - gracefully recovering from a potential error by erring on the side of least privilege

crates/bitwarden-policies/src/filter.rs
Hinton: Once you add tsify/wasm_bindgen it will require it through a compile error.

crates/bitwarden-policies/src/filter.rs
```diff
@@ -0,0 +1,282 @@
...
+pub fn filter<'a, P: PolicyDefinition>(
```
dani-garcia, line 69: Small suggestion: 

Free functions like this can be a bit hard to discover, but you can also use an extension trait:
```rust
trait PolicyDefinitionExt: PolicyDefinition {
    fn filter<'a>(
        &'a self,
        policies: &'a [RawPolicy],
        organizations: &'a [ProfileOrganization],
    ) -> Vec<&'a RawPolicy> {
        // Same code as before
    }
}

impl<T: PolicyDefinition> PolicyDefinitionExt for T {}

// Then you can do this, which is usually easier for the IDE to complete:
TestPolicy.filter(policies, organizations);
```

dani-garcia (approved):
LGTM, just one small suggestion that we can leave as a follow up

PR #922, opened 2026-04-09
[PM-34829] Refactor FoldersClient to be tested

Hinton:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-34829
## 📔 Objective

We've made recent changes to https://contributing.bitwarden.com/architecture/sdk/internal/#client-structure suggesting that thin pass-through functions should be eliminated. This updates folders to follow this pattern to serve as an example of how things should work.

This also aims to make sure all our public interfaces are extensively tested since we've identified existing gaps resulting in unexpected breaking changes or broken functionality. Testing directly on the exposed functions ensures we test from a slightly broader perspective.

---

- Adds a `from_api_client` helper to `bitwarden-core` to eliminate common boilerplate when initializing `ApiClient` contexts. 
- Adds `require` helper for going `Option<Arc<Repository<T>>` -> `Result<Arc<Repository<T>>, RepositoryError>` which assist with returning the appropriate error when unwrapping the optional repositories.


crates/bitwarden-vault/src/vault_client.rs
```diff
@@ -45,9 +45,8 @@ impl VaultClient {
 
     /// Folder related operations.
     pub fn folders(&self) -> FoldersClient {
-        FoldersClient {
-            client: self.client.clone(),
-        }
+        FoldersClient::from_client(&self.client)
+            .expect("Failed to construct FoldersClient from Client")
```
Hinton: I suppose we could split `FoldersClient` and have V2 require repository 🤔. This would avoid this being a breaking change.

What do you think @dani-garcia @coroiu?

dani-garcia (approved):

nikwithak (approved):

PR #925, opened 2026-04-09
Reexport new auth registration in uniffi

dani-garcia:
## 🎟️ Tracking

https://bitwarden.slack.com/archives/C054ZQSBS49/p1775586953115189

## 📔 Objective

Export the auth registration client from Uniffi. Because of uniffi error handling constraints that all the clients deal with regarding error handling, the registration API is currently duplicated in the uniffi crate. To avoid duplicate names, I've removed the uniffi::export usages in `bitwarden-auth`

### DIFF
--- a/crates/bitwarden-uniffi/src/auth/mod.rs
+++ b/crates/bitwarden-uniffi/src/auth/mod.rs
@@ -17,8 +17,7 @@ pub struct AuthClient(pub(crate) bitwarden_core::Client);
 
 #[uniffi::export(async_runtime = "tokio")]
 impl AuthClient {
-    // Reexport the registration client from the bitwarden-auth AuthClient until everything else
-    // from here is migrated over.
+    /// Client for initializing user account cryptography and unlock methods after JIT provisioning
     pub fn registration(&self) -> RegistrationClient {
         RegistrationClient(self.0.clone())
     }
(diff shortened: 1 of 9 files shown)


crates/bitwarden-uniffi/src/auth/mod.rs
```diff
@@ -15,6 +16,12 @@ pub struct AuthClient(pub(crate) bitwarden_core::Client);
 
 #[uniffi::export(async_runtime = "tokio")]
 impl AuthClient {
+    // Reexport the registration client from the bitwarden-auth AuthClient until everything else
+    // from here is migrated over.
+    pub fn registration(&self) -> RegistrationClient {
+        self.0.auth_new().registration()
```
Hinton: This client should be moved to this create to ensure it uses BitwardenError right?

Hinton (changes requested):

crates/bitwarden-uniffi/src/auth/mod.rs
dani-garcia: We only just talked about this a couple of days and I already forgot 😅 . Fixed in https://github.com/bitwarden/sdk-internal/pull/925/changes/2012dd0366f7f7be74afe7cdbfb06bb40b6955c5

Hinton:
We should also remove `uniffi::export` from the registration client to prevent there being two clients.

dani-garcia:
Hmm I wasn't aware that names had to be globally unique, that seems like a swift only thing. Fixed

Hinton (approved):

addisonbeck (approved):

PR #926, opened 2026-04-09
[BRE-1670] Replace PAT with app token

AmyLGalles:
## 🎟️ Tracking

[BRE-1670](https://bitwarden.atlassian.net/browse/BRE-1670?atlOrigin=eyJpIjoiYWQ0N2Q0OGIzODE5NDU3ZDg3OWI4OGE4ZTQ0ZTRhOGEiLCJwIjoiaiJ9)

## 📔 Objective

Updating github workflows that use PAT to use app token or GITHUB_TOKEN instead.

In some cases, the PAT token was being used for simple repo actions that didn't require further authentication, such as reading from or cloning a public repository.  

GPG Keys:
Github offers native signing of commits when done through the API, so a step importing the GPG key is not needed for signed commits that use the API-commit step. If a commit is later merged by a human reviewing a PR, it doesn't need to be signed as the resulting squashed commit will be signed.

## 🚨 Breaking Changes

[BRE-1670]: https://bitwarden.atlassian.net/browse/BRE-1670?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/workflows/build-swift.yml
+++ b/.github/workflows/build-swift.yml
@@ -104,7 +104,6 @@ jobs:
     runs-on: ubuntu-24.04
     needs: build
     permissions:
-      contents: 
       actions: write
     steps:
       - name: Trigger Swift release
--- a/.github/workflows/build-swift.yml
+++ b/.github/workflows/build-swift.yml
@@ -113,31 +113,11 @@ jobs:
           subscription_id: ${{ secrets.AZURE_SUBSCRIPTION_ID }}
           tenant_id: ${{ secrets.AZURE_TENANT_ID }}
           client_id: ${{ secrets.AZURE_CLIENT_ID }}
-
-      - name: Retrieve github secrets
-        id: retrieve-secret
-        uses: bitwarden/gh-actions/get-keyvault-secrets@main
-        with:
-          keyvault: gh-org-bitwarden
-          secrets: "BW-GHAPP-ID,BW-GHAPP-KEY"
-
-      - name: Log out from Azure
-        uses: bitwarden/gh-actions/azure-logout@main
-
-      - name: Generate GH App token
-        uses: actions/create-github-app-token@29824e69f54612133e76f7eaac726eef6c875baf # v2.2.1
-        id: app-token
-        with:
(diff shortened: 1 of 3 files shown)


.github/workflows/build-swift.yml
```diff
@@ -115,19 +115,27 @@ jobs:
           client_id: ${{ secrets.AZURE_CLIENT_ID }}
 
       - name: Retrieve github PAT secrets
```
dani-garcia: I think we can probably remove the PAT from the name here, as we're using app tokens now
```suggestion
      - name: Retrieve github secrets
```

.github/workflows/build-swift.yml
```diff
@@ -115,19 +115,27 @@ jobs:
...
       - name: Log out from Azure
         uses: bitwarden/gh-actions/azure-logout@main
 
+      - name: Generate GH App token
+        uses: actions/create-github-app-token@29824e69f54612133e76f7eaac726eef6c875baf # v2.2.1
+        id: app-token
+        with:
+          app-id: ${{ steps.retrieve-secret.outputs.BW-GHAPP-ID }}
+          private-key: ${{ steps.retrieve-secret.outputs.BW-GHAPP-KEY }}
+          owner: ${{ github.repository_owner }}
```
dani-garcia: This token is only used to trigger anothe workflow in this same repo, so we can probably scope down the token to only this repo:
```suggestion
          app-id: ${{ steps.retrieve-secret.outputs.BW-GHAPP-ID }}
          private-key: ${{ steps.retrieve-secret.outputs.BW-GHAPP-KEY }}
          owner: ${{ github.repository_owner }}
          repositories: sdk-internal
          permission-actions: write
```

dani-garcia (commented):
LGTM, just gotta run `npm run prettier` to fix the failing format CI

fntyler (approved):

dani-garcia (approved):

PR #931, opened 2026-04-10
[PM-34815] Add rotate_user_keys

Thomas-Avery:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-34815

## 📔 Objective

Adds the new client method `rotate_user_keys` for SDK user key rotation without password change. This currently only supports master password users. TDE and key connector to come later.


crates/bitwarden-user-crypto-management/src/key_rotation/rotate_user_keys.rs
```diff
@@ -19,3 +37,414 @@ pub enum KeyRotationMethod {
     /// RotateUserKeysError::UnimplementedKeyRotationMethod error if used.
     Tde,
 }
+
+#[derive(Serialize, Deserialize, Clone)]
+#[cfg_attr(feature = "wasm", derive(Tsify), tsify(into_wasm_abi, from_wasm_abi))]
+#[cfg_attr(feature = "uniffi", derive(uniffi::Record))]
+pub struct RotateUserKeysRequest {
+    pub key_rotation_method: KeyRotationMethod,
+    pub trusted_emergency_access_public_keys: Vec<PublicKey>,
```
quexten, line 46: Non-blocking thought: @Thomas-Avery Do you have any thoughts about the UI interaction between SDK and clients? We could theoretically allow the SDK to call out into clients, and show prompts on-demand, which would make these interfaces presumably simpler, but I'm not sure whether such callback and interop code maybe makes things more complicated in the long run.

The interop is more complicated, but the publicly facing "structure" is simpler since you only have one function for the operation and don't first need to get the keys via the SDK that you have to approve.

crates/bitwarden-user-crypto-management/src/key_rotation/rotate_user_keys.rs
```diff
@@ -19,3 +37,414 @@ pub enum KeyRotationMethod {
...
+        internal_rotate_user_keys(key_store, api_client, request).await
+    }
+}
+
+async fn internal_rotate_user_keys(
+    key_store: &KeyStore<KeySlotIds>,
+    api_client: &bitwarden_api_api::apis::ApiClient,
+    request: RotateUserKeysRequest,
+) -> Result<(), RotateUserKeysError> {
+    let _span = info_span!("rotate_user_keys").entered();
```
quexten: Issue: This span is held across await points, which is dangerous / results in incorrect traces: https://docs.rs/tracing/latest/tracing/span/struct.Span.html#in-asynchronous-code
You must use the trace only within a sync chunk (i.e before the await) and drop it before the await.

In this case I suggest doing the async request to `sync` first, *then* entering the trace, so that we still get the trace for the more interesting (and sync) work that follows.

Alternatively, the tracing macro `#[tracing::instrument(level = "info")]` would generate the correct code that enters the span in every sync chunk of the underlying state machine that gets generated by the async function.

@dani-garcia A question for you; I recall we have a dylint for this for the key-store. I'm not very familiar with dylints. Could we add one for spans too?

crates/bitwarden-user-crypto-management/src/key_rotation/rotate_user_keys.rs
```diff
@@ -19,3 +37,414 @@ pub enum KeyRotationMethod {
...
+        }
+        KeyRotationMethod::Password { .. } => {}
+    }
+
+    let sync = sync_current_account_data(api_client)
+        .await
+        .map_err(|_| RotateUserKeysError::ApiError)?;
+
+    // Create a separate scope so that the mutable context is not held across the await point
+    let post_request = {
```
quexten, line 85: I.e add the span at the top of this block instead of the top of the function.

quexten (changes requested):

crates/bitwarden-user-crypto-management/src/key_rotation/rotate_user_keys.rs
dani-garcia: It's actually an existing clippy lint, but you have to configure the types yourself: 
https://github.com/bitwarden/sdk-internal/blob/b7278278dc4addcb315c6169857161fc69012ae9/clippy.toml#L5-L11

We should be able to just add `tracing::span::EnteredSpan` to that list.

quexten: Nice, thank you! I made a PR here https://github.com/bitwarden/sdk-internal/pull/970 and tested it locally for such a case.

crates/bitwarden-user-crypto-management/src/key_rotation/rotate_user_keys.rs
Thomas-Avery: In general I'm not a fan of callbacks for UI interactions. Typically it swaps compile time errors to run time errors and loosen/hinders readability of the contract between consumers and the library. It also makes the main flow here harder to unit test, but there are ways around that. 

I would like to hear more on the concerns with the current API. I think there's room to improve the design without introducing callbacks. A typed session builder, for example, could improve both readability and discoverability.



```
let pendingRotation = await ref.value.user_crypto_management().begin_key_rotation({
    old_password: currentMasterPassword,
    password: newMasterPassword,
    hint: hint,
  });

let trustedMembership = [];
pendingRotation.untrustedMemberships.forEach((untrustedMembership) => {
  if (promptUserToTrustMembership(untrustedMembership)) {
    trustedMembership.push(untrustedMembership);
  }
});

await ref.value
  .user_crypto_management()
  .execute_key_rotation(pendingRotation.data, trustedMembership);
```

crates/bitwarden-user-crypto-management/src/key_rotation/rotate_user_keys.rs
Thomas-Avery: Great catch fixed in https://github.com/bitwarden/sdk-internal/pull/931/changes/73f6e2763d017270fe2286baeae0a57e4a9fd890

crates/bitwarden-user-crypto-management/src/key_rotation/rotate_user_keys.rs
Thomas-Avery: Fixed in https://github.com/bitwarden/sdk-internal/pull/931/changes/73f6e2763d017270fe2286baeae0a57e4a9fd890

crates/bitwarden-user-crypto-management/src/key_rotation/rotate_user_keys.rs
quexten: I don't necessarily have concerns, I just want to discuss our strategy on this. If we don't want to bring UI into the SDK that makes sense, it just means we need to design multi-step APIs sometimes.

quexten (approved):

crates/bitwarden-user-crypto-management/src/key_rotation/rotate_user_keys.rs
Thomas-Avery: I would say lets not bring UI into the SDK, and it's okay to design multi-step APIs like this. If we expect many other teams to be consumers and it to be in a hot path we should invest in making it as "pit of success" as possible.

PR #944, opened 2026-04-13
[PM-35036] Rename keyid to keyslotid

quexten:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-35036
https://bitwarden.slack.com/archives/C090UKMPK6G/p1770896132851739

## 📔 Objective

Currently there is KeyId (The ID of a specific key as encoded on the CoseKey, and in the "encrypted by" and "contained key id" COSE Headers), and the KeyId which refers to the key function in the key context / key store. The latter will be renamed to KeySlotId, since it does not refer to a specific key, but a "slot" which a specific key can be set to.

This de-tangles concepts, and also makes it so that we can publicly expose KeyId, which currently has a naming collision.

### DIFF
--- a/crates/bitwarden-core/src/key_management/master_password.rs
+++ b/crates/bitwarden-core/src/key_management/master_password.rs
@@ -5,7 +5,9 @@ use bitwarden_api_api::models::{
     MasterPasswordUnlockDataRequestModel,
     master_password_unlock_response_model::MasterPasswordUnlockResponseModel,
 };
-use bitwarden_crypto::{EncString, Kdf, KeyIds, KeyStoreContext, MasterKey, SymmetricCryptoKey};
+use bitwarden_crypto::{
+    EncString, Kdf, KeySlotIds, KeyStoreContext, MasterKey, SymmetricCryptoKey,
+};
 use bitwarden_encoding::B64;
 use bitwarden_error::bitwarden_error;
 use serde::{Deserialize, Serialize};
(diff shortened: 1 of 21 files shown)


dani-garcia:
There's still a bunch of usages of `key_id` on other places:
- The `key_id!` macro itself: https://github.com/bitwarden/sdk-internal/blob/1116a4599f1cb196835ba9a2e1fcab028ba42796/crates/bitwarden-crypto/src/traits/keyslot_ids.rs#L83
- The usage of the macro: https://github.com/bitwarden/sdk-internal/blob/1116a4599f1cb196835ba9a2e1fcab028ba42796/crates/bitwarden-core/src/key_management/mod.rs#L119. This one is used all over the place.
- A lot of local variables in tests, just have to search for `key_id `, for example: https://github.com/bitwarden/sdk-internal/blob/1116a4599f1cb196835ba9a2e1fcab028ba42796/crates/bitwarden-collections/src/collection.rs#L176

We'll probably want to do them in parts though, some of those changes have a pretty wide blast radius.

quexten:
@dani-garcia Part 2 is here: https://github.com/bitwarden/sdk-internal/pull/955, however this includes so many teams as reviewers that I don't want to include it in this PR, given that this PR is blocking for other work.

dani-garcia (approved):
Ah nice, yeah doing it in parts sounds good to me!

quexten:
Ah, oops, i didn't realize auto-merge would ignore the KM review, but it makes sense considering the ownership.

mzieniukbw:
@quexten just checked it out, 👍 approved by KM.

quexten:
Thank you @mzieniukbw !

PR #947, opened 2026-04-13
[PM-31061] Expose getting the key-id

quexten:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-31061

Clients change: https://github.com/bitwarden/clients/pull/20127

## 📔 Objective

Exposes a function to get key ids for any symmetric key. This is required to update the biometrics subsystem after a user-key change. The current biometrics key id is stored without biometrics protection, such that the client can known whether it needs to invoke a UV prompt to overwrite the existing key.


crates/bitwarden-core/src/key_management/crypto_client.rs
```diff
@@ -189,6 +190,15 @@ impl CryptoClient {
     pub async fn get_user_encryption_key(&self) -> Result<B64, CryptoClientError> {
         get_user_encryption_key(&self.client).await
     }
+
+    /// Takes a raw key and returns the corresponding key id. This is used for the biometrics
+    /// subsystem and should be removed after moving over biometric management to the SDK.
+    pub fn get_key_id_for_symmetric_key(
```
quexten, line 196: Note: This is expanding core slightly, but I don't think we can make a client for this just yet. We should either move out the complete crypto client (somewhere)? Or even move the entire biometric subsystem to be accessed from the SDK directly.

The latter is probably the better approach but a fairly large scope creep, and not necessary just yet.

mzieniukbw (approved):

PR #949, opened 2026-04-13
[PM-34060] Support Bank Account Types

nick-livefront:
## 🎟️ Tracking

[PM-34060](https://bitwarden.atlassian.net/browse/PM-34060)
Server PR: https://github.com/bitwarden/server/pull/7112
Clients PR: https://github.com/bitwarden/clients/pull/19302

## 📔 Objective

Adds support for Bank Account types in the SDK

## 🚨 Breaking Changes

- New `BankAccount` variant added to `CipherType`, `CipherListViewType`, and `CipherViewType` — clients must handle

[PM-34060]: https://bitwarden.atlassian.net/browse/PM-34060?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


crates/bitwarden-vault/src/cipher/bank_account.rs
```diff
@@ -0,0 +1,254 @@
...
+            pin: bank_account.pin.map(|n| n.to_string()),
+            swift_code: bank_account.swift_code.map(|n| n.to_string()),
+            iban: bank_account.iban.map(|n| n.to_string()),
+            bank_contact_phone: bank_account.bank_contact_phone.map(|n| n.to_string()),
+        }
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
quexten, line 163: A suggestion: Why not record a real test json cipher for this and decrypt it in the test? This would guaranteed ensure that compatibility is kept. Currently, a field rename with your ide would not break tests, but would break compatibility.

quexten (dismissed):
Looks good; It's a bit unfortunate that this work has to be duplicated between clients and sdk currently. I've left a comment here on that: https://github.com/bitwarden/clients/pull/20147/changes#r3098476864

One suggestion (non blocking, but recommended): Make a test against a recorded test vector (recorded *real* json string). You can even make a test to generate the test vector, and then commit it to code like this:

https://github.com/bitwarden/sdk-internal/blob/b7278278dc4addcb315c6169857161fc69012ae9/crates/bitwarden-vault/src/cipher/blob/v1.rs#L289

crates/bitwarden-vault/src/cipher/bank_account.rs
```diff
@@ -0,0 +1,254 @@
...
+use super::cipher::CipherKind;
+use crate::{Cipher, VaultParseError, cipher::cipher::CopyableCipherFields};
+
+#[derive(Serialize, Deserialize, Debug, Clone)]
+#[serde(rename_all = "camelCase")]
+#[cfg_attr(feature = "uniffi", derive(uniffi::Record))]
+#[cfg_attr(feature = "wasm", derive(Tsify), tsify(into_wasm_abi, from_wasm_abi))]
+pub struct BankAccount {
+    pub bank_name: Option<EncString>,
+    pub name_on_account: Option<EncString>,
```
Hinton, line 20: This is commonly called "account holder", worth pondering if it should be changed before we're stuck.

crates/bitwarden-exporters/src/cxf/export.rs
Hinton, line 1: **blocking:** Are there tickets for capturing the export work? This work is required before new item types can be shipped (otherwise there would need to be UI work to show a warning), and should be implemented for both Credential Exchange and our internal json format.

nick-livefront: I wrote up a ticket to account for the UI warning needed: [PM-35325](https://bitwarden.atlassian.net/browse/PM-35325)

crates/bitwarden-vault/src/cipher/bank_account.rs
nick-livefront: Great idea, thank you for the suggestion! [8aea536](https://github.com/bitwarden/sdk-internal/pull/949/commits/8aea53648c4691db11577876f7bc9b1e3e14a757)

quexten (dismissed):

crates/bitwarden-vault/src/cipher/bank_account.rs
nick-livefront: For posterity, an offline discussion took place and `Name on Account` was used intentionally and will leave as is.

crates/bitwarden-vault/src/cipher/cipher_view_type.rs
```diff
@@ -101,4 +104,11 @@ impl CipherViewType {
             _ => None,
         }
     }
+
```
jengstrom-bw, line 114: 🎨  Do we also want a `&mut self` version like the other types above?

nick-livefront: Nice catch! Better to be consistent now: [b1f9a98](https://github.com/bitwarden/sdk-internal/pull/949/commits/b1f9a988fb4220d71581b5f62dd1ae1a9a2d1390)

jengstrom-bw (approved):

mcamirault (approved):

quexten (approved):

dereknance (approved):

PR #950, opened 2026-04-13
Remove concept of SDK SME ownership group

trmartin4:
## 📔 Objective

As we mature in our usage of the SDK, and we have more PRs coming in, we run the risk of overwhelming those few engineers who are in the `team-sdk-sme` group. 

This PR proposes that we remove this phase of the SDK onboarding process, and teams will move straight from partnership with Platform as co-owners to full owners of their domain.

Note that this does add `team-platform-dev` as shared codeowners for the `/key-management` directory in `bitwarden-core`.  The reasoning there is that Platform should be aware and working with KM to actively shrink the footprint of code in `bitwarden-core` and not in another, KM-owned, crate.


trmartin4:
@Hinton I talked to @dani-garcia about this, and he already gets notified of `sdk-internal` PRs from Github directly through email notifications, and I proposed that we remove this group to avoid any confusion around whether teams need to wait on them for review, as more teams come on board to the SDK.  If you are deriving value from it we can keep it, but we felt like it might not be worth it at this time.

dereknance:
I resolved conflicts I caused by approving #1019, which added the `bitwarden-collections` crate to Admin Console's ownership (plus Platform) and sorted the `Team-owned crates` section alphabetically.

Hinton (approved):

dani-garcia (approved):

PR #952, opened 2026-04-13
[PM-35077] Fix add emergency access filtering for key rotation

Thomas-Avery:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-35077

## 📔 Objective

Add filtering out `Invited` and `Accepted` emergency access as they don't have public keys and this currently breaks SDK rotation as is.

TS clients do this here https://github.com/bitwarden/clients/blob/2c2067388aabdce5c5d9a7212a94c38ca91df403/apps/web/src/app/auth/emergency-access/services/emergency-access.service.ts#L379-L393


eligrubb (approved):
Looks good!

PR #958, opened 2026-04-14
[PM-34716] Implement legacy CLI generate command within Rust CLI

harr1424:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-34716

## 📔 Objective

Implement `bw generate` within the Rust CLI to have full parity with existing CLI.


crates/bw/src/tools/mod.rs
```diff
@@ -40,25 +50,44 @@ pub struct GenerateArgs {
...
 }
 
 impl GenerateArgs {
     pub fn run(mut self, client: &PasswordManagerClient) -> CommandResult {
         let result = if self.passphrase {
+            // If a user specifies less than three words, the number of words will be silently
+            // increased to three, matching behavior in the legacy CLI
+            if self.words < 3 {
+                self.words = 3;
+            }
```
Hinton: Do you know if these are CLI concerns or if they occur for all clients?

It might be nicer to do `cmp::min(self.words, 3)` further down when you call `client.generator()` since you can drop `mut` on `self`.

dani-garcia: This is also something that the current passphrase generator code returns an error from: 
https://github.com/bitwarden/sdk-internal/blob/b7278278dc4addcb315c6169857161fc69012ae9/crates/bitwarden-generators/src/passphrase.rs#L68

Maybe we can just make the constants used there public just to make sure they're not out of sync? Or if we don't care about erroring for this, increasing the minimum there.

harr1424: @Hinton  this behavior is present across all clients: requesting a passphrase with fewer than three words, or more than twenty words, will result in the limit being silently enforced with no message to the user.

harr1424: Please see 3469bc82be08fca6f77f5e401e3b8e225c6f71aa

dani-garcia (approved):

adudek-bw (approved):

PR #961, opened 2026-04-15
Add missing functions to Send API calls

adudek-bw:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-31068

## 📔 Objective

Adds missing functions required for PM-31068.

## 🚨 Breaking Changes

https://github.com/bitwarden/clients/pull/20170


Hinton:
File uploading would only work for self-hosted servers since cloud uses direct upload to Azure. There are plans to build a proper abstraction around this in https://bitwarden.atlassian.net/browse/PM-7594 but that's a bit out still.

crates/bitwarden-send/src/access.rs
```diff
@@ -0,0 +1,290 @@
...
+#[derive(Debug, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+#[cfg_attr(feature = "wasm", derive(Tsify), tsify(into_wasm_abi))]
+pub struct SendAccessFileView {
+    /// The file ID
+    pub id: Option<String>,
+    /// Encrypted file name
+    pub file_name: Option<String>,
+    /// File size in bytes as a string
+    pub size: Option<String>,
```
harr1424, line 61: It looks like this is already set to match changes in [server #7474](https://github.com/bitwarden/server/pull/7474/changes)  ✅

crates/bitwarden-send/src/access.rs
harr1424, line 1: 🎨  I like the comment style, especially indicating the request path on the server
❓  Would it make sense to add some test coverage here?

crates/bitwarden-send/src/create_file_send.rs
harr1424, line 1: ❓ This logic might be more involved to mock, but at some point I think test coverage would be useful

crates/bitwarden-send/src/remove_password.rs
```diff
@@ -0,0 +1,193 @@
...
+    /// Remove the password from a [Send], saving the updated state to the server and local state.
+    pub async fn remove_password(
+        &self,
+        send_id: SendId,
+    ) -> Result<SendView, RemoveSendPasswordError> {
+        let key_store = self.client.internal.get_key_store();
+        let config = self.client.internal.get_api_configurations();
+        let repository = self.get_repository()?;
+
+        remove_send_password(key_store, &config.api_client, repository.as_ref(), send_id).await
```
harr1424, line 57: This is more applicable to the API, non-blocking, and only adds clarity, but `PutRemovePassword` only calls `PutRemoveAuth` which will remove all auth from a Send (not just password, but email OTP as well). 

https://github.com/bitwarden/server/blob/44b818acc69ec7eb52d2cdb89ae3d3bec46a29f4/src/Api/Tools/Controllers/SendsController.cs#L447-L463

✅  Test coverage already checks this logic

harr1424 (dismissed):
No blockers found, please consider the comments involving test coverage, and API method naming changes.

crates/bitwarden-send/src/access.rs
ike-kottlowski, line 1: 🎨 : Accessing Send data within is outside of the scope of Auth module. The only action that should be accomplished here is to fetch and manage issuing a token to the consuming client.

The methods fetching encrypted data from the server should be moved to tools ownership. From an Auth perspective, we only want to handle requests to the `token/` endpoint. 

cc: @JaredSnider-Bitwarden

ike-kottlowski (dismissed):
Changes look good, but I have a concern around code organization within the Auth Module.

crates/bitwarden-send/src/access.rs
```diff
@@ -0,0 +1,290 @@
...
+    size: Option<String>,
+    size_name: Option<String>,
+}
+
+#[derive(Debug, Deserialize)]
+#[serde(rename_all = "camelCase")]
+struct SendFileDownloadDataApiResponse {
+    id: Option<String>,
+    url: Option<String>,
+}
```
coroiu: questions: are these api models not already available in the auto-generated bindings? e.g. crates/bitwarden-api-api/src/models/send_file_model.rs

crates/bitwarden-send/src/access.rs
```diff
@@ -0,0 +1,290 @@
...
+        .map_err(ApiError::from)?;
+
+    let status = response.status();
+    if !status.is_success() {
+        let message = response.text().await.unwrap_or_default();
+        return Err(ApiError::ResponseContent { status, message }.into());
+    }
+
+    let raw: SendAccessApiResponse = response.json().await.map_err(ApiError::from)?;
+    Ok(raw.into())
```
coroiu: question: can we not use the automated API bindings? E.g. `crates/bitwarden-api-api/src/apis/sends_api.rs`

crates/bitwarden-send/src/create_file_send.rs
```diff
@@ -0,0 +1,236 @@
...
+    pub async fn create_file_send(
+        &self,
+        request: SendAddRequest,
+    ) -> Result<CreateFileSendView, CreateFileSendError> {
+        let key_store = self.client.internal.get_key_store();
+        let config = self.client.internal.get_api_configurations();
+        let repository = self.get_repository()?;
+
+        create_file_send(key_store, &config.api_client, repository.as_ref(), request).await
+    }
```
coroiu, line 167: issue: Do not delegate method bodies to free functions. This splits the implementation away from the API surface, makes the client harder to navigate, and obscures what the method actually does in generated documentation.

See: [Anti-pattern: thin passthroughs](https://contributing.bitwarden.com/architecture/sdk/client-patterns#per-method-files-or-subdirectories)

crates/bitwarden-auth/src/send_access/client.rs
```diff
@@ -26,6 +27,71 @@ impl SendAccessClient {
...
+    /// The returned [SendAccessView] contains encrypted fields that must be decrypted
+    /// client-side using the key derived from the URL fragment.
+    pub async fn access_send(
+        &self,
+        access_token: String,
+    ) -> Result<SendAccessView, AccessSendError> {
+        let configurations = self.client.internal.get_api_configurations();
+        crate::send_access::access_send::access_send(&configurations.api_config, &access_token)
+            .await
+    }
```
coroiu: question: How and where are you envisioning that these are going to be used? Currently you are using the `api_client` that is already constructed within the `Client` (i.e. `self.client.internal.get_api_configurations()`) which will already have an `api_url` configured, which means that you won't be able to access sends from other environments

crates/bitwarden-send/src/access.rs
adudek-bw: The auto-generated bindings for these endpoints (access, access_using_auth, get_send_file_download_data, get_send_file_download_data_using_auth) all return Result<(), Error> via process_with_empty_response, discarding the response body entirely. We need the actual JSON response (the encrypted send metadata) to be useful. There are also no auto-generated response models for the send access response shape — SendFileModel covers file metadata in the authenticated send management context, but the send access response is a different shape that isn't in the OpenAPI spec.

crates/bitwarden-send/src/access.rs
adudek-bw: Same reason as above — the generated API calls all use process_with_empty_response for these endpoints, so the response body is discarded. Additionally, the V1 endpoint (access) uses AuthRequired::Bearer in the generated code, which would attach the user's bearer token — but send access is unauthenticated (it uses either a Send-Id header or a send access token, not a user session token). The V1 variant also requires a Send-Id header that the generated code doesn't include.

ike-kottlowski (commented):
Auth concerns have been removed. Thanks for moving the code to Tools ownership!

crates/bitwarden-send/src/access.rs
Hinton: That seems to be an error with the controllers? They should output correct request/responses and you should not need to hand write APIs.

crates/bitwarden-auth/src/send_access/client.rs
adudek-bw: These are intended for use by the Bitwarden web/browser clients to access sends on the server they're already configured against — the api_url in the client is set to the correct server before these methods are called. For self-hosted sends, the client would be initialized with the self-hosted api_url. This is the same pattern used by all other sends operations (create, edit, delete) in SendClient.

crates/bitwarden-send/src/access.rs
Hinton: After further discussions with @adudek-bw, I believe we can update the server bindings to include the correct server result. We do need to create our own instance of the `ApiClient` since we have to pass in a different auth header which might mean creating a new token handler which uses a static value.

coroiu (commented):
This code seems out of date? Please re-request review when it has been adapted to the server PR (or if I'm wrong and this is ready for review 🙂)

theMickster:
Howdy @adudek-bw et. al 👋🏼

Many Thanks! 


---

<details>
<summary>Opus - Code Review: Add missing functions to Send API calls (#961)</summary>

## Summary

| Severity     | Count |
| ------------ | ----- |
| 🛑 Blocker   | 0     |
| ⚠️ Important | 4     |
| ♻️ Refactor  | 6     |

This PR adds send-access, file-upload, delete, and remove-password operations to the
`bitwarden-send` crate. The structure (error types, impl separation, test coverage) is solid. The
four important findings are: two URL-path injection gaps in `access.rs` where `send_id`/`file_id`
are embedded without percent-encoding (inconsistent with the same PR's `upload_send_file`); an
ordering bug in `create_file_send` where a repository write precedes a decrypt that can fail; and a
misuse of the `View` suffix on types that carry still-encrypted vault data. The refactor findings
center on `access.rs` diverging from the crate's established HTTP pattern and introducing raw
primitive types where typed enums already exist.

## Findings

### ⚠️ Important

#### `SendAccessView`/`SendAccessTextView`/`SendAccessFileView` misuse the `View` suffix convention

`crates/bitwarden-send/src/access.rs:52-68` **Caught by:** Architecture agent

The sdk-internal CLAUDE.md Data Model Conventions table reserves the `View` suffix for **decrypted
DTOs returned to clients** (`CipherView`, `SendView`). The convention is followed consistently in
this crate: `SendFile` (encrypted, `EncString` fields) vs `SendFileView` (decrypted, `String`
fields).

The three new types carry encrypted content — their own doc comments confirm this:

- `SendAccessView.name` — "Encrypted send name"
- `SendAccessTextView.text` — "Encrypted text content"
- `SendAccessFileView.file_name` — "Encrypted file name"

A developer reading `SendAccessView` will reasonably assume its fields are plaintext per the rest of
the crate, and may pass them directly to UI or log them without decrypting first. In a
security-sensitive codebase where the zero-knowledge invariant depends on correct handling of
encrypted vs decrypted field state, this naming mismatch creates a genuine mishandling risk for
downstream consumers (Vault Data classification).

**Suggested fix:** Rename to `SendAccessResponse`/`SendAccessTextResponse`/`SendAccessFileResponse`
per the `Response` suffix ("Public output DTO from SDK to client") in the conventions table. Reserve
a future `SendAccessView` for the post-decryption result produced after the caller applies the
URL-fragment key.

Note: `SendFileDownloadData` (no suffix) is already correctly named — it's not a View.

</details>

---

#### `send_id` interpolated into URL path without `urlencode` in V1 access

`crates/bitwarden-send/src/access.rs:137` **Caught by:** Security & logic agent

`access_send_v1` constructs:

```rust
let url = format!("{}/sends/access/{}", api_config.base_path, send_id);
```

without calling `bitwarden_api_base::urlencode()`.

`upload_send_file` in the same PR correctly encodes both segments:

```rust
let url = format!(
    "{}/sends/{}/file/{}",
    config.api_config.base_path,
    bitwarden_api_base::urlencode(send_id.to_string()),
    bitwarden_api_base::urlencode(&file_id),
);
```

The `send_id` parameter is a `String` accepted from WASM callers. A value containing `/`, `?`, or
`#` (e.g., `"abc/../../admin"`) would alter the request path, potentially redirecting it to a
different endpoint or breaking the `Send-Id` header/path correlation the server expects. Implicates
SC (Secure Channels — request must reach the intended resource) and P05.

**Suggested fix:**

```rust
let url = format!("{}/sends/access/{}", api_config.base_path, bitwarden_api_base::urlencode(send_id));
```

</details>

---

#### `send_id` and `file_id` interpolated into URL path without `urlencode` in V1 file download

`crates/bitwarden-send/src/access.rs:194-197` **Caught by:** Security & logic agent

`get_file_download_data_v1` constructs:

```rust
let url = format!(
    "{}/sends/{}/access/file/{}",
    api_config.base_path, send_id, file_id
);
```

Both `send_id` and `file_id` are interpolated without `urlencode()`. A malformed `file_id`
containing `/` could escape into a sibling path segment; a `#` would truncate the path entirely.
Same inconsistency with `upload_send_file` as sec-1 above. Implicates SC and P05.

**Suggested fix:**

```rust
let url = format!(
    "{}/sends/{}/access/file/{}",
    api_config.base_path,
    bitwarden_api_base::urlencode(send_id),
    bitwarden_api_base::urlencode(file_id)
);
```

</details>

---

#### Repository written before decrypt — failed decrypt leaves orphan in client store

`crates/bitwarden-send/src/create_file_send.rs:132-134` **Caught by:** Security & logic agent

In `create_file_send`, the current order is:

```rust
let send_id = require!(send.id);
repository.set(send_id, send.clone()).await?;   // line 132 — writes to store
let send_view = key_store.decrypt(&send)?;       // line 134 — can fail
```

If `key_store.decrypt` fails (key store in unexpected state, key slot missing, transient crypto
error), the function returns `Err(CryptoError)` to the caller. The caller receives no `send_id`, has
no recovery path, and cannot clean up — but the local repository now contains a Send the server
successfully created and the store durably holds. This orphan cannot be detected or deleted without
a full resync.

**Suggested fix:** Decrypt before writing to the repository:

```rust
let send_view = key_store.decrypt(&send)?;       // fail fast before mutation
let send_id = require!(send.id);
repository.set(send_id, send.clone()).await?;    // only write if decrypt succeeded
```

Implicates P06 (minimizing impact of failures).

</details>

---

### ♻️ Refactor

#### `access.rs` bypasses generated API client while every sibling file in this PR uses `api_client.sends_api()`

`crates/bitwarden-send/src/access.rs:132-225` **Caught by:** Architecture agent

Every other new file in this PR uses the generated trait:

- `delete.rs` → `config.api_client.sends_api().delete(...)`
- `remove_password.rs` → `config.api_client.sends_api().put_remove_password(...)`
- `create_file_send.rs` → `config.api_client.sends_api().post_file(...)`

`access.rs` uniquely calls `api_config.client.post(&url)` directly, hand-formats headers and
response parsing, and re-declares `SendAccessApiResponse` / `SendAccessFileApiResponse` /
`SendFileDownloadDataApiResponse` as private structs (models the auto-generated client would own).

The PR simultaneously adds an `x-explicit-bearer-token` extension to
`support/openapi-template/reqwest-trait/api.mustache` — the stated purpose of which is to generate
bearer-only endpoint methods with an explicit `access_token` parameter. However `access.rs` never
uses the generated client, so the template change has no consumer in this PR. The two approaches
(template-generated and hand-rolled) are both present but only the hand-rolled one is wired up,
leaving the template change dead.

**Suggested path:** Either generate the access endpoints through the API client (apply
`x-explicit-bearer-token` to the OpenAPI spec and regenerate), or if regeneration is out of scope
for this PR, defer the template change until a consumer exists and document why `access.rs` bypasses
the generated client.

</details>

---

#### Four hand-rolled HTTP request functions in `access.rs` duplicate the same boilerplate

`crates/bitwarden-send/src/access.rs:132-225` **Caught by:** Code quality agent

`access_send_v1`, `access_send`, `get_file_download_data_v1`, and `get_file_download_data` follow an
identical structure: format URL → `client.post(&url)` → attach one header (either `Send-Id` or
`Authorization: Bearer`) → optional JSON body → `.send().await` → check `status.is_success()` →
`response.text()` on failure → `response.json()` on success.

The full pattern is repeated approximately 25 lines × 4 = ~100 lines in a single file. The only
variation per function is: URL template, auth header name and value, and whether a
`SendAccessRequestModel` body is attached.

If the recommended path (above) of routing through `sends_api()` is taken, this duplication
disappears. If the bypass is retained, factoring a single helper reduces maintenance surface.

</details>

---

#### `SendAccessView.type_` uses raw `i32` instead of the typed `SendType` enum

`crates/bitwarden-send/src/access.rs:57` **Caught by:** Code quality agent

`pub type_: Option<i32>` with the doc comment `(0 = Text, 1 = File)` reimplements what `SendType`
already encodes. The rest of the crate (`SendView::type_`, `SendAddRequest`) uses `SendType`, which
is a `serde_repr`-derived enum and deserializes directly from the same integer wire format. Exposing
a raw `i32` here forces every consumer to reproduce the `0 = Text, 1 = File` mapping from a comment
rather than from the type system, and inconsistently with every other send-related struct in the
crate.

**Suggested fix:** Use `pub type_: Option<SendType>` — `SendType` already derives `Deserialize` via
`serde_repr` and maps the same wire integers.

</details>

---

#### `SendAccessView.expiration_date` uses `Option<String>` instead of `Option<DateTime<Utc>>`

`crates/bitwarden-send/src/access.rs:65` **Caught by:** Code quality agent

Every other public Send DTO in this crate models dates as `DateTime<Utc>` (e.g.,
`SendView::expiration_date: Option<DateTime<Utc>>`, `SendAddRequest::expiration_date`). The raw
`Option<String>` here:

- Forces every caller to re-parse the ISO 8601 string.
- Silently accepts malformed date strings at the API boundary.
- Is inconsistent with the `chrono` usage already pulled in by this crate.

**Suggested fix:** Use `Option<DateTime<Utc>>` with chrono's serde feature
(`#[serde(with = "chrono::serde::ts_seconds_option")]` or the ISO string counterpart already used
elsewhere in the crate).

</details>

---

#### `CreateFileSendView.file_upload_type` uses raw `i32` instead of `FileUploadType` enum

`crates/bitwarden-send/src/create_file_send.rs:30` **Caught by:** Code quality agent

`pub file_upload_type: i32` with doc comment
`(0 = Direct to Bitwarden server, 1 = Azure Blob Storage)` duplicates
`bitwarden_api_api::models::FileUploadType`, which the API response already returns. The constructor
coerces and silently defaults:

```rust
let file_upload_type = resp
    .file_upload_type
    .map(|t| t.as_i64() as i32)
    .unwrap_or(0);  // silently falls back to Direct upload
```

Defaulting to `0` (Direct) when the server omits the field could cause a file to be uploaded to the
wrong storage backend without any error or warning.

**Suggested fix:** Expose `file_upload_type: Option<FileUploadType>` (or `FileUploadType` with an
explicit fallback in the caller, not silently in the constructor). This preserves type safety and
makes the default explicit.

</details>

---

#### Authorization header for send access bypass uses `format!("Bearer {}")` instead of middleware pattern

`crates/bitwarden-send/src/access.rs:169-172` **Caught by:** Security & logic agent

The rest of the SDK attaches bearer auth via `with_extension(AuthRequired::Bearer)`, which is
applied by the `bitwarden-api-base` middleware layer. `access_send` and `get_file_download_data`
instead do:

```rust
.header(reqwest::header::AUTHORIZATION, format!("Bearer {}", access_token))
```

While reqwest validates header values before sending (control characters would raise an error rather
than silently leaking the token), the direct header construction bypasses any redaction or
sanitization the middleware provides. This is primarily a code-consistency issue (pattern
established for how bearer auth is applied in this SDK), not an active token-leak vulnerability. The
correct pattern is `with_extension(AuthRequired::Bearer)` — though the V2 access endpoints take an
explicit per-call token rather than the ambient client token, so the middleware approach would need
to be combined with the new `x-explicit-bearer-token` template extension or an equivalent mechanism.

</details>

---

## Reviewed and Dismissed

<details><summary>🔍 3 initial findings dismissed after validation</summary>

#### New modules not declared in `lib.rs`

`crates/bitwarden-send/src/lib.rs:13-20` **Caught by:** Architecture agent **Original severity:** ⚠️
Important **Original confidence:** 90/100 **Dismissed at:** Step 4 validation **Dismissed because:**
The PR diff explicitly includes `mod` declarations and `pub use` re-exports for all four new modules
(`access`, `create_file_send`, `delete`, `remove_password`) in `lib.rs`. The premise of the finding
is false.

---

#### Raw response body included verbatim in error message on non-2xx

`crates/bitwarden-send/src/access.rs:150-152` **Caught by:** Security & logic agent **Original
severity:** ♻️ Refactor **Original confidence:** 80/100 **Dismissed at:** Step 4 validation
**Dismissed because:** Per the finding's own detail, this exactly matches the pattern used by the
auto-generated `bitwarden-api-api` clients throughout the SDK. It is a pre-existing concern not
introduced or worsened by this change.

---

#### `SendAccessView`/`SendAccessTextView`/`SendAccessFileView` violate the `View` suffix convention (duplicate)

`crates/bitwarden-send/src/access.rs:52-68` **Caught by:** Code quality agent **Original severity:**
♻️ Refactor **Original confidence:** 80/100 **Dismissed at:** Step 5 severity audit **Dismissed
because:** Duplicate of arch-1 (Architecture agent finding), which covers the same issue at higher
confidence and with a more complete analysis. Consolidated under that finding to avoid
double-reporting.

</details>

</details>

---

<details>
<summary> Sonnet Code Review: Add missing functions to Send API calls (#961)</summary>

## Summary

| Severity     | Count |
| ------------ | ----- |
| 🛑 Blocker   | 0     |
| ⚠️ Important | 1     |
| ♻️ Refactor  | 6     |

This PR adds four new modules to `bitwarden-send` covering send access, file send creation/upload,
deletion, and password removal. The `delete.rs` and `remove_password.rs` files follow established
crate conventions well and have solid test coverage. The main concerns are in `access.rs`: the four
Send access functions bypass the auto-generated `SendsApi` trait (issuing raw `reqwest` HTTP calls
instead), define hand-rolled wire-format DTOs that should live in `bitwarden-api-api`, and use the
`*View` suffix for structs that carry encrypted — not decrypted — data, inverting the SDK's
documented naming contract. One concrete URL-encoding inconsistency in the V1 functions is a latent
correctness risk.

## Findings

### ⚠️ Important

#### `send_id` and `file_id` interpolated into V1 URL paths without percent-encoding

`crates/bitwarden-send/src/access.rs:126-147` **Caught by:** Security & logic agent

`access_send_v1` and `get_file_download_data_v1` construct URL paths using `format!()` with
`send_id` and `file_id` interpolated directly:

```rust
let url = format!("{}/sends/access/{}", api_config.base_path, send_id);
let url = format!("{}/sends/{}/access/file/{}", api_config.base_path, send_id, file_id);
```

In the **same PR**, `upload_send_file` in `create_file_send.rs` explicitly encodes the same URL
slots:

```rust
let url = format!(
    "{}/sends/{}/file/{}",
    config.api_config.base_path,
    bitwarden_api_base::urlencode(send_id.to_string()),
    bitwarden_api_base::urlencode(&file_id),
);
```

`send_id` in the V1 functions is a caller-supplied `String` (access ID from a Send URL), and
`file_id` is a server-returned `Option<String>`. While current IDs are base64url or UUIDs
(URL-safe), both slots lack encoding. If the server ever returns a `file_id` containing `/` or other
path-special characters, the request routes to an unintended endpoint. The divergence from
`upload_send_file` within the same PR is concrete and unambiguous.

**Fix:** Apply `bitwarden_api_base::urlencode()` to all path parameters in all four functions in
`access.rs`, matching the pattern from `create_file_send.rs`:

```rust
let url = format!("{}/sends/access/{}", api_config.base_path,
    bitwarden_api_base::urlencode(send_id));
```

  </details>

### ♻️ Refactor

#### Raw reqwest HTTP calls in `access.rs` bypass the auto-generated `SendsApi` trait

`crates/bitwarden-send/src/access.rs:99-200` **Caught by:** Architecture agent

The four functions `access_send_v1`, `access_send`, `get_file_download_data_v1`, and
`get_file_download_data` construct URLs by hand and issue raw HTTP calls via
`api_config.client.post(&url).send().await`, bypassing `bitwarden_api_api::apis::SendsApi`. Every
other operation in this PR — and in `create.rs`, `edit.rs`, `get_list.rs` — goes through
`config.api_client.sends_api().*()`.

The mismatch means that changes to these send-access endpoints in the OpenAPI spec won't
automatically propagate to the SDK via regeneration; they require manual updates to the hand-rolled
URL strings and response types in `access.rs`.

**Fix:** Update the OpenAPI spec to model the response types for the Send access endpoints,
regenerate `bitwarden-api-api`, and route these calls through `config.api_client.sends_api().*()`
like every other operation in the crate.

  </details>

#### Hand-written wire-format DTOs duplicate work that belongs in auto-generated `bitwarden-api-api` models

`crates/bitwarden-send/src/access.rs:16-39` **Caught by:** Architecture agent

`SendAccessApiResponse` and `SendFileDownloadDataApiResponse` are hand-written deserialization
shapes for server responses. Per `CLAUDE.md`: _"bitwarden-api-api, bitwarden-api-identity:
Auto-generated API clients — DO NOT edit — regenerate from OpenAPI specs."_ Every other API response
model in the workspace lives under `crates/bitwarden-api-api/src/models/`. The corresponding request
model (`SendAccessRequestModel`) already lives there — only the response side is missing.

Defining these locally means: (a) future server-side field additions won't be picked up
automatically; (b) the `From<…ApiResponse> for …View` conversion impls are additional maintenance
surface; (c) contributors looking for the response shape in `bitwarden-api-api/src/models` won't
find it.

**Fix:** Add `[ProducesResponseType]` annotations (or equivalent) to the server-side controllers,
regenerate `bitwarden-api-api`, and map directly from the generated models. This also eliminates the
need for the raw HTTP calls flagged above.

  </details>

#### `SendAccessView` / `SendAccessTextView` / `SendAccessFileView` use the `*View` suffix but contain encrypted data

`crates/bitwarden-send/src/access.rs:44-77` **Caught by:** Architecture agent and Security & logic
agent

The project `CLAUDE.md` documents the SDK data-model layer conventions explicitly:

- **View** — Decrypted DTO returned to clients (examples: `CipherView`, `SendView`)
- **Response** — Public output DTO from SDK to client

The new types use the `*View` suffix, but their struct doc comments state the content is still
encrypted: _"Name, text, and file fields are encrypted and must be decrypted client-side using the
key derived from the URL fragment."_ The fields `name`, `text.text`, and `file.file_name` are
encrypted ciphertext wrapped in `Option<String>` rather than the SDK's usual `EncString` type.

This inverts the naming contract every other type in the codebase follows — `SendView` and
`SendTextView` hold plaintext after decryption, while these new `*View` types hold ciphertext before
decryption. TypeScript and mobile callers who see `SendAccessView` will naturally assume its fields
are plaintext, consistent with every other `*View` in the SDK.

**Fix:** Rename to `SendAccessResponse`, `SendAccessTextResponse`, `SendAccessFileResponse` to match
the `Response` convention for output DTOs that carry still-encrypted data. This is already a
breaking-change PR, so there is no additional cost to renaming.

  </details>

#### `SendAccessView::type_` uses raw `Option<i32>` instead of the existing `SendType` enum

`crates/bitwarden-send/src/access.rs:55-57` **Caught by:** Architecture agent

`send.rs` defines `pub enum SendType { Text = 0, File = 1 }` as a `#[repr(u8)]` enum wired up to
`Tsify` and `uniffi::Enum`, so TypeScript and mobile callers see a real enum. This is used
consistently in `SendView` and `SendListView`. The new field `pub type_: Option<i32>` duplicates the
same semantics as prose in a doc comment, weakening the WASM/UniFFI contract (TypeScript callers see
`number | null` instead of the `SendType` enum).

**Fix:** Change the field to `pub type_: Option<SendType>`.

  </details>

#### `file_upload_type` erases the typed `FileUploadType` enum to a raw `i32`

`crates/bitwarden-send/src/create_file_send.rs:96-99` **Caught by:** Code quality agent

```rust
let file_upload_type = resp
    .file_upload_type
    .map(|t| t.as_i64() as i32)
    .unwrap_or(0);
```

`FileUploadType` is an enum with `Direct` (0), `Azure` (1), and `__Unknown(i64)` variants. The
conversion discards all type-system information, leaving `CreateFileSendView.file_upload_type: i32`
— an opaque integer callers must interpret manually. Comparable types elsewhere in the workspace
preserve the enum. Callers who need to dispatch on Direct vs Azure upload paths now re-implement the
variant mapping themselves.

**Fix:** Change `CreateFileSendView.file_upload_type` to `bitwarden_api_api::models::FileUploadType`
and assign directly:

```rust
let file_upload_type = resp.file_upload_type.unwrap_or_default();
```

  </details>

#### `*View` naming on encrypted DTOs violates the SDK's documented decryption-state contract

`crates/bitwarden-send/src/access.rs:52-108` **Caught by:** Security & logic agent

This finding overlaps with the arch-4 naming finding above but is retained because it identifies the
security-layer implication: the naming mismatch creates a documentation-level confusion between
encrypted ciphertext (what these types hold) and decrypted plaintext (what the `*View` suffix
promises) specifically for cross-platform callers consuming WASM/UniFFI-generated bindings who may
not read Rust source doc comments.

The doc comments on each struct warn about the encryption, which mitigates the worst-case risk, but
the type naming is the primary contract seen in generated TypeScript and Kotlin interfaces.

**Fix:** Same as the arch-4 finding — rename to `*Response`, and consider using `EncString` fields
for `name`, `text.text`, and `file.file_name` to make the encrypted/decrypted boundary explicit at
the type level, consistent with how `Send` (the domain model) uses `EncString`.

  </details>

## Reviewed and Dismissed

  <details><summary>🔍 3 initial findings dismissed after validation</summary>

#### Adding `reqwest` as a direct dependency to a Feature Implementation crate

`crates/bitwarden-send/Cargo.toml:38` **Caught by:** Architecture agent **Original severity:** ♻️
Refactor **Original confidence:** 85/100 **Dismissed at:** Step 5 severity audit **Dismissed
because:** `reqwest` is already a direct workspace dependency in `bitwarden-vault`,
`bitwarden-auth`, and `bitwarden-generators`. No documented architectural rule prohibits feature
crates from depending on it; the finding treated an established codebase pattern as a violation.

---

#### `UploadSendFileError::Reqwest` variant is missing `#[error(...)]` attribute

`crates/bitwarden-send/src/create_file_send.rs:55-58` **Caught by:** Code quality agent **Original
severity:** 🛑 Blocker **Original confidence:** 95/100 **Dismissed at:** Step 4 validation
**Dismissed because:** The actual diff shows `#[error(transparent)]` IS present on the
`Reqwest(reqwest::Error)` variant. The missing `#[from]` on that variant is intentional — the error
is wrapped manually via `.map_err(UploadSendFileError::Reqwest)`. This is not a compile error.

---

#### `test_remove_send_password` and `test_remove_send_password_http_error` are empty stubs

`crates/bitwarden-send/src/remove_password.rs:176-177` **Caught by:** Code quality agent **Original
severity:** ⚠️ Important **Original confidence:** 97/100 **Dismissed at:** Step 4 validation
**Dismissed because:** The actual diff shows fully implemented tests with mock wiring
(`ApiClient::new_mocked`), concrete assertions (`assert!(result.is_ok())`,
`assert_eq!(view.id, ...)`, `assert!(!view.has_password)`,
`assert_eq!(view.auth_type, AuthType::None)`), and error path verification. The `{ /* ... */ }`
bodies seen during analysis were a diff-truncation artifact, not the real content.

  </details>


</details>

---

<details>
<summary>Code Review Comparison: PR #961 — Opus vs Sonnet</summary>

## Counts

| | Opus | Sonnet |
|---|---|---|
| 🛑 Blocker | 0 | 0 |
| ⚠️ Important | 4 | 1 |
| ♻️ Refactor | 6 | 6 |
| Dismissed | 3 | 3 |

---

## Findings Agreement

Both models caught the same core issues:

- URL encoding missing on V1 path params (Opus split to 2 findings, Sonnet merged to 1)
- `*View` suffix on encrypted types (disagreed on severity — see below)
- `access.rs` bypassing the generated `SendsApi` trait
- Hand-rolled DTOs that belong in `bitwarden-api-api`
- `type_: Option<i32>` instead of `Option<SendType>`
- `file_upload_type: i32` instead of `FileUploadType` enum

---

## What Opus Found That Sonnet Missed

**1. Repository written before decrypt — functional bug** (`create_file_send.rs:132-134`, ⚠️ Important)

The most significant miss. In `create_file_send`, `repository.set()` runs before `key_store.decrypt()`. If decrypt fails, the local store contains an orphaned Send the caller can't identify or delete — it's invisible without a full resync. Sonnet did not flag this at all.

```rust
repository.set(send_id, send.clone()).await?;  // written
let send_view = key_store.decrypt(&send)?;     // can fail — store is now dirty
```

**2. `expiration_date: Option<String>` instead of `Option<DateTime<Utc>>`** (`access.rs:65`, ♻️ Refactor)

Every other public Send DTO uses `DateTime<Utc>`. This forces callers to re-parse, silently accepts malformed dates, and is inconsistent with the crate's own `chrono` usage. Sonnet missed it.

**3. Template change has no consumer** (arch-1 detail)

Opus observed that the `x-explicit-bearer-token` mustache change adds code-gen infrastructure that `access.rs` never uses — the hand-rolled HTTP calls and the template extension coexist without connecting. The template change is effectively dead until the raw calls are replaced. Sonnet described the same architectural divergence but didn't make this connection.

**4. Boilerplate duplication across 4 HTTP functions** (`access.rs:132-225`, ♻️ Refactor)

Four functions with identical 25-line patterns (format URL → attach one header → optional body → send → check status → parse). Sonnet didn't surface this.

**5. Manual `format!("Bearer {}")` instead of middleware pattern** (`access.rs:169-172`, ♻️ Refactor)

The SDK uses `with_extension(AuthRequired::Bearer)` everywhere else. `access.rs` hardcodes the header string directly. Sonnet didn't flag this.

---

## What Sonnet Found That Opus Didn't

Nothing substantive. Sonnet's extra dismissed finding (`reqwest` as a direct dep in a feature crate) was correctly dismissed when the validator confirmed `reqwest` is already a direct dep in other feature crates. Sonnet also kept the `*View` naming as two separate findings (arch-4 + sec-1) rather than consolidating, which created redundancy.

---

## Severity Disagreement

**`*View` suffix on encrypted types**

- Opus: ⚠️ Important — downstream callers (TypeScript, mobile via UniFFI) will assume plaintext per every other `*View` in the SDK; mishandling encrypted vault data is a real VD/P01 risk.
- Sonnet: ♻️ Refactor — mitigated by doc comments, so it's a convention violation rather than an exploitable issue.

Opus is right. Doc comments don't appear in TypeScript type definitions or Kotlin/Swift generated bindings. Callers consuming `SendAccessView` via WASM or UniFFI have no runtime signal that `name` is ciphertext.

---

## False Positives

| Model | FP | What |
|---|---|---|
| Opus | 1 | "New modules not declared in lib.rs" — lib.rs changes were in the diff |
| Sonnet | 2 | "`#[error(transparent)]` missing" (it was present); "empty test stubs" (tests were complete — Sonnet's prompt truncation caused this) |

Sonnet's quality agent submitted a confidence-95 Blocker and a confidence-97 Important, both wrong. The blocker was a diff-reading error; the Important was caused by Sonnet's own truncated prompt being mistaken for reality. Opus made one false positive from a simpler misread.

---

## Summary Assessment

Opus outperformed on this PR. The single most important difference is the **orphan repository state bug** — a real functional issue that Opus caught and Sonnet didn't. Opus also found three additional refactor-level issues Sonnet skipped and correctly rated the `*View` naming as Important rather than Refactor. Sonnet's false-positive rate was higher (2 vs 1) and both were more embarrassing: the blocker was a diff-reading error, and the Important was triggered by context the model itself truncated. On shared findings, the analysis quality was comparable. For a security-sensitive Rust codebase, Opus produced a more complete and better-calibrated review.


</details>

crates/bitwarden-send/src/access.rs
adudek-bw: Alright, after talking to Oscar we have a server change to update the automated API bindings and use them. The updated SDK code will be added once the server code is merged :)

adudek-bw:
Hey y'all, currently waiting on https://github.com/bitwarden/server/pull/7556 to finish going through QA and merge. Then this will be ready for re-review :)

adudek-bw:
Hi all, I've updated the API bindings and rewrote a lot of this to use the new bindings. This should be good for re-review :)

crates/bitwarden-send/src/access.rs
adudek-bw: Added

harr1424:
It looks like a different PR is responsible, but as an FYI I ran the workflow to check for breaking changes and noticed [quite a few errors during the TypeScript compatibility check](https://github.com/bitwarden/clients/actions/runs/26254746670/job/77274611172).

harr1424:
@claude please give this a review

harr1424 (dismissed):

coroiu (commented):
We still seem to be using a custom middleware, didn't we make changes on the server to avoid that? Is that what the `api.mustache` changes are?

adudek-bw:
> We still seem to be using a custom middleware, didn't we make changes on the server to avoid that? Is that what the `api.mustache` changes are?

Good catch — the api.mustache change adds support for a new x-explicit-bearer-token vendor extension that, when applied to an endpoint, generates a signature taking the bearer token as a direct access_token: &str parameter and attaches the header inline (no middleware). That was the long-term plan we discussed in April.

The catch: the template was updated, but the send-access endpoints on the server's OpenAPI spec aren't marked with x-explicit-bearer-token yet, so when the bindings were regenerated in https://github.com/bitwarden/sdk-internal/pull/1110 they still came out with the middleware-based shape (async fn access_using_auth(&self) -> _ + AuthRequired::Bearer extension). You can confirm with grep x-explicit-bearer-token crates/bitwarden-api-api/ — no hits.

The SendAccessTokenHandler middleware is the bridge until that spec change lands. I'll open a follow-up to mark the send-access endpoints in the spec, regenerate the bindings, and drop the middleware in favor of the inline access_token parameter once that's in.

coroiu (commented):
@adudek-bw 

Reading through this, I think there might be a circular dependency in the PR that's making the diagnosis in your last comment hard to verify. Let me walk through what I'm seeing.

**The PR contains two things that work against each other.**

The first is the change to `api.mustache` (the one that teaches the codegen to put the bearer token into the generated method signature, so it produces `access_using_auth(&self, access_token: &str)` instead of `access_using_auth(&self)` plus an auth-extension marker).

The second is the new `SendAccessTokenHandler` middleware in the SDK code. That middleware only makes sense if the generated method doesn't take a token, because its whole job is to inject one via a request interceptor.

So one part of this PR adds the feature, and another part of this PR works around the absence of the feature. They can't both be load-bearing at the same time, which suggests one of them is doing work the other was supposed to make unnecessary.

**About the "still came out with the middleware-based shape" evidence.**

PR #1110 was merged on 2026-05-20, and the bindings it regenerated were produced against `main`. But the `api.mustache` change is only on this branch, it has never been on `main`. So the codegen inside #1110 ran with the old template, which has no handling for `x-explicit-bearer-token` (zero references to it).

That matters: with the old template, the bindings would emerge in the middleware shape regardless of whether the server spec carried the extension or not. So the `grep x-explicit-bearer-token crates/bitwarden-api-api/` result returning zero hits is consistent with both "the spec is missing something" and "the spec is fine, but the old template can't read the extension yet." It doesn't actually distinguish between the two.

**What that means for the diagnosis.**

Until somebody runs the new template (the one in this PR's branch) against the current server spec, either of these could be true:

1. The new template by itself is enough. The bindings come out with `access_token: &str` in the signature, no middleware needed, the `SendAccessTokenHandler` can be deleted.
2. The new template isn't enough on its own. The codegen needs the extension to live somewhere else in the spec (probably on the operations rather than only on the security scheme definition), and a small server-side follow-up really is needed.

Your comment lines up with case 2, and that might well be the right call, but I think the regen in #1110 doesn't give us enough information to know yet. Could be I'm missing something here, in which case happy to be corrected.

**Two ways to untangle this.**

**Option A (recommended): regenerate the bindings yourself, in this PR.**

The nice thing about this option is that you don't need to merge anything to find out whether the template change works. You can run the codegen locally first and see what comes out.

Here's the idea:

1. Run the codegen on your machine with the new template from this branch, pointing it at the same server commit #1110 already used (`25e78ceba`). Same server input as before means the only thing that should be different in the output is what *your template change* causes, nothing else gets pulled in.
2. Look at the new `sends_api.rs`. It tells you definitively which case we're in:
   - If `access_using_auth` now takes an `access_token: &str` parameter, the template change is enough. The middleware isn't needed, you can delete `SendAccessTokenHandler` and commit the regenerated bindings.
   - If it still emerges as `access_using_auth(&self)` plus the `AuthRequired::Bearer` extension, then yes, the server spec does need additional changes. But now it's empirical, not a hypothesis, and you can spin a targeted follow-up.
3. Commit the regenerated bindings into this PR, replacing the ones from #1110. The PR then carries: the template change, the bindings that prove it works, and the SDK code shaped to match. It all ships in one go.

The win: you find out *before* anything merges. No middleware on faith, no cleanup PR after the fact, no waiting for CI to pick it up next week.

**Option B: split the PR.**

Pull the `api.mustache` change into its own small PR and merge it first. After that lands, trigger a bindings regen manually using [`update-api-bindings.yml`](https://github.com/bitwarden/sdk-internal/actions/workflows/update-api-bindings.yml). It has `workflow_dispatch`, so you don't have to wait for the Monday cron. Then rebase this PR onto the new bindings.

Either way, the question is the same: what does the new template actually produce against the current spec? Once that's answered, it'll be clear whether the middleware needs to ship or whether it's working around a problem the PR itself fixes.

adudek-bw:
@coroiu Ahhh I see what you're saying! Working on option A now!

adudek-bw:
@coroiu alright I regenerated it and removed `SendAccessTokenHandler`! Thanks for the call out!

coroiu (approved):

harr1424 (approved):

PR #964, opened 2026-04-16
[PM-34152] Add PolicyRegistry for policy filtering

eliykat:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-34152

## 📔 Objective

Introduces `PolicyRegistry`, a type-erased registry that maps organization policy types to their filtering logic. This will support an FFI used by native callers to filter applicable policies without needing to know about the corresponding `Policy` implementation, which is a rust concept. (rust callers will likely skip this and just specify the `Policy` directly.)

  - Consumers register `Policy` implementations via `PolicyRegistry::builder().register(...).build()`.
  - At filter time, `filter_by_type` looks up the registered definition for a given policy type and applies its filtering rules (exempted roles, org membership requirements, etc.).
  - If no definition is registered for a type, it falls back to `DefaultPolicy`, which applies reasonable defaults that work for most policy types.

Usage example:

```rust
  let registry = PolicyRegistry::builder()
      .register(MasterPasswordPolicy)
      .register(TwoFactorAuthenticationPolicy)
      .build();

  let enforced = registry.filter_by_type(&raw_policies, &organizations, RawPolicyType(0));
```

## 🚨 Breaking Changes

No breaking changes - code is not wired up yet.

### DIFF
--- a/crates/bitwarden-policies/src/filter.rs
+++ b/crates/bitwarden-policies/src/filter.rs
@@ -14,25 +14,29 @@ use serde::{Deserialize, Serialize};
 use uuid::Uuid;
 
 /// A newtype representing the policy type.
-#[derive(PartialEq, Serialize, Deserialize, Debug)]
-pub struct PolicyType(pub u8);
+#[derive(PartialEq, Eq, Hash, Serialize, Deserialize, Debug, Copy, Clone)]
+pub struct PolicyType(
+    /// The raw integer value as defined by the server.
+    pub u8,
+);
 
 /// An organization policy.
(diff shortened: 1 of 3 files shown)


eliykat:
Builds are broken because `main` is broken - looks like it is unrelated to this PR.

@dani-garcia I rebased but the only substantive change since you took a quick look yesterday should be https://github.com/bitwarden/sdk-internal/pull/964/commits/dda4ff521a106175788e9a06e701a4b166fec36e.

dani-garcia (approved):

PR #965, opened 2026-04-16
[PM-31878] Persist UserId, AccountCrypto, and OrgSharedKeys

Hinton:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-31878
## 📔 Objective

To support session rehydration, write cryptographic state to the SDK's Settings and Repository stores as it is established at runtime:

- `init_user_id` is now async and writes the user ID to the `USER_ID` setting. The `USER_ID` key type is changed from `String` to `UserId` to avoid an unnecessary round-trip through string conversion.

- `initialize_user_crypto` writes `account_cryptographic_state` to the `ACCOUNT_CRYPTO_STATE` setting after the key store is populated.

- `initialize_org_crypto` writes each org key to the `OrganizationSharedKey` repository after the key store is populated.

All three writes are best-effort: failures emit a `tracing::warn` and do not propagate an error to the caller. The registry returning `DatabaseNotInitialized` (e.g. in tests without a backing store) is also silently ignored.

Tests are added for each persistence point:
- `initializing_user_multiple_times` now reads back the `USER_ID` setting and asserts it matches the initialised value.
- `test_initialize_user_crypto_persists_account_crypto_state` verifies the account crypto state round-trips through the setting store.
- `test_initialize_org_crypto_persists_org_keys` verifies that org keys appear in the `OrganizationSharedKey` repository after initialization.

### DIFF
--- a/crates/bitwarden-core/src/key_management/crypto.rs
+++ b/crates/bitwarden-core/src/key_management/crypto.rs
@@ -31,7 +31,11 @@ use crate::key_management::wasm_unlock_state::{
 };
 use crate::{
     Client, NotAuthenticatedError, OrganizationId, UserId, WrongPasswordError,
-    client::{LoginMethod, UserLoginMethod, encryption_settings::EncryptionSettingsError},
+    client::{
+        LoginMethod, UserLoginMethod,
+        encryption_settings::EncryptionSettingsError,
+        persisted_state::{ACCOUNT_CRYPTO_STATE, OrganizationSharedKey},
+    },
     error::StatefulCryptoError,
     key_management::{
         MasterPasswordError, PrivateKeySlotId, SecurityState, SignedSecurityState,
@@ -193,7 +197,7 @@ pub(super) async fn initialize_user_crypto(
     use crate::auth::{auth_request_decrypt_master_key, auth_request_decrypt_user_key};
 
     if let Some(user_id) = req.user_id {
-        client.internal.init_user_id(user_id)?;
+        client.internal.init_user_id(user_id).await?;
     }
 
     tracing::Span::current().record(
@@ -358,6 +362,12 @@ pub(super) async fn initialize_user_crypto(
         }))
         .await;
 
+    if let Ok(setting) = client.internal.state_registry.setting(ACCOUNT_CRYPTO_STATE)
+        && let Err(e) = setting.update(req.account_cryptographic_state).await
+    {
+        tracing::warn!("Failed to persist account crypto state: {e}");
+    }
+
     info!("User crypto initialized successfully");
 
     Ok(())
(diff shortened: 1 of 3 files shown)


crates/bitwarden-core/src/key_management/crypto.rs
```diff
@@ -378,8 +388,27 @@ pub(super) async fn initialize_org_crypto(
...
 ) -> Result<(), EncryptionSettingsError> {
-    let organization_keys = req.organization_keys.into_iter().collect();
-    client.internal.initialize_org_crypto(organization_keys)?;
+    let organization_keys: Vec<_> = req.organization_keys.into_iter().collect();
+    client
+        .internal
+        .initialize_org_crypto(organization_keys.clone())?;
+
+    // Persist org keys for rehydration
+    if let Ok(repo) = client
```
quexten, line 397: Currently these are "initialized" with both organization keys, but also hackily re-encrypted provider-shared-org-keys. We should not persist these to the org key state, and need to do work to split these out, and store them separately.

Two questions here @Hinton 
- Are we enabling this soon for `clients`, or just the rust-based cli? In the former case we'd need to do work first to no longer pass in these re-encrypted provider-shared-org-keys
- Do we have a framework for migrations for the SDK state? I'm a bit hesitant to commit to specific on-disk formats so easily if we don't have a migration path. I'm not sure that "OrganizationSharedKey" should be the key we index by, but instead there may be a proper per-organization struct that stores more than just the shared key. I haven't yet mapped out what needs to be stored for V2 encryption, but ideally we would not have separate on-disk keys, just different format versions?

quexten (commented):
I have two questions on this.

crates/bitwarden-core/src/key_management/crypto.rs
Hinton: 1. No plans from our side to enable this in clients. This is purely to support rehydration.
2. No official framework, for any migration ATM would require one-off solutions. Until we decide to ship an alpha release we're free to make any breaking changes though.

quexten (approved):
Okay, I don't want to be blocking where it doesn't need to be. Changes seem reasonable for now.

dereknance (approved):

crates/bitwarden-core/src/key_management/crypto.rs
dani-garcia: Agreed, there's no plans to use this on `clients` or mobile at this time. Before we decide to do that, we'll probably need to go through some process like auth is doing where we move all this code out of core into a KM owned crate so at that point we can work on a better design.

No migrations framework yet, but it's on the todo list: https://bitwarden.atlassian.net/browse/PM-32271 For now we're good to do breaking changes, but we'll definitely have some basic migration path ready for release at least.

dani-garcia (approved):

PR #970, opened 2026-04-17
Add clippy lint for holding tracing span across await points

quexten:
## 📔 Objective

https://github.com/bitwarden/sdk-internal/pull/931#discussion_r3098595725
https://docs.rs/tracing/latest/tracing/span/struct.Span.html#in-asynchronous-code
Holding spans across await points leads to incorrect traces and should be prevented by clippy.


crates/bitwarden-user-crypto-management/src/key_rotation/password_change_and_rotate_user_keys.rs
```diff
@@ -48,12 +48,17 @@ impl UserCryptoManagementClient {
     }
 }
 
+#[instrument(
+    name = "password_change_and_rotate_user_keys",
```
quexten, line 52: This span was incorrect and lead to incorrect traces. Instrument correctly emits code that is safe for use with async.

mzieniukbw (approved):

dani-garcia (approved):

PR #972, opened 2026-04-20
[poc] Client managed value state

quexten:
## 📔 Objective

Currently, the bitwarden-state crate only allows client-managed state via repositories. This works for sets of data, like ciphers, collections, folders. This does not work for single-data such as the user-key, or pin ephemeral state. This PR introduces a second type of client managed state - Value state - which just directly takes the value instead of wrapping it into a key-value store.

This de-risks and speeds up migrations for existing state.

### DIFF
--- a/crates/bitwarden-core/src/client/test_accounts.rs
+++ b/crates/bitwarden-core/src/client/test_accounts.rs
@@ -5,7 +5,7 @@ use std::collections::HashMap;
 #[cfg(feature = "test-fixtures")]
 use bitwarden_crypto::{EncString, Kdf};
 #[cfg(feature = "test-fixtures")]
-use bitwarden_test::MemoryRepository;
+use bitwarden_test::{MemoryRepository, MemoryValue};
 
 #[cfg(feature = "test-fixtures")]
 use crate::{
@@ -26,9 +26,7 @@ impl Client {
         client
             .platform()
             .state()
-            .register_client_managed(std::sync::Arc::new(
-                MemoryRepository::<UserKeyState>::default(),
-            ));
+            .register_client_managed_value(std::sync::Arc::new(MemoryValue::<UserKeyState>::default()));
         client
             .platform()
             .state()
--- a/crates/bitwarden-core/src/key_management/mod.rs
+++ b/crates/bitwarden-core/src/key_management/mod.rs
@@ -68,7 +68,7 @@ pub struct UserKeyState {
     decrypted_user_key: B64,
 }
 
-bitwarden_state::register_repository_item!(String => UserKeyState, "UserKey");
+bitwarden_state::register_value_item!(UserKeyState, "UserKey");
 
 /// Represents the local user data key, wrapped by user key.
 /// This key is used to encrypt local user data (e.g., password generator history).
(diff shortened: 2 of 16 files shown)


Hinton:
Without looking at the PR. Are there any reasons we can't use the settings interface assuming we introduce some nice interfaces on the JS side. (Settings are an abstraction on top of repositories, so you could implement a single repository with key individual mappers)

quexten:
@Hinton There is two use-cases here:
- Fresh new state
- Old state that needs interop and needs to be accessed from the SDK now

For either of these we need state that is in-memory only (but surviving SDK reload), and state that is persisted to disk.

Ideally, we would re-think state storage, and as far as I understand the settings interface you propose is best for either fresh new state, or state that is migrated to the settings pattern. However, it is not a drop-in way to expose existing client state to the SDK.

Migrations always carry risk, and increase scope. This leads to features not being fully SDK migrated immediately, which then in turn has been the cause for real production bugs. I'm trying to explore how to decrease risk and accelerate movement to the SDK here.

So it would help me if you could roughly describe what this would look like for:
- Migrating existing state
  - For in-memory state
  - For persisted state
- New state
  - For in-memory state
  - For persisted state

Hinton:
Right now all settings would have to be client owned since they are only exposed through a repository which fits nicely into your requirements since there is no way to define where state should live (memory or persisted), although it's being built out.

I think the flow would be the same be it memory or persisted since it's managed by the client and they would own the `StateDefinition`. It would also be identical for existing and new since you're always consuming client managed state.

This is just some pseudo code of how it could work, but **platform** would own this and should do the actual planning and breakdown of how it would work.

```ts
export interface SdkSettingMapper<ClientType, SdkType> {
  userKeyDefinition(): UserKeyDefinition<ClientType>;
  toSdk(value: ClientType): SdkType;
  fromSdk(value: SdkType): ClientType;
}

export class SettingValue<ClientType, SdkType> implements Setting<SdkType> {
  constructor(
    private userId: UserId,
    private stateProvider: StateProvider,
    private mapper: SdkSettingMapper<ClientType, SdkType>,
  ) {}

  async get(): Promise<SdkType | null> {
    const raw = await firstValueFrom(this.state().state$);
    return raw == null ? null : this.mapper.toSdk(raw);
  }

  async set(value: SdkType): Promise<void> {
    await this.state().update(() => this.mapper.fromSdk(value));
  }

  async remove(): Promise<void> {
    await this.state().update(() => null);
  }

  private state() {
    return this.stateProvider.getUser(this.userId, this.mapper.userKeyDefinition());
  }
}
```

Image that you have `SHOW_INLINE_MENU` which you want to consume from the SDK.

```ts
// libs/common/src/autofill/state/autofill-settings.state.ts
export const AUTOFILL_SETTINGS_DISK = new StateDefinition(
  "autofillSettingsDisk",
  "disk",
  { web: "disk-local", browser: "disk-backup-local-storage" },
);

export const SHOW_INLINE_MENU = new UserKeyDefinition<boolean>(
  AUTOFILL_SETTINGS_DISK,
  "showInlineMenu",
  { deserializer: (v) => v, clearOn: ["logout"] },
);

export class ShowInlineMenuMapper implements SdkSettingMapper<boolean, boolean> {
  userKeyDefinition() { return SHOW_INLINE_MENU; }
  toSdk(v: boolean) { return v; }
  fromSdk(v: boolean) { return v; }
}
```

Eventually as we desire that the SDK own state we would need to flip the relationship.

PR #982, opened 2026-04-22
Make get_key_id_for_symmetric_key follow client convention

quexten:
## 📔 Objective

As per sync discussion, we want the function to be implemented on the client struct instead of being static.


Thomas-Avery (approved):
LGTM

PR #983, opened 2026-04-22
[PM-31876] Migrate Flags to Setting

djsmith85:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-31876

## 📔 Objective

- Migrate Flags to Setting - 51dcc04c89a765de1ac5e73f209d70bd8a0d955f
  - Update `load_flags` and `get_flags` to read and persist to `state_registry.setting(FLAGS)`
  - Remove [expect(clippy::unused_async)] attributes from `load_flags` and `get_flags` as they are now awaited
- Add a unit test to verify saving and retrieving flags works as expected - 9b8b2a98d2c046424e161f8764a65e7a60ab0887
- Removed the Flags import and the flags: RwLock::new(Flags::default() initialization - d486eaf83b4c764cd71b74ea1415601134cedebb

### DIFF
--- a/crates/bitwarden-core/src/client/internal.rs
+++ b/crates/bitwarden-core/src/client/internal.rs
@@ -107,6 +107,8 @@ pub struct InternalClient {
     #[cfg(feature = "internal")]
     pub(crate) security_state: RwLock<Option<SecurityState>>,
 
+    // TODO: Flags have been migrated to Setting but this will have to stay temporarily until the feature flags are removed.
+    #[allow(dead_code)]
     pub(crate) state_registry: StateRegistry,
 }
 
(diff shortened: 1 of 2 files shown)


crates/bitwarden-core/src/client/internal.rs
```diff
@@ -108,26 +107,43 @@ pub struct InternalClient {
     #[cfg(feature = "internal")]
     pub(crate) security_state: RwLock<Option<SecurityState>>,
 
-    // TODO(PM-31876): Remove once Flags are migrated to Setting, which will add an in-crate
-    // read path and satisfy the dead_code lint without suppression.
-    #[allow(dead_code)]
```
dani-garcia, line 113: You'll need to add this allow back to get CI passing, it seems the comment above wasn't entirely correct and we're not using it when neither of the feature flags are enabled

dani-garcia (commented):
LGTM, just some unused code warnings that will need to be silenced again

dani-garcia (approved):

PR #984, opened 2026-04-22
Configurable log level for uniffi

mzieniukbw:
## 📔 Objective

Configurable log level for uniffi. 
Optional, defaults to info, as before.

Why the change ? The `RUST_LOG` env variable is not feasable to be set at runtime for android.

## 🚨 Breaking Changes

The `init_logger` function is not used explicitly anywhere, but indirectly when new uniffi `Client` is created. Since the function can take effect only once, on first usage, the function would need to be explicitly called before `Client` creation and set appropriate log level - unless it's added as a parameter for `Client::new` or into `ClientSettings`.

On android likely in https://github.com/bitwarden/android/blob/main/app/src/main/kotlin/com/x8bit/bitwarden/data/platform/manager/SdkClientManagerImpl.kt#L49

### DIFF
--- a/crates/bitwarden-uniffi/examples/callback_demo.rs
+++ b/crates/bitwarden-uniffi/examples/callback_demo.rs
@@ -49,7 +49,7 @@ fn main() {
 
     // Initialize logger with callback BEFORE creating any clients
     println!("Step 1: Initialize SDK logger with callback...\n");
-    bitwarden_uniffi::init_logger(Some(callback));
+    bitwarden_uniffi::init_logger(Some(callback), None);
 
     println!("Step 2: Create SDK client...\n");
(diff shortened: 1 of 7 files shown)


dani-garcia (approved):
LGTM, seems to match pretty well with what we're doing for WASM already:

https://github.com/bitwarden/sdk-internal/blob/85ee2d6ec2f51ae391ac44fb9d1d2b7203006b35/crates/bitwarden-wasm-internal/src/init.rs#L36-L40

On your comment on the breaking changes section, the expectation as you mention is that this function should be called by the mobile apps only once before the clients get initialized. This is what we do on WASM, and the `init_logger` call inside `Client::new` is for backwards compatibility support.

PR #987, opened 2026-04-23
[PM-35786] Use CipherView encryption for Create/Edit requests

shane-melton:
## 🎟️ Tracking

[PM-35786](https://bitwarden.atlassian.net/browse/PM-35786)

## 📔 Objective

Consolidate `CiphersClient::create` and `CiphersClient::edit` (plus their admin-client equivalents) onto a single encrypt pipeline by routing everything through `CipherView` and its `CompositeEncryptable` impls. This will allow us to implement blob encryption for a single flow, and allow it to be re-used for any cipher request models.

Today there are three `CompositeEncryptable` impls that all independently re-implement field-level encryption for every cipher field (`name`, `notes`, `login`, `card`, `identity`, …):

- `CipherView → Cipher`
- `CipherCreateRequestInternal → CipherRequestModel`
- `CipherEditRequestInternal → CipherRequestModel`

To reduce this repetition, introduce the following:

- `convert_request_to_cipher_view` for `CipherCreateRequest` and `CipherEditRequest` — the public request DTOs stay as validation and avoid breaking changes to the existing SDK interface.
- `TryFrom<Cipher> for CipherRequestModel` — pure structural mapping from the encrypted cipher to the API request model, no encryption.

Then rewrite each write entry point as:

```
request → view → encrypt → API request model
```

Delete `CipherCreateRequestInternal`, `CipherEditRequestInternal`, and their duplicate `CompositeEncryptable` impls.

[PM-35786]: https://bitwarden.atlassian.net/browse/PM-35786?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


crates/bitwarden-vault/src/cipher/cipher_client/admin/create.rs
```diff
@@ -87,29 +89,31 @@ impl CipherAdminClient {
...
         // TODO: Once this flag is removed, the key generation logic should
-        // be moved closer to the actual encryption logic.
-        if self
-            .client
-            .internal
-            .get_flags()
-            .await
-            .enable_cipher_key_encryption
+        // be moved directly into the CompositeEncryptable implementation.
+        if view.key.is_none()
```
nikwithak: non-blocking nit: Since this is `create`, I believe there never be a `view.key` value here - it's not something the user can specify in the request.

crates/bitwarden-vault/src/cipher/cipher_client/create.rs
```diff
@@ -269,56 +145,52 @@ async fn create_cipher<R: Repository<Cipher> + ?Sized>(
...
         // TODO: Once this flag is removed, the key generation logic should
-        // be moved closer to the actual encryption logic.
-        if self
-            .client
-            .internal
-            .get_flags()
-            .await
-            .enable_cipher_key_encryption
+        // be moved directly into the CompositeEncryptable implementation.
+        if view.key.is_none()
```
nikwithak: Here too

nikwithak (dismissed):

crates/bitwarden-vault/src/cipher/cipher_client/admin/create.rs
shane-melton: Good point, fixed in [b5e83ad](https://github.com/bitwarden/sdk-internal/pull/987/commits/b5e83ad9b26255561665d2240ce8f2530ee620d1)

nikwithak (approved):

PR #992, opened 2026-04-24
[PM-35828] Add move_many and bulk_update_collections operations to CiphersClient

nikwithak:
## 🎟️ Tracking

## 📔 Objective

This implements some of the missing Cipher Operations in the SDK. Adds functions to `CiphersClient` that orchestrates the calls to the API and updates the Repository state for:

`move_many` (change folder on multiple ciphers) and `bulk_update_collections`.

As a part of this, I also updated the `CiphersClient` to use the new pattern demonstrated by `FoldersClient` for easy testing without the standalone functions.


crates/bitwarden-vault/src/cipher/cipher_client/bulk_update_collections.rs
```diff
@@ -0,0 +1,236 @@
...
+                        cipher
+                            .collection_ids
+                            .retain(|id| !collection_ids.contains(id));
+                    } else {
+                        cipher.collection_ids = cipher
+                            .collection_ids
+                            .into_iter()
+                            .chain(collection_ids.iter().copied())
+                            .collect::<HashSet<_>>()
+                            .into_iter()
```
gbubemismith: ⚠️ I think the HashSet in rust yields elements in random order, so this will not preserve the insertion order

crates/bitwarden-vault/src/cipher/cipher_client/bulk_update_collections.rs
```diff
@@ -0,0 +1,236 @@
...
+            .ciphers_api()
+            .post_bulk_collections(Some(CipherBulkUpdateCollectionsRequestModel {
+                organization_id: Some(organization_id.into()),
+                cipher_ids: Some(cipher_ids.iter().map(|id| (*id).into()).collect()),
+                collection_ids: Some(collection_ids.iter().map(|id| (*id).into()).collect()),
+                remove_collections: Some(remove_collections),
+            }))
+            .await?;
+
+        if let Some(repository) = &self.repository {
```
gbubemismith: ❓ what happens when repository is None, I think we normally use 

``` rust
let repository = self.repository.require()?;
```

which returns an error when state it not set up

crates/bitwarden-vault/src/cipher/cipher_client/move_many.rs
```diff
@@ -0,0 +1,190 @@
...
+        self.api_configurations
+            .api_client
+            .ciphers_api()
+            .move_many(Some(CipherBulkMoveRequestModel {
+                ids: cipher_ids.iter().map(|id| id.to_string()).collect(),
+                folder_id: folder_id.map(|id| id.to_string()),
+            }))
+            .await?;
+
+        if let Some(repository) = &self.repository {
```
gbubemismith: ❓ what happens when repository is None, I think we normally use 

``` rust
let repository = self.repository.require()?;
```

which returns an error when state it not set up

crates/bitwarden-vault/src/vault_client.rs
```diff
@@ -38,9 +38,7 @@ impl VaultClient {
 
     /// Cipher related operations.
     pub fn ciphers(&self) -> CiphersClient {
-        CiphersClient {
-            client: self.client.clone(),
-        }
+        CiphersClient::from_client(&self.client)
```
gbubemismith, line 41: 👍🏼

gbubemismith (commented):
I just added some comments, changes look good generally

crates/bitwarden-vault/src/cipher/cipher_client/bulk_update_collections.rs
nikwithak: Easy fix, I'll make the change. Out of curiosity, are there known use cases where the insertion order is important to preserve?

gbubemismith: I think we use it in the collection badge, we could always sort on the client. So this is not a blocker, I'll leave the call to you

gbubemismith (approved):

PR #993, opened 2026-04-25
[PM-34151] Add PolicyClient

eliykat:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-34151

## 📔 Objective

Add the client struct for the policies crate to expose it to native callers. Make some minor adjustments to types and serialization to get it working.

I wired this up to my local web client and verified that it works.

### DIFF
--- a/crates/bitwarden-policies/src/policy_client.rs
+++ b/crates/bitwarden-policies/src/policy_client.rs
@@ -1,0 +1,153 @@
+//! [`PolicyClient`] and its associated extension trait.
+
+use bitwarden_core::Client;
+use bitwarden_organizations::ProfileOrganization;
+#[cfg(feature = "wasm")]
+use wasm_bindgen::prelude::wasm_bindgen;
+
+use crate::{
+    filter::{PolicyType, PolicyView},
+    registry::PolicyRegistry,
+};
+
+fn build_policy_registry() -> PolicyRegistry {
+    // Policy definitions will be registered here in PM-34154
+    PolicyRegistry::builder().build()
+}
+
+/// Client for policy domain operations.
+///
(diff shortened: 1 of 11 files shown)


crates/bitwarden-policies/src/policy_client.rs
```diff
@@ -0,0 +1,153 @@
...
+}
+
+impl PoliciesClientExt for Client {
+    fn policies(&self) -> PolicyClient {
+        PolicyClient::new()
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
eliykat, line 74: Not sure how much testing is needed here. The docs say [we should test](https://contributing.bitwarden.com/architecture/sdk/client-patterns#testing), but no real functionality is actually implemented here.

JaredScar (approved):

eliykat:
@claude review

crates/bitwarden-policies/src/policy_client.rs
```diff
@@ -0,0 +1,159 @@
...
+        self.registry
+            .filter_by_type(&policies, &organizations, PolicyType(policy_type))
+            .into_iter()
+            .cloned()
+            .collect()
+    }
+}
+
+/// Extension trait that adds a [`policies`](PoliciesClientExt::policies) method to [`Client`].
+#[allow(missing_docs)]
```
dani-garcia, line 62: You already have a comment here, so you probably don't need the missing_docs allow

dani-garcia (approved):
LGTM just one small comment

crates/bitwarden-policies/src/policy_client.rs
eliykat: I'll get this in the next PR!

PR #995, opened 2026-04-27
[PM-31059] Add password protected key envelope and symmetric crypto key FFI support

quexten:
## 🎟️ Tracking

## 📔 Objective

Instead of passing around untyped `Vec<u8>`'s or `String`s across the FFI boundary, we want to be able to pass typed wrappers around these. This adds them for PasswordProtectedKeyEnvelope, and SymmetricCryptoKey, as part of managing PIN settings via the SDK.

Note: Ideally, `SymmetricKey` will not be passed across the FFI boundary, but for some KM processes, this is necessary for the time being, such as setting the user-key to cilent-managed state. This is accepted, and already being done via `Vec<u8>` / `String`.


crates/bitwarden-crypto/src/keys/symmetric_crypto_key.rs
```diff
@@ -503,6 +511,23 @@ pub fn derive_symmetric_key(name: &str) -> Aes256CbcHmacKey {
     derive_shareable_key(secret, name, None)
 }
 
+#[cfg(feature = "wasm")]
```
dani-garcia: I believe what we should be doing for WASM support here is implement WasmDescribe and FromWasmAbi instead, like we do for EncStrings. I don't think just having From and Into for `JsValue` allows it to be used at the API layer.

https://github.com/bitwarden/sdk-internal/blob/7f71304be076d49f93c9447914c52f9995583886/crates/bitwarden-crypto/src/enc_string/symmetric.rs#L84-L100

quexten: The follow-up PR handled this by doing the JSValue conversion in the consumer. But you're right, this is cleaner. Switched this and the follow-up PR.

dani-garcia (approved):

PR #996, opened 2026-04-27
[PM-31059] Add key-management state bridge

quexten:
## 🎟️ Tracking

## 📔 Objective

Adds an implementation of the state bridge. The state bridge is an interface that is externally implemented by the host application (in kotlin/swift/typescript). The clients can more easily set state of non-repository-shaped data without having to write a migration.

This will be used to immediately access state for all KM work, without risky and scope-increasing migrations.


dani-garcia (dismissed):

crates/bitwarden-core/src/key_management/state_bridge/mod.rs
```diff
@@ -0,0 +1,159 @@
...
+#[cfg(feature = "wasm")]
+mod wasm;
+
+/// Host-provided storage bridge for key-management state.
+///
+/// SDK consumers register an implementation that persists or caches sensitive
+/// account state across unlock flows.
+#[cfg_attr(target_arch = "wasm32", async_trait(?Send))]
+#[cfg_attr(not(target_arch = "wasm32"), async_trait)]
+pub trait StateBridgeImpl: Send + Sync {
```
mzieniukbw: If you remove `Send + Sync` for wasm target, the `WasmStateBridge` becomes much easier, since it would no longer require `ThreadBoundRunner`.
(The `Arc<Mutex<` implementation would likely have to change too for wasm)

crates/bitwarden-core/src/key_management/state_bridge/wasm.rs
```diff
@@ -0,0 +1,186 @@
...
+use crate::key_management::state_bridge::{StateBridgeClient, StateBridgeImpl};
+
+#[wasm_bindgen::prelude::wasm_bindgen(typescript_custom_section)]
+const TS_CUSTOM_TYPES: &'static str = r#"
+/**
+ * Typescript interface that the state bridge needs to implement. The state bridge
+ * is a temporary layer that allows quickly transitioning non-repository shaped
+ * state to be accessible from within the SDK.
+ */
+export interface WasmStateBridge {
```
mzieniukbw: Is this all worth it ?
I do think maybe we should re-visit it, in similar way as we do with `register_client_managed` (instead of JS-to-SDK way like here), but simpler for KM with a `StateBridgeClient` that just wraps some of that complexity behind.

quexten: This was the suggested approach by platform. https://bitwarden.slack.com/archives/C0321H9M0KS/p1776783574345289?thread_ts=1776697321.749269&cid=C0321H9M0KS

I don't want to shift approaches again, especially considering we will eventually transition to the approach platform gives us. Yes, the trade-off here is a bit of boiler-plate code. We can reduce it by using more macros, I suppose.

crates/bitwarden-core/src/key_management/state_bridge/mod.rs
quexten: Yeah, we can remove the ThreadBound runner, but the non-wasm impl still requires the `Arc<Mutex<...>>` though.. 

I don't know if we need to introduce compile time features around this.

Cargo.toml
```diff
@@ -72,6 +72,7 @@ hmac = "0.13.0"
 http = ">=1.4.0, <2.0"
 js-sys = { version = ">=0.3.72, <0.4" }
 mockall = { version = ">=0.13.1, <0.15" }
+paste = ">=1.0, <2.0"
```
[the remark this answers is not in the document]
quexten: @claude Can you point to where this requirement is prescribed?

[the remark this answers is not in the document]
quexten: Removed paste.

crates/bitwarden-core/src/key_management/state_bridge.rs
```diff
@@ -0,0 +1,124 @@
...
+use bitwarden_crypto::{EncString, SymmetricCryptoKey, safe::PasswordProtectedKeyEnvelope};
+#[cfg(feature = "wasm")]
+use wasm_bindgen::prelude::*;
+
+use crate::Client;
+
+/// Thread-safe wrapper around the registered [`StateBridgeImpl`] instance.
+#[derive(Clone)]
+pub struct StateBridge {
+    implementation: Arc<Mutex<Option<Arc<dyn StateBridgeImpl + Send + Sync>>>>,
```
mzieniukbw: ❓ Is the outer `Arc` needed ?

crates/bitwarden-state-bridge-macro/src/lib.rs
```diff
@@ -0,0 +1,271 @@
...
+        let clear_doc = format!("Clears the `{n}` value.");
+        quote! {
+            #[doc = #set_doc]
+            pub async fn #set(&self, value: &#ty) {
+                let implementation = self
+                    .implementation
+                    .lock()
+                    .expect("Mutex is not poisoned")
+                    .as_ref()
+                    .expect("StateBridge not registered")
```
mzieniukbw, line 102: ❓ Are we ok with this panicking when not registered ?

mzieniukbw (dismissed):

crates/bitwarden-state-bridge-macro/src/lib.rs
quexten: I think so. I'd rather not pollute the return error types with also returning the error case of a dependency (state bridge) not being set up, when this should never occur during runtime, outside of development.

A feature should only use the state bridge if it either checks with:
```
pub fn is_registered(&self) -> bool {
```
or if all clients that have the feature available already implement the state bridge.

crates/bitwarden-core/src/key_management/state_bridge.rs
quexten: It was because we derived Clone, but Clone was not necessary, so Arc can also be dropped. Done.

mzieniukbw (approved):

Thomas-Avery (approved):
Approving codeowners changes

dani-garcia (approved):

PR #997, opened 2026-04-27
[PM-31059] Implement Pin lock system & Pin lock management & Pin State unlock

quexten:
## 🎟️ Tracking

## 📔 Objective

Adds the core pin lock implementation. This lives in `bitwarden-core` until unlock is moved to a separate crate entirely. Additionally, adds the PIN settings client, which just routes calls through to the core pin lock system. Finally, a new unlock method - PinState - is added.

### DIFF
--- a/crates/bitwarden-core/src/key_management/crypto.rs
+++ b/crates/bitwarden-core/src/key_management/crypto.rs
@@ -47,6 +47,7 @@ use crate::{
             get_local_user_data_key_from_state, initialize_local_user_data_key_into_state,
         },
         master_password::{MasterPasswordAuthenticationData, MasterPasswordUnlockData},
+        pin_lock_system::{PinLockSystem, UnlockError},
     },
 };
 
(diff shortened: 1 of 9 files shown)


crates/bitwarden-core/src/key_management/crypto.rs
```diff
@@ -273,6 +280,30 @@ pub(super) async fn initialize_user_crypto(
...
+            PinLockSystem::with_client(client)
+                .unlock(pin.as_str())
+                .await
+                .map_err(|err| match err {
+                    UnlockError::PinWrong => EncryptionSettingsError::WrongPin,
+                    _ => EncryptionSettingsError::CryptoInitialization,
+                })?;
+            // Note: PinLockSystem sets the user-key to state, and this section is reading it from
+            // state, then re-setting it via `initialize_user_crypto_decrypted_key`.
+            // This is not ideal and should be refactored in the future.
```
quexten, line 293: Not done here because I don't think it's reasonable to refactor unlock here.

eligrubb (approved):
LGTM

PR #1000, opened 2026-04-27
Update CX create and improve editable field parsing

Hinton:
## 📔 Objective

Updates the `credential-exchange-format` dependency to the in-progress `Expected<T>` design (git rev `482506ea`) and reworks the CXF import path so a payload whose `fieldType` disagrees with the spec's expected slot is preserved coherently instead of being silently mistyped.

### What changed in upstream

`EditableField<T>.value` is now `Expected<T>` rather than the bare `T`. Deserialization no longer fails on a type mismatch; instead it captures the actual incoming value in `UnexpectedField` and the consumer chooses how to react via `as_expected()` / `field_type()` / `From<Expected<T>> for String`.

### What this PR does in `bitwarden-exporters`

- **`editable_field.rs`** — `field_type()` is now derived from the value's *runtime* CXF type rather than from the statically-expected `T`. A payload that arrives as `boolean` in a slot statically typed as `string` now reports `FieldType::Boolean` (with the raw value preserved) instead of `Text`. The per-type `to_field_value` impls collapse to a default that delegates to upstream's `Into<String>`; only `EditableFieldWifiNetworkSecurityType` keeps a custom impl because vault display form differs from the spec wire form.
- **`import.rs`** — adds explicit arms for `EditableFieldValue::Email` and `Number` in `custom_fields_to_fields`. Previously they fell into the wildcard "Unknown Field" branch and lost their label and value.
- **`card.rs`, `identity.rs`, `login.rs`, `note.rs`, `export.rs`** — call-site updates to go through `as_expected()` / `String::from(value.clone())` instead of reaching into `.value.0`. `sanitize_brand` switched to taking `String` to clean up the `&String::from(b)` pattern at the call site.
- **Tests** — added coverage for mismatched-type fallback (`boolean` in a `string` slot, `concealed-string` in a `string` slot), unknown CXF types collapsing to `Text`, and email/number round-trip through `custom_fields_to_fields`.


crates/bitwarden-exporters/Cargo.toml
```diff
@@ -35,7 +36,7 @@ bitwarden-fido = { workspace = true }
 bitwarden-ssh = { workspace = true }
 bitwarden-vault = { workspace = true }
 chrono = { workspace = true, features = ["std"] }
-credential-exchange-format = { version = ">=0.3, <0.4" }
+credential-exchange-format = { git = "https://github.com/bitwarden/credential-exchange", rev = "482506ea422c720246dbe481dde271137f42b3ac" }
```
Hinton: Todo: release a new version.

harr1424:
@claude please review

harr1424 (dismissed):

harr1424 (approved):

djsmith85 (approved):

PR #1001, opened 2026-04-27
Add warning message for breaking changes

trmartin4:
## 📔 Objective

Add more stern warning message to breaking change handling on SDK PRs.


addisonbeck (approved):

trmartin4:
@addisonbeck thinking this through, I don't actually think we want this assertion at the repo level, do we?  We can't expect that there's a PR ready for the `android` repo when this change merges; that's not how we work.

This feels like we may need a more nuanced approach that's specific to how `clients` expectations are different from `android` and `ios` 🤔 .

PR #1003, opened 2026-04-27
[PM-34154] Add policy overrides to match clients

eliykat:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-34154

## 📔 Objective

Now that we have basic policy filtering in the SDK, implement `Policy` traits for each policy type that requires custom filtering rules. This matches [isExemptFromPolicy](https://github.com/bitwarden/clients/blob/main/libs/common/src/admin-console/services/policy/default-policy.service.ts#L272) in `clients`, except for:

- single org, which is being refactored in https://bitwarden.atlassian.net/browse/PM-34159 so the override is not required
- the Master Password policy, which includes the changes being added in https://github.com/bitwarden/clients/pull/20398

### DIFF
--- a/crates/bitwarden-policies/src/policy_overrides.rs
+++ b/crates/bitwarden-policies/src/policy_overrides.rs
@@ -1,0 +1,256 @@
+//! Custom policy implementations that override the default rules.
+
+use bitwarden_organizations::OrganizationUserType;
+
+use crate::filter::{Policy, PolicyType};
+
+/// Master Password policy (type 1).
+///
+/// Applies to **everyone**, including Owners and Admins.
+pub struct MasterPasswordPolicy;
+
+impl Policy for MasterPasswordPolicy {
+    fn policy_type(&self) -> PolicyType {
+        PolicyType(1)
+    }
+
+    fn exempt_roles(&self) -> &[OrganizationUserType] {
+        &[]
+    }
+}
+
+/// Password Generator policy (type 2).
+///
+/// Applies to **everyone**, including Owners and Admins.
+pub struct PasswordGeneratorPolicy;
+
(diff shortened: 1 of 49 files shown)


eliykat:
Rebasing accidentally blasted some unrelated teams with review requests, sorry about that. Please ignore if your review is not currently requested.

eliykat:
@claude review

crates/bitwarden-policies/src/policy_overrides.rs
```diff
@@ -0,0 +1,256 @@
+//! Custom policy implementations that override the default rules.
+
+use bitwarden_organizations::OrganizationUserType;
+
+use crate::filter::{Policy, PolicyType};
+
+/// Master Password policy (type 1).
```
dani-garcia, line 7: Idea for the future, as we add more overrides here, we could create a small macro to reduce the duplication:

```rust
/// Password Generator policy (type 2).
///
/// Applies to **everyone**, including Owners and Admins.
define_policy!(PasswordGeneratorPolicy, 2, []);

/// Maximum Vault Timeout policy (type 9).
///
/// Applies to everyone **except Owners**. Admins are not exempt.
define_policy!(MaximumVaultTimeoutPolicy, 9, [OrganizationUserType::Owner]);
```

dani-garcia (dismissed):

crates/bitwarden-policies/src/policy_overrides.rs
eliykat: Interesting! I haven't read up on macros yet, I'll make sure to take a look and keep this in mind for the future.

dani-garcia (approved):

jrmccannon (approved):

PR #1007, opened 2026-04-28
[PM-36021] Remove hyphenated words from passphrase generator input/output

harr1424:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-36021

## 📔 Objective

This PR removes hyphenated words from the slice of words used to generate passphrases. This accomplishes the following:
- avoid obscure test failures 
- avoid passphrase output that appears to violate requested word count

As an example, the word list formerly contained the word "drop-down" which resulted in the following test failure:

```
failures:

---- test_generate_passphrase_default_words stdout ----

thread 'test_generate_passphrase_default_words' (26459) panicked at crates/bw/tests/generate.rs:208:5:
assertion `left == right` failed: Default passphrase should have 6 words, got: safeness-drop-down-proud-endpoint-siesta-bobble
  left: 7
 right: 6
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

⚠️  4 words out of 7776 in the wordlist will effectively be filtered from the input of words used to generate a random passphrase, slightly reducing the amount of entropy present in this process.


quexten (changes requested):
I don't think we should base changes to cryptographic entropy on the fact that test failures occur. However, I would argue that felt, tip, are two separate words juts as well as felt-tip is one, so in the context of `felt-tip-drool`, it is unclear whether the user has chosen three words or two. In the context of our application use of joining with `-`, these words create inconsistent output generation (as discovered by the test).

If we were to do this, we must add a proper comment. We no longer use the EFF wordlist. We should rename the const to `EFF_WORDLIST_MODIFIED`.

However, this currently introduces a breaking change that makes cryptographic fingerprints incompatible with the old set of fingerprints. The unit tests also confirm this.

https://github.com/bitwarden/sdk-internal/blob/309618d0a5429311cd379445086ceb46ccf6d7da/crates/bitwarden-crypto/src/fingerprint.rs#L33


Because of this, my suggestion is to actually introduce a tools-owned function to filter the EFF wordlist to exclude these specific words, and to use that in your code.

crates/bitwarden-generators/src/passphrase.rs
```diff
@@ -102,9 +102,14 @@ fn passphrase_with_rng(mut rng: impl Rng, options: ValidPassphraseGeneratorOptio
 }
 
 fn gen_words(mut rng: impl Rng, num_words: u8) -> Vec<String> {
+    let words: Vec<_> = EFF_LONG_WORD_LIST
```
quexten, line 108:
```suggestion
    // A passphrase consists of words joined by hyphens. The `EFF_LONG_WORD_LIST` contains some words that contain hyphens, which creates ambiguous encodings with more hyphen-separated-segments than expected.
    let words: Vec<_> = EFF_LONG_WORD_LIST
```

mcamirault (dismissed):

crates/bitwarden-generators/src/passphrase.rs
```diff
@@ -134,10 +142,10 @@ mod tests {
         let mut rng = rand_chacha::ChaCha8Rng::from_seed([0u8; 32]);
         assert_eq!(
             &gen_words(&mut rng, 4),
-            &["crust", "subsystem", "undertook", "protector"]
+            &["crust", "substance", "undertook", "protector"]
```
harr1424, line 145: The changes to `gen_words()` effectively reduce the size of the wordlist used, but only local to that function call. The tests in this file have been updated so that word removals have been accounted for.

mcamirault (approved):

quexten (approved):

PR #1012, opened 2026-04-29
reverse recent changes that broke the workflow

AmyLGalles:
## 🎟️ Tracking

the API-commit action requires improvement before the last change can be implemented, for now it's blocking publishing of the SDK.


Hinton (approved):

AmyLGalles:
This revert won't work because the GPG key has since been deleted.  I'm working on another solution to get this working again

PR #1013, opened 2026-04-29
Remove blank line in PR comment that breaks breaking change comment

trmartin4:
## Tracking

[bitwarden.atlassian.net/browse/PM-27825](https://bitwarden.atlassian.net/browse/PM-27825)

## 📔 Objective

Fixes broken breaking change detection comments, which were added in https://github.com/bitwarden/sdk-internal/pull/892.

See this comment on this PR for an example of the broken behavior: https://github.com/bitwarden/sdk-internal/pull/1013#issuecomment-4347222956.

The  HTML comments on separate lines inside the markdown table caused GitHub to terminate the table early, rendering data rows as plain text. Row identification now matches on the client label in the first table cell instead.

### Validation                                                      
   
Tested the `awk` logic locally against simulated comment bodies:     
                                                            
  **Test 1: Update an existing row while leaving others intact**     
  ```bash                                                   
  BODY='| Client | Status | Details |                                
  |--------|--------|---------|                             
  | typescript | ⏳ In progress | link |         
  | android | ✅ No breaking changes | link |
  ---'                                                               
   
  echo "$BODY" | awk -v label="typescript" -v newrow="| typescript | 
  ❌ Breaking changes | new link |" '                       
    /^\|/ && index($0, "| " label " |") { print newrow; next }
    { print }                                                        
  '
```  
Result: typescript row replaced, android row untouched.            
                                                            
 **Test 2: Append a new client row before the footer**
```bash
  BODY='| Client | Status | Details |                                
  |--------|--------|---------|      
  | typescript | ⏳ In progress | link |                             
  ---'                                                      
                                                 
  echo "$BODY" | awk -v newrow="| android | ⏳ In progress | link |"
  '                                                                  
    /^---$/ && !inserted { print newrow; inserted=1 }
    { print }                                                        
  '
```                                                         
  Result: android row appended before ---, table structure preserved.
                                                                     
**Test 3: Verify header row is not a collision risk**                  
 ```bash
 BODY='| Client | Status | Details |                                
  |--------|--------|---------|                                      
  | typescript | ⏳ In progress | link |                             
  ---'                                                      
                                                 
  echo "$BODY" | awk -v label="Client" -v newrow="| Client | OOPS | 
  replaced |" '                                                      
    /^\|/ && index($0, "| " label " |") { print newrow; next }
    { print }                                                        
  '           
```                                              
 Result: header row matches as expected, but this is a non-issue since the workflow only passes typescript or android as the label.

### DIFF
--- a/.github/workflows/detect-breaking-changes.yml
+++ b/.github/workflows/detect-breaking-changes.yml
@@ -195,23 +195,21 @@ jobs:
             echo "comment_id=$EXISTING_ID" >> $GITHUB_OUTPUT
           else
             echo "Creating new comment"
-            NEW_COMMENT=$(cat << EOF
-          ${COMMENT_MARKER}
-          ## 🔍 SDK Breaking Change Detection
-
-          **SDK Version:** \`${SDK_VERSION}\`
-
-          > ⚠️ **If breaking changes are detected, you MUST have a corresponding pull request addressing the breaking changes ready for merge in the affected client repository.**
-
-          | Client | Status | Details |
-          |--------|--------|---------|
-          ${ROW_MARKER}
-          ${OUR_ROW}
-
-          ---
-          *Results update as workflows complete.*
-          EOF
-            )
+            NEW_COMMENT="$(printf '%s\n' \
+              "${COMMENT_MARKER}" \
+              "## 🔍 SDK Breaking Change Detection" \
+              "" \
+              "**SDK Version:** \`${SDK_VERSION}\`" \
+              "" \
+              "> ⚠️ **If breaking changes are detected, you MUST have a corresponding pull request addressing the breaking changes ready for merge in the affected client repository.**" \
+              "" \
+              "| Client | Status | Details |" \
+              "|--------|--------|---------|" \
+              "${ROW_MARKER}" \
+              "${OUR_ROW}" \
+              "---" \
+              "*Results update as workflows complete.*" \
+            )"
 
             COMMENT_ID=$(gh api --method POST "repos/$GITHUB_REPOSITORY/issues/$PR_NUMBER/comments" \
               --field body="$NEW_COMMENT" --jq '.id')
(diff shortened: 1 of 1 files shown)


vvolkgang:
> [...] and causing  HTML comments to render as visible text.

That's not happening. That row comment is also breaking the table rendering, it's happening in this [PR](https://github.com/bitwarden/sdk-internal/pull/1013#issuecomment-4347222956). 


image
image


Removing the row comments:

image

trmartin4:
> > [...] and causing  HTML comments to render as visible text.
> 
> That's not happening. That row comment is also breaking the table rendering, it's happening in this [PR](https://github.com/bitwarden/sdk-internal/pull/1013#issuecomment-4347222956).
> 
> image image
> Removing the row comments:
> 
> image

@vvolkgang I'm sorry, I missed that completely.  I think it's been fixed now.

theMickster:
Hi @trmartin4  👋🏼

When we have identified breaking change via this workflow, what are your thoughts about adding a clarifying statement/caveat to the messaging that states to all who read the messaging that the source of the breaking change may not in-fact be the PR at-hand?  I ask this because of my confusion in #1027 lead me to an incorrect assumption that the PR had problem, when in-fact, some other PR was the cause. 

Also, I am seeing an opportunity to reduce confusion via the label's description to state something similar 🤔

image

theMickster (commented):
In a future PR and work item, please refactor the sticky comment logic that exists in this workflow to, instead, use the composite action found here -> [gh-actions repo + update-pr-comment](https://github.com/bitwarden/gh-actions/tree/main/update-pr-comment).  We wrote that especially for the need to maintain sticky comments. 

I think that changed will reduce the complexity in this workflow and let this workflow focus on what it does best, locate breaking changes 🤓

trmartin4:
@theMickster addressed the request to clarify the meaning of breaking change detection with [`1c43b8f` (this PR)](https://github.com/bitwarden/sdk-internal/pull/1013/commits/1c43b8ffff8fff52eb1235733fe7364ce6e6a6dd).

theMickster (approved):
Thank you! Approving from my perspective.

addisonbeck (approved):

PR #1014, opened 2026-04-29
[PM-28346] SDK delete attachment operations

gbubemismith:
## 🎟️ Tracking
https://bitwarden.atlassian.net/browse/PM-28346

## 📔 Objective
Change the admin `delete_attachment` to return a response instead of `()`

### DIFF
--- a/crates/bitwarden-vault/src/cipher/cipher_client/admin/delete_attachment.rs
+++ b/crates/bitwarden-vault/src/cipher/cipher_client/admin/delete_attachment.rs
@@ -1,10 +1,13 @@
-use bitwarden_core::ApiError;
+use bitwarden_core::{ApiError, MissingFieldError};
 use bitwarden_error::bitwarden_error;
 use thiserror::Error;
 #[cfg(feature = "wasm")]
 use wasm_bindgen::prelude::wasm_bindgen;
 
-use crate::{CipherId, cipher_client::admin::CipherAdminClient};
+use crate::{
+    Cipher, CipherId, VaultParseError, cipher::cipher::PartialCipher,
+    cipher_client::admin::CipherAdminClient,
+};
 
 #[allow(missing_docs)]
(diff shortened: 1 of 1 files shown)


crates/bitwarden-vault/src/cipher/cipher_client/admin/delete_attachment.rs
```diff
@@ -33,39 +52,54 @@ impl CipherAdminClient {
         &self,
         cipher_id: CipherId,
         attachment_id: String,
-    ) -> Result<(), DeleteAttachmentAdminError> {
-        Ok(delete_attachment(
+    ) -> Result<Cipher, DeleteAttachmentAdminError> {
```
nikwithak, line 55: Should we return the `Cipher` or decrypt it and return the `CipherView`?

I think long-term we want the `CipherView` and the SDK is an interface into the encrypted state to abstract the decrypt/encrypt concerns away from the clients entirely.

That said, we do have a couple of operations that return the encrypted Cipher right now though - thoughts on decrypting here and returning the `CipherView` for the caller?

gbubemismith: I agree long term. For this specific operation though, neither callers nor the SDK itself really need the decrypted view, the one place we consume the returned cipher today is the upgrade old attachments flow. Once all attachments have been upgraded, that consumer goes away and we wouldn't need to return an encrypted or decrypted cipher at all from this method

nikwithak (approved):

crates/bitwarden-vault/src/cipher/cipher_client/admin/delete_attachment.rs
nikwithak: Makes sense - I'm sure we'll have some iterations on several of these SDK operations anyhow.

PR #1015, opened 2026-04-29
[PM-35828] Add CipherAdminClient::list_assigned_org_ciphers

nikwithak:
## 🎟️ Tracking

## 📔 Objective

In tandem with #992 , adds the missing API operations to the SDK. This function call will replace the logic in `getManyFromApiForOrganization()` in the TS clients


crates/bitwarden-vault/src/cipher/cipher_client/admin/get.rs
```diff
@@ -12,6 +14,48 @@ use crate::{
...
+        Self::Api(value.into())
+    }
+}
+
+/// Fetches and decrypts all assigned ciphers for an organization.
+pub async fn list_assigned_org_ciphers(
+    org_id: OrganizationId,
+    api_client: &bitwarden_api_api::apis::ApiClient,
+    key_store: &KeyStore<KeySlotIds>,
+) -> Result<ListOrganizationCiphersResult, GetAssignedOrgCiphersAdminError> {
```
nikwithak: Re-uses the result type from the `list_org_ciphers`, which returns both the CipherListView AND the raw `Ciphers` for applications that need more than the CipherListView provides.

jengstrom-bw (approved):
Looks good!

PR #1016, opened 2026-04-30
[PM-35204] Create GlobalClient

dani-garcia:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-35204

## 📔 Objective

Create a `GlobalClient` that will be the foundation for unauthenticated operations, like receiving sends. The global client will make use of the systemwide `HostPlatformInfo`, rather than accepting it as a parameter.


crates/bitwarden-core/src/global/global_client.rs
```diff
@@ -0,0 +1,78 @@
...
+
+/// The entry point for unauthenticated SDK operations.
+///
+/// `GlobalClient` internally contains an [`Arc`] so is cheap to clone.
+///
+/// Headers are sourced from the process-wide
+/// [`crate::HostPlatformInfo`] singleton, which must be initialized via
+/// [`crate::init_host_platform_info`] before constructing a `GlobalClient`.
+#[derive(Clone)]
+pub struct GlobalClient {
```
quexten, line 16: @dani-garcia A question, should registration operations for unregistered flows (that is, password registration specifically, since TDE, KC, MP-JIT are authenticated) be moved to this client?


(In general we have quite a few different client concepts now, which are difficult to deal with. On KM we already have to write helper functions to dynamically get you a register client / user client. We should probably sync up on this sometime..)

coroiu (approved):
Nice

crates/bitwarden-core/src/global/global_client.rs
dani-garcia: @quexten Yeah, I think the idea is that we'll slowly pull all the unauthenticated functionality from extending a `PasswordManagerClient` or using their own ad-hoc clients and instead be moved under this. So we'd have some `GlobalClient.login()`, `GlobalClient.register()`, `GlobalClient.sends()` for receiving sends, etc. That should give a more clear idea of the entrypoints on the SDK, and knowing which ones are authenticated and which ones aren't.

This is being created to allow tools to handle receiving sends from the CLI, and I'm going to be working with auth into looking how we can move login over and integrate it with rehydration. My hope is we can expand the use of the `FromClient` macro to this new global client as well, so for cases like registration where you may have both authenticated and unauthenticated cases, we can automatically derive those helper functions for you.

That said that is all still in flight so if you have any pain points figured out here we can try to work on them.

quexten: Thank you!

PR #1020, opened 2026-05-01
Fix swapped private and public keys

quexten:
## 🎟️ Tracking

https://bitwarden.slack.com/archives/C054ZQSBS49/p1776869829733399

## 📔 Objective

I'm not sure how this happened since we regression tested the TDE flows on clients, however the private and public key in the request model conversions were swapped, leading to a broken state.

This swaps them back.

### DIFF
--- a/crates/bitwarden-auth/src/registration.rs
+++ b/crates/bitwarden-auth/src/registration.rs
@@ -546,7 +546,7 @@ mod tests {
                     let body = body.unwrap();
                     assert!(matches!(
                         EncString::from_str(body.encrypted_private_key.as_str()).unwrap(),
-                        EncString::Aes256Cbc_B64 { .. }
+                        EncString::Aes256Cbc_HmacSha256_B64 { .. }
                     ));
                     assert!(matches!(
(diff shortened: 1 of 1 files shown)


crates/bitwarden-auth/src/registration.rs
```diff
@@ -248,20 +248,20 @@ async fn internal_post_keys_for_tde_registration(
...
                 request.device_identifier.as_str(),
-                Some(DeviceKeysRequestModel::new(
-                    tde_registration_crypto_result
+                Some(DeviceKeysRequestModel {
+                    encrypted_user_key: tde_registration_crypto_result
                         .trusted_device_keys
                         .protected_user_key
                         .to_string(),
-                    tde_registration_crypto_result
+                    encrypted_public_key: tde_registration_crypto_result
```
quexten, line 209: Switched to a more explicit syntax, since the unnamed parameters to the `new` function is what lead to the bug in the first place.

ike-kottlowski (approved):
Good catch!

PR #1021, opened 2026-05-01
[PM-34165] Introduce InviteKeyBundle for supporting future AC auto-confirm workflow

eligrubb:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-34165

## 📔 Objective

This PR introduces the `InviteKeyBundle` struct in the newly created `bitwarden-organization-crypto` crate. `InviteKeyBundle` generates a new invite key, wraps it using the provided organization key, then returns. Callers can access either the wrapped key or the raw key data (serialized in `B64Url` format).

Additionally, the `bitwarden-organization-crypto/examples` directory was added to demonstrate how to properly use the new interface.


crates/bitwarden-organization-crypto/examples/make_invite_key_bundle.rs
```diff
@@ -0,0 +1,68 @@
...
+    let bytes = bundle.raw_invite_key();
+    let organization_wrapped_invitation_key = bundle.organization_wrapped_invite_key();
+
+    let unwrapped_key_id = ctx
+        .unwrap_symmetric_key(organization_key, organization_wrapped_invitation_key)
+        .expect("unwrapping should work");
+
+    // Testing purposes only
+    // DO NOT REPLICATE
+    #[allow(deprecated)]
```
dani-garcia: Non-blocking suggestion: Rather than adding a comment explaining why we allow the deprecated warning, we can use the `allow` `reason`  field for that:
```suggestion
    #[allow(deprecated, reason = "This is done for testing purposes only, DO NOT REPLICATE")]
```

dani-garcia (dismissed):

mzieniukbw (commented):
Are WASM bindings and function to create the `InviteKeyBundle` to be created in separate PR ?

r-tome:
> Are WASM bindings and function to create the InviteKeyBundle to be created in separate PR ?

We're expecting the KM team to handle that part as well, since we need the bindings before we can pick up some of the AC client tasks. Happy to sync up if it'd help!

crates/bitwarden-organization-crypto/examples/make_invite_key_bundle.rs
eligrubb: Very cool, done!

eligrubb:
Thank you @Thomas-Avery, @r-tome, the rest Admin Console team for syncing up on this topic. 

The plan is: this PR and crate will handle generating the invite crypto bundle structure. A separate, Admin-Console-owned, sdk client feature crate will call this crate to generate the bundle and be responsible for exposing WASM bindings to clients. AC is planning on writing the sdk client feature crate, KM will participate in the review and is available to answer any questions that come up.

theMickster:
Howdy @eligrubb 👋🏼
Will you please add a .claude\CLAUDE.md? I think a great note to start with is the last sentence in the README.md. 

Many Thanks!!

crates/bitwarden-organization-crypto/src/invite_key_envelope.rs
```diff
@@ -0,0 +1,168 @@
...
+
+        Ok(Self {
+            raw_key_data,
+            wrapped_invite_key,
+        })
+    }
+
+    /// Get the raw invite key bytes in B64Url encoding
+    /// CRITICAL: this data MUST NOT be sent to the server
+    pub fn raw_invite_key(&self) -> B64Url {
```
quexten: Question: Why not just return SymmetricCryptoKey here? It's supported across FFI boundaries, and has more type safety.

We should be concerned since this is a *public* API surface we are offering, and that should generally not work on keys in B64 encoding.

crates/bitwarden-organization-crypto/examples/make_invite_key_bundle.rs
```diff
@@ -0,0 +1,69 @@
...
+
+    let unwrapped_key_id = ctx
+        .unwrap_symmetric_key(organization_key, organization_wrapped_invitation_key)
+        .expect("unwrapping should work");
+
+    #[allow(
+        deprecated,
+        reason = "DO NOT REPLICATE, this is done for testing purposes only"
+    )]
+    let decrypted_bytes = B64Url::from(
```
quexten: Thought: We should add a function `assert_symmetric_keys_equal()` to the key context. That would make this a lot easier.

crates/bitwarden-organization-crypto/src/invite_key_bundle.rs
```diff
@@ -0,0 +1,168 @@
...
+    fn from(key: &InviteKeyData) -> Self {
+        B64Url::from(key.0.to_encoded().as_ref())
+    }
+}
+
+/// A struct for holding the invitation key and the invitation key wrapped by
+/// the organization key
+pub struct InviteKeyBundle {
+    raw_key_data: InviteKeyData,
+    wrapped_invite_key: EncString,
```
quexten: ~~Please use https://github.com/bitwarden/sdk-internal/blob/main/crates/bitwarden-crypto/src/safe/symmetric_key_envelope.rs instead of encstring. Since the Invite key is a cose key, we can use the key envelope, which gives us richer metadata and safety.~~

Nevermind, it's wrapped with the org key which is not cose. I'll leave the review up for anyone else who gets confused.

quexten (dismissed):

theMickster:
Howdy @eligrubb et. al 👋🏼

Many Thanks! 

---

<details>
<summary>Opus Code Review: [PM-34165] Introduce InviteKeyBundle for supporting future AC auto-confirm workflow (#1021)</summary>

## Summary

| Severity     | Count |
| ------------ | ----- |
| 🛑 Blocker   | 0     |
| ⚠️ Important | 5     |
| ♻️ Refactor  | 1     |

The new crate establishes organization-specific crypto APIs but introduces two interconnected design
problems with security implications. First, it clones raw key material out of the key store using a
`#[deprecated]` API that both the deprecation notice and `CLAUDE.md` prohibit calling from outside
`bitwarden-crypto`. Second, it uses the legacy `EncString` wrapper (no namespace tag) instead of the
existing `safe::SymmetricKeyEnvelope`, bypassing the cryptographic separation that prevents
key-confusion attacks. Findings arch-1/sec-3 and arch-2/sec-1 target the same two root causes from
complementary angles; fixing the root causes resolves all five Important findings at once.

## Findings

### ⚠️ Important

#### Raw key material cloned out of key store using deprecated API in production code

`crates/bitwarden-organization-crypto/src/invite_key_envelope.rs:45` **Caught by:** Security & logic
agent

`InviteKeyBundle::make()` calls `ctx.dangerous_get_symmetric_key(key_id)?.clone()` and stores the
resulting `SymmetricCryptoKey` in `InviteKeyData(SymmetricCryptoKey)`. Two project rules are
violated simultaneously:

1. `dangerous_get_symmetric_key` carries
   `#[deprecated(note = "This function should ideally never be used outside this crate")]` — this is
   non-test production code in an external crate.
2. The `bitwarden-crypto` `CLAUDE.md` is explicit: **"Do not expose key material from the crypto
   crate, use key references in the key store instead."**

The key store exists precisely to hold all plaintext key material in a single, audited location with
controlled lifetime and zeroization. Cloning the `SymmetricCryptoKey` out creates a second copy that
lives outside the store's lifecycle — any future memory dump, refactor, or copy of `InviteKeyBundle`
now exposes the key twice. Implicates **EK** ("encryption keys must never be exported") and **P05**
(Controlled Access).

_(Note: shares the same root code path as arch-2; fixing either resolves both.)_

**Suggested fix:** Remove `InviteKeyData(SymmetricCryptoKey)` and keep only the key slot ID
reference inside `InviteKeyBundle`. Pass the `KeyId` to callers who need to use the key via
high-level encryption helpers, rather than handing out raw bytes.

```rust
// Instead of storing the raw key:
pub struct InviteKeyBundle {
    invite_key_id: Ids::Symmetric,   // keep as a key reference
    wrapped_invite_key: EncString,   // or SymmetricKeyEnvelope — see sec-3
}
```

  </details>

---

#### `raw_invite_key()` exports plaintext key bytes with documentation-only enforcement

`crates/bitwarden-organization-crypto/src/invite_key_envelope.rs:62` **Caught by:** Security & logic
agent

`pub fn raw_invite_key(&self) -> B64Url` returns the base64url-encoded raw bytes of the underlying
`SymmetricCryptoKey`. The only guard is a doc comment:
`/// CRITICAL: this data MUST NOT be sent to the server`.

Once the bytes are in a `B64Url`, they can be:

- Logged (via `Display`/`Debug`)
- Serialized into a JSON response to the server
- Copied as a string across the WASM boundary
- Stored in an unencrypted field

There is no compile-time or runtime mechanism preventing a consumer from sending these bytes to the
Bitwarden API. This directly implicates **P01** (servers are zero-knowledge — "no possibility for an
attacker or Bitwarden employee to access unencrypted vault data") and **EK** ("must never be
exported").

**Suggested fix:** If the invite key genuinely must reach a recipient out-of-band, model it as a
"Data Exporting" operation per the security vocabulary — use an explicit, visibly-unsafe API name
(e.g. `dangerous_export_raw_invite_key_bytes`) with documentation of the threat model, or transport
the key wrapped under a recipient-specific public key so raw material never leaves the secure
environment unprotected.

  </details>

---

#### Wrapping uses legacy `EncString` instead of `safe::SymmetricKeyEnvelope`, losing namespace separation

`crates/bitwarden-organization-crypto/src/invite_key_envelope.rs:51` **Caught by:** Security & logic
agent

`ctx.wrap_symmetric_key(organization_key, key_id)` produces a bare `EncString` with no object-type
or content namespace tag. The `bitwarden-crypto` safe module README is explicit:

> "When items are signed / encrypted under the same key, these may be swapped. If the context or
> type of these objects is different, the consuming code may misinterpret these, leading to security
> vulnerabilities. **A simpler approach is strong cryptographic namespace separation, which prevents
> this** by adding metadata about where an object appropriately can be used."

`safe::SymmetricKeyEnvelope` already exists for exactly this case (sealing one symmetric key with
another) and tags the envelope with `SafeObjectNamespace::SymmetricKeyEnvelope` plus a
caller-supplied `SymmetricKeyEnvelopeNamespace` discriminant. Without a namespace, an `EncString`
invite-key-wrapped-by-org-key is bit-identical to any other symmetric key wrapped under the same org
key — a malicious or compromised server could substitute one for another and the client cannot
detect the swap. Implicates **EK** and **SC** (Secure Channel / authentication of key material).

_(Note: shares the same root cause as arch-1; both concern the missing namespace.)_

**Suggested fix:**

1. Add `OrganizationInviteKey` to `SymmetricKeyEnvelopeNamespace` in `bitwarden-crypto`.
2. Use
   `SymmetricKeyEnvelope::seal(invite_key_id, org_key_id, SymmetricKeyEnvelopeNamespace::OrganizationInviteKey, &ctx)`.
3. Return the serialized `SymmetricKeyEnvelope` instead of `EncString`.

  </details>

---

#### `InviteKeyBundle::make` bypasses `bitwarden_crypto::safe` module, reinventing key wrapping without namespace protection

`crates/bitwarden-organization-crypto/src/invite_key_envelope.rs:44` **Caught by:** Architecture
agent

The repo's `CLAUDE.md` (Cryptography section) is unambiguous: **"Use `bitwarden_crypto::safe` module
first (password-protected key envelope, data envelope) instead of more low-level primitives."**

`InviteKeyBundle::make` bypasses the safe module entirely. `safe::SymmetricKeyEnvelope` was designed
for exactly the operation this code performs — sealing a symmetric key under another symmetric key.
The safe module enforces:

- A two-layer namespace: object-level (`SafeObjectNamespace::SymmetricKeyEnvelope`) and
  content-level (`SymmetricKeyEnvelopeNamespace::*`).
- COSE_Encrypt0 format, making the object type and namespace part of the authenticated ciphertext.
- Cryptographic modularity (the safe README: "can we switch to new primitives / algorithms
  easily?").

The new crate's description ("surfaces APIs for safely manipulating organization-specific
cryptography") implies adherence to this pattern, not divergence from it.

_(This is the same root cause as sec-3, viewed as an architectural rule violation rather than a
security property.)_

**Suggested fix:** Replace `wrap_symmetric_key` / `EncString` with `SymmetricKeyEnvelope::seal` /
`unseal` and define a new content-namespace variant for invite keys. This aligns with existing
conventions, adds namespace protection at no extra cost, and removes the bespoke
`InviteKeyEnvelopeError` type in favor of `SymmetricKeyEnvelopeError`.

  </details>

---

#### `raw_key_data: InviteKeyData` stores exported key material, violating key store ownership rule

`crates/bitwarden-organization-crypto/src/invite_key_envelope.rs:59` **Caught by:** Architecture
agent

The `InviteKeyBundle` struct holds `raw_key_data: InviteKeyData(SymmetricCryptoKey)` as a field.
This is the result of calling the `#[deprecated]` `dangerous_get_symmetric_key` and cloning the key.
The `CLAUDE.md` Cryptography rule: "Do not expose key material from the crypto crate, use key
references in the key store instead."

Architectural concern beyond the security angle: storing a `SymmetricCryptoKey` in a feature crate's
struct creates an ownership and lifetime pattern that tensions the key store's invariant. If
`InviteKeyBundle` is moved, or if `InviteKeyData` is ever made `Clone` by a future author (it
already derives `Clone`), the key material travels with it, outside the store's audit surface.

**Suggested fix:** Remove `InviteKeyData` entirely and keep only the sealed envelope. If the raw
bytes must be accessible to the caller for transmission to a recipient, redesign as an explicit
export operation gated on a clear interface (see sec-2). Do not store the raw `SymmetricCryptoKey`
in a struct that lives in a feature crate.

  </details>

---

### ♻️ Refactor

#### Module, error, and type names diverge — `InviteKeyBundle` inside `invite_key_envelope.rs` with `InviteKeyEnvelopeError`

`crates/bitwarden-organization-crypto/src/invite_key_envelope.rs:1` **Caught by:** Architecture
agent

The module is named `invite_key_envelope`, the error type is `InviteKeyEnvelopeError` (doc: "Errors
that can occur when creating an invite key envelope"), but the only exported type is
`InviteKeyBundle`. There is no `InviteKeyEnvelope`.

In this codebase, **"envelope" is a term of art** with a specific, documented meaning. The safe
module contains `SymmetricKeyEnvelope`, `PasswordProtectedKeyEnvelope`, and `DataEnvelope` — all
COSE-encrypted, namespaced containers with a defined serialization contract. A contributor searching
for "envelope" types will land here and incorrectly expect that contract.

**Concrete evidence:** All `*Envelope` types in this codebase (3 of them in
`bitwarden-crypto/src/safe/`) are COSE-encrypted, namespaced containers. This type is none of those
things.

**Suggested fix:** Rename `invite_key_envelope.rs` → `invite_key_bundle.rs` and
`InviteKeyEnvelopeError` → `InviteKeyBundleError`. Update `lib.rs` re-exports accordingly. (If the
⚠️ Important findings are addressed by adopting `safe::SymmetricKeyEnvelope`, the naming will become
partially accurate again — but the module-level name should still match the primary type it houses.)

  </details>

</details>


---

<details>
<summary>Sonnet Code Review: [PM-34165] Introduce InviteKeyBundle for supporting future AC auto-confirm workflow (#1021)</summary>

## Summary

| Severity     | Count |
| ------------ | ----- |
| 🛑 Blocker   | 0     |
| ⚠️ Important | 0     |
| ♻️ Refactor  | 4     |

This PR introduces a well-structured new crate with a clear purpose and appropriate crate-level
design (no `bitwarden-core` dependency, correct CODEOWNERS placement, Foundation-Layer positioning).
Four refactor-level findings remain: three relate to the intentional use of
`dangerous_get_symmetric_key` and raw key materialization — patterns present elsewhere in the SDK
but flagged here for documented convention compliance — and one is a Cargo dependency placement fix.
No blockers or important issues were found.

## Findings

### ♻️ Refactor

#### `uuid` declared as runtime dependency but only used in examples

`crates/bitwarden-organization-crypto/Cargo.toml:21` **Caught by:** Architecture agent

`uuid = { workspace = true }` appears under `[dependencies]`, but the production source file
`src/invite_key_envelope.rs` makes no reference to `uuid` — it only appears in
`examples/make_invite_key_bundle.rs`. In Cargo, `[dependencies]` are propagated to downstream
crates, while `[dev-dependencies]` are not. As a planned Foundation-Layer crate, adding an
unnecessary runtime dependency edge is avoidable.

**Suggested fix:** Move `uuid` to a `[dev-dependencies]` section:

```toml
[dev-dependencies]
uuid = { workspace = true }
```

  </details>

---

#### `InviteKeyBundle::make` uses deprecated `dangerous_get_symmetric_key` in production code, violating CLAUDE.md rules

`crates/bitwarden-organization-crypto/src/invite_key_envelope.rs:43-48` **Caught by:** Code quality
agent

`make()` calls `ctx.dangerous_get_symmetric_key(key_id)` and `.clone()`s the returned
`&SymmetricCryptoKey` into `InviteKeyData`, storing raw key material outside the key store in a
named struct field. This directly violates two documented rules in the sdk-internal `CLAUDE.md`:

1. _"Use `bitwarden_crypto::safe` module first (password-protected key envelope, data envelope)
   instead of more low-level primitives"_
2. _"Do not expose key material from the crypto crate, use key references in the key store instead"_

The deprecation note on the function states: _"This function should ideally never be used outside
this crate."_ This is not a test — it is in the production `make()` path.

**Mitigating factors:** `SymmetricCryptoKey` has `ZeroizeOnDrop`, so the key is zeroed when
`InviteKeyBundle` drops. The existing codebase has established uses of `dangerous_get_symmetric_key`
in `bitwarden-vault`, `bitwarden-core`, and `bitwarden-wasm-internal`.

**Suggested fix:** Add an inline comment at the `#[allow(deprecated)]` site documenting why raw key
extraction is required for the AC auto-confirm protocol and what lifecycle constraint justifies the
exception (e.g., that `InviteKeyBundle` is dropped after the invite bytes are transmitted). This
makes the deliberate exception legible to future reviewers rather than appearing as an accidental
`#[allow(deprecated)]`.

  </details>

---

#### Raw key material extracted into non-key-store struct escapes key store lifecycle (EK)

`crates/bitwarden-organization-crypto/src/invite_key_envelope.rs:43-48` **Caught by:** Security &
logic agent

The same `dangerous_get_symmetric_key` + `.clone()` pattern has a security dimension: by
materializing the raw `SymmetricCryptoKey` in `InviteKeyData`, the crate creates an out-of-store
copy of the invite key that is invisible to the key store lifecycle. Any future protections added at
the store level (memory locking, hardware-backed storage, key rotation hooks) will not apply to the
copy held in `InviteKeyData`.

**Security category:** EK — encryption keys must be protected in memory. The cloned key escapes the
store's protection boundary.

**Downgraded from ⚠️ Important to ♻️ Refactor** because: (a) `SymmetricCryptoKey` has
`ZeroizeOnDrop` so memory is zeroed on drop; (b) this pattern is already present across multiple
other SDK crates; and (c) there is no demonstrated exploit path from the isolated key copy to a
concrete breach.

  </details>

---

#### `raw_invite_key()` returns `B64Url` holding unzeroized key bytes (EK, P02)

`crates/bitwarden-organization-crypto/src/invite_key_envelope.rs:20-23` **Caught by:** Security &
logic agent

`raw_invite_key()` calls `B64Url::from(&self.raw_key_data)`, which executes:

1. `key.0.to_encoded()` → allocates `BitwardenLegacyKeyBytes` (`Vec<u8>`, no `ZeroizeOnDrop`)
2. `B64Url::from(…)` → allocates a second `Vec<u8>` inside `B64Url` (no `ZeroizeOnDrop`)

The `SymmetricCryptoKey` inside `InviteKeyData` WILL be zeroized via `ZeroizeOnDrop`. But the
intermediate `BitwardenLegacyKeyBytes` and the returned `B64Url` hold unzeroized copies of the raw
key bytes.

**Category:** EK (key bytes unprotected in intermediaries), P02 (weakened for the post-generation
window).

**Downgraded from ⚠️ Important to ♻️ Refactor** because the raw bytes are intentionally being
serialized for transmission — this is the entire purpose of `raw_invite_key()`. P02 governs
post-lock state, not an active key-export operation. The unzeroized intermediaries are short-lived
return values; practical risk is low.

**Optional improvement:** Document on `raw_invite_key()` that the returned `B64Url` holds unzeroized
key material and should be dropped as soon as consumed. If `bitwarden-encoding` gains
`ZeroizeOnDrop` on `B64Url` in the future, this concern dissolves.

  </details>

## Reviewed and Dismissed

  <details><summary>🔍 2 initial findings dismissed after validation</summary>

#### Architecture Overview in sdk-internal CLAUDE.md not updated for new Foundation-Layer crate

`.claude/CLAUDE.md:17-26` **Caught by:** Architecture agent **Original severity:** ♻️ Refactor
**Original confidence:** 85/100 **Dismissed at:** Step 4 validation **Dismissed because:**
Documentation-only gap with no functional impact. The CLAUDE.md architecture crate list is
explanatory prose, not an enforced registry. A missing entry for a new crate is a doc update a
senior engineer would note informally but would not raise as reviewable technical debt.

---

#### Commented-out dead code left in `test_into_base64_url`

`crates/bitwarden-organization-crypto/src/invite_key_envelope.rs:123` **Caught by:** Code quality
agent **Original severity:** ♻️ Refactor **Original confidence:** 82/100 **Dismissed at:** Step 5
severity audit **Dismissed because:** A single commented-out line in a test does not meet the
Refactor bar — it creates no maintenance burden, violates no documented convention, and is something
a senior engineer would mention informally at most.

  </details>


</details>

---

<details>
<summary>Model Comparison: PR #1021 Code Review</summary>

TL;DR: Same issues found, opposite severity calls. Opus flagged 5 Important; Sonnet called them all Refactor. The critical gap: Sonnet missed the EncString vs SymmetricKeyEnvelope namespace separation finding entirely — the only finding with real key-confusion attack surface. Sonnet found one thing Opus missed (uuid in wrong Cargo section). Sonnet's downgrade reasoning leans too hard on "pattern exists elsewhere." Use Opus's findings for this PR.


**Date:** 2026-05-07 | **PR:** [PM-34165] Introduce InviteKeyBundle

## Severity Summary

| Model  | 🛑 Blocker | ⚠️ Important | ♻️ Refactor | Dismissed |
| ------ | ---------- | ------------ | ----------- | --------- |
| Opus   | 0          | 5            | 1           | 0         |
| Sonnet | 0          | 0            | 4           | 2         |

Both found the same core issues. They disagree entirely on severity.

---

## What Each Found Exclusively

### Opus only

**EncString vs `safe::SymmetricKeyEnvelope` (namespace separation)** — The most architecturally
significant finding. `wrap_symmetric_key` returns a bare `EncString` with no object-type tag; a
namespaced `SymmetricKeyEnvelope` already exists for exactly this use case. Without a namespace, a
malicious server could swap one wrapped key for another and the client can't detect it. Sonnet
missed this entirely.

**Naming incoherence** — `invite_key_envelope.rs` / `InviteKeyEnvelopeError` when the exported type
is `InviteKeyBundle`. "Envelope" is a term of art in this codebase (COSE-encrypted, namespaced
containers). Sonnet missed this.

### Sonnet only

**`uuid` in `[dependencies]` instead of `[dev-dependencies]`** — Concrete, correct, actionable. Opus
missed this.

---

## Where They Agree on the Issue, Not the Severity

Both found:

1. `dangerous_get_symmetric_key` called in production code from an external crate (deprecated,
   rule-violating)
2. Raw `SymmetricCryptoKey` stored outside the key store in `InviteKeyData`
3. `raw_invite_key()` returns unzeroized key bytes with documentation-only enforcement

**Opus:** Important (security violations, EK + P05 implications) **Sonnet:** Refactor (downgraded)

### Sonnet's downgrade reasoning, evaluated

| Argument                                 | Valid?                                                                       |
| ---------------------------------------- | ---------------------------------------------------------------------------- |
| `SymmetricCryptoKey` has `ZeroizeOnDrop` | Yes — genuine mitigation                                                     |
| Pattern exists elsewhere in the SDK      | Weak — precedent for a bad pattern isn't justification                       |
| No demonstrated exploit path             | Reasonable for avoiding Blocker; doesn't override documented rule violations |

Sonnet's downgrades are defensible but lean heavily on "it's done elsewhere" rather than assessing
whether the rules should apply here.

---

## Verdict

**Opus was more rigorous.** It found the most significant gap (missing namespace separation) that
Sonnet missed entirely, correctly treated rule violations in new production code as Important rather
than Refactor, and caught the naming incoherence.

**Sonnet was not wrong, just lenient.** Its mitigating factors are real, its `uuid` finding is
correct, and its dismissals (CLAUDE.md update, commented-out test line) were the right call. But it
passed over the EncString/namespace issue — the one finding in this PR with genuine key-confusion
attack surface — without comment.

**For this PR: use Opus's findings.** The two root causes Opus identified (no `safe` module usage,
raw key material outside the store) are violations of documented rules in a _new_ crate that claims
to "safely manipulate organization-specific cryptography." Sonnet's severity floor was set too low
for code making that claim.

</details>


[PM-34165]: https://bitwarden.atlassian.net/browse/PM-34165?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[PM-34165]: https://bitwarden.atlassian.net/browse/PM-34165?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

crates/bitwarden-organization-crypto/src/invite_key_bundle.rs
```diff
@@ -0,0 +1,168 @@
...
+    /// Get the raw invite key bytes in B64Url encoding
+    /// CRITICAL: this data MUST NOT be sent to the server
+    pub fn raw_invite_key(&self) -> &InviteKeyData {
+        &self.raw_key_data
+    }
+
+    /// Get the invite key wrapped using the organization key
+    pub fn organization_wrapped_invite_key(&self) -> &EncString {
+        &self.wrapped_invite_key
+    }
```
quexten: I think this is a testing only API and not something we want to publically support as an interface. Consumers shouldn't care that we use an encstring under the hood. You use it in the examples, so one of the solutions here is to add a compile-time flag, which is enabled for examples.

```suggestion
    #[cfg(feature = "test-support")]
    pub fn organization_wrapped_invite_key(&self) -> &EncString {
        &self.wrapped_invite_key
    }
```

You can then use the `require-features` field to make sure your example only runs when the feature is enabled https://doc.rust-lang.org/cargo/reference/cargo-targets.html#the-required-features-field

quexten (changes requested):
Looks good, I just want to make sure the internal details of the bundle are not public api surface that teams somehow depend on.

Once we start using it, we should make functions to operate on it (rotate, accept_org_invite) but not just give out the internal cryptographic details to public consumers.

Feel free to ping me after resolving the one concern, and I'll approve.

crates/bitwarden-organization-crypto/src/invite_key_bundle.rs
eligrubb: This method's purpose was for AC to get the organization wrapped invite key so it could be stored server side. However, I agree that the design leads to unnecessary public API dependency on EncString and removes type safety. I've added a wrapper type called `InviteKeyEnvelope` which hopefully limits these issues while fulfilling the original goal. The `InviteKeyEnvelope` inner type is private, so in the future we can swap `EncString` to migrate to safer underlying primitives. 

For example, `InviteKeyEnvelope` could look like this once org keys are upgraded:
```
pub struct InviteKeyEnvelope(InnerKeyEnvelope);

enum InnerKeyEnvelope {
  EncString(EncString),
  SymmetricKeyEnvelope(SymmetricKeyEnvelope),
}
```

So we can control migrations without breaking the API.

crates/bitwarden-organization-crypto/src/invite_key_envelope.rs
eligrubb: The main reason I don't want to return `SymmetricCryptoKey` directly here is default serialization and WASM encoding use `base64` for the `SymmetricCryptoKey` struct. However, for the raw invite key bytes we want `B64Url` encoding: the only consumer is AC using the encoded form to build URLs. 

My solution was struct-specific wasm bindings and serde serialization, all relying on `From<&InviteKeyData> for String` and `FromStr` to do the actual encoding. The new struct provide the same support across FFI boundaries with increased type safety. Additionally, returning the `InviteKeyData` wrapper type means we get type separation preventing users from taking a `SymmetricCryptoKey` generated using this primitive and applying it elsewhere.

crates/bitwarden-organization-crypto/examples/make_invite_key_bundle.rs
eligrubb: Since the API exposed to the examples folder has shrunk, I removed this check altogether. However, I noticed you added an `assert_symmetric_keys_equal()` function to the KeyContext, so I'm using it in the unit tests for `InviteKeyBundle`.

eligrubb:
I reconfigured the design of `InviteKeyBundle` to hide away unnecessary cryptographic complexity and internal types. I also reorganized the `examples/make_invite_key_bundle.rs` example to better highlight AC's use cases outlined in the Jira ticket. Hopefully this better illustrates what the API should and shouldn't be doing. Let me know if you have any questions!

eligrubb:
The sonarqube security failure appears to be a false positive. The failing check asserts the issue is there's no `Cargo.lock` checked in along with the new `crates/bitwarden-organization-crypto/Cargo.toml`, meaning dependencies are unpredictable. However, my understanding is the workspace `Cargo.lock` handles this already, and my changes include updates to the workspace `Cargo.lock` specifically addressing the new crate's dependencies.

quexten (dismissed):

dani-garcia (dismissed):

SaintPatrck (dismissed):

eligrubb:
Fixed a merge conflict affecting `.github/CODEOWNERS`. All other changes are the same as last reviewed.

dani-garcia (approved):

Thomas-Avery (approved):
Approving for tech leads CODEOWNERS changes.

quexten (approved):

SaintPatrck (approved):

PR #1022, opened 2026-05-01
[PM-35828]  Add missing drivers_license and passport fields to Cipher in tests

nikwithak:
## 🎟️ Tracking

## 📔 Objective

#992 merged after adding `drivers_license` and `passport` to `Cipher`, which broke a couple of the tests. This adds those back in to fix the build pipelines.


nick-livefront (approved):
Thank you!

PR #1028, opened 2026-05-04
[PM-34239] Fix dylint warning on cipher/blob/sealed.rs

dani-garcia:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-34239

## 📔 Objective

Rename the error variants to not have an `Error` suffix. This should be caught by our custom lints, but they're not currently enabled in CI so I'm trying to fix them and enable a CI check. The work is split per team to avoid too many conflicts.

https://github.com/bitwarden/sdk-internal/blob/a7c9289e54a930f02f79a11fa133100e0ecf1b9e/support/lints/error_enum/src/lib.rs#L12-L42


nick-livefront (approved):
✅

PR #1032, opened 2026-05-04
[PM-36551] Add uniffi bindings to policies crate

eliykat:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-36551

## 📔 Objective

Add uniffi bindings to the policies crate for consumption by mobile.


eliykat:
Android breaking changes appear to be due to Vault's work adding new item types, unrelated to this PR.

sven-bitwarden (dismissed):

coroiu (commented):
@eliykat should policies be owned by AC?

eliykat:
@coroiu The policies crate is joint owned by AC and Platform per Phase 1 of the [SDK ownership transition](https://bitwarden.atlassian.net/wiki/spaces/EN/pages/2035318803/Team+domain+SDK+ownership+transition). I think that's appropriate because this is the first substantive work we've done in rust, but let me know if you disagree. Hopefully we can upskill and drop off your code review TODO list soon.

Or if you're talking about the lib in the `bitwarden-uniffi` crate - how should this crate be structured? All other crates are structured by domain, but this crate appears to have team folders, and yet `CODEOWNERS` doesn't actually seem to grant per-team codeownership over them. Let me know and I can update as required.

crates/bitwarden-uniffi/src/policies.rs
```diff
@@ -0,0 +1,22 @@
...
+impl PoliciesClient {
+    /// Filter policies of the given type for the current user.
+    ///
+    /// Returns the subset of `policies` that should be enforced against the user,
+    /// based on their organization memberships and roles.
+    pub fn filter_by_type(
+        &self,
+        policies: Vec<PolicyView>,
+        organizations: Vec<ProfileOrganization>,
+        policy_type: u8,
```
quexten: @eliykat you should consider exposing an enum here instead of a u8. Enums are supported across the FFI boundary. If an enum is not possible, you should at least consider a [newtype](https://doc.rust-lang.org/rust-by-example/generics/new_types.html) around the policy type. `u8` should generally not be directly used.

Doing so gives you three benefits:
- Type safety across FFI boundaries
- Easier mental model
- Ability to add validation later that is stricter than u8 / makes illegal states unrepresentable

quexten: Actually, reading more thoroughly, why not just use `PolicyType` here? It is already effectively a newtype around u8, why not expose it across the FFI boundary here too considering you are generating the uniffi support code for it?

eliykat: Yes, this should've been using `PolicyType`! Good pick up, thanks.

quexten (approved):

sven-bitwarden (approved):

coroiu (approved):

PR #1033, opened 2026-05-05
[PM-35205] Implement CLI client state machine

dani-garcia:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-35205

## 📔 Objective

Introduce a compile-time client-state machine for the `bw` CLI so commands declare the auth/lock state they need (`LoggedOut`, `LoggedIn`, `Locked`, `Unlocked`, `AnyState`) instead of checking it manually.

- New `BwCommand` trait + `BwCommandExt::dispatch` extension trait. Each command picks a   `Client` state; the dispatcher extracts it from a single `ClientContext` via `TryFrom`   and rejects mismatched states with a user-friendly error before the command runs.
- `ClientState` is sealed so that new states can't be added outside the file.
- Migrated `sync`, `encode`, `config`, and `completion` to the new trait. Each got its own module under `platform/` so the command logic lives with the team that owns it.
- Added `PasswordManagerClient::is_unlocked()` so the `Locked`/`Unlocked` extractors can check key-store state. Note, this should probably be moved over to the crate that deals with unlock, (maybe after [PM-34417](https://bitwarden.atlassian.net/browse/PM-34417)?)
- Added `init_cli_platform_info()` to seed `HostPlatformInfo` with the CLI's device type, user agent, and version on startup.

## 🚨 Breaking Changes

[PM-34417]: https://bitwarden.atlassian.net/browse/PM-34417?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


djsmith85 (approved):
Changes are looking great!

Having a dispatch makes this easy to use.
Moving config into it's own files makes this cleaner.

PR #1034, opened 2026-05-06
Add more KM state to state bridge

quexten:
## 🎟️ Tracking

[no ticket]

## 📔 Objective

Adds the UpgradeToken, MPUnlockData, AccountCryptographicState to the state bridge. 
Mainly, the upgrade token is required for the PIN migration after key rotation.

Since this seems trivial, there is no ticket


djsmith85 (approved):
Approving for platform
KM can decide if a ticket is needed for tracking and possibly QA

quexten:
@djsmith85 It seems the co-ownership lead to your approval being enough to auto-merge. Apologies :(

I'll DM a KM reviewer to still double check.

PR #1035, opened 2026-05-06
[PM-31056] Migrate PIN unlock after V2 upgrade

quexten:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-31056

## 📔 Objective

Implements PIN migration to v2 on unlock. Note: Since not all clients have the state bridge implemented yet, this bails out without error if the state bridge is not implemented.


crates/bitwarden-core/src/key_management/pin_lock_system.rs
```diff
@@ -125,6 +125,100 @@ impl PinLockSystem<'_> {
             .map_err(|_| UnlockError::InternalError)
     }
 
+    /// After a V2 upgrade, when a V2 upgrade token is present and the persistent
+    /// PIN envelope is still encrypted with the V1 user key, this function migrates
+    /// the persistent PIN enrollment to be encrypted with the current user-key.
+    async fn migrate_pin_envelope_if_needed(&self) {
+        let Some(envelope) = self
```
quexten, line 152: Will be removed when the bridge is merged on all clients.

crates/bitwarden-core/src/key_management/pin_lock_system.rs
```diff
@@ -125,6 +125,100 @@ impl PinLockSystem<'_> {
...
+        // Do a fresh enrollment with the new user-key
+        if self
+            .set_pin(pin, PinLockType::BeforeFirstUnlock)
+            .await
+            .is_err()
+        {
+            warn!("Failed to re-seal PIN envelope after migration, unenrolling PIN");
+            self.unset_pin().await;
+        }
+    }
```
dani-garcia, line 216: Seems like every error path is doing `self.unset_pin().await;` I wonder if it would be easier for the function to return the error and then just do:

```rust
if let Err(e) = self.migrate_pin_envelope_if_needed().await {
    self.unset_pin().await;
}
```

dani-garcia (dismissed):

crates/bitwarden-core/src/key_management/pin_lock_system.rs
quexten: Nice. I've made the pin migration return an error enum with the different error states, and have a central log on it (as we do want to have this in e.g. flight recorder), and unenrolling centrally as per your suggestion.

quexten:
Macos build errors are unrelated and happening on other PRs.

eligrubb (approved):

PR #1037, opened 2026-05-06
[PM-36565] creating new item from new dialog is not working

JaredScar:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-36565

## 📔 Objective

Fixes issue where collections decrypt fails because it expects a Number type of u8, but clients code actually utilized string for the collection type.


eliykat (changes requested):
This is the wrong way round - we should be using the numeric format because that is the standard wire format for our enums. (e.g. this is how it's stored in the db, sent from the server, and sent by mobile clients to the SDK today which already use this code.)

Please review how other enums are sent across the boundary by `clients` and follow that pattern to send the numeric value here.

Hinton:
@eliykat It's weird that the JS types exposes strings as the type. I wonder if we should move to using `wasm_bindgen` instead of tsify.

Current types:

```js
/**
 * Type of collection
 */
export type CollectionType = "SharedCollection" | "DefaultUserCollection";
```

If we changed it to:

```rust
#[derive(Serialize_repr, Deserialize_repr, Debug, Clone, Copy, Eq, PartialEq)]
#[cfg_attr(feature = "uniffi", derive(uniffi::Enum))]
#[repr(u8)]
```
it gets exposed as:
```js
/**
 * Type of collection
 */
export enum CollectionType {
  /**
   * Default collection type. Can be assigned by an organization to user(s) or group(s)
   */
  SharedCollection = 0,
  /**
   * Default collection assigned to a user for an organization that has
   * OrganizationDataOwnership (formerly PersonalOwnership) policy enabled.
   */
  DefaultUserCollection = 1,
}
```

djsmith85 (commented):
Please ping for re-review when comments from Thomas and Oscar have been addressed

eliykat:
@Hinton good point, I didn't realise it was exposed like that. I agree with your recommendation, and that should probably be our standard practice for exposing enums to TS.

PR #1038, opened 2026-05-06
[PM-31050] Add generating a v2_upgrade_token on key rotation

Thomas-Avery:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-31053

## 📔 Objective

Adds the ability to create an upgrade token for v1 -> v2 userKey rotations. Consumers will pass in `UpgradeTokenAction:CreateIfNeeded` when they want an upgrade token generated. 

Note I've made the `UpgradeTokenAction` optional so this isn't a breaking change. When a value isn't provided we default to skipping upgrade token generation. With PM-30213 in QA making this a breaking change would be clunky.

To make this non-optional in the future we can: 

- Merge this in
- Merge PM-30213 after QA is complete
- Update clients to pass in the `UpgradeTokenAction`
- Come back around and make the SDK `UpgradeTokenAction` non-optional


crates/bitwarden-user-crypto-management/src/key_rotation/unlock.rs
```diff
@@ -599,4 +638,78 @@ mod tests {
...
+        let unlock_data = reencrypt_master_password_change_unlock_data(
+            input,
+            current_user_key_id,
+            new_user_key_id,
+            &mut ctx,
+        )
+        .expect("should be ok");
+
+        assert!(
+            unlock_data.v2_upgrade_token.is_some(),
```
quexten: Could we verify that this is actually a parseable upgrade token by deserializing it into an upgrade token?

We've recently had the switched pub / private key, which makes me think an "is_some()" assertion is not enough.

quexten (dismissed):

crates/bitwarden-user-crypto-management/src/key_rotation/unlock.rs
Thomas-Avery: Coverage should be increased in https://github.com/bitwarden/sdk-internal/pull/1038/changes/a2b19e49f0f2fa0c2a055d291025faa23f829a7c

crates/bitwarden-user-crypto-management/src/key_rotation/rotate_user_keys.rs
```diff
@@ -38,13 +38,28 @@ pub enum KeyRotationMethod {
     Tde,
 }
 
+#[derive(Serialize, Deserialize, Clone)]
+#[cfg_attr(feature = "wasm", derive(Tsify), tsify(into_wasm_abi, from_wasm_abi))]
+#[cfg_attr(feature = "uniffi", derive(uniffi::Enum))]
+pub enum UpgradeTokenAction {
+    /// Skip creating and sending an upgrade token to the server. This will be the default behavior
```
quexten, line 45: Nice, way better than the boolean we had discussed in the 1:1. I like this.

crates/bitwarden-user-crypto-management/src/key_rotation/rotate_user_keys.rs
```diff
@@ -38,13 +38,28 @@ pub enum KeyRotationMethod {
...
 #[derive(Serialize, Deserialize, Clone)]
 #[cfg_attr(feature = "wasm", derive(Tsify), tsify(into_wasm_abi, from_wasm_abi))]
 #[cfg_attr(feature = "uniffi", derive(uniffi::Record))]
 pub struct RotateUserKeysRequest {
     pub key_rotation_method: KeyRotationMethod,
     pub trusted_emergency_access_public_keys: Vec<PublicKey>,
     pub trusted_organization_public_keys: Vec<PublicKey>,
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    #[cfg_attr(feature = "wasm", tsify(optional))]
+    pub upgrade_token_action: Option<UpgradeTokenAction>,
```
quexten, line 62: Question: I'm confused, why are we making it an optional with `None` being equivalent to `Skip`? Can we not just always require the action?

Is this just a temporary addition to not make the changes breaking?

Thomas-Avery: See comment in the PR. 

Note I've made the UpgradeTokenAction optional so this isn't a breaking change. When a value isn't provided we default to skipping upgrade token generation. With PM-30213 in QA making this a breaking change would be clunky.

To make this non-optional in the future we can:

Merge this in
Merge PM-30213 after QA is complete
Update clients to pass in the UpgradeTokenAction
Come back around and make the SDK UpgradeTokenAction non-optional

quexten: Ah, sorry, I missed this.

quexten (approved):

PR #1040, opened 2026-05-07
[PM-31113] Migrate local user data key after V2 upgrade

quexten:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-31113

## 📔 Objective

Implements re-wrapping of the local-user-data-key after an upgrade of the user account to v2 encryption.


crates/bitwarden-core/src/key_management/crypto.rs
```diff
@@ -2396,6 +2403,253 @@ mod tests {
...
+                        kdf: Kdf::PBKDF2 {
+                            iterations: 600_000.try_into().unwrap(),
+                        },
+                        master_key_wrapped_user_key: TEST_ACCOUNT_USER_KEY.parse().unwrap(),
+                        salt: TEST_USER_EMAIL.to_string(),
+                    },
+                },
+                upgrade_token: Some(upgrade_token),
+            },
+        )
```
Thomas-Avery: Looks like we are duplicating this. Might be worth making a helper test function.

crates/bitwarden-core/src/key_management/local_user_data_key_state.rs
```diff
@@ -37,6 +38,74 @@ pub(crate) async fn initialize_local_user_data_key_into_state(
...
+///
+/// No-op when:
+/// - state bridge is not registered by the host
+/// - upgrade token is not available
+/// - no `LocalUserDataKeyState` is present for `user_id` yet (first-time init handles creation),
+pub(crate) async fn migrate_local_user_data_key_for_user_key_upgrade(
+    client: &Client,
+    user_id: UserId,
+) -> Result<(), MigrateLocalUserDataKeyForUserKeyUpgradeError> {
+    // Todo: Remove when all host clients implement the state bridge
```
Thomas-Avery: Can we link a JIRA ticket for tracking here?

crates/bitwarden-core/src/key_management/local_user_data_key_state.rs
```diff
@@ -37,6 +38,74 @@ pub(crate) async fn initialize_local_user_data_key_into_state(
...
+
+/// Re-wraps a persisted [`WrappedLocalUserDataKey`] with the current user key after a V1→V2
+/// user-key upgrade. Preserves the inner-key plaintext so local data encrypted before the
+/// upgrade remains decryptable.
+///
+/// No-op when:
+/// - state bridge is not registered by the host
+/// - upgrade token is not available
+/// - no `LocalUserDataKeyState` is present for `user_id` yet (first-time init handles creation),
+pub(crate) async fn migrate_local_user_data_key_for_user_key_upgrade(
```
Thomas-Avery, line 47: Would be nice to have unit tests for this.

crates/bitwarden-core/src/client/encryption_settings.rs
```diff
@@ -55,6 +55,10 @@ pub enum EncryptionSettingsError {
     /// The local user data key could not be loaded into the key store context.
     #[error("Unable to load local user data key into key store")]
     LocalUserDataKeyLoadFailed,
+
+    // The migration (re-wrap) of the local user data key failed after a user key upgrade
```
Thomas-Avery:
```suggestion
    /// The migration (re-wrap) of the local user data key failed after a user key upgrade
```

crates/bitwarden-core/src/key_management/local_user_data_key_state.rs
```diff
@@ -37,6 +38,74 @@ pub(crate) async fn initialize_local_user_data_key_into_state(
...
+pub(crate) struct MigrateLocalUserDataKeyForUserKeyUpgradeError;
+
+/// Re-wraps a persisted [`WrappedLocalUserDataKey`] with the current user key after a V1→V2
+/// user-key upgrade. Preserves the inner-key plaintext so local data encrypted before the
+/// upgrade remains decryptable.
+///
+/// No-op when:
+/// - state bridge is not registered by the host
+/// - upgrade token is not available
+/// - no `LocalUserDataKeyState` is present for `user_id` yet (first-time init handles creation),
```
Thomas-Avery: There looks to be more no-ops here. Specifically `wrappedLocalUserDataKey is not a V1 wrapped key` and when `token.unwrap_v1` fails `Upgrade token does not apply to current user key`. 

This is documenting specifics of the "what" of this function. Typically this is hard to keep in sync with code and the recommendation is to instead communicate the "why". I would suggest something more concise. 


```
/// Re-wraps a persisted [`WrappedLocalUserDataKey`] with the current user key after a V1→V2
/// user-key upgrade, preserving the inner-key plaintext so local data encrypted before the
/// upgrade remains decryptable. No-ops when migration is unnecessary or impossible.
```

Thomas-Avery (changes requested):

crates/bitwarden-core/src/key_management/local_user_data_key_state.rs
quexten: DM'd. I added tests, and made them follow a new pattern which I think is easier to follow. LMK what you think.

dereknance (dismissed):

crates/bitwarden-crypto/src/store/context.rs
```diff
@@ -809,6 +809,23 @@ impl<Ids: KeySlotIds> KeyStoreContext<'_, Ids> {
...
+    /// A test helper to assert that the symmetric keys corresponding to the given identifiers are
+    /// equal. DO NOT USE THIS FOR NON-TEST CODE
+    pub fn assert_symmetric_keys_equal(&self, key_id_1: Ids::Symmetric, key_id_2: Ids::Symmetric) {
+        let key_1 = self
+            .get_symmetric_key(key_id_1)
+            .expect("Key 1 should exist in context");
+        let key_2 = self
+            .get_symmetric_key(key_id_2)
+            .expect("Key 2 should exist in context");
+        if key_1 != key_2 {
```
Thomas-Avery, line 823: 💭
I'm not a fan of coupling test code and implementation code like this. Relying on code comments for enforcement of what to use is a concern and not scalable.
Suggestions to consider:

- Crate feature flag for test utilities
- Stand alone test utility crate

## Crate feature flag
```
#[cfg(any(test, feature = "test-utils"))]
impl<Ids: KeySlotIds> KeyStoreContext<'_, Ids> {
    /// Test-only helper that panics if the symmetric keys stored under the two identifiers are
    /// not equal.
    pub fn assert_symmetric_keys_equal(
        &self,
        key_id_1: Ids::Symmetric,
        key_id_2: Ids::Symmetric,
    ) {
        let key_1 = self
            .get_symmetric_key(key_id_1)
            .expect("Key 1 should exist in context");
        let key_2 = self
            .get_symmetric_key(key_id_2)
            .expect("Key 2 should exist in context");
        assert_eq!(
            key_1, key_2,
            "Symmetric keys with ids {key_id_1:?} and {key_id_2:?} are not equal",
        );
    }
}
```

Consumer crates need to add
```
[dev-dependencies]
bitwarden-crypto = { workspace = true, features = ["test-utils"] }
```


This would keep `assert_symmetric_keys_equal` out of production code and enforced via compiler.

## Stand alone
Just have this as a dev dependency where you want it. This would require the utilities to not access the internals of the class which is a more traditional approach.

It would probably be good to discuss this with a wider audience and I'm good with having some duplicated test helpers in each file's`#[cfg(test)]` to not block this PR.

Thomas-Avery (changes requested):

crates/bitwarden-crypto/src/store/context.rs
quexten: I'm just adding test-utils as per your first suggestion. It is likely that we will want more test only code and as you mention the flag is the best approach for scaling this, so may as well add it here.

quexten:
Please note the macos build failures are unrelated and seem to happen on other PR's too:
https://github.com/bitwarden/sdk-internal/pull/1076
https://github.com/bitwarden/sdk-internal/pull/1075

dereknance (approved):

crates/bitwarden-crypto/src/store/context.rs
```diff
@@ -809,6 +809,24 @@ impl<Ids: KeySlotIds> KeyStoreContext<'_, Ids> {
             self,
         )
     }
+
+    /// A test helper to assert that the symmetric keys corresponding to the given identifiers are
+    /// equal.
+    #[cfg(any(test, feature = "test-utils"))]
+    pub fn assert_symmetric_keys_equal(&self, key_id_1: Ids::Symmetric, key_id_2: Ids::Symmetric) {
```
Thomas-Avery, line 816:
```suggestion
    #[doc(hidden)]
    pub fn assert_symmetric_keys_equal(&self, key_id_1: Ids::Symmetric, key_id_2: Ids::Symmetric) {
```

As discussed I think I would go with `#[doc(hidden)]` here. But this is non blocking.

Thomas-Avery (approved):
LGTM

PR #1018, opened 2026-04-30
[PM-24221] Reorganize registration mod and add uniffi bindings

eligrubb:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-24221

## 📔 Objective

This PR builds on #728 reorganizes the registration client mod in the sdk to better follow [our file organization for client patterns](https://contributing.bitwarden.com/architecture/sdk/client-patterns/#file-organization). 

In addition, this PR also adds uniffi bindings for mobile clients for the `post_keys_for_user_password_registration` method on the registration client.

No functionality or APIs are changed (beyond the aforementioned uniffi addition). Auto-merge is disabled to ensure both auth and platform can review.

## 🚨 Breaking Changes

This PR changes the name of a WASM api method introduced in #728. This api is not currently used by any clients and there is no risk to the change.

PR #1041, opened 2026-05-07
[PM-36839] Replace explicit API error variants by .into()

dani-garcia:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-36839

## 📔 Objective

We currently have two `ApiError` enums: one in `bitwarden-api-base` and another one in `bitwarden-core`, which wraps the former and exposes it through UniFFI. 

I'd like to consolidate both of these into one (which would remove a fair amount of error mapping that we're doing), but doing so requires quite extensive changes.

This PR starts building on that by removing explicit usages of the error type and replacing them by `.into()` calls with type inference. This should make the error migration easier, and it also makes the errors simpler to read.

### DIFF
--- a/crates/bitwarden-auth/src/registration.rs
+++ b/crates/bitwarden-auth/src/registration.rs
@@ -820,9 +820,7 @@ mod tests {
                 .expect_post_keys()
                 .once()
                 .returning(move |_body| {
-                    Err(bitwarden_api_api::apis::Error::Serde(
-                        serde_json::Error::io(std::io::Error::other("API error")),
-                    ))
+                    Err(serde_json::Error::io(std::io::Error::other("API error")).into())
                 });
             // Subsequent API calls should not be made if post_keys fails
             mock.organization_users_api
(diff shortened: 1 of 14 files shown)


Thomas-Avery (approved):
KM changes LGTM

enmande (approved):
Auth changes LGTM; however, I think due to the reorganization of the `registration` mod, there might be a merge conflict race in progress with https://github.com/bitwarden/sdk-internal/pull/1018. 🤔 Due to the surface, this might do well to land first?

jaasen-livefront (approved):

dereknance (approved):

harr1424 (approved):
Tools owned changes look good

crates/bitwarden-auth/src/registration.rs
eligrubb, line 1041: It looks like `bitwarden_api_key_connector` uses a similar pattern, should replacing this variant be an extension of this work?

dani-garcia: Yeah, `bitwarden_api_key_connector` is currently using their own Error type, so my first goal is to consolidate the `core` Error type with the `api_base` error type (which will require removing the generic bound from the `api_base` type).

Once that's done we should be able to move `bitwarden_api_key_connector` to use the shared error type. I figured because that's all on KM's domain I can do it all in one PR.

PR #1047, opened 2026-05-07
Add post_keys_for_user_password_registration in uniffi bindings for registration client

eligrubb:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-24221
Originally part of #1018 - separated out to decouple review dependencies.

## 📔 Objective

This PR adds uniffi bindings for the newly created `post_keys_for_user_password_registration` method in the registration client, allowing mobile to access the new v2 registration flow.

This PR stacks on top of #1046 so it can use the correct method API.


djsmith85 (approved):
Changes are looking good!

PR #1049, opened 2026-05-08
[PM-36839] Prepare API base Error

dani-garcia:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-36839

## 📔 Objective

Prepare the Error type in `bitwarden-api-base` so it can be merged with the one in `bitwarden-core`. This requires:
- UniFFI support
- ResponseContent variant as it's own type, rather than inline in the enum, and without a generic type. Also renamed the variant to not have the `Error` suffix

Once this change is done, we can proceed with merging both error types, which is done on a separate PR as it involves basically every team:  https://github.com/bitwarden/sdk-internal/pull/1050


crates/bitwarden-api-base/src/error.rs
```diff
@@ -1,20 +1,30 @@
...
-pub struct ResponseContent<T = ()> {
+#[cfg_attr(feature = "uniffi", derive(uniffi::Record))]
+pub struct ResponseContent {
     /// HTTP status code of the response.
-    pub status: reqwest::StatusCode,
-    /// Raw response body content.
-    pub content: String,
-    /// Deserialized entity from the response.
-    pub entity: Option<T>,
+    pub status: u16,
```
djsmith85: ❓ With the removal `reqwest::StatusCode` we are losing type information. I think this would benefit from having a different type than just u16, to clarify these can only be http status codes?

djsmith85 (commented):
Changes are looking great, just one question regarding removing type information.

crates/bitwarden-api-base/src/error.rs
dani-garcia: Hmm good point, I removed the statuscode type as we can't pass it as is through UniFFI or WASM, but it's probably worth it for us to do a small wrapper that is FFI compatible

dani-garcia: I've reverted it back to use the actual `StatusCode` and implemented UniFFI type conversion and serde support (for WASM, required ~the http-serde crate~ a serializer module)

crates/bitwarden-api-base/Cargo.toml
```diff
@@ -14,12 +14,18 @@ repository.workspace = true
 license-file.workspace = true
 keywords.workspace = true
 
+[features]
+uniffi = ["dep:bitwarden-uniffi-error", "dep:uniffi"]
+
 [dependencies]
+bitwarden-uniffi-error = { workspace = true, optional = true }
 http = ">=1.4.0, <2"
+http-serde = { workspace = true }
```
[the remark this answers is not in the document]
dani-garcia: Replaced the dep with a simple hand rolled implementation in https://github.com/bitwarden/sdk-internal/pull/1049/changes/547fc7b321e67ed586e490ce7fbd083c77030253

djsmith85 (dismissed):
Changes are looking great !

mzieniukbw (dismissed):

djsmith85 (approved):

JaredSnider-Bitwarden (approved):

mzieniukbw (approved):

PR #1051, opened 2026-05-08
[PM-35207] Implement `bw lock`

dani-garcia:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-35207

## 📔 Objective

Implement `bw lock` for the Rust CLI.

Locking on the CLI is a persistence operation: because the process exits between invocations, locking just deletes the persisted session key from the database rather than clear in-memory key store (as `lock()` does on long-lived clients like mobile and desktop).

This adds:
- `PasswordManagerClient::invalidate_session_key()`, which removes `SESSION_PROTECTED_USER_KEY` from the settings repository. This will likely be moved to the `bitwarden-unlock` crate once that's done ([PM-34417](https://bitwarden.atlassian.net/browse/PM-34417))
- `LockArgs` in `crates/bw/src/key_management/lock.rs`, a `BwCommand` that calls `invalidate_session_key()` and reports `"Your vault is now locked."`.
- CODEOWNERS entry for `crates/bw/src/key_management/` under `team-key-management-dev`.

The command requires `LoggedIn` rather than `Unlocked` to match the legacy node CLI, which accepts `bw lock` even when the vault is already locked.

## 🚨 Breaking Changes

None.

## 🚨 Breaking Changes

[PM-34417]: https://bitwarden.atlassian.net/browse/PM-34417?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Thomas-Avery:
Added KM as reviewers because of CODEOWNERS assignment.

dereknance (approved):

quexten (approved):

quexten:
No specific comments from my end, seems straight forward.

Thomas-Avery (approved):
Approving for tech leads.

PR #1052, opened 2026-05-08
[PM-36565] Refactor CollectionType enum to use wasm_bindgen for better WASM integration

JaredScar:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-36565

## 📔 Objective

Fixes up the collection.rs file so the generating via wasm is correct and uses u8 rather than strings for TypeScript


eliykat (approved):

djsmith85 (approved):
Originally got tagged to review https://github.com/bitwarden/sdk-internal/pull/1037 which had some offline discussion and got closed to create this PR and the follow-up https://github.com/bitwarden/clients/pull/20560 in clients

Approving for platform

PR #1069, opened 2026-05-12
[PM-34165] Add Aes256CbcHmac support to SymmetricKeyEnvelope

eligrubb:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-34165

## 📔 Objective

We have two concurrent goals: move our APIs over to new, "`safe`", structures and move off of not-as-secure cryptographic protocols. While the second goal (for organization keys) is blocked by our switch to organization signing v2, we can make progress towards the first goal by supporting legacy keys in the new `safe` APIs.

This PR adds support for `SymmetricCryptoKey::Aes256CbcHmac` as the `sealing_key` for `SymmetricKeyEnvelope`. This will allow us to use the `SymmetricKeyEnvelope` for new APIs like the `InviteKeyBundle`. This is done by introducing the `SymmetricCoseEncrypt0EnvelopeKey` trait to simplify the logic flow.

### DIFF
--- a/crates/bitwarden-crypto/src/cose.rs
+++ b/crates/bitwarden-crypto/src/cose.rs
@@ -6,7 +6,7 @@
 use std::fmt::Debug;
 
 use coset::{
-    CborSerializable, ContentType, CoseEncrypt0, CoseEncrypt0Builder, Header, Label,
+    CborSerializable, ContentType, Header, Label,
     iana::{self, CoapContentFormat, KeyOperation},
 };
 use hybrid_array::Array;
@@ -28,6 +28,7 @@ use crate::{
 /// to be able to randomly generate nonces, and to not have to worry about key wearout. Since
 /// the draft was never published as an RFC, we use a private-use value for the algorithm.
 pub(crate) const XCHACHA20_POLY1305: i64 = -70000;
+pub(crate) const AES256_CBC_HMAC: i64 = -70001;
 pub(crate) const ALG_ARGON2ID13: i64 = -71000;
 
 // Custom labels for COSE headers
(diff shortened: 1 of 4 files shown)


crates/bitwarden-crypto/src/cose.rs
```diff
@@ -28,6 +28,7 @@ use crate::{
 /// to be able to randomly generate nonces, and to not have to worry about key wearout. Since
 /// the draft was never published as an RFC, we use a private-use value for the algorithm.
 pub(crate) const XCHACHA20_POLY1305: i64 = -70000;
+pub(crate) const AES256_CBC_HMAC: i64 = -70001;
```
quexten, line 31: https://datatracker.ietf.org/doc/html/rfc8152#section-5.3
https://datatracker.ietf.org/doc/html/rfc8152#section-5.4

~~AES256_CBC_HMAC is not an AEAD cipher but an AE cipher, and cannot be used in COSE to encrypt objects with a protected header. We do not have a user-case that does not have a protected header and therefore AES256_CBC_HMAC is not compatible with COSE.~~

This PR moves the protected fields into unprotected.

crates/bitwarden-crypto/src/safe/symmetric_key_envelope/symmetric_cose_encrypt0_key.rs
```diff
@@ -0,0 +1,288 @@
...
+        cose_ciphertext: &CoseEncrypt0,
+        namespace: SymmetricKeyEnvelopeNamespace,
+    ) -> Result<SymmetricCryptoKey, SymmetricKeyEnvelopeError> {
+        // verify the protected field is empty
+        if !cose_ciphertext.protected.is_empty() {
+            return Err(SymmetricKeyEnvelopeError::IncorrectCiphertextStructure);
+        }
+
+        validate_safe_namespaces(
+            &cose_ciphertext.unprotected,
```
quexten, line 196: This breaks all security guarantees  the namespacing security property that the symmetric key envelope is supposed to give. The unprotected header is not included in any authentication.

quexten (changes requested):
I have concerns with this approach. Given that the namespace is supposed to be placed in the protected header (and thus AAD of the cipher), and given that in cose, an AE cipher MUST have an empty protected header, that makes these approaches fundamentally incompatible.

Currently, this approach just places the namespace in the unprotected header, which renders it's purpose invalid.

Provided we do want to use the safe crypto primitive for the org invite work, I see two paths forward.
1. We upgrade AES256_CBC_HMAC to be an AEAD cipher in the context of COSE
  - For this, a AES256_CBC_HMAC encryption is done as:
    HMAC(IV, AAD_LENGTH, AAD, CBC(DATA))
  - This creates effectively an AAD version of AES256_CBC_HMAC, however do note that this is different from the regular AES256_CBC_HMAC encstrings
2. We make the consumer always have a COSE compatible key that wraps whichever object you want to have. In the case of the org, the org would either need to upgrade (which is not feasible), or gain a new ORG invite management key which is a cose key. However, this approach only pushes the problem one layer up the stack.

PR #1072, opened 2026-05-12
[PM-28192] Migrate Cipher Attachment Operations to use SDK

gbubemismith:
## 🎟️ Tracking
https://bitwarden.atlassian.net/browse/PM-28192

## 📔 Objective
Bring attachment management into the SDK (create, download, renew, upgrade, delete).

What changed:
  - Refactor: Moved attachment code out of CiphersClient into a new attachment_client/ module under bitwarden-vault.
  - New operations:
    - create_attachment - upload a new attachment to a cipher
    - upgrade - re-encrypt legacy attachments to the current crypto format
    - renew_file_upload_url - refresh an expired upload URL
    - get_attachment_download_url - fetch a download URL (plus admin/emergency-access variants)


crates/bitwarden-vault/src/cipher/attachment_client/upgrade.rs
```diff
@@ -0,0 +1,359 @@
...
+    ///
+    /// Returns the data needed to push the re-encrypted bytes to the upload URL. The caller
+    /// is responsible for the byte push (see [`AttachmentUpgrade`]) and for deleting the
+    /// legacy attachment via [`Self::delete_attachment`] once the push succeeds.
+    pub async fn prepare_attachment_upgrade(
+        &self,
+        cipher_id: CipherId,
+        attachment_id: String,
+    ) -> Result<AttachmentUpgrade, CipherPrepareAttachmentUpgradeError> {
+        let http_client = reqwest::Client::new();
```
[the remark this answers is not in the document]
gbubemismith: Resolved 8cae91ddbfd510d6f8be52f29be75c2b8bbd4fe9. Thanks

crates/bitwarden-vault/src/cipher/attachment_client/upgrade.rs
```diff
@@ -0,0 +1,358 @@
...
+    /// Server-assigned ID of the new attachment.
+    pub attachment_id: String,
+    /// URL to push the encrypted bytes to (presigned blob URL or local server endpoint).
+    pub upload_url: String,
+    /// Where to push the bytes — local Bitwarden server (`Direct`) or Azure Blob storage.
+    pub file_upload_type: FileUploadType,
+    /// Encrypted file name in `EncString` format.
+    pub encrypted_file_name: String,
+    /// Encrypted file bytes ready to push to `upload_url`.
+    pub encrypted_contents: Vec<u8>,
```
quexten: A thought: Attachments can be 500MiB. The WASM SDK will never release this memory as wasm does not support releasing the memory. Have we consider writing this as a streaming interface? Otherwise we will increase the memory usage of our clients by up to half a gigabyte until restart.

One way of doing this would be to pass in the attachment upgrade request along with a FileReader.

```rust
struct Chunk {
   data: Vec<u8>
   is_done: bool
}

#[wasm_bindgen]
extern "C" {
    pub type FileReader;

    #[wasm_bindgen(structural, method)]
    pub fn read(this: &FileReader) -> Chunk;
}
```

Then, you could write any upgrade requests to take the metadata as a struct, but the file contents as a file reader for which you implement the above interface in TS to stream from disk.

This approach scales to any filesize without increasing memory usage.

gbubemismith: TIL WASM linear memory only grows, good catch 🙌🏼 , I agree. One thing tho, I don't think the legacy format  actually forces buffering. Since the upgrade isn't called frequently, I don't think the streaming interface will be worth it for this PR. I'd rather drop the `prepare_attachment_upgrade` and let TS client wire the migrated operations together. 

If we ever move actual file uploads to the  SDK, this streaming interface becomes unavoidable. So it's worth keeping the design in mind

gbubemismith: @quexten, I tried giving this a go and noticed the existing crypto functions (`encrypt_aes256_hmac /decrypt_aes256_hmac`) take and return the entire payload as one buffer, which is what's blowing up memory on big attachments. There's no way to make those functions process a chunk at a time without changing their signature. I have a draft PR here https://github.com/bitwarden/sdk-internal/pull/1093 and they touched key management files. I'll appreciate key management's review on this

crates/bitwarden-vault/src/cipher/attachment_client/admin/download_url.rs
```diff
@@ -0,0 +1,183 @@
...
+            bitwarden_api_api::apis::Error::ResponseError(content)
+                if content.status == StatusCode::NOT_FOUND =>
+            {
+                CipherAdminGetAttachmentDownloadUrlError::NotFound
+            }
+            other => other.into(),
+        })?;
+
+    response.url.ok_or_else(|| MissingFieldError("url").into())
+}
```
nikwithak: I think the new guidance is to avoid the standalone URLs and implement the logic directly in the client function, and testing by mocking the attributes of the client directly.

crates/bitwarden-vault/src/cipher/attachment_client/create.rs
```diff
@@ -0,0 +1,504 @@
...
+
+/// Metadata for opening a new attachment slot on the server.
+#[derive(Clone, Debug, Serialize, Deserialize)]
+#[cfg_attr(feature = "wasm", derive(Tsify), tsify(into_wasm_abi, from_wasm_abi))]
+#[serde(rename_all = "camelCase")]
+pub struct CreateAttachmentRequest {
+    /// Encrypted attachment key in `EncString` format.
+    pub key: String,
+    /// Encrypted file name in `EncString` format.
+    pub file_name: String,
```
nikwithak: Is there a reason not define these as `EncString` type?

crates/bitwarden-vault/src/cipher/attachment_client/create.rs
```diff
@@ -0,0 +1,504 @@
...
+    /// Size of the encrypted file in bytes.
+    pub file_size: i64,
+    /// Revision date of the cipher when the request was prepared, used by the server
+    /// to detect concurrent modifications.
+    pub last_known_revision_date: String,
```
nikwithak: Should this be a `DateTime<Utc>` instead?

crates/bitwarden-vault/src/cipher/attachment_client/create.rs
```diff
@@ -0,0 +1,504 @@
...
+#[derive(Clone, Debug, Serialize, Deserialize)]
+#[cfg_attr(feature = "wasm", derive(Tsify), tsify(into_wasm_abi, from_wasm_abi))]
+#[serde(rename_all = "camelCase")]
+pub struct CreateAttachmentRequest {
+    /// Encrypted attachment key in `EncString` format.
+    pub key: String,
+    /// Encrypted file name in `EncString` format.
+    pub file_name: String,
+    /// Size of the encrypted file in bytes.
+    pub file_size: i64,
```
nikwithak: nit: Could this be a `u64` since we can't have a negative file size?

crates/bitwarden-vault/src/cipher/attachment_client/create.rs
nikwithak: Actually since this is the request passed to the SDK from the client, could we make these the unencrypted values to match the patterns in `CiphersClient`? For CiphersClient operations we pass the unencrypted data to the SDK, and the SDK handles the encryption before converting it to the request, so the client doesn't have to care about it.

crates/bitwarden-vault/src/cipher/attachment_client/create.rs
```diff
@@ -0,0 +1,504 @@
...
+            .cipher_mini_response
+            .ok_or(MissingFieldError("cipher_mini_response"))?;
+        Some((*cipher_mini).merge_with_cipher(None)?)
+    } else {
+        let cipher_response = response
+            .cipher_response
+            .ok_or(MissingFieldError("cipher_response"))?;
+        let merged = (*cipher_response).merge_with_cipher(existing_cipher)?;
+        repository.set(cipher_id, merged).await?;
+        None
```
nikwithak: Should we still return the `merged` cipher, here?

crates/bitwarden-vault/src/cipher/attachment_client/delete.rs
```diff
@@ -53,22 +53,20 @@ pub async fn delete_attachment<R: Repository<Cipher> + ?Sized>(
...
         attachment_id: String,
     ) -> Result<Cipher, CipherDeleteAttachmentError> {
-        let configs = self.client.internal.get_api_configurations();
         delete_attachment(
             cipher_id,
             &attachment_id,
-            &configs.api_client,
-            &*self.get_repository()?,
+            &self.api_configurations.api_client,
+            self.repository.require()?.as_ref(),
```
nikwithak: :art: Consider moving the logic out of the standalone function into the `AttachmentsClient` here as part of this change, to match the recommended pattern.

crates/bitwarden-vault/src/cipher/attachment_client/upgrade.rs
nikwithak: Same as above - for the `CiphersClient` we pass the decrypted data back and forth, and allow the SDK to manage encrypting the data before sending it to the server. Is there a reason not to keep that same pattern here, and offload the necessity to consciously encrypt data from the client?

crates/bitwarden-vault/src/cipher/attachment_client/admin/download_url.rs
nikwithak: (Applies here and the other operations)

nikwithak (changes requested):

crates/bitwarden-vault/src/cipher/attachment_client/create.rs
```diff
@@ -0,0 +1,531 @@
...
+}
+
+/// Server data for a newly created attachment slot. The caller uploads the
+/// encrypted bytes to [`Self::upload_url`] using [`Self::file_upload_type`]
+#[derive(Clone, Debug, Serialize, Deserialize)]
+#[cfg_attr(feature = "wasm", derive(Tsify), tsify(into_wasm_abi, from_wasm_abi))]
+#[serde(rename_all = "camelCase")]
+pub struct CreatedAttachment {
+    /// Server-assigned attachment ID.
+    pub attachment_id: String,
```
nikwithak, line 105: Could we add a `AttachmentId` type? There's a `uuid_newtype!` macro to make it a oneliner: https://github.com/bitwarden/sdk-internal/blob/main/crates/bitwarden-vault/src/cipher/cipher.rs#L43

crates/bitwarden-vault/src/cipher/attachment_client/create.rs
```diff
@@ -0,0 +1,531 @@
...
+    /// Upload target for the encrypted bytes
+    pub upload_url: String,
+    /// Bitwarden server or Azure Blob Storage.
+    pub file_upload_type: AttachmentFileUploadType,
+    /// Cipher returned by the server. For non-admin requests, this is also
+    /// written to the local repository
+    pub cipher: Cipher,
```
nikwithak, line 112: (Most of) the `CIphersClient` operations return the decrypted cipher in responses - is there a reason not to do the same here?

crates/bitwarden-vault/src/cipher/attachment_client/create.rs
```diff
@@ -0,0 +1,531 @@
...
+}
+
+#[cfg_attr(feature = "wasm", wasm_bindgen)]
+impl AttachmentsClient {
+    /// Creates a new attachment slot on the server and updates local repository
+    /// state with the merged cipher returned by the server.
+    ///
+    /// The caller must upload the encrypted bytes to [`CreatedAttachment::upload_url`]
+    /// using the transport in [`CreatedAttachment::file_upload_type`].
```
nikwithak, line 121: Is there an opportunity to let the SDK handle this? i.e. the client calls `AttachmentClient::create_attachment` with the `CreateAttachmentRequest` details AND the attachment raw bytes (decrypted), and then write the SDK in a way that it orchestrates the multiple API requests to create the attachment, encrypt the bytes, and upload it to the file endpoint?

crates/bitwarden-vault/src/cipher/attachment_client/create.rs
```diff
@@ -0,0 +1,531 @@
...
+        response: bitwarden_api_api::models::AttachmentUploadDataResponseModel,
+        existing_cipher: Option<Cipher>,
+        cipher_id: CipherId,
+        as_admin: bool,
+    ) -> Result<CreatedAttachment, CipherCreateAttachmentError> {
+        let cipher = if as_admin {
+            let cipher_mini = response
+                .cipher_mini_response
+                .ok_or(MissingFieldError("cipher_mini_response"))?;
+            (*cipher_mini).merge_with_cipher(None)?
```
nikwithak: Should this be `merge_with_cipher(existing_cipher)`?

crates/bitwarden-vault/src/cipher/attachment_client/download_url.rs
```diff
@@ -0,0 +1,468 @@
...
+                    bitwarden_api_api::apis::Error::Response(content)
+                        if content.status == StatusCode::NOT_FOUND =>
+                    {
+                        CipherGetAttachmentDownloadUrlError::NotFound
+                    }
+                    other => other.into(),
+                })?;
+
+            return response.url.ok_or_else(|| MissingFieldError("url").into());
+        }
```
nikwithak, line 53: 🎨 Consider making this its own unique function

crates/bitwarden-vault/src/cipher/attachment_client/upgrade.rs
```diff
@@ -0,0 +1,852 @@
...
+                atts.iter()
+                    .find(|a| a.id.as_deref() == Some(&attachment_id))
+            })
+            .ok_or(CipherUpgradeAttachmentError::NotFound)?;
+
+        if attachment.key.is_some() {
+            return Err(CipherUpgradeAttachmentError::AlreadyUpgraded);
+        }
+
+        // Encrypted size stays the same; both formats use the same AES-CBC-HMAC wire layout.
```
nikwithak: Is this guaranteed? If anything changes about the encryption implementaiton/algo in the future this won't necessarily be true, correct?

crates/bitwarden-vault/src/cipher/attachment_client/upgrade.rs
```diff
@@ -0,0 +1,852 @@
...
+            .map_err(DecryptError::from)?;
+        let attachment_view = cipher_view
+            .attachments
+            .as_ref()
+            .and_then(|atts| {
+                atts.iter()
+                    .find(|a| a.id.as_deref() == Some(&attachment_id))
+            })
+            .cloned()
+            .ok_or(CipherUpgradeAttachmentError::NotFound)?;
```
nikwithak: :art: `Attachment` implements `Decryptable` on its own, so you could call `self.key_store.decrypt(&attachment)` without needing to decrypt the full cipher.

crates/bitwarden-vault/src/cipher/attachment_client/upgrade.rs
```diff
@@ -0,0 +1,852 @@
...
+        plaintext_size_hint: u64,
+        download_url: &str,
+        created: &CreatedAttachment,
+    ) -> Result<(), CipherUpgradeAttachmentError> {
+        let response = self
+            .http_client
+            .get(download_url)
+            .send()
+            .await
+            .map_err(|_| CipherUpgradeAttachmentError::Download)?;
```
nikwithak, line 200: Is it worthwhile to pass through the `reqwest::Error` in this case?

nikwithak (changes requested):

crates/bitwarden-vault/src/cipher/attachment_client/create.rs
gbubemismith: 🤔 I think our attachment Ids are not like UUID's e.g kb7g6gztrgfdy6f57o4396bafcuzh0kw is the attachment id

``` json
{"kb7g6gztrgfdy6f57o4396bafcuzh0kw":{"Size":"5345633","FileName":"","Key":"","ContainerName":"attachments-v2","Validated":false,"TempMetadata":null}}
```

crates/bitwarden-vault/src/cipher/attachment_client/create.rs
gbubemismith: That's a direction I wanted to go , but I realized moving the upload into the SDK means reworking the client's upload‑progress reporting. I'll rather create a new ticket for that work

crates/bitwarden-vault/src/cipher/attachment_client/create.rs
gbubemismith: It's intentional here, in the admin branch `existing_cipher` is None. I'll add a comment to clarify it 👍🏼

crates/bitwarden-vault/src/cipher/attachment_client/create.rs
gbubemismith: Good point 👍🏼

crates/bitwarden-vault/src/cipher/attachment_client/upgrade.rs
gbubemismith: I noticed `reqwest::Error` carries the presigned URL, I didn't want to risk leaking that

crates/bitwarden-vault/src/cipher/attachment_client/create.rs
gbubemismith: Fixed here f31d69d928e88c68441cc14b8e48069225271ad0

crates/bitwarden-vault/src/cipher/attachment_client/download_url.rs
gbubemismith: Fixed here f31d69d928e88c68441cc14b8e48069225271ad0

crates/bitwarden-vault/src/cipher/attachment_client/upgrade.rs
gbubemismith: Fixed here f31d69d928e88c68441cc14b8e48069225271ad0

crates/bitwarden-vault/src/cipher/attachment_client/upgrade.rs
```diff
@@ -0,0 +1,880 @@
...
+            {
+                tracing::warn!(
+                    "failed to roll back orphaned attachment slot {} on cipher {cipher_id}: {rollback_err:?}",
+                    created.attachment_id,
+                );
+            }
+            return Err(e);
+        }
+
+        let upgraded_cipher = self.delete_attachment(cipher_id, attachment_id).await?;
```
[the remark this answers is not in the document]
gbubemismith: Good catch, but I think this is fine to leave as-is. A failed legacy delete isn't silently lost,  the cipher still reports `hasOldAttachments`. Re-running the upgrade is user-driven from the attachment UI.

crates/bitwarden-vault/src/cipher/attachment_client/create.rs
nikwithak: Looks like this one was missed: https://github.com/bitwarden/sdk-internal/pull/1072/changes#diff-c266e89834c4c37facb8dc19a3b216546c47c1dd1cc65b18d7da9ba0c8ab3ed9R112

crates/bitwarden-vault/src/cipher/attachment_client/create.rs
nikwithak: I think it owuld better to just blindly pass the parameter in this case - since it's a helper function that takes the `Option<Cipher>` as an input already, the logic to decide `Some(_)` vs `None` is already handled by the caller, here: https://github.com/bitwarden/sdk-internal/pull/1072/files/0bcc661ea2c20694b2a23f53b5a19a603a486395#diff-c266e89834c4c37facb8dc19a3b216546c47c1dd1cc65b18d7da9ba0c8ab3ed9

crates/bitwarden-vault/src/cipher/attachment_client/create.rs
gbubemismith: @nikwithak I intentionally left this as `cipher` because this also covers the admin path, where the org cipher may not be decryptable in the keystore. The client only uses it in the admin branch

crates/bitwarden-vault/src/cipher/attachment_client/create.rs
gbubemismith: Fixed 89f4e427b51f2ea693911488fcf9136a06438c08

nikwithak (dismissed):

dereknance (dismissed):
✅ Platform

crates/bitwarden-vault/src/cipher/attachment_client/upgrade.rs
```diff
@@ -0,0 +1,885 @@
...
+    /// Upgrades a legacy v1 attachment to `CipherKey(AttachmentKey(Contents))`.
+    ///
+    /// Downloads and re-encrypts the attachment, creates a new slot, uploads the
+    /// new bytes, then deletes the old attachment. If the upload fails, it tries
+    /// to delete the new slot before returning the error. Returns the decrypted
+    /// cipher view.
+    pub async fn upgrade_attachment(
+        &self,
+        cipher_id: CipherId,
+        attachment_id: String,
```
quexten, line 86: Nit/Question: Any reason not to use a newtype for attachmentid?

crates/bitwarden-vault/src/cipher/attachment_client/upgrade.rs
```diff
@@ -0,0 +1,885 @@
...
+
+#[cfg_attr(feature = "wasm", wasm_bindgen)]
+impl AttachmentsClient {
+    /// Upgrades a legacy v1 attachment to `CipherKey(AttachmentKey(Contents))`.
+    ///
+    /// Downloads and re-encrypts the attachment, creates a new slot, uploads the
+    /// new bytes, then deletes the old attachment. If the upload fails, it tries
+    /// to delete the new slot before returning the error. Returns the decrypted
+    /// cipher view.
+    pub async fn upgrade_attachment(
```
quexten, line 83: A thought (non-blocking): Why even offer the API at this level? You are having to expose error types. A simpler API would be: "upgrade_attachment(cipher_id)", or even "upgrade_attachments". That way there is less degrees of freedom to pass incorrect things to the public API of the SDK.

crates/bitwarden-vault/src/cipher/attachment_client/upgrade.rs
```diff
@@ -0,0 +1,885 @@
...
+            })
+            .ok_or(CipherUpgradeAttachmentError::NotFound)?;
+
+        if attachment.key.is_some() {
+            return Err(CipherUpgradeAttachmentError::AlreadyUpgraded);
+        }
+
+        // Used only to pre-size the encryptor buffer. The legacy encrypted size is
+        // a safe upper bound here.
+        let plaintext_size_hint: u64 = attachment
```
quexten, line 112: Note: Can be dropped once KM implements new encryption format for attachments.

crates/bitwarden-vault/src/cipher/attachment_client/upgrade.rs
```diff
@@ -0,0 +1,885 @@
...
+                Cipher::decrypt_cipher_key(&mut ctx, cipher.key_identifier(), &cipher.key)?;
+            attachment
+                .decrypt(&mut ctx, cipher_key)
+                .map_err(DecryptError::from)?
+                .file_name
+                .ok_or(MissingFieldError("file_name"))?
+        };
+
+        let download_url = self
+            .get_attachment_download_url(cipher_id, attachment_id.clone(), None)
```
quexten, line 130: Rust nit: If you make the params of your functions take refs, you can avoid all of the clones of these non-copy types.

crates/bitwarden-vault/src/cipher/attachment_client/upgrade.rs
```diff
@@ -0,0 +1,885 @@
...
+        &self,
+        cipher: &Cipher,
+        file_name_plain: &str,
+    ) -> Result<ReencryptionMaterial, CipherUpgradeAttachmentError> {
+        let mut ctx = self.key_store.context();
+
+        let cipher_key =
+            Cipher::decrypt_cipher_key(&mut ctx, cipher.key_identifier(), &cipher.key)?;
+
+        let new_attachment_key = SymmetricCryptoKey::make(SymmetricKeyAlgorithm::Aes256CbcHmac);
```
quexten: Can we please make a task to extract this logic out? I'm concerned that if we also implement regular upload creation, then the implementations will diverge. Generally, we only ever want one implementation of something, otherwise we get divergence bugs like with cipher keys.

crates/bitwarden-vault/src/cipher/attachment_client/upgrade.rs
```diff
@@ -0,0 +1,885 @@
...
+}
+
+impl<T> From<bitwarden_api_api::apis::Error<T>> for CipherUpgradeAttachmentError {
+    fn from(value: bitwarden_api_api::apis::Error<T>) -> Self {
+        Self::Api(value.into())
+    }
+}
+
+/// Data needed to re-encrypt an attachment
+struct ReencryptionMaterial {
```
quexten: As per the other comment, this should just be "AttachmentMaterial" or something, and then be used *both* for upgrade and creation.

crates/bitwarden-vault/src/cipher/attachment_client/upgrade.rs
```diff
@@ -0,0 +1,885 @@
...
+        let attachment = cipher
+            .attachments
+            .as_ref()
+            .and_then(|atts| {
+                atts.iter()
+                    .find(|a| a.id.as_deref() == Some(&attachment_id))
+            })
+            .ok_or(CipherUpgradeAttachmentError::NotFound)?;
+
+        if attachment.key.is_some() {
```
quexten: Suggestion: Implement something like:
```
impl Attachment {
  fn encryption_version() -> AttachmentEncryptionVersion
}

enum AttachmentEncryptionVersion -> {
  LegacyNoKey_V1,
  AttachmentKey_V2,
  FUTURE_VERSION_V3 // doesn't need to be added here but is forseeable.
}
```

Then, you can just "if matches!(attachment.encryption_version(), AttachmentEncryptionVersion::AttachmentKey_V2)", which is way easier to follow. As a new unfamiliar reader I would not know that no key = legacy version since it's not documented here.

crates/bitwarden-vault/src/cipher/attachment_client/upgrade.rs
```diff
@@ -0,0 +1,885 @@
...
+        // a safe upper bound here.
+        let plaintext_size_hint: u64 = attachment
+            .size
+            .as_ref()
+            .and_then(|s| s.parse().ok())
+            .ok_or(MissingFieldError("attachment.size"))?;
+
+        let file_name_plain = {
+            let mut ctx = self.key_store.context();
+            let cipher_key =
```
quexten, line 120: Thought (no action needed): We (vault or KM) should make helper functions to load a cipher key context into local context scope.

crates/bitwarden-vault/src/cipher/attachment_client/upgrade.rs
```diff
@@ -0,0 +1,885 @@
...
+        Ok(ReencryptionMaterial {
+            new_attachment_key,
+            wrapped_new_attachment_key,
+            encrypted_file_name,
+        })
+    }
+
+    /// Downloads the legacy ciphertext and re-encrypts it into memory.
+    ///
+    /// Download and decrypt still stream, but the encrypted output is buffered
```
quexten: This is not great. We made streaming decryption so that we don't have to buffer, and now we are buffering again. @gbubemismith is all you need here the total filesize? We can calculate it for you on the KM side and give it to you, based on the plaintext size. Does it need to be exact (byte accurate) or an estimate?

crates/bitwarden-vault/src/cipher/attachment_client/upgrade.rs
```diff
@@ -0,0 +1,885 @@
...
+    /// because `StreamingAttachmentEncryptor` needs the full payload before it
+    /// can write, and wasm `reqwest` only supports buffered request bodies.
+    async fn download_and_reencrypt(
+        &self,
+        legacy_key_slot: SymmetricKeySlotId,
+        new_attachment_key: SymmetricCryptoKey,
```
quexten, line 191: Why pass in one as a key slot and one as a key?

crates/bitwarden-vault/src/cipher/attachment_client/upgrade.rs
```diff
@@ -0,0 +1,885 @@
...
+            wrapped_new_attachment_key,
+            encrypted_file_name,
+        })
+    }
+
+    /// Downloads the legacy ciphertext and re-encrypts it into memory.
+    ///
+    /// Download and decrypt still stream, but the encrypted output is buffered
+    /// because `StreamingAttachmentEncryptor` needs the full payload before it
+    /// can write, and wasm `reqwest` only supports buffered request bodies.
```
quexten:
```suggestion
    /// Wasm `reqwest` only supports buffered request bodies so the output is buffered.
```

The implementation detail of StreamingAttachmentEncryptor *may* change to not buffer, and should not be documented here.

quexten (changes requested):
Apologies took a while to circle to this.

crates/bitwarden-vault/src/cipher/attachment_client/upgrade.rs
gbubemismith: The buffering here is the legacy format, download, and decrypt still stream, but the re-encrypted output gets buffered because `AES-CBC-HMAC` prefixes the HMAC, so the encryptor has to materialize the entire ciphertext before emitting it. And since streamed uploads still have browser gaps, I decided not to follow through on the streaming re-encrypt path for now and just buffer

quexten: > The buffering here is the legacy format, download, and decrypt still stream, but the re-encrypted output gets buffered because AES-CBC-HMAC prefixes the HMAC, so the encryptor has to materialize the entire ciphertext before emitting it. 

Yes, I know, I implemented this >.<

However, my point is that this is implementation detail of the encryptor, and vault code does not do the first buffering. The encryptor does. Since it is implementation detail of the encryptor, and may be changed by km, and is not even implemented here, this is not the correct place to place this comment. It should just be removed.

> And since streamed uploads still have browser gaps, I decided not to follow through on the streaming re-encrypt path for now and just buffer

That's fine, but this is a *separate* buffer from the first buffer. Assuming we do not accept the memory increase, we do have a way to circumvent it - the same external buffer approach used in the implementation of the encryptor - here: https://github.com/bitwarden/sdk-internal/blob/main/crates/bitwarden-crypto/src/stream/large_memory_buffer.rs

crates/bitwarden-vault/src/cipher/attachment_client/upgrade.rs
gbubemismith: Oh, attachment Ids are not UUID's,  so they can't use the `uuid_newtype!`

crates/bitwarden-vault/src/cipher/attachment_client/upgrade.rs
gbubemismith: For this I was trying to be consistent with the attachment methods (delete/create/renew) we expose through wasm as they all take owned

crates/bitwarden-vault/src/cipher/attachment_client/upgrade.rs
gbubemismith: `legacy_key_slot` already lives in the key store so we just pass it, for the `new_key` the context is gone by the time we re-encrypt, so its slot would be dead

crates/bitwarden-vault/src/cipher/attachment_client/upgrade.rs
gbubemismith: Very clean 🙌🏼

crates/bitwarden-vault/src/cipher/attachment_client/upgrade.rs
gbubemismith: This was more of a drop-in to the existing per-attachment path on the client side, so I kept the signature matching that. I'll create a ticket to revisit folding the loop into the SDK (upgrade_attachment(cipher_id) / upgrade_attachments) once this is merged and stabilized.

crates/bitwarden-vault/src/cipher/attachment_client/upgrade.rs
gbubemismith: Fixed a1f355f81130c450153ab16c77094cb160101e06

crates/bitwarden-vault/src/cipher/attachment_client/upgrade.rs
gbubemismith: Fixed a1f355f81130c450153ab16c77094cb160101e06

crates/bitwarden-vault/src/cipher/attachment_client/upgrade.rs
gbubemismith: Fixed a1f355f81130c450153ab16c77094cb160101e06

crates/bitwarden-vault/src/cipher/attachment_client/upgrade.rs
quexten: Ah, i think the correct way to frame this is: "This is an async function, and  so we cannot pass a key context in. The legacy key store is a persistent key slot, and thus can survive awaits, the key context that we would put a temporary key into would not survive the await, so we must pass a direct reference".

Thanks @gbubemismith my concern is resolved, but on the KM side, I think we are seeing more of these async workarounds, and KM probably needs to work on this.

crates/bitwarden-vault/src/cipher/attachment_client/upgrade.rs
```diff
@@ -0,0 +1,885 @@
...
+    use bitwarden_state::repository::Repository;
+    use bitwarden_test::MemoryRepository;
+
+    use super::*;
+    use crate::{Attachment, CipherRepromptType, CipherType};
+
+    const TEST_CIPHER_ID: &str = "5faa9684-c793-4a2d-8a12-b33900187097";
+    const OLD_ATTACHMENT_ID: &str = "uf7bkexzag04d3cw04jsbqqkbpbwhxs0";
+    const NEW_ATTACHMENT_ID: &str = "newatt9999999999999999999999999";
+    const TEST_CIPHER_NAME: &str = "2.pMS6/icTQABtulw52pq2lg==|XXbxKxDTh+mWiN1HjH2N1w==|Q6PkuT+KX/axrgN9ubD5Ajk2YNwxQkgs3WJM0S0wtG8=";
```
quexten, line 349: Thought: Could we record a real key + attachment buffer value, and add it as a test value here? That would catch a lot of regressions.

quexten (approved):

dereknance (approved):

nikwithak (approved):

PR #1075, opened 2026-05-12
[PM-33906] Add PureCrypto::new_guid

dereknance:
## 🎟️ Tracking

[PM-33906](https://bitwarden.atlassian.net/browse/PM-33906)

## 📔 Objective

Expose `PureCrypto::new_guid`

## 🚨 Breaking Changes

[PM-33906]: https://bitwarden.atlassian.net/browse/PM-33906?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1316,6 +1316,7 @@ dependencies = [
  "tracing-subscriber",
  "tracing-web",
  "tsify",
+ "uuid",
  "wasm-bindgen",
  "wasm-bindgen-futures",
 ]
--- a/crates/bitwarden-wasm-internal/Cargo.toml
+++ b/crates/bitwarden-wasm-internal/Cargo.toml
@@ -53,6 +53,7 @@ tracing = { workspace = true }
 tracing-subscriber = { workspace = true }
 tracing-web = { version = "0.1.3" }
 tsify = { workspace = true }
+uuid = { workspace = true }
 wasm-bindgen = { workspace = true }
 wasm-bindgen-futures = { workspace = true }
 
--- a/crates/bitwarden-wasm-internal/src/pure_crypto.rs
+++ b/crates/bitwarden-wasm-internal/src/pure_crypto.rs
@@ -419,6 +419,12 @@ impl PureCrypto {
         let mut rng = rand::rng();
         rng.random_range(min..=max)
     }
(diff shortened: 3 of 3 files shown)


djsmith85 (approved):
Changes are looking great!

quexten (approved):
From the code directly this looks good. However, PureCrypto is a temporary shim that is intended to get crypto into the SDK quicker, without having the high level features ported yet. KM does not intend to keep it long term and will remove it as soon as the consumers are moved off to higher level clients (like vault's crypto client).

If this is supposed to live in the SDK long-term, and be a long-term API, then KM would actually recommend that Platform makes a client, and makes a static function impl on the client.

Something like:
```rust
// file owned by platform
struct GuidClient {
}

impl GuidClient {
  pub fn new_guid() -> ... {
  }
}
```

then you can call it from typescript:
```ts
await SdkLoadService.Ready;
GuidClient.new_guid(); 
```


I'm also not quite sure if guid is actually a crypto function, it seems one layer above crypto.

From the epic, it seems platform intends to not replace the Utils.newGuid function implementation with a PureCrypto call, but to instead directly spread the PureCrypto calls through domain code, which is not a supported pattern. PureCrypto *MUST* be limited to a single typescript shim that is easy to drop later.

> /// This module represents a stopgap solution to provide access to primitive crypto functions for JS
/// clients. It is not intended to be used outside of the JS clients and this pattern should not be
/// proliferated. It is necessary because we want to use SDK crypto prior to the SDK being fully
/// responsible for state and keys.


If you make a GuidClient that is platform owned like above, you can directly use it, however you have to await the SdkLoadService ready promise.


The much better solution of course is to move the high level caller into the SDK.

quexten (changes requested):
Actually, setting the review to blocking and discussing with KM.

quexten (approved):
Okay, after reading the epic, while I do have my reservations, as per @Thomas-Avery comment here, this is accepted by KM as a temporary state, so approving: https://bitwarden.atlassian.net/browse/PM-33906?focusedCommentId=147877

PR #1076, opened 2026-05-13
[PM-32695] Use blob cipher encryption in CiphersClient encrypt functions

nikwithak:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-33107

## 📔 Objective

This adds support for Cipher Blob Encryption to the encrypt functionality on `CiphersClient`, including when handling API requests.

Note that these are breaking / incompatible changes until (a) the `decrypt` functions are also handled [PM-32696](https://bitwarden.atlassian.net/browse/PM-32696), and (b) changes are needed in `server` to support these changes (server is currently crashing when local testing).

## 🚨 Breaking Changes

[PM-32696]: https://bitwarden.atlassian.net/browse/PM-32696?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


crates/bitwarden-vault/src/cipher/cipher_client/create.rs
```diff
@@ -153,47 +123,61 @@ impl CiphersClient {
...
-        // be moved directly into the CompositeEncryptable implementation.
-        if self
-            .client
-            .internal
-            .get_flags()
-            .await
-            .enable_cipher_key_encryption
-        {
-            let key = view.key_identifier();
-            view.generate_cipher_key(&mut key_store.context(), key)?;
```
nikwithak, line 178: Now inside `CiphersClient::encrypt()`'s existing impl on L135

crates/bitwarden-vault/src/cipher/cipher_client/edit.rs
```diff
@@ -32,6 +28,8 @@ pub enum EditCipherError {
     #[error(transparent)]
     Crypto(#[from] CryptoError),
     #[error(transparent)]
+    BlobEncryption(#[from] crate::blob::BlobEncryptionError),
```
nikwithak, line 31: Note: Since this isn't directly translatable to `CryptoError`, it will prevent us from putting the BlobEncryption / BlobDecryption logic directly in the `Encryptable`/`Decryptable` implementations.

PR #1077, opened 2026-05-13
chore: add description to policies and server-communication-config

addisonbeck:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-37252

## 📔 Objective

`bitwarden-policies` and `bitwarden-server-communication-config` are part of the `bitwarden-core` release set. The `description` field is required by crates.io for publishing. Both crates are internal and not intended for direct use, so the description follows the established convention for internal crates.


djsmith85 (approved):
Changes are looking good. No idea why the workflows would be failing and I assume BRE is working on them.

jrmccannon (approved):

PR #1078, opened 2026-05-13
[PM-37481] fix: sdk-swift release workflow fails to commit 

vvolkgang:
## 🎟️ Tracking

PM-37481

## 📔 Objective

Context:

1. https://github.com/bitwarden/sdk-internal/pull/926/changes 
2. https://github.com/bitwarden/sdk-internal/pull/1012  

In order to restore iOS SDK updates, we're partially reverting the api-commit step, without introducing GPG commit signing given that the key was recently deleted. We'll be able to re-introduce the api-commit action when sub-folder support is introduced. 

Test Runs:

1. https://github.com/bitwarden/sdk-internal/actions/runs/25829532270
    * Created: https://github.com/bitwarden/ios/pull/2646


AmyLGalles (approved):
LGTM

PR #1081, opened 2026-05-14
[PM-37488] Fix state bridge and add integration tests

quexten:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-37488

## 📔 Objective

The state bridge return types were compiling correctly, but wrong at runtime. This is a limitation to wasm-bindgen that caused failures when used in the clients. Instead, we return `JsValue` here, changing nothing about the TS interface.

To detect issues that cross the FFI boundary, this PR adds an integration test module, that initializes a basic account and enrolls it in PIN. This verifies that the state bridge is working correctly across the FFI boundary.

The hope is that we can expand the integration tests to cover more issues that would otherwise only be caught when using in clients.

### DIFF
--- a/crates/bitwarden-wasm-internal/integration-tests/package.json
+++ b/crates/bitwarden-wasm-internal/integration-tests/package.json
@@ -1,0 +1,19 @@
+{
+  "name": "sdk-internal-integration-tests",
+  "version": "0.0.0",
+  "private": true,
+  "type": "module",
+  "scripts": {
+    "test": "NODE_OPTIONS=--experimental-vm-modules jest"
+  },
+  "dependencies": {
+    "@bitwarden/sdk-internal": "file:../npm/"
+  },
+  "devDependencies": {
+    "@types/jest": "^29.5.0",
+    "@types/node": "^22.0.0",
+    "jest": "^29.7.0",
+    "ts-jest": "^29.1.0",
+    "typescript": "^5.6.0"
+  }
+}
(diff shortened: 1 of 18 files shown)


crates/bitwarden-wasm-internal/integration-tests/pin-lock.test.ts
```diff
@@ -0,0 +1,137 @@
...
+    get_masterpassword_unlock_data: async () => masterPasswordUnlockData,
+    clear_masterpassword_unlock_data: async () => void (masterPasswordUnlockData = undefined),
+  };
+}
+
+const TEST_EMAIL = "test@bitwarden.com";
+const TEST_PASSWORD = "asdfasdfasdf";
+const TEST_PIN = "1234";
+
+const PRIVATE_KEY =
```
quexten: Note: It probably makes sense to extract these to a helper file that all integration tests can use. However, given that this is the first and only use so far, I'm not doing that here.

dani-garcia (approved):
LGTM, love the integration tests!

crates/bitwarden-wasm-internal/integration-tests/package.json
```diff
@@ -0,0 +1,19 @@
...
+  "private": true,
+  "type": "module",
+  "scripts": {
+    "test": "NODE_OPTIONS=--experimental-vm-modules jest"
+  },
+  "dependencies": {
+    "@bitwarden/sdk-internal": "file:../npm/"
+  },
+  "devDependencies": {
+    "@types/jest": "^29.5.0",
```
mzieniukbw, line 13: I think this needs to be pinned.

mzieniukbw (approved):

PR #1084, opened 2026-05-14
[PM-18786] Include passkeys and password history in exports and adjust timestamps to match web export

harr1424:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-18786

## 📔 Objective

When exporting a vault from iOS, there are discrepancies compared to the web export. Specifically:
- fido2Credentials array is empty in iOS exports but populated in web exports.
- Password history is always null in iOS exports.
- File size difference: iOS export is 10.8MB, while web export is 11.1MB, which suggests missing or altered data.
- Timestamps inconsistency: revisionDate and creationDate fields have slight differences in their fractional seconds.

This PR introduces code changes to achieve parity with the web vault export, as described above.


crates/bitwarden-exporters/src/json.rs
```diff
@@ -1,11 +1,30 @@
...
+/// Serialize an optional `DateTime<Utc>` with millisecond precision when present.
+fn serialize_opt_date_millis<S: Serializer>(
+    date: &Option<DateTime<Utc>>,
+    s: S,
+) -> Result<S::Ok, S::Error> {
+    match date {
+        Some(d) => s.serialize_str(&d.to_rfc3339_opts(SecondsFormat::Millis, true)),
+        None => s.serialize_none(),
+    }
+}
```
Hinton, line 26: **suggestion:** You can use https://docs.rs/chrono/latest/chrono/serde/ts_milliseconds/index.html which we already import.

harr1424: ❓ I don't think that the `chrono` method will be equivalent to `to_rfc3339_opts(SecondsFormat::Millis, ...)`

One of the acceptance criteria for this story was aligning the mobile export timestamp to match that of the web vault: 

```
// sample revisionDate from a web export
      "revisionDate": "2026-05-14T14:04:58.756Z"
```

https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=9bbea70e5e9f4b59723ef2b2d1a1ac80

Hinton: Ah that's true. FYI it seems we may be switching date time lib to jiff in the near future.

Calling it `rfc3339` or `rfc3339_millis` might be more clear what the format is.

harr1424: Agreed, I've renamed the functions ✅

itsadrago (dismissed):

crates/bitwarden-exporters/src/json.rs
Hinton: For consistency with other serializers in bitwarden-exporters it should be named `rfc3339_millis_serialize`.

Any reason there are two millis in the name?

dani-garcia (approved):

PR #1086, opened 2026-05-14
feat(bitwarden-core): introduce Client::save/load_from_state

addisonbeck:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-31879

## 📔 Objective

Adds `save_to_state` and `load_from_state` to `Client` in `bitwarden-core`, enabling the CLI to reconstruct a locked client from a persistent `StateRegistry` without re-authenticating.

`PasswordManagerClient` exposes thin delegation wrappers for `load_from_state` and `save_to_state` for callers working at the PM layer.

Some supporting types were added in `crates/bitwarden-core/src/client/rehydration.rs`: `RehydrationError` and `SaveStateData`.

All foundation (`StateRegistry`, setting keys, `ClientBuilder::with_state`, crypto persistence) was committed to main in prior PRs (#871, #917, #906, #912, #957, #965, #983, #973). This PR adds the rehydration entry points on top of that infrastructure.

### DIFF
--- a/crates/bitwarden-core/src/client/rehydration.rs
+++ b/crates/bitwarden-core/src/client/rehydration.rs
@@ -22,9 +22,6 @@ pub enum RehydrationError {
     /// An error occurred accessing or updating a setting in the state registry.
     #[error("State access error: {0}")]
     State(#[from] bitwarden_state::SettingsError),
-    /// A cryptographic operation failed.
-    #[error("Cryptographic operation failed during rehydration")]
-    Crypto,
     /// The client already has a user ID set that conflicts with the rehydrated user ID.
     #[error("User ID conflict during rehydration")]
     UserIdAlreadySet,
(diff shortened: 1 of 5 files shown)


crates/bitwarden-core/src/client/rehydration.rs
```diff
@@ -0,0 +1,305 @@
...
+            .with_settings(settings)
+            .with_token_handler(token_handler)
+            .with_state(registry)
+            .build();
+
+        client
+            .internal
+            .init_user_id(user_id)
+            .await
+            .map_err(|_| RehydrationError::UserIdAlreadySet)?;
```
dani-garcia: Now that I think about it, we could probably get away with ignoring this error: we just create the client and know no one has initialized this first, so the error should never happen.

dani-garcia (dismissed):
LGTM, one small suggestion that could simplify the error a bit, but not required.

crates/bitwarden-core/src/client/rehydration.rs
addisonbeck: Dropped this catch on https://github.com/bitwarden/sdk-internal/pull/1086/commits/e2053bf981905fd9bf65986c1601f2f475858f8a

dani-garcia (approved):

PR #1088, opened 2026-05-15
[PM-31059] Fix pin state unlock and add tests

quexten:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-31059

## 📔 Objective

During adding more integration tests I uncovered that the PinState regressed and is no longer working. This PR fixes it, and adds unit tests in rust, along with integration tests in TS, to ensure the unlock flows will not regress.

### DIFF
--- a/crates/bitwarden-wasm-internal/integration-tests/tests/utils.ts
+++ b/crates/bitwarden-wasm-internal/integration-tests/tests/utils.ts
@@ -122,11 +122,23 @@ export function initializeCryptoDefault(client: PasswordManagerClient) {
     kdfParams: TEST_KDF_PARAMS,
     email: TEST_EMAIL,
     accountCryptographicState: { V1: { private_key: encstring(PRIVATE_KEY) } },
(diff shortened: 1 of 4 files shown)


quexten:
Macos failures are unrelated to the PR.

crates/bitwarden-wasm-internal/integration-tests/tests/utils.ts
```diff
@@ -109,24 +110,51 @@ export async function makeInitializedPasswordmanagerClient(
...
+          kdf: TEST_KDF_PARAMS,
         },
       },
     },
-  };
+  });
+}
 
-  await client.crypto().initialize_user_crypto(req);
+export function initializeUserCrypto(
```
eligrubb, line 138: Nit, non-blocking: this helper hard codes V1 crypto state, should that be reflected in the method call? `initializeCryptoDefault` has a similar problem but the `Default` kind of implies v1.

Could also be a problem for when more v2 integration tests are added. The new tests are super cool, plus if mocking server APIs is solved... 🚀

eligrubb (approved):
One non-blocking comment. Looks good!

crates/bitwarden-wasm-internal/integration-tests/tests/utils.ts
quexten: I agree, we should add V2 support, I think It's out of scope for this first iteration, but we should definitely add it as a configurable option for test writers.

coroiu (approved):

PR #1091, opened 2026-05-15
[PM-37616] Update reqwest to 0.13

dani-garcia:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-37616

## 📔 Objective
Upgrade the HTTP client stack:

- `reqwest`: 0.12.28 → 0.13.3
- `reqwest-middleware`: 0.4.2 → 0.5.1 (to match the `reqwest` dependency)
- 
### Notable changes

- **Reqwest feature rename**: 0.13 dropped `rustls-tls-manual-roots`. Switched to `rustls-no-provider`, which activates reqwest's rustls connector + `rustls-platform-verifier` without forcing reqwest's new aws-lc-rs default. We continue to provide the `ClientConfig` ourselves via `tls_backend_preconfigured`.
- **Crypto provider stays on ring**: explicitly enable `std, tls12, logging, ring` on `rustls` (its previous defaults minus `aws_lc_rs`, which used to be enabled by reqwest for us).
- **Reqwest-middleware features**: added `form` and `query`, which 0.13 stopped enabling by default but which we use.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1577,12 +1577,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "9330f8b2ff13f34540b44e946ef35111825727b38d33286ef986142615121801"
 
 [[package]]
-name = "cfg_aliases"
-version = "0.2.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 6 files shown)


rr-bw (dismissed):

dereknance (approved):

rr-bw:
@dani-garcia Looks like I'm not needed for review here after the rebase 👍.

PR #1093, opened 2026-05-15
[PM-28192] Stream legacy attachment upgrade to bound WASM heap usage

gbubemismith:
## 🎟️ Tracking
https://bitwarden.atlassian.net/browse/PM-28192

## 📔 Objective
Replace the buffered legacy-attachment upgrade flow with a chunked streaming pipeline so WASM heap usage stays bounded by chunk size regardless of attachment size.


crates/bitwarden-crypto/src/safe/streaming_attachment_cipher.rs
```diff
@@ -0,0 +1,72 @@
...
+/// Streaming AES-256-CBC + HMAC-SHA256 decryptor.
+pub struct StreamingAttachmentDecryptor(StreamingAes256CbcHmacDecryptor);
+
+impl StreamingAttachmentDecryptor {
+    pub(crate) fn new(inner: StreamingAes256CbcHmacDecryptor) -> Self {
+        Self(inner)
+    }
+
+    /// Feeds a ciphertext chunk; returns any plaintext bytes ready to emit (unauthenticated
+    /// until [`Self::finalize`] returns `Ok`).
```
quexten:
```suggestion
    /// Feeds a ciphertext chunk; returns any plaintext bytes ready to emit (WARNING: unauthenticated
    /// until [`Self::finalize`] returns `Ok`. You must fully read the decrypted stream and discard the result if [`Self::finalize`] does not return `Ok`
```

We must add a bigger warning here since this interface is really dangerous.

KM should spend some time to implement a proper encryption scheme that is better suited for streaming decryption, that makes this impossible.

crates/bitwarden-crypto/src/safe/streaming_attachment_cipher.rs
```diff
@@ -0,0 +1,72 @@
...
+
+use crate::{
+    CryptoError,
+    aes::{
+        StreamingAes256CbcHmacDecryptor, StreamingAes256CbcHmacEncryptor, StreamingEncryptFinal,
+    },
+};
+
+/// IV + MAC + trailing ciphertext produced by [`StreamingAttachmentEncryptor::finalize`].
+pub struct StreamingAttachmentFinal {
```
quexten: I don't think this should live in the `safe` module. We have a `safe` primitive planned for streaming file encryption, but this encryptor is not `safe`.

Namely, from `Readme.md`:

When adding a new primitive there are a few considerations to make:

- Does this serve a new purpose that is not already fulfilled better otherwise
  - Replacing insecure functionality is a valid reason here
- Is it easy to use for developers that are not cryptography experts
- Does the API prevent (accidental) mis-use by developers that are not cryptography experts
  - Avoid exposing many settings/configurations to API consumers
    ^ This it does not fulfill
- Is the format extensible and cover the use-case adequately
- Does the new object have adequate security analysis performed?
  ^ This it does not fulfill
- Do we have cryptographic modularity?
  - That is, can we switch to new primitives / algorithms easily?
  ^ This it does not fulfill

Further, each new object should be validated against existing known attack classes.

Cryptographic namespaces are also not provided.
The correct safe primitive would be the `FileBuffer` described here https://bitwarden.atlassian.net/wiki/spaces/~712020c1551f20895b498884b767b55a92b68d/pages/2562457606/Safe+Crypto+Module+-+Technical+Vision#Module-Catalogue, but we have not implemented it yet.

I would suggest to just move the streaming attachment decryption to `bitwarden-crypto/src/`

crates/bitwarden-crypto/src/store/context.rs
```diff
@@ -537,6 +537,37 @@ impl<Ids: KeySlotIds> KeyStoreContext<'_, Ids> {
         self.get_symmetric_key(key_id)
     }
 
+    /// Builds a streaming AES-256-CBC + HMAC-SHA256 decryptor for the
+    /// [`Aes256CbcHmacKey`] at `key_id`. See [`crate::safe::StreamingAttachmentDecryptor`]
+    /// for the unverified-plaintext caveat.
+    pub fn streaming_decrypt_aes_cbc_hmac(
```
quexten: This should not live on the context. The relationship should be inverted and you should rather have `StreamingAttachmentDecryptor::new(&filereader, &key_id, &ctx)` and `SteramingAttachmentEncryptor::new(&filewriter, &key_id, &ctx)

crates/bitwarden-crypto/src/aes.rs
```diff
@@ -93,6 +99,138 @@ fn encrypt_aes256_internal(
...
+/// authenticated** and must be treated as untrusted until `finalize` returns `Ok`.
+pub(crate) struct StreamingAes256CbcHmacDecryptor {
+    decryptor: cbc::Decryptor<aes::Aes256>,
+    hmac: PbkdfSha256Hmac,
+    /// Always retains at least one full block so `finalize` can apply PKCS7 unpadding.
+    pending: Vec<u8>,
+}
+
+impl StreamingAes256CbcHmacDecryptor {
+    pub(crate) fn new(iv: &[u8; 16], mac_key: &Array<u8, U32>, enc_key: &Array<u8, U32>) -> Self {
```
quexten:
```suggestion
    pub(crate) fn new(iv: &[u8; 16], key: &Aes256CbcHmacKey) -> Self {
```

Passing mac key and enc key is not something we should do. Please pass `Aes256CbcHmacKey` instead

crates/bitwarden-crypto/src/keys/symmetric_crypto_key.rs
```diff
@@ -124,6 +124,19 @@ impl PartialEq for Aes256CbcHmacKey {
     }
 }
 
+impl Aes256CbcHmacKey {
+    pub(crate) fn streaming_decryptor(
```
quexten: The relationship should be inverted. The decryptor should implement new, taking a key. The key should not have making a decryptor as a public interface.

quexten (changes requested):

crates/bitwarden-crypto/src/store/context.rs
```diff
@@ -537,6 +537,18 @@ impl<Ids: KeySlotIds> KeyStoreContext<'_, Ids> {
...
+    pub(crate) fn get_aes256_cbc_hmac_key(
+        &self,
+        key_id: Ids::Symmetric,
+    ) -> Result<&crate::keys::Aes256CbcHmacKey> {
+        match self.get_symmetric_key(key_id)? {
+            SymmetricCryptoKey::Aes256CbcHmacKey(key) => Ok(key),
+            _ => Err(CryptoError::InvalidKey),
+        }
+    }
+
```
quexten, line 551:
```suggestion

```

This feels like a weird one-off API and should be done by the caller instead by using `get_symmetric_key` + match.

crates/bitwarden-vault/src/cipher/attachment_client/upgrade.rs
```diff
@@ -111,107 +110,227 @@ pub async fn prepare_attachment_upgrade<R: Repository<Cipher> + ?Sized>(
...
+/// them).
+///
+/// WASM heap usage is bounded by chunk size — the streaming primitives never hold the
+/// full payload in linear memory.
+#[cfg(feature = "wasm")]
+pub async fn prepare_attachment_upgrade<R: Repository<Cipher> + ?Sized>(
+    cipher_id: CipherId,
+    attachment_id: &str,
+    reader: AttachmentByteReader,
+    writer: AttachmentByteWriter,
```
quexten, line 140:
```suggestion
    legacy_format_attachment_reader: AttachmentByteReader,
    new_format_attachment_writer: AttachmentByteWriter,
```

TBH, it would make more sense to have an API that just takes *any* attachment (not just legacy formatted ones), downloads, decrypts, reencrypts and uploads them. It is forseeable that we'll switch formats again, so it feels weird to couple to specific versions here.

crates/bitwarden-vault/src/cipher/attachment_client/upgrade.rs
```diff
@@ -111,107 +110,227 @@ pub async fn prepare_attachment_upgrade<R: Repository<Cipher> + ?Sized>(
...
-        .map_err(|_| CipherPrepareAttachmentUpgradeError::Download)?;
+    Ok((cipher, file_name))
+}
 
-    if !response.status().is_success() {
-        return Err(CipherPrepareAttachmentUpgradeError::Download);
+/// Prepares a legacy attachment for upgrade to the per-attachment-key format using a
+/// chunked streaming pipeline. The host feeds the entire downloaded body — including
+/// the `encType (1) || IV (16) || MAC (32)` header — to `reader`; the SDK parses the
+/// 49-byte header internally and pushes re-encrypted body chunks to `writer`.
```
quexten, line 126:
```suggestion
/// chunked streaming pipeline. It reads the body from the reader, decrypts it, re-encrypts it and feeds it into the writer.
```

quexten (changes requested):
I'll circle back around to this with a concrete proposal for the KM-side API for this. Currently it exposes too many low level crypto details that we don't want to be public API.

crates/bitwarden-crypto/src/streaming_attachment_cipher.rs
```diff
@@ -0,0 +1,148 @@
...
+#![allow(async_fn_in_trait)]
+
+use crate::{
+    CryptoError, KeyStoreContext,
+    aes::{StreamingAes256CbcHmacDecryptor, StreamingAes256CbcHmacEncryptor},
+    traits::KeySlotIds,
+};
+
+/// Host-supplied source of ciphertext chunks consumed by [`StreamingAttachmentDecryptor`].
+pub trait ChunkReader {
```
quexten, line 19: This should use AsyncRead and AsyncWrite instead of custom traits. that way your entire vault code could be one line using: https://dtantsur.github.io/rust-openstack/tokio/io/fn.copy.html and is standard / can use standard functions

crates/bitwarden-crypto/src/streaming_attachment_cipher.rs
```diff
@@ -0,0 +1,148 @@
...
+
+impl<R: ChunkReader> StreamingAttachmentDecryptor<R> {
+    /// Resolves the [`crate::Aes256CbcHmacKey`] at `key_id` from `ctx` and primes the
+    /// decryptor with `iv`. Returns [`CryptoError::InvalidKey`] when the slot holds a key
+    /// of a different type.
+    pub fn new<Ids: KeySlotIds>(
+        reader: R,
+        key_id: Ids::Symmetric,
+        ctx: &KeyStoreContext<'_, Ids>,
+        iv: &[u8; 16],
```
quexten, line 59: Iv should never be passed in externally.

crates/bitwarden-vault/src/cipher/attachment_client/upgrade.rs
```diff
@@ -57,49 +62,43 @@ impl<T> From<bitwarden_api_api::apis::Error<T>> for CipherPrepareAttachmentUpgra
...
     pub file_upload_type: FileUploadType,
-    /// Encrypted file name in `EncString` format.
+    /// EncString-format encrypted file name.
     pub encrypted_file_name: String,
-    /// Encrypted file bytes ready to push to `upload_url`.
-    pub encrypted_contents: Vec<u8>,
+    /// 16-byte AES-CBC IV.
+    #[cfg_attr(feature = "wasm", serde(with = "serde_bytes"))]
+    #[cfg_attr(feature = "wasm", tsify(type = "Uint8Array"))]
+    pub iv: Vec<u8>,
```
quexten, line 83: Crypto internal detail that leaks out of the crypto crate.
```suggestion
```

crates/bitwarden-vault/src/cipher/attachment_client/upgrade.rs
```diff
@@ -57,49 +62,43 @@ impl<T> From<bitwarden_api_api::apis::Error<T>> for CipherPrepareAttachmentUpgra
...
+    /// 32-byte HMAC-SHA256 tag over IV + ciphertext.
+    #[cfg_attr(feature = "wasm", serde(with = "serde_bytes"))]
+    #[cfg_attr(feature = "wasm", tsify(type = "Uint8Array"))]
+    pub mac: Vec<u8>,
```
quexten, line 87: Crypto internal detail that leaks out of the crypto crate.
```suggestion
```

crates/bitwarden-vault/src/cipher/attachment_client/upgrade.rs
```diff
@@ -111,107 +110,227 @@ pub async fn prepare_attachment_upgrade<R: Repository<Cipher> + ?Sized>(
...
+    key_store: &KeyStore<KeySlotIds>,
+) -> Result<AttachmentUpgrade, CipherPrepareAttachmentUpgradeError> {
+    let (cipher, decrypted_file_name) =
+        lookup_upgrade_target(cipher_id, attachment_id, repository, key_store).await?;
+
+    let LegacyHeader {
+        iv: legacy_iv,
+        mac: legacy_mac,
+        leftover,
+    } = read_legacy_header(&reader).await?;
```
quexten, line 152: Crypto internal detail that leaks out of the crypto crate.
```suggestion
```

crates/bitwarden-vault/src/cipher/attachment_client/upgrade.rs
```diff
@@ -111,107 +110,227 @@ pub async fn prepare_attachment_upgrade<R: Repository<Cipher> + ?Sized>(
...
-        .as_ref()
-        .ok_or(MissingFieldError("file_name"))?
-        .to_string();
-
-    let last_known_revision_date = cipher
-        .revision_date
-        .to_rfc3339_opts(SecondsFormat::Millis, true);
-
+    // encType (1) || IV (16) || MAC (32) — the header the caller prepends before upload.
+    const ENCSTRING_HEADER_BYTES: usize = 1 + 16 + 32;
```
quexten, line 185: Crypto internal detail that leaks out of the crypto crate.
```suggestion
```

crates/bitwarden-vault/src/cipher/attachment_client/upgrade.rs
```diff
@@ -111,107 +110,227 @@ pub async fn prepare_attachment_upgrade<R: Repository<Cipher> + ?Sized>(
...
+        mac: final_.mac.to_vec(),
+    })
+}
+
+/// Legacy attachment wire format: `encType (1) || IV (16) || MAC (32)` then payload.
+#[cfg(feature = "wasm")]
+const LEGACY_HEADER_BYTES: usize = 1 + 16 + 32;
+/// `EncString::Aes256Cbc_HmacSha256_B64` — the only encryption type the upgrade flow accepts.
+#[cfg(feature = "wasm")]
+const LEGACY_ATTACHMENT_ENC_TYPE: u8 = 2;
```
quexten, line 212: Crypto internal detail that leaks out of the crypto crate.
```suggestion
```

crates/bitwarden-vault/src/cipher/attachment_client/upgrade.rs
```diff
@@ -111,107 +110,227 @@ pub async fn prepare_attachment_upgrade<R: Repository<Cipher> + ?Sized>(
...
 
-#[cfg_attr(feature = "wasm", wasm_bindgen)]
+#[cfg(feature = "wasm")]
+struct LegacyHeader {
+    iv: [u8; 16],
+    mac: [u8; 32],
+    /// Payload bytes that arrived in the same chunk as the header; must be fed to the
+    /// decryptor before any further reads.
+    leftover: Vec<u8>,
+}
```
quexten, line 284: Crypto internal detail that leaks out of the crypto crate.
```suggestion
```

crates/bitwarden-vault/src/cipher/attachment_client/upgrade.rs
```diff
@@ -111,107 +110,227 @@ pub async fn prepare_attachment_upgrade<R: Repository<Cipher> + ?Sized>(
...
+
+    if buf[0] != LEGACY_ATTACHMENT_ENC_TYPE {
+        return Err(CipherPrepareAttachmentUpgradeError::InvalidLegacyHeader);
+    }
+    let iv: [u8; 16] = buf[1..17].try_into().expect("slice of length 16");
+    let mac: [u8; 32] = buf[17..49].try_into().expect("slice of length 32");
+    let leftover = buf.split_off(LEGACY_HEADER_BYTES);
+
+    Ok(LegacyHeader { iv, mac, leftover })
+}
```
quexten, line 310: Crypto internal detail that leaks out of the crypto crate.
```suggestion
```

crates/bitwarden-vault/src/cipher/attachment_client/upgrade.rs
```diff
@@ -111,107 +110,227 @@ pub async fn prepare_attachment_upgrade<R: Repository<Cipher> + ?Sized>(
...
-
     let created = create_attachment(cipher_id, request, api_client, repository).await?;
 
     Ok(AttachmentUpgrade {
         attachment_id: created.attachment_id,
         upload_url: created.upload_url,
         file_upload_type: created.file_upload_type,
+        encrypted_file_name: encrypted_file_name.to_string(),
+        iv: final_.iv.to_vec(),
+        mac: final_.mac.to_vec(),
```
quexten, line 203: IV, mac, are already written to the output attachment file, there is no reason to return them here.
```suggestion
```

crates/bitwarden-vault/src/cipher/attachment_client/upgrade.rs
```diff
@@ -111,107 +110,227 @@ pub async fn prepare_attachment_upgrade<R: Repository<Cipher> + ?Sized>(
...
+        &cipher,
+        &decrypted_file_name,
+        &legacy_iv,
+        reader,
+        writer,
+        leftover,
+        key_store,
+    )?;
+
+    while let Some(plaintext) = decryptor.next_chunk().await.map_err(DecryptError::from)? {
```
quexten, line 169: This entire loop should be replaced with https://dtantsur.github.io/rust-openstack/tokio/io/fn.copy.html after replacing the non-standard ad-hoc stream reader / writer implementations with standard `AsyncRead` / `AsyncWrite`

crates/bitwarden-crypto/src/streaming_attachment_cipher.rs
```diff
@@ -0,0 +1,148 @@
+//! Streaming AES-256-CBC + HMAC primitives for the attachment-upgrade pipeline.
```
quexten, line 1: This entire file should be replaced, with readers / writers that implement the standard `AsyncRead` / `AsyncWrite` traits. The constructor should take only a `&SymmetricCryptoKey` for both the reader and the writer, and the reader / writer should then read the header.

Internally the nicest way to implement this is to represent this through a state machine / enum:
```
enum DecryptionState {
  Initial,
  Streaming { hmac: ..., buffer: ..., header: Aes256CbcHmacHeader { mac, iv } }
  Done,
}
```

crates/bitwarden-crypto/src/aes.rs
```diff
@@ -93,6 +99,138 @@ fn encrypt_aes256_internal(
...
+/// authenticated** and must be treated as untrusted until `finalize` returns `Ok`.
+pub(crate) struct StreamingAes256CbcHmacDecryptor {
+    decryptor: cbc::Decryptor<aes::Aes256>,
+    hmac: PbkdfSha256Hmac,
+    /// Always retains at least one full block so `finalize` can apply PKCS7 unpadding.
+    pending: Vec<u8>,
+}
+
+impl StreamingAes256CbcHmacDecryptor {
+    pub(crate) fn new(iv: &[u8; 16], key: &Aes256CbcHmacKey) -> Self {
```
quexten, line 113:
```suggestion
    pub(crate) fn try_new(key: &SymmetricCryptoKey) -> Result<Self, WrongKeyType> {
```

crates/bitwarden-crypto/src/aes.rs
```diff
@@ -93,6 +99,138 @@ fn encrypt_aes256_internal(
     (iv, data)
 }
 
+/// Streaming AES-256-CBC + HMAC-SHA256 decryptor. The HMAC is verified only at
+/// [`Self::finalize`]; bytes returned from [`Self::update`] are decrypted but **not yet
+/// authenticated** and must be treated as untrusted until `finalize` returns `Ok`.
+pub(crate) struct StreamingAes256CbcHmacDecryptor {
+    decryptor: cbc::Decryptor<aes::Aes256>,
```
quexten, line 106: This struct should be replaced with an enum, since we have multiple states:
- No header read yet
- Streaming
- Done

PR #1096, opened 2026-05-18
Enable codable on swift bindings

quexten:
## 🎟️ Tracking

https://github.com/bitwarden/ios/pull/2648#issuecomment-4460930000

## 📔 Objective

As per mobile's request, we are enabling codable for the uniffi bindings so that they can directly serialize the structs they get returned from the SDK without having to maintain a seprate copy. Maintaining a separate copy in the past has lead to bugs / incorrect state being saved (https://github.com/bitwarden/ios/pull/2303).

This is safer than mobiles current approach of persisting server request / response models, which do not match what the SDK returns.

### DIFF
--- a/crates/bitwarden-core/uniffi.toml
+++ b/crates/bitwarden-core/uniffi.toml
@@ -9,3 +9,4 @@ omit_checksums = true
 ffi_module_name = "BitwardenCoreFFI"
 module_name = "BitwardenCore"
 generate_immutable_records = true
+generate_codable_conformance = true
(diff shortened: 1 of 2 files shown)


djsmith85 (approved):

dani-garcia:
Is this something we could also enable for all other crates too? Seems pretty safe to add in general and it seems like it could also be beneficial for ios devs as they integrate the SDK more. Plus it might be a bit strange that some SDK items implement codable but others don't.

quexten:
@dani-garcia Sure, I think that makes sense. I just did the minimum amount of change to get KM's structs covered, and did not investigate the user-cases / appropriateness on other crates.

PR #1097, opened 2026-05-18
[PM-37661] Add uniffi support for state bridge and integration tests

quexten:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-37661

## 📔 Objective

Adds uniffi bindings for the KM state bridge, and adds integration tests that cross the FFI boundary to prove that the bindings work correctly, and to have an example for how to implement it in mobile.


dani-garcia (approved):

eligrubb (approved):
Changes looks good. One thought: the description of these new integration tests should be added to the `.claude/CLAUDE.md` file. E.g. how to run them, when to add new ones, how to determine if an interface is testable, avoiding tests that would require mocking for now, etc.

PR #1098, opened 2026-05-18
[PM-27801] Refresh tokens on 401

dani-garcia:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-27801

## 📔 Objective

Retry once with a freshly refreshed token when an authenticated request comes back with a `401 Unauthorized`. Until now the middleware attached whatever token `get_token` returned, including a still-valid-by-clock token that the server had since invalidated, and surfaced the 401 to the caller.

Changes:
- `MiddlewareExt::get_token` takes a `force: bool` flag. When `true`, `PasswordManagerTokenHandler` and `SecretsManagerTokenHandler` bypass the local expiration check and hit `/connect/token` unconditionally.
- `MiddlewareWrapper::handle` clones the request before sending (via `try_clone`, so streaming bodies cleanly opt out) and, on a 401 from a request that required auth, re-injects a forced-refresh token and retries exactly once.
- Token injection moves into a small `inject_token_if_required` helper so the initial-send and retry paths share the same logic. It returns whether the request specified an auth method, which is what gates the retry.

Bounded by construction: at most one retry per request, the retry only happens when the original request actually carried an `AuthRequired` extension, and `try_clone` failing (streaming body) skips the retry instead of erroring. Concurrent refreshes are unchanged from existing behavior and called out in the comment inside `get_token`. We have a tracking issue to follow up on that: https://bitwarden.atlassian.net/browse/PM-32346

Tests cover the three meaningful paths in both handlers: retry succeeds with a renewed token, retry succeeds when the initial `get_token` failed but the forced refresh works, and no retry occurs when the request had no auth extension.

## 🚨 Breaking Changes

### DIFF
--- a/crates/bitwarden-auth/src/token_management/middleware.rs
+++ b/crates/bitwarden-auth/src/token_management/middleware.rs
@@ -18,7 +18,9 @@ pub(crate) struct MiddlewareWrapper<T>(pub(crate) T);
 #[cfg_attr(not(target_arch = "wasm32"), async_trait::async_trait)]
 #[cfg_attr(target_arch = "wasm32", async_trait::async_trait(?Send))]
 pub(crate) trait MiddlewareExt: 'static + Send + Sync {
-    async fn get_token(&self) -> Result<Option<String>, LoginError>;
+    /// Retrieves the current access token, renewing it if necessary. If `force` is true,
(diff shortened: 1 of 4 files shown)


JaredSnider-Bitwarden (approved):
Nice functionality addition! LGTM!

PR #1099, opened 2026-05-18
[Shared Unlock] [PM-35083] Fix biometrics IPC, add user key return

quexten:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-35083

## 📔 Objective

Biometrics IPC is broken as proven by the integration tests. The biggest reason is that the tokio timer implementation is not working on wasm. This PR adds an implementation on top of the gloo timers that we already use.

Further, this adds integration tests to show that this works from within the wasm environment.

Finally, this adds returning the user-key to the requesting sender. While this is not the long-term solution after shared unlock is rolled out, it is required for the transition period.


crates/bitwarden-threading/src/time.rs
```diff
@@ -12,6 +12,75 @@ pub async fn sleep(duration: Duration) {
...
+pub async fn timeout<F: std::future::Future>(
+    duration: Duration,
+    future: F,
+) -> Result<F::Output, ElapsedError> {
+    // Wrap the !Send `gloo_timers` future and the caller's future in a single
+    // state machine that we then assert as `Send`. wasm32-unknown-unknown is
+    // single-threaded, so the future is never actually moved across threads
+    // and the trait bound on `CryptoProvider::send` is satisfied.
+    wasm_send::WasmSend(async move {
+        let sleep_fut = gloo_timers::future::sleep(duration);
```
quexten: @coroiu LMK if there is a better way to do this / preferrable way from platforms perspective. This allowed adding no new deps at least.

coroiu: Probably a nitpick, but maybe use the `sleep` function instead of `gloo_timers::sleep` ?

crates/bitwarden-wasm-internal/integration-tests/tests/biometrics-ipc.test.ts
coroiu, line 1: Nice! How about adding this to CI?

crates/bitwarden-threading/src/time.rs
quexten: Updated as per @coroiu 's suggestion in DM

coroiu (dismissed):
Sorry, forgot to commit review 🙈

crates/bitwarden-wasm-internal/integration-tests/tests/biometrics-ipc.test.ts
quexten: It's a job in `build-wasm-internal` and already runs in the CI :)
https://github.com/bitwarden/sdk-internal/blob/8e643e943d56a3df42449d27b1d0cf39ac82e33b/.github/workflows/build-wasm-internal.yml#L137

coroiu: Oh, I missed this, nice! 👍

coroiu (approved):

eligrubb (approved):

PR #1102, opened 2026-05-18
[PM-29547] Add html passwordrules attribute parser

adudek-bw:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-29547

## 📔 Objective

Add HTML passwordrules parser to the generator client

## 🚨 Breaking Changes

No breaking changes, all additive


crates/bitwarden-generators/src/password.rs
```diff
@@ -81,10 +104,29 @@ impl Default for PasswordGeneratorRequest {
...
+            custom_allowed_chars: None,
+            max_consecutive: None,
         }
     }
 }
 
+/// Returns true if `c` is an ASCII-printable, non-whitespace character — the set of
+/// characters that the HTML `passwordrules` spec allows inside a custom class.
+pub(crate) fn is_ascii_printable_non_whitespace(c: char) -> bool {
+    c.is_ascii_graphic()
```
harr1424: ❓ Why is this function `pub(crate)` and could `is_ascii_graphic()` just be called inline instead? 

Not a blocker and it's just a personal preference, but it seems off to alias `char.is_ascii_graphic()` with a new function `is_ascii_printable_non_whitespace()` public to the crate.

harr1424 (commented):
Looks like the lint workflow is failing, did you anticipate breaking changes to the SDK?

adudek-bw:
> Looks like the lint workflow is failing, did you anticipate breaking changes to the SDK?

Ah, I thought that with the changes being only additions it wouldn't cause that, let me look into it.

crates/bitwarden-generators/src/password.rs
adudek-bw: Yes good call!

harr1424 (dismissed):

crates/bitwarden-generators/src/generator_client.rs
```diff
@@ -47,6 +47,17 @@ impl GeneratorClient {
         password(input)
     }
 
+    /// Parses an HTML `passwordrules` attribute string into a [`PasswordGeneratorRequest`].
+    ///
+    /// The returned request can be passed to [`GeneratorClient::password`] to produce a
+    /// password that satisfies the website's declared constraints.
+    pub fn password_rules(
```
fedemkr, line 54: 🤔 Maybe I'm overlooking something but I think this should be also in `GeneratorClients` under `bitwarden-uniffi->src->tool->mod` to be able to be used in mobile with UniFFI.

adudek-bw: Good catch! updated :)

harr1424 (dismissed):

Hinton:
Why not just use the battle tested https://crates.io/crates/password-rules-parser/1.1.0?

adudek-bw:
> Why not just use the battle tested https://crates.io/crates/password-rules-parser/1.1.0?

Updated with said crate!

crates/bitwarden-generators/src/password.rs
```diff
@@ -53,6 +53,35 @@ pub struct PasswordGeneratorRequest {
...
+    /// character classes (e.g. `allowed: [-_.]`).
+    #[cfg_attr(feature = "uniffi", uniffi(default = None))]
+    #[cfg_attr(feature = "wasm", tsify(optional))]
+    pub custom_allowed_chars: Option<String>,
+
+    /// The maximum number of consecutive identical characters allowed in the generated password,
+    /// as expressed by the HTML `passwordrules` `max-consecutive` property.
+    ///
+    /// Currently parsed and stored, but not yet enforced by the generator.
+    // TODO: enforce `max_consecutive` in the generator.
```
itsadrago: Can we implement this as a part of this work?

crates/bitwarden-generators/src/password.rs
```diff
@@ -147,6 +189,9 @@ struct PasswordGeneratorOptions {
     pub(super) upper: (CharSet, usize),
     pub(super) number: (CharSet, usize),
     pub(super) special: (CharSet, usize),
+    /// Custom required characters from `passwordrules` `required: [...]` entries. At least one
+    /// char from this set is guaranteed to appear in the output when the set is non-empty.
+    pub(super) custom: (CharSet, usize),
```
itsadrago: Do we need the `pub(super)` visibility modifiers if the struct is private?

crates/bitwarden-generators/Cargo.toml
```diff
@@ -15,17 +15,14 @@ license-file.workspace = true
...
-    "dep:wasm-bindgen",
-] # WebAssembly bindings
+uniffi = ["dep:uniffi"]
+wasm = ["bitwarden-core/wasm", "dep:tsify", "dep:wasm-bindgen"]
 
 [dependencies]
 bitwarden-core = { workspace = true, features = ["internal"] }
 bitwarden-crypto = { workspace = true }
 bitwarden-error = { workspace = true }
+password-rules-parser = { workspace = true }
```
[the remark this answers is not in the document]
adudek-bw: Yes: https://bitwarden.atlassian.net/browse/VULN-599

crates/bitwarden-generators/src/password.rs
```diff
@@ -158,8 +207,27 @@ impl PasswordGeneratorRequest {
...
         // TODO: Add password generator policy checks
 
-        // We always have to have at least one character set enabled
-        if !self.lowercase && !self.uppercase && !self.numbers && !self.special {
+        // Sanitize custom char lists defensively: the parser already filters non-printable
+        // characters, but the request type is `pub` so callers can construct it directly.
+        let custom_required = self
+            .custom_required_chars
+            .as_deref()
+            .map(sanitize_custom_chars)
```
dani-garcia, line 215: Question: This will filter out required characters that aren't ASCII, potentially generating a password that is not valid. Is this expected?

For example, if the required character set contains `á`, `¿` and `ñ`. This check will filter them out and generate a password without any of them.

crates/bitwarden-generators/src/password.rs
```diff
@@ -411,4 +583,118 @@ mod test {
...
+                run = 1;
+                prev = Some(c);
+            }
+            longest = longest.max(run);
+        }
+        longest
+    }
+
+    #[test]
+    fn test_password_gen_honors_max_consecutive() {
```
dani-garcia, line 604: We have tests for `max_consecutive` now, we should also add some tests for the `custom_required` and `custom_allowed` options.

dani-garcia (commented):
Looks pretty good to me, just some small comments

crates/bitwarden-generators/src/password.rs
adudek-bw: That is a good question! Now that I think about it I'm not sure what the "proper" thing to do would be. I'd assume that we don't want any non-ASCII characters and we should fail loudly with an error or something. @fedemkr what do you think?

fedemkr: I think we could keep the initial version with only ASCII characters as that's what we support in-app. However, I'll ask Max about this and see what is preferred.

fedemkr: He agrees, so we're good as is; and yes it can just throw an error when getting non-ascii characters.

dani-garcia (approved):

PR #1103, opened 2026-05-19
[Shared Unlock] [PM-35083] Shared unlock fixes & integration tests

quexten:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-35083

## 📔 Objective

Fixes FFI return types of shared unlock and adds integration tests proving it works.


quexten:
@copilot resolve the merge conflicts in this pull request

crates/bitwarden-shared-unlock/src/wasm/drivers.rs
```diff
@@ -40,7 +40,7 @@ extern "C" {
     async fn get_user_key(
         this: &RawJsSharedUnlockDriver,
         user_id: UserId,
-    ) -> Result<Option<SymmetricCryptoKey>, JsValue>;
+    ) -> Result<JsValue, JsValue>;
```
quexten, line 43: Wasm bindgen does not support this as a return type so we have to manually use JSValue

dereknance (dismissed):
✅ Changes to WASM integration tests approved for Platform

crates/bitwarden-wasm-internal/integration-tests/tests/shared-unlock.test.ts
```diff
@@ -0,0 +1,106 @@
...
+describe("shared unlock ipc", () => {
+  it("unlocks the leader when the follower reports a manual unlock", async () => {
+    const { follower, leaderDriver: leaderHandle } = await setupSharedUnlockPair({
+      leader: { initialStates: USER_A_LOCKED_STATE },
+      follower: { initialStates: USER_A_LOCKED_STATE },
+    });
+
+    await follower.handle_device_event(UNLOCK_EVENT);
+    await sleep(20);
+    expect(leaderHandle.getUserKey(USER_A)).toBe(USER_KEY);
```
mzieniukbw, line 26: ❓  Could we instead poll every 1 second, up to 20 seconds and only throw if the expect still fails ?
That way we don't have to wait for arbitrary number of seconds.

mzieniukbw (approved):

crates/bitwarden-wasm-internal/integration-tests/tests/shared-unlock.test.ts
quexten: I think there may be confusion here, 20 is in milliseconds not seconds, unless I'm missing the intent of the comment?

dereknance (approved):

PR #1104, opened 2026-05-19
[PM-33322] Implement generate_session_key() and unlock() with session key

coroiu:
## 📔 Objective

Adds session-key rehydration unlock to `PasswordManagerClient` in a new KM-owned `bitwarden-pm/src/key_management/` directory.

- `SessionKey` (`new()`, `Display`, `FromStr`) wraps a XChaCha20Poly1305 key the caller persists outside the SDK.
- `PasswordManagerClient::generate_session_key()` mints a session key, seals the user key with it via `SymmetricKeyEnvelope` (SessionKey namespace), and persists the envelope under `SESSION_PROTECTED_USER_KEY`.
- `PasswordManagerClient::unlock(UnlockMethod::SessionKey(_))` reads the envelope + `ACCOUNT_CRYPTO_STATE`, unseals the user key, calls `initialize_user_crypto_decrypted_key`, then restores org keys from the `OrganizationSharedKey` repository.
- `UnlockError::Unknown` is a uniform opaque failure; details are emitted via `tracing::error!` on each error path.

Supporting changes in `bitwarden-core`:
- `SESSION_PROTECTED_USER_KEY` is now typed as `SymmetricKeyEnvelope` (was `EncString`) to match the seal format. The setting key is pre-release, so no migration is required.
- `InternalClient::initialize_user_crypto_decrypted_key` is now `pub` so callers in `bitwarden-pm` can reuse the unlock primitive.


crates/bitwarden-pm/src/key_management/session_key.rs
```diff
@@ -0,0 +1,27 @@
+use bitwarden_crypto::SymmetricCryptoKey;
+
+/// A symmetric key that wraps the user key in the persisted state, allowing a
+/// rehydrated client to unlock without re-deriving the user key from a master
+/// password or other primary unlock factor.
+///
+/// Callers are responsible for storing this key in a secure location outside
+/// the SDK (e.g. the OS keychain) and providing it back to
+/// [`PasswordManagerClient::unlock`](crate::PasswordManagerClient::unlock) when
+/// reconstructing the client.
```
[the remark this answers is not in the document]
coroiu: That will be handled in a follow up PR

crates/bitwarden-unlock/src/lib.rs
```diff
@@ -0,0 +1,11 @@
+#![doc = include_str!("../README.md")]
+
+// Enable uniffi scaffolding when the "uniffi" feature is enabled.
+#[cfg(feature = "uniffi")]
+uniffi::setup_scaffolding!();
```
dani-garcia, line 5: This crate is missing the uniffi.toml file, to correctly set the package name for swift/kotlin

crates/bitwarden-unlock/src/unlock_client.rs
```diff
@@ -0,0 +1,507 @@
...
+                LockingError::Unknown
+            })?
+            .ok_or_else(|| {
+                tracing::error!("Missing session_protected_user_key in database");
+                LockingError::Unknown
+            })?;
+
+        let account_crypto_state = state
+            .setting(ACCOUNT_CRYPTO_STATE)
+            .map_err(|e| {
```
dani-garcia, line 113: This is unrelated to this PR, but this code is fairly hard to read, I wonder if we should be considering changes to the repository/setting API:
- have `.setting()` not return a result anymore, now that we have an in memory database as fallback
- have `.get()` return `Result<T, Error>` rather than `Result<Option<T>, Error>`, and make the no value an error. We can then provide a utility `.get_opt()` that returns `Result<Option<T>, Error>`. Feels like in most cases we handle the absence of value as an error anyway.

Nothing to do here though, I'll think about it more another day.

crates/bitwarden-locking/src/session_key.rs
```diff
@@ -0,0 +1,27 @@
...
+
+impl SessionKey {
+    /// Mint a new random session key.
+    #[allow(clippy::new_without_default)]
+    pub fn new() -> Self {
+        Self(SymmetricCryptoKey::make_xchacha20_poly1305_key())
+    }
+}
+
+impl PartialEq for SessionKey {
```
dani-garcia: Can't we just `#[derive(PartialEq)]` for this?

crates/bitwarden-locking/README.md
dani-garcia, line 1: We should probably set KM as codeowners of this crate?

dani-garcia (commented):
Looks pretty good overall, just some small changes

crates/bitwarden-locking/src/session_key.rs
coroiu: 🤦 yeah, we probably can

crates/bitwarden-unlock/src/unlock_client.rs
coroiu: I'm not sure about using memory as a fallback, seems like it might hide some real bugs, but I do agree with this being hard to manage. `No value = Error` sounds like a reasonable alternative yeah

.github/CODEOWNERS
```diff
@@ -43,6 +43,7 @@ crates/bitwarden-generators/** @bitwarden/team-tools-dev @bitwarden/team-platfor
 crates/bitwarden-organizations/** @bitwarden/team-admin-console-dev @bitwarden/team-platform-dev
 crates/bitwarden-policies/** @bitwarden/team-admin-console-dev @bitwarden/team-platform-dev
 crates/bitwarden-send/** @bitwarden/team-tools-dev @bitwarden/team-platform-dev
+crates/bitwarden-locking/** @bitwarden/team-key-management-dev
```
quexten: Thought: Can we name this `unlock`? 

We don't have any crates that are named in the present participle form of a verb, so if anything, it should be `lock`. However, generally we consider the opposite -  `unlock` not `lock` in most parts of the app. `UnlockService`.

It's confusing to have two namings here.

crates/bitwarden-locking/src/locking_client.rs
```diff
@@ -0,0 +1,507 @@
...
+#[derive(Debug, thiserror::Error)]
+pub enum LockingError {
+    /// An unknown error occurred. See logs for details.
+    #[error("An unknown error occurred while unlocking the client")]
+    Unknown,
+}
+
+/// Client for minting session keys and unlocking the vault with one.
+#[derive(Clone)]
+pub struct LockingClient {
```
quexten: Same thing applies here, let's name it `UnlockClient`. The session key is just another (local) unlock method.

crates/bitwarden-locking/src/session_key.rs
```diff
@@ -0,0 +1,20 @@
...
+/// the SDK (e.g. the OS keychain) and providing it back to
+/// [`LockingClient::unlock`](crate::LockingClient::unlock) when reconstructing
+/// the client.
+#[derive(PartialEq)] // This is ok because SymmetricCryptoKey implements PartialEq with constant-time equality checks.
+pub struct SessionKey(pub(crate) SymmetricCryptoKey);
+
+impl SessionKey {
+    /// Mint a new random session key.
+    #[allow(clippy::new_without_default)]
+    pub fn new() -> Self {
```
quexten: `make` is what we use for cryptographic keys. Also, it is forseeable that this will take a parameter, either directly a crypto algorithm, or at least a: `enum CryptoSuite { FIPS, Standard }` enum in the future.
```suggestion
    pub fn make() -> Self {
```

crates/bitwarden-locking/src/locking_client.rs
```diff
@@ -0,0 +1,507 @@
...
+    /// Mint a new session key and persist the user key wrapped by it.
+    ///
+    /// Requires the client to be unlocked (the user key must be present in the
+    /// key store). The returned [`SessionKey`] should be stored outside the SDK
+    /// by the caller and provided back to [`LockingClient::unlock`] on the next
+    /// rehydrated client.
+    pub async fn generate_session_key(&self) -> Result<SessionKey, LockingError> {
+        use bitwarden_core::key_management::SymmetricKeySlotId;
+
+        let session_key = SessionKey::new();
```
quexten: We introduced an entire new type here. I want to make the following proposal:

There should be two functions as an impl on SessionKey:

```
from_context(...) -> (SymmetricKeyEnvelope, SessionKey)
unwrap_to_context(&self, key_envelope, ctx);
```

which contain this implementation instead of having the client contain the logic. This feels more appropriate since now we have the new concept owning what it is used for / does / not leak accidentally usage (the symmetric crypto key) outside of it's code.

crates/bitwarden-unlock/src/unlock_client.rs
```diff
@@ -0,0 +1,507 @@
...
+    /// the state registry, unwraps the user key, initializes the user's crypto
+    /// state, and restores any persisted organization keys.
+    pub async fn unlock(&self, unlock: UnlockMethod) -> Result<(), LockingError> {
+        let UnlockMethod::SessionKey(session_key) = unlock;
+
+        let state = self.client.platform().state();
+
+        let session_protected_user_key = state
+            .setting(SESSION_PROTECTED_USER_KEY)
+            .map_err(|e| {
```
quexten, line 96: A general thought: Is there any way we could use macros to make this less painful to read? This is 16 lines for a single state read which is a lot.

crates/bitwarden-unlock/src/unlock_client.rs
```diff
@@ -0,0 +1,507 @@
...
+            .to_base64()
+            .to_string()
+    }
+
+    /// Mint a fresh `(user_key, account_crypto_state)` pair where the wrapped
+    /// state was sealed with `user_key`.
+    fn make_test_user_crypto() -> (SymmetricCryptoKey, WrappedAccountCryptographicState) {
+        let store: KeyStore<KeySlotIds> = KeyStore::default();
+        let mut ctx = store.context_mut();
+        let user_key_id = ctx.make_symmetric_key(SymmetricKeyAlgorithm::XChaCha20Poly1305);
```
quexten, line 281: Thought (non-blocking): IT feels weird to re-implement initialization of v2 crypto here. In fact, this is not compliant with what a v2 account is. Ed25519 is no longer the default signature algorithm and will never be rolled out to real users.

While out of scope here, KM should provide test fixtures in the core crate (or a separate crate if not prevented currently by dep cycles), behind a compile time flag, which centralize this for all tests to use.

quexten (changes requested):

crates/bitwarden-pm/src/lib.rs
```diff
@@ -154,19 +156,9 @@ impl PasswordManagerClient {
...
-    /// key rather than mutate in-memory state.
-    pub async fn invalidate_session_key(&self) -> Result<(), bitwarden_state::SettingsError> {
-        self.0
-            .platform()
-            .state()
-            .setting(SESSION_PROTECTED_USER_KEY)?
-            .delete()
-            .await
+    /// Lock and unlock operations
+    pub fn locking(&self) -> bitwarden_locking::LockingClient {
```
quexten: Should we compile time flag this for the CLI? Given that settings do not work on `clients` / `mobile` yet (given they don't have sdk managed state), this client is not usable on non-cli.

.github/CODEOWNERS
coroiu: This was actually named `Unlock` first but changed it to `Locking` after the following slack convo 
image

But this is ultimately KMs crate so I'll leave the final decision to you

crates/bitwarden-unlock/src/unlock_client.rs
coroiu: @dani-garcia commented on the same thing above, we should probably address this somehow but keep it out-of-scope for this PR

crates/bitwarden-pm/src/lib.rs
coroiu: Good point, I'll do that!

crates/bitwarden-unlock/src/unlock_client.rs
dani-garcia: Sorry I think I wasn't super clear with the db fallback, I meant that now that we have the in-memory DB for tests and SM, we're always initializing the database in the `StateRegistry` as part of the `Client` construction.

That means we always have a DB to pull the Settings/Repositories from, so I don't think we'd be hitting the error case anymore.

But yeah that's a more minor thing, I feel like making `no value = error` and maybe moving the error logging as part of the state framework would simplify this a lot

.github/CODEOWNERS
coroiu: I think I've addressed most of the feedback except this, I'll hold off on making this change until I'm able to get your input :)

crates/bitwarden-unlock/src/unlock_client.rs
coroiu: Created [PM-37793 Improve usage of bitwarden-state settings](https://bitwarden.atlassian.net/browse/PM-37793) for this

.github/CODEOWNERS
quexten: Sorry, I didn't circle back around on this until now. I've posted a poll in `km` and will report back the findings here, but so far I favor `bitwarden-lock-state` if `unlock` doesn't work.

quexten: @coroiu A poll in the KM dev channel unanimously arrived at `bitwarden-unlock`. Please rename it to that. I'll give it a final pass tomorrow morning it aside from that I think in happy to put the approval

crates/bitwarden-unlock/src/session_key.rs
```diff
@@ -0,0 +1,54 @@
...
+/// the SDK (e.g. the OS keychain) and providing it back to
+/// [`UnlockClient::unlock`](crate::UnlockClient::unlock) when reconstructing
+/// the client.
+#[derive(PartialEq)] // This is ok because SymmetricCryptoKey implements PartialEq with constant-time equality checks.
+pub struct SessionKey(pub(crate) SymmetricCryptoKey);
+
+impl SessionKey {
+    /// Mint a new random session key.
+    pub fn make() -> Self {
+        Self(SymmetricCryptoKey::make_xchacha20_poly1305_key())
```
quexten, line 20:
```suggestion
        Self(SymmetricCryptoKey::make(SymmetricKeyAlgorithm::XChaCha20Poly1305)
```

KM note: We should deprecate the above function / make it non public.

A thought: We probably want to change the make function to work on the key context instead. That doesn't have to be in this PR, and it's OK to leave it up to KM for later, but we will need to switch to XAES-GCM instead for FIPS based on the cipher suite stored in the key context. [Breakdown pending].

cc @eligrubb This work is in flight just want to put it on your radar that this call site will change

crates/bitwarden-unlock/uniffi.toml
```diff
@@ -0,0 +1,11 @@
+[bindings.kotlin]
+package_name = "com.bitwarden.locking"
```
quexten: Package name is out of date

quexten (commented):
Two things, then I'm happy to approve.

dani-garcia (approved):

crates/bitwarden-pm/Cargo.toml
```diff
@@ -55,14 +59,20 @@ bitwarden-send = { workspace = true }
...
 reqwest-middleware = { workspace = true }
+thiserror = { workspace = true }
+tracing = { workspace = true }
 tsify = { workspace = true, optional = true }
 uniffi = { workspace = true, optional = true, features = ["tokio"] }
 wasm-bindgen = { workspace = true, optional = true }
 wasm-bindgen-futures = { workspace = true, optional = true }
 
+[dev-dependencies]
+tokio = { workspace = true, features = ["rt", "macros"] }
```
Thomas-Avery, line 75: Looking at the code changes to `lib.rs` I don't see why these new dependencies are needed in this cargo.toml. Might be missing something.

Thomas-Avery (approved):
LGTM, one non blocking thing I noticed.

crates/bitwarden-pm/Cargo.toml
dani-garcia: I believe this is coming from an earlier version of the PR where the unlock code and its tests were in bitwarden-pm rather than it's own separate crate. I'm surprised CI didn't detect it as unused here. 

I'll clean it up in a follow up PR.

quexten (approved):

PR #1105, opened 2026-05-19
Add support for depending on other ClientExt

Hinton:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-37745

## 📔 Objective

Give feature clients a way to expose themselves as a mockable trait, and let other clients hold that trait as an `Arc<dyn FooTrait>` field that `#[derive(FromClient)]` populates automatically.

This PR adds two pieces of infrastructure. Nothing in production code uses either yet — they're available for feature crates to opt into.

**`FromClientShared` trait and blanket `FromClientPart<Arc<T>>` impl** ([crates/bitwarden-core/src/client/from_client_part.rs](crates/bitwarden-core/src/client/from_client_part.rs))

Implementing `FromClientShared` on `dyn FooTrait` makes `Arc<dyn FooTrait>` extractable through the existing `FromClientPart` machinery, so `#[derive(FromClient)]` can populate trait-object fields:

```rust
#[derive(FromClient)]
struct OuterClient {
    inner: Arc<dyn FooTrait>,
}
```

The indirection exists because Rust's orphan rule rejects `impl FromClientPart<Arc<dyn FooTrait>> for Client` from any crate other than `bitwarden-core` (both types are foreign and `Arc<_>` isn't a fundamental wrapper). `impl FromClientShared for dyn FooTrait` is allowed because `dyn FooTrait` is a local type.

**`#[client_trait(via = <expression>)]` attribute macro** ([crates/bitwarden-core-macro/src/lib.rs](crates/bitwarden-core-macro/src/lib.rs))

Applied to a hand-written trait declaration, the macro re-emits the trait unchanged and adds the four-line `FromClientShared` bridge using the provided expression:

```rust
#[client_trait(via = client.folders())]
#[cfg_attr(any(test, feature = "test-fixtures"), mockall::automock)]
#[async_trait::async_trait]
pub trait FoldersClientTrait: Send + Sync {
    async fn get(&self, id: FolderId) -> Result<FolderView, FolderError>;
}
```

The expression has `client: &Client` in scope. The macro deliberately doesn't generate the trait surface, forwarding impl, or mockall attribute — those are written or applied by the user, so the pattern works for clients whose inherent methods span multiple `impl` blocks across multiple files (e.g. `CiphersClient`).

**End-to-end test** ([crates/bitwarden-core/tests/from_client_shared.rs](crates/bitwarden-core/tests/from_client_shared.rs))

Demonstrates the full pattern: hand-written trait + `#[mockall::automock]` + `#[async_trait]` + `#[client_trait(via = ...)]` + concrete struct + extension trait. Three tests cover real-client wiring through `OuterClient::from_client(&client)`, mock substitution for sync methods, and mock substitution for async methods.

---

For clients that split implementation in multiple files this won't work out of the box and will instead encourage patterns where they have to defer their implementation to another method.

## 🚨 Breaking Changes

None. Purely additive: a new trait, a new blanket impl, and a new attribute macro. No changes to existing public APIs or behavior.


crates/bitwarden-core/tests/from_client_shared.rs
```diff
@@ -0,0 +1,98 @@
...
+    #[allow(clippy::unused_async)]
+    async fn fetch(&self, key: String) -> String {
+        format!("{}={}", key, self.server_url())
+    }
+}
+
+#[derive(FromClient)]
+struct OuterClient {
+    inner: Arc<dyn InnerClientTrait>,
+}
```
coroiu, line 55: question/suggestion: A few things rolled together about the intended usage pattern, since the PR ships no production usage yet:

**1. Naming convention.** Is `Arc<dyn VaultClientTrait>` the intended convention (struct `VaultClient` + accessor `vault()` + mockable trait `VaultClientTrait`, field `vault_client`)? `vault_client: Arc<dyn VaultClientTrait>` repeats "client"/"Client"/"Trait" three ways — did you consider `vault: Arc<dyn Vault>` or similar? A worked fido→vault example would help lock the convention down.

**2. Parallel surface / drift.** This gives `fido()` a hand-curated subset of vault's API that has to be declared on the trait, implemented on `VaultClient`, and kept in sync as vault evolves. So `pm.vault()` (full concrete surface) and `fido().vault_client` (the trait subset) end up with deliberately different interfaces. Is that maintenance cost intended, and how do we keep the trait from drifting from the concrete client?

**3. Granularity.** `VaultClient`'s methods mostly return concrete sub-clients (`ciphers()`, `folders()`). Where do you expect the trait boundary to sit — leaf methods fido needs, or the sub-client accessors? If it returns the concrete sub-clients, mocking `VaultClientTrait` doesn't actually make the calls underneath mockable, which seems to undercut the goal.

**4. Documentation.** The rustdoc on `FromClientShared` and `client_trait` is good, but the *conventions* above (naming, where the trait boundary sits, the mockability rationale, when to hand-write a `FromClientShared` impl vs. use the macro) live only in this PR description right now. Could we codify them so they outlive the PR? `bitwarden-core` has a `README.md` but no `CLAUDE.md` — this pattern seems worth both: the README for the architectural narrative, and a `CLAUDE.md` rule so the convention is enforced going forward. The macro crate (`bitwarden-core-macro`) has neither and could use at least a short pointer.

Hinton: 1. I was leaning towards `VaultClientTrait` we could call it `Vault` but what would the implementation be, `VaultImpl` if so we should call it `VaultTrait` which is more consistent with rust from what I've seen.
2. There are some different paths here, I think most would prefer a single trait over granularity but it does mean we have to move back into a single file again 😢.
3. I would expect you to define if you want cipher or folder, and not use the "group"?
4. I was planning on updating contributing docs, but maybe we should document it here to.

Hinton: Investigate if we can use automock on struct and combine with double to avoid defining our own trait and to be able to split clients into multiple files.

coroiu (dismissed):
Fix the small conflict and I'll approve as soon as I can

crates/bitwarden-core/tests/from_client_shared.rs
Hinton: Turned out to not be possible, and we're stuck with the whole impl living inside a single file.

coroiu (approved):

PR #1106, opened 2026-05-19
[PM-32346] Serialize concurrent token renewals

dani-garcia:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-32346

## 📔 Objective

Serialize calls to the auth middleware's `get_token` so that concurrent
requests don't race the renewal path. Previously, if N requests arrived
simultaneously while the access token was expired, each would independently
fire a renewal against the identity server. After the 401-retry path landed
in #1098, a concurrent renewal could also cause an in-flight request to go
out with a token that another task had just invalidated.

`MiddlewareWrapper` now holds a `tokio::sync::Mutex` around the inner
`MiddlewareExt`, and `get_token` takes `&mut self` to encode the
exclusive-access invariant at the trait level. The handlers' internal
`RwLock<Inner>` is unchanged and still protects in-memory state from
sync access paths that don't go through the middleware.

### DIFF
--- a/crates/bitwarden-auth/Cargo.toml
+++ b/crates/bitwarden-auth/Cargo.toml
@@ -50,6 +50,7 @@ serde = { workspace = true }
 serde_bytes = { workspace = true }
 serde_repr = { workspace = true }
 thiserror = { workspace = true }
+tokio = { workspace = true, features = ["sync"] }
 tracing = { workspace = true }
 tsify = { workspace = true, optional = true }
 uniffi = { workspace = true, optional = true }
--- a/crates/bitwarden-auth/src/token_management/middleware.rs
+++ b/crates/bitwarden-auth/src/token_management/middleware.rs
@@ -90,10 +90,7 @@ impl<T: MiddlewareExt> MiddlewareWrapper<T> {
 
     /// Retry resolution after a 401: if a concurrent retry already renewed (the stored token
     /// differs from `previous`), reuse that result. Otherwise renew.
-    async fn resolve_retry(
-        &self,
-        previous: Option<String>,
-    ) -> Result<Option<String>, LoginError> {
+    async fn resolve_retry(&self, previous: Option<String>) -> Result<Option<String>, LoginError> {
         let mut handler = self.0.lock().await;
         if let Some((access_token, _)) = handler.current_token().await
             && let Some(prev) = &previous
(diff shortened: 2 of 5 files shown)


JaredSnider-Bitwarden (commented):
All of this looks great to me, but Claude is saying these particular test gaps might be worth filling?

1. **Retry runs exactly once.** No test confirms that a second 401 surfaces to the caller instead of triggering a third request.
2. **Non-401 responses don't retry.** A 500 on an `AuthRequired` request should pass through unchanged — currently untested.
3. **Streaming bodies skip the retry.** The `try_clone`-fails path is called out in the PR description but never exercised.
4. **Concurrent 401 retries collapse to one renewal.** The current concurrency test only covers the *proactive* renewal path (token already expired locally). The retry path — apparently-valid tokens that the server has invalidated — isn't covered.

Thoughts?

dani-garcia:
@JaredSnider-Bitwarden Good catches, added tests for all four gaps:
- `surfaces_second_401_without_third_attempt`: confirms the second 401 surfaces to the caller with no third attempt
- `does_not_retry_on_non_401_status`: confirms 500 passes through without renewal
- `does_not_retry_when_body_cannot_be_cloned`: exercises the `try_clone` → `None` path
- `concurrent_401s_trigger_a_single_renewal`: covers the 401-retry coalescing path (both handlers)

Also while trying to cleanly ensure serialized requests on retry, I took the opportunity to clean up the middleware internals and split `get_token` into two methods:
- `current_token()`: returns the stored token and its expiry
- `renew_token()`: performs the renewal

This allows the middleware to own all renewal policy. `MiddlewareWrapper` wraps `T` in a `tokio::sync::Mutex` to ensure serialization, and two private methods do the work:
- `resolve_initial()`: acquires the lock, checks expiry against the renewal margin, calls `renew_token()` only if needed
- `resolve_retry()`: acquires the lock after a 401, checks whether the stored token already differs from what was sent (meaning a concurrent retry already renewed), calls `renew_token()` only if it hasn't

Because both paths hold the mutex for the full duration of the renewal, concurrent requests queue up rather than racing. As another benefit of the middleware owning all the renewal policy, I've also eliminated some test duplication between password manager and secrets manager.

JaredSnider-Bitwarden (approved):
Thank you for adding those tests + the refactor looks great!

PR #1112, opened 2026-05-20
[PM-37570] Make UpgradeTokenAction required

Thomas-Avery:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-37570

## 📔 Objective

Make UpgradeTokenAction required for user key rotation. This depends on https://github.com/bitwarden/clients/pull/20711 merging into clients before we merge this in. There are currently no mobile consumers for user key rotation.

## 🚨 Breaking Changes

This is a breaking change of requiring the UpgradeTokenAction for user key rotation. Clients will already be passing this in after https://github.com/bitwarden/clients/pull/20711 is merged.


crates/bitwarden-user-crypto-management/src/key_rotation/rotate_user_keys.rs
```diff
@@ -303,6 +298,7 @@ mod tests {
             Err(RotateUserKeysError::UnimplementedKeyRotationMethod)
         ));
         if let ApiClient::Mock(mut mock) = api_client {
+            mock.sync_api.checkpoint();
```
eligrubb: ❓ what is the purpose of the new `checkpoint()` call?

eligrubb (dismissed):
One question, otherwise looks good!

crates/bitwarden-user-crypto-management/src/key_rotation/rotate_user_keys.rs
Thomas-Avery: Good catch looks like this was removed in the upstream and I missed it when doing the merge conflict.

Fixed in https://github.com/bitwarden/sdk-internal/pull/1112/commits/b7e2a6bb763f491f111ea2c985b06daf9d29de4d

eligrubb (approved):

PR #1121, opened 2026-05-21
[PM-32696] Blob aware decrypt

shane-melton:
## 🎟️ Tracking

[PM-32696](https://bitwarden.atlassian.net/browse/PM-32696)

## 📔 Objective

Routes blob-encrypted ciphers through the blob unseal flow transparently from the `Decryptable` impls on `Cipher` and `StrictDecrypt<Cipher>`, so external callers (`bitwarden-exporters`, key rotation, tests) keep
  calling `key_store.decrypt(&cipher)` unchanged while supporting both legacy field-level and new blob storage formats in the same `Vec<Cipher>`.

 ### Key changes

  - `Cipher::decrypt` (for both `CipherView` and `CipherListView`) now inspects `blob::is_blob_encrypted(&Cipher)` and dispatches per-cipher: blob ciphers go through `decrypt_blob_cipher`; legacy ciphers go through the
   existing lenient field-level path. `StrictDecrypt<Cipher>` mirrors the same dispatch (blob path is all-or-nothing; strict/lenient only governs the legacy branch).
  - Added `CipherView::to_list_view` so the blob path can unseal once to `CipherView` and project to `CipherListView` without re-decrypting. Required exposing per-type subtitle helpers (`build_subtitle_card`,
  `build_subtitle_identity`, `build_subtitle_drivers_license`, `build_subtitle_passport`) and a new `LoginView::to_list_view` that re-encrypts `totp` under the cipher key to preserve the `LoginListView` API contract.
  - Existing field-level decryption bodies extracted into `lenient_decrypt_cipher_*` / `strict_decrypt_cipher_*` free functions (no behavior change in the legacy branch).
  - Simplified `CiphersClient` decrypt methods now that wrapper selection is the only branching they do; refactored `CipherAdminClient::list_all` similarly.
  - New `crates/bitwarden-vault/src/cipher/README.md` documenting the type layers, storage formats, dispatch matrix, and rollout state (including the `StrictDecrypt` / `PM-34500-strict_cipher_decryption` removal
  tracked under PM-34531).

[PM-32696]: https://bitwarden.atlassian.net/browse/PM-32696?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


crates/bitwarden-vault/src/cipher/cipher_client/admin/get.rs
```diff
@@ -78,27 +117,14 @@ impl CipherAdminClient {
...
-            list_views,
-        })
+        list_org_ciphers(
+            org_id,
+            include_member_items,
+            &self.api_configurations.api_client,
+            &self.key_store,
+            self.is_strict_decrypt().await,
+        )
+        .await
```
nikwithak, line 127: Was there a specific reason for moving this into a standalone function? I believe the current guidance is to avoid the standalone functions and do everything within the Client implementation.

nikwithak: Note that doing so also lets us re-use any logic added in `CiphersClient::decrypt`/`::encrypt`, so we don't have to fragment the logic for thing like `strict_decrypt`

nikwithak (changes requested):
Just one note about the ciphersclient functions, but otherwise looks good - great work!

crates/bitwarden-vault/src/cipher/cipher_client/admin/get.rs
shane-melton: There wasn't a specific reason and was a leftover from an earlier commit/implementation attempt. I'm good to roll it back into the client; though then it will be the odd one out since our other operations `create.rs`, `edit.rs`, etc. are using standalone functions.

As for re-using the `CiphersClient::decrypt`/`::encrypt` logic, I don't think we'll be able to re-use it in the admin methods since they're different underlying clients. We'd have to construct a normal `CipherClient` (which may be more expensive with the repository inside) to get access to that shared logic as well.

nikwithak: Good callout, I missed the fact that this was the `admin` client which doesn't have those functions.

It's no big lift either way - I'm planning to refactor these as a BEEEP initiative at some point anyway, so it'll roll up into that eventually either way.

nikwithak (dismissed):

crates/bitwarden-vault/src/cipher/cipher.rs
```diff
@@ -1030,93 +1038,388 @@ impl CipherView {
...
+                        b.pin.as_ref().map(|_| CopyableCipherFields::BankAccountPin),
+                        b.iban
+                            .as_ref()
+                            .map(|_| CopyableCipherFields::BankAccountIban),
+                    ]
+                    .into_iter()
+                    .flatten()
+                    .collect()
+                })
+                .unwrap_or_default(),
```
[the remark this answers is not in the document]
shane-melton: Fixed in [e112551](https://github.com/bitwarden/sdk-internal/pull/1121/commits/e112551dece49ed8e11047ba5aadf9302c3fbb23) and tests added in [2bda8b8](https://github.com/bitwarden/sdk-internal/pull/1121/commits/2bda8b8c729133d4e85a733c94e8b0f817c5c201) to catch it in the future.

crates/bitwarden-vault/src/cipher/cipher.rs
```diff
@@ -1030,93 +1038,397 @@ impl CipherView {
...
+        let attachments_count = self
+            .attachments
+            .as_ref()
+            .map(|a| a.len() as u32)
+            .unwrap_or(0);
+        let has_old_attachments = self
+            .attachments
+            .as_ref()
+            .map(|a| a.iter().any(|att| att.key.is_none()))
+            .unwrap_or(false);
```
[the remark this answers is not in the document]
shane-melton: Fixed in [f8fbdd2](https://github.com/bitwarden/sdk-internal/pull/1121/commits/f8fbdd218d6cdab31541a06035c3210baff94701)

nikwithak (approved):

PR #1122, opened 2026-05-22
[PM-32695] Cipher blob encryption

shane-melton:
## 🎟️ Tracking

[PM-32695](https://bitwarden.atlassian.net/browse/PM-32695)

## 📔 Objective

Potential alternative to #1076. 

Routes blob-encrypted ciphers through the encrypt path symmetrically to the decrypt-side dispatch added in PM-32696. A new `EncryptMode<CipherView>` selects per-cipher whether to seal sensitive content into the blob
  format or take the legacy field-level path, with `CiphersClient` choosing based on the user's security state version. Builds on  #1121.

 ### Key changes

  - New `EncryptMode<T>` enum (`Blob(T)` / `Legacy(T)`) with `CompositeEncryptable<…, Cipher>` impl on `EncryptMode<CipherView>`. The blob arm calls `encrypt_blob_cipher`; the legacy arm delegates to the existing
  `CompositeEncryptable` impl on `CipherView`. `IdentifyKey` delegates to the inner view so `encrypt_list` still picks the correct scope key.
  - `CiphersClient::should_use_blob_encryption` (lifted to a free function so admin paths can share it) gates dispatch: personal-vault ciphers qualify once the user's security state reaches `BLOB_SECURITY_VERSION`;
  organization ciphers stay legacy until PM-32430.
  - All encrypt entry points now wrap views in `EncryptMode` before calling `key_store.encrypt`: `CiphersClient::encrypt`, `encrypt_list`, `encrypt_cipher_for_rotation`, `create_cipher` / `edit_cipher`, and the admin
  equivalents.
  - New `encrypt_blob_cipher_with_wrapping_key` variant for `encrypt_cipher_for_rotation`, which installs the new user key under a `Local` slot rather than the view's natural `User`/`Organization` slot.
  - `Cipher.name` becomes `Option<EncString>` — blob-encrypted ciphers don't carry a top-level name (it lives inside the sealed blob). Strict-decrypt and legacy paths still treat a missing name as an error.
  - Decrypt-side dispatch refactored: `is_blob_encrypted` removed; replaced by `try_parse_blob` that parses the `SealedCipherBlob` once and passes it into `decrypt_blob_cipher`, so the legacy/blob branch and the unseal
   share work.

[PM-32695]: https://bitwarden.atlassian.net/browse/PM-32695?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


crates/bitwarden-vault/src/cipher/cipher.rs
```diff
@@ -1701,15 +1717,59 @@ fn strict_decrypt_cipher_list_view(
...
+/// "blob operation failed".
+pub(crate) fn blob_err_to_crypto(err: BlobEncryptionError) -> CryptoError {
+    tracing::warn!(error = %err, error_debug = ?err, "blob decryption failed");
     match err {
         BlobEncryptionError::Crypto(c) => c,
-        BlobEncryptionError::SealedBlob(_) | BlobEncryptionError::NoBlobData => {
-            CryptoError::Decrypt
+        BlobEncryptionError::SealedBlob(_) => CryptoError::Decrypt,
+    }
+}
```
nikwithak: This would work better as `impl From<BlobEncryptionError> for CryptoError{...}`, which also would let you just use `?` anywhere you need this conversion.

crates/bitwarden-vault/src/cipher/cipher_client/mod.rs
```diff
@@ -36,6 +39,24 @@ mod move_many;
...
+) -> bool {
+    if organization_id.is_some() {
+        return false;
+    }
+    client
+        .internal
+        .get_key_store()
+        .context()
+        .get_security_state_version()
+        >= BLOB_SECURITY_VERSION
```
nikwithak, line 56: Would it work to put this logic inside the `CompositeEncryptable` impl for `Cipher`? Since it uses the `KeyStoreContext` internally and not feature flags this logic would still be compatible without needing `CiphersClient` internals.

crates/bitwarden-vault/src/cipher/cipher.rs
shane-melton: Good idea! Converted to a `From` impl in [f7d9d8f](https://github.com/bitwarden/sdk-internal/pull/1122/commits/f7d9d8fa9dd17de445f32bf4a947c18a555794b9)

crates/bitwarden-vault/src/cipher/cipher.rs
```diff
@@ -147,7 +147,11 @@ impl TryFrom<EncryptionContext> for CipherWithIdRequestModel {
             favorite: cipher.favorite.into(),
             reprompt: Some(cipher.reprompt.into()),
             key: cipher.key.map(|k| k.to_string()),
-            name: cipher.name.to_string(),
+            name: cipher
+                .name
+                .as_ref()
+                .map(ToString::to_string)
+                .unwrap_or_default(),
```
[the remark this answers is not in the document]
shane-melton: Fixed in [86c0f3a](https://github.com/bitwarden/sdk-internal/pull/1122/commits/86c0f3a996f4c6c351361bfe51eb42e0f9c16be1)

shane-melton:
@claude Review this again

crates/bitwarden-vault/src/cipher/blob/encryption.rs
```diff
@@ -132,18 +148,18 @@ pub(crate) fn encrypt_blob_cipher(
...
-/// Decrypts a blob-encrypted `Cipher` into a `CipherView`.
-///
-/// Unseals the blob data, decrypts attachments and local data, then applies
-/// the blob content fields onto the view.
+/// Decrypts a pre-parsed blob-encrypted `Cipher` into a `CipherView`. Callers
+/// should obtain the [`SealedCipherBlob`] via [`try_parse_blob`].
+#[instrument(err, skip_all, fields(cipher_id = ?cipher.id, org_id = ?cipher.organization_id))]
 pub(crate) fn decrypt_blob_cipher(
     cipher: &Cipher,
+    sealed: SealedCipherBlob,
```
quexten: nit:
```suggestion
    sealed: &SealedCipherBlob,
```

no need to pass ownership. Otherwise this will lead to expensive clones. For large vaults in a decrypt loop, we'd want these to be zero copy, otherwise decrypt will be slow(er).

quexten (dismissed):

crates/bitwarden-vault/src/cipher/blob/encryption.rs
shane-melton: Changed in [9c98988](https://github.com/bitwarden/sdk-internal/pull/1122/commits/9c989889f9323b9429fd77e811ec7fcd14f5f820)

quexten (dismissed):

crates/bitwarden-vault/src/cipher/cipher.rs
```diff
@@ -1456,7 +1456,7 @@ impl Decryptable<KeySlotIds, SymmetricKeySlotId, CipherView> for Cipher {
         key: SymmetricKeySlotId,
     ) -> Result<CipherView, CryptoError> {
         match try_parse_blob(self) {
-            Some(sealed) => decrypt_blob_cipher(self, sealed, ctx).map_err(blob_err_to_crypto),
+            Some(sealed) => decrypt_blob_cipher(self, sealed, ctx).map_err(CryptoError::from),
```
nikwithak: I think you can get away with `decrypt_blob_cipher(self, sealed, ctx)?` now that you have the `From` impl

nikwithak (dismissed):
A couple of minor comments but nothing blocking.

nick-livefront (dismissed):

nick-livefront (approved):

djsmith85 (approved):

adudek-bw (approved):
LGTM on the tools side :)

jlf0dev (approved):

PR #1124, opened 2026-05-22
[PM-37933] refactor: Introduce OrganizationUserPolicyContext model

eliykat:
## 🎟️ Tracking

[PM-37933](https://bitwarden.atlassian.net/browse/PM-37933)

## 📔 Objective

The policies crate requires the full `ProfileOrganization` sync response model to be passed in for policy evaluation. However, that model is different between mobile and TS clients - the mobile sync model is much smaller because they need fewer admin-related properties. This tight coupling would require mobile to expand their organization model just to meet the interface, even though the logic doesn't actually need all these properties.

The fix is to introduce a domain-specific `PolicyOrganizationContext` in `bitwarden-policies` and switch the filtering APIs and FFI wrapper to use it instead of the 61-field `ProfileOrganization`. The policy domain only needs 6 fields; this gives consumers a small, purpose-built input type to construct at the FFI boundary. This decoupling generally seems like better design for a library crate.

## 🚨 Breaking Changes

The FFI signature of `filter_by_type` now takes `PolicyOrganizationContext` instead of `ProfileOrganization`. **No client is consuming this API yet**, so there is no migration impact in practice.

[PM-37933]: https://bitwarden.atlassian.net/browse/PM-37933?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/crates/bitwarden-policies/src/filter.rs
+++ b/crates/bitwarden-policies/src/filter.rs
@@ -57,15 +57,17 @@ pub trait PolicyFilter: Policy {
     /// This evaluates common business rules (e.g. the policy is enabled),
     /// as well as policy-specific rules according to its [`Policy`].
     ///
-    /// If a policy's organization is not present in `organizations`, the policy is enforced by
-    /// default.
+    /// If a policy's organization is not present in `organization_user_policy_contexts`, the policy
+    /// is enforced by default.
     fn filter<'a>(
         &self,
         policies: &'a [PolicyView],
-        organizations: &[PolicyOrganizationContext],
+        organization_user_policy_contexts: &[PolicyOrganizationContext],
     ) -> Vec<&'a PolicyView> {
-        let org_map: HashMap<&Uuid, &PolicyOrganizationContext> =
-            organizations.iter().map(|o| (&o.id, o)).collect();
+        let org_map: HashMap<&Uuid, &PolicyOrganizationContext> = organization_user_policy_contexts
+            .iter()
+            .map(|o| (&o.id, o))
+            .collect();
 
         policies
             .iter()
--- a/crates/bitwarden-policies/src/filter.rs
+++ b/crates/bitwarden-policies/src/filter.rs
@@ -11,7 +11,7 @@ use bitwarden_organizations::{OrganizationUserStatusType, OrganizationUserType};
 use uuid::Uuid;
 
 use crate::{
-    models::{PolicyOrganizationContext, PolicyView},
+    models::{OrganizationUserPolicyContext, PolicyView},
     policy_type::PolicyType,
 };
 
@@ -62,12 +62,13 @@ pub trait PolicyFilter: Policy {
     fn filter<'a>(
         &self,
         policies: &'a [PolicyView],
-        organization_user_policy_contexts: &[PolicyOrganizationContext],
+        organization_user_policy_contexts: &[OrganizationUserPolicyContext],
     ) -> Vec<&'a PolicyView> {
-        let org_map: HashMap<&Uuid, &PolicyOrganizationContext> = organization_user_policy_contexts
-            .iter()
-            .map(|o| (&o.id, o))
-            .collect();
+        let org_map: HashMap<&Uuid, &OrganizationUserPolicyContext> =
+            organization_user_policy_contexts
+                .iter()
+                .map(|o| (&o.id, o))
+                .collect();
 
         policies
             .iter()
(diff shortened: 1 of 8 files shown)


crates/bitwarden-policies/src/policy_type.rs
```diff
@@ -50,7 +50,7 @@ pub enum PolicyType {
     /// Prevents members from unlocking the app with a PIN.
     RemoveUnlockWithPin = 14,
     /// Restricts the item types that members can create.
-    RestrictedItemTypesPolicy = 15,
+    RestrictedItemTypes = 15,
```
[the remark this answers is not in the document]
eliykat: @claude fixed

crates/bitwarden-policies/src/models.rs
```diff
@@ -0,0 +1,45 @@
...
+
+use crate::policy_type::PolicyType;
+
+/// An organization policy.
+#[allow(missing_docs)]
+#[derive(Serialize, Deserialize, Debug, Clone)]
+#[serde(rename_all = "camelCase")]
+#[cfg_attr(feature = "uniffi", derive(uniffi::Record))]
+#[cfg_attr(feature = "wasm", derive(Tsify), tsify(into_wasm_abi, from_wasm_abi))]
+pub struct PolicyView {
```
eliykat, line 20: Moved as-is from `filter.rs` because it didn't really belong in there.

crates/bitwarden-policies/src/models.rs
```diff
@@ -0,0 +1,45 @@
...
+    pub r#type: PolicyType,
+    /// The policy's raw configuration data as a JSON string, if any.
+    pub data: Option<String>,
+    pub enabled: bool,
+    pub revision_date: Option<DateTime<Utc>>,
+}
+
+/// Minimal organization data needed to evaluate organization policies for a
+/// user.
+#[allow(missing_docs)]
```
quexten:
```suggestion
```

Nit: https://contributing.bitwarden.com/contributing/code-style/rust#documentation in rust, unless we have a good reason we should have docs for the various fields. In this case, there is several fields for which I would not know what they mean by just the names alone, so documentation here is preferred.

crates/bitwarden-policies/src/models.rs
```diff
@@ -0,0 +1,45 @@
...
+    pub organization_id: Uuid,
+    pub r#type: PolicyType,
+    /// The policy's raw configuration data as a JSON string, if any.
+    pub data: Option<String>,
+    pub enabled: bool,
+    pub revision_date: Option<DateTime<Utc>>,
+}
+
+/// Minimal organization data needed to evaluate organization policies for a
+/// user.
```
quexten:
```suggestion
/// A minimal set of data for a user in an organization. This provides
/// the context needed to evaluate the policies that are applied to the
/// user by the organization.
```

just as a non-domain expert, the above comment is a bit clearer.

crates/bitwarden-policies/src/filter.rs
```diff
@@ -80,9 +62,9 @@ pub trait PolicyFilter: Policy {
     fn filter<'a>(
         &self,
         policies: &'a [PolicyView],
-        organizations: &[ProfileOrganization],
+        organizations: &[PolicyOrganizationContext],
```
quexten:
```suggestion
        organization_user_policy_contexts: &[PolicyOrganizationContext],
```

As far as i understand, the PolicyOrganizationContext is a set of data unique to a (org, user) pair. This is currently not clear from the variable naming.

crates/bitwarden-policies/src/registry.rs
```diff
@@ -50,7 +48,7 @@ impl PolicyRegistry {
     pub(crate) fn filter_by_type<'a>(
         &self,
         policies: &'a [PolicyView],
-        organizations: &[ProfileOrganization],
+        organizations: &[PolicyOrganizationContext],
```
quexten:
```suggestion
        organization_user_policy_contexts: &[PolicyOrganizationContext],
```

crates/bitwarden-policies/src/policy_client.rs
```diff
@@ -51,7 +53,7 @@ impl PolicyClient {
     pub fn filter_by_type(
         &self,
         policies: Vec<PolicyView>,
-        organizations: Vec<ProfileOrganization>,
+        organizations: Vec<PolicyOrganizationContext>,
```
quexten:
```suggestion
        organization_user_policy_contexts: &[PolicyOrganizationContext],
```

quexten (commented):
Looks good. A few comments for clarity from the perspective of a non-domain-owner.

crates/bitwarden-policies/src/filter.rs
eliykat: Good point. I have renamed the struct as well for consistency.

quexten (approved):

r-tome (approved):

dereknance (approved):

PR #1127, opened 2026-05-22
[PM-37971] Add wipe support to the StateRegistry

dani-garcia:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-37971

## 📔 Objective

Add a `wipe()` method to `StateRegistry` for use during logout, ensuring all SDK-managed state is removed when a user signs out.

- `StateRegistry::wipe()` clears the client-managed pool first, then wipes the SDK-managed database, including the sqlite files themselves.
- Each `Database` backend implements `wipe()` and transitions to a closed state under a lock; subsequent calls return `DatabaseError::Closed` so outstanding repository handles fail cleanly instead of operating on stale storage.
  - **Memory**: drops the in-memory map.
  - **SQLite**: drops the connection under the lock, then unlinks the `.sqlite`, `.sqlite-wal`, and `.sqlite-shm` files.
  - **IndexedDB**: wraps the database in a `DatabaseHandle` whose `Drop` calls `close()`, then awaits `Factory::delete_database`. Verified with a browser integration test that reopening yields a fresh DB.

This is intended to be invoked right before the `Client` is dropped, this is due to the fact that it leaves the `StateRegistry` in an inoperable state, where operations will return an error from now on. As such, this should be executed right at the end of the logout process only.

The PR is divided in two commits, the first one is just a preparation refactor that has no functionality changes. The second commit contains the actual wipe implementation and some browser tests.


crates/bitwarden-state/tests/indexed_db.rs
coroiu, line 1: suggestion: would it be possible to add this to the CI test runner
https://github.com/bitwarden/sdk-internal/blob/77391b703a519f8aa18c61d9cf6ba69b75850266/.github/workflows/rust-test.yml#L83-L84

coroiu (dismissed):

crates/bitwarden-state/tests/indexed_db.rs
dani-garcia: I wasn't sure if the CI runner had a browser and driver setup, but turns out it does! I've added this test to CI now.

coroiu (approved):

PR #1129, opened 2026-05-24
[PM-37875] Implement set state after key rotation

quexten:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-37875

## 📔 Objective

In order to not log out or lock the rotating device, we need to set at least:
- User key
- Account cryptographic state
- V2 upgrade token

If we do not do this, then there is a possibility of race conditions between the sync coming in, and the sdk initializing, with the automatic migration scheduler, triggering a double migration.

Please note all unlock methods should also be updated after the rotation. This is handled by a full sync after the rotation for now.


crates/bitwarden-user-crypto-management/src/key_rotation/crypto.rs
```diff
@@ -32,32 +32,24 @@ pub(super) fn rotate_account_cryptographic_state_to_request_model(
     Ok(account_keys_model)
 }
 
-/// Rotates an account cryptographic state and upgrades it to V2 if necessary.
+/// Converts a rotated account cryptographic state to a wrapped model for the request.
 /// This function fails and logs an error via tracing if the passed keys are invalid, or if the
 /// account cryptographic state is malformed.
-pub(super) fn rotate_account_cryptographic_state_to_wrapped_model(
+pub(super) fn account_cryptographic_state_to_wrapped_model(
```
quexten, line 38: @Thomas-Avery I have sent a a DM about this. IMO in the long term (not here directly?) we should refactor this to act on the data models, performing the full rotations on just the data models. Then we can call:
- sync_to_server(data model)
  - Converts data models to requests models
- set to state(data model)

that way we don't need to do back conversions, and we don't couple cryptographic logic to synchronization logic

coroiu (dismissed):

crates/bitwarden-user-crypto-management/src/key_rotation/rotate_user_keys.rs
```diff
@@ -223,6 +258,20 @@ async fn internal_rotate_user_keys(
         .await
         .map_err(|_| RotateUserKeysError::Api)?;
     info!("Successfully rotated user account keys and data");
+
+    if let Some(upgrade_token) = state_bridge_update.upgrade_token.as_ref() {
```
mzieniukbw, line 250: 🎨 Test coverage is low here.

mzieniukbw (dismissed):

crates/bitwarden-user-crypto-management/src/key_rotation/rotate_user_keys.rs
quexten: Added tests!

mzieniukbw (approved):

coroiu (approved):

PR #1139, opened 2026-05-26
Add bitwarden-logging-macro with safer instrument wrapper

coroiu:
## 📔 Objective

`tracing::instrument` records every argument that implements `Display`/`Debug` by default, so forgetting `skip_all` on a function like `derive_master_key` would log the user's password. This addresses Architecture's logging concerns.

Introduces `#[bitwarden_logging::instrument]`, a drop-in wrapper that rewrites to `#[tracing::instrument(skip_all, ...)]`, making field logging opt-in via `fields(...)`. User-supplied `skip(...)` or `skip_all` are rejected at compile time so the safety property is enforced and explicit.

A `tracing_instrument` dylint backstops the convention by warning on both `#[tracing::instrument]` and bare `#[instrument]` (which can hide the foot-gun behind a `use tracing::instrument`). Engineers should always write the fully-qualified `#[bitwarden_logging::instrument]` so the safety property is visible at every call site.

Migration of existing call sites is intentionally left for a follow-up.

### DIFF
--- a/crates/bitwarden-logging-macro/README.md
+++ b/crates/bitwarden-logging-macro/README.md
@@ -6,8 +6,8 @@ By default, `tracing::instrument` records every function argument that implement
 `Debug` as a span field. In a vault-handling SDK this is a foot-gun: forgetting `skip_all` on a
 function like `derive_master_key(password, email, ...)` would log the user's password.
 
-The `#[bitwarden_logging::instrument]` attribute re-emits as `#[tracing::instrument(skip_all, ...)]`,
-making field logging opt-in via `fields(...)`.
+The `#[bitwarden_logging::instrument]` attribute re-emits as
+`#[tracing::instrument(skip_all, ...)]`, making field logging opt-in via `fields(...)`.
 
 ```rust,ignore
 use bitwarden_logging::instrument;
@@ -21,23 +21,23 @@ fn sensitive(password: &str) { /* ... */ }
 fn less_sensitive(id: Uuid, password: &str) { /* ... */ }
 ```
 
-User-supplied `skip(...)` and `skip_all` are rejected at compile time, since `skip_all` is
-already enforced and `fields(...)` is the way to opt in.
+User-supplied `skip(...)` and `skip_all` are rejected at compile time, since `skip_all` is already
+enforced and `fields(...)` is the way to opt in.
 
 ## Lint backstop
 
-A companion dylint rule (`tracing_instrument`) catches any use of `tracing::instrument` that
-slips past the convention. It identifies the macro by its definition in the `tracing_attributes`
-crate, so all of these are caught:
+A companion dylint rule (`tracing_instrument`) catches any use of `tracing::instrument` that slips
+past the convention. It identifies the macro by its definition in the `tracing_attributes` crate, so
+all of these are caught:
 
 - `#[tracing::instrument]`
 - `use tracing::instrument; #[instrument]`
 - `use tracing::instrument as foo; #[foo]`
 
 The wrapper opts out of the lint at its own emission site by including
-`#[allow(unknown_lints, tracing_instrument)]` next to the `#[tracing::instrument]` it
-generates, so the lint can stay general.
+`#[allow(unknown_lints, tracing_instrument)]` next to the `#[tracing::instrument]` it generates, so
+the lint can stay general.
 
-The lint currently defaults to `allow` so existing workspace call sites have time to migrate.
-Crates that have been swept can opt in with `#![warn(tracing_instrument)]` (or `deny`) at the
-crate root. The default will flip to `warn` once the workspace is clean.
+The lint currently defaults to `allow` so existing workspace call sites have time to migrate. Crates
+that have been swept can opt in with `#![warn(tracing_instrument)]` (or `deny`) at the crate root.
+The default will flip to `warn` once the workspace is clean.
--- a/crates/bitwarden-logging-macro/README.md
+++ b/crates/bitwarden-logging-macro/README.md
@@ -34,8 +34,9 @@ crate, so all of these are caught:
 - `use tracing::instrument; #[instrument]`
 - `use tracing::instrument as foo; #[foo]`
 
-Expansions emitted by `bitwarden_logging::instrument` itself (which internally re-emits
-`tracing::instrument`) are filtered out.
+The wrapper opts out of the lint at its own emission site by including
+`#[allow(unknown_lints, tracing_instrument)]` next to the `#[tracing::instrument]` it
+generates, so the lint can stay general.
 
 The lint currently defaults to `allow` so existing workspace call sites have time to migrate.
 Crates that have been swept can opt in with `#![warn(tracing_instrument)]` (or `deny`) at the
--- a/crates/bitwarden-logging-macro/README.md
+++ b/crates/bitwarden-logging-macro/README.md
@@ -10,26 +10,33 @@ The `#[bitwarden_logging::instrument]` attribute re-emits as `#[tracing::instrum
 making field logging opt-in via `fields(...)`.
 
 ```rust,ignore
+use bitwarden_logging::instrument;
+
 // All args skipped by default.
-#[bitwarden_logging::instrument]
+#[instrument]
 fn sensitive(password: &str) { /* ... */ }
 
 // Explicit opt-in.
-#[bitwarden_logging::instrument(fields(user_id = ?id))]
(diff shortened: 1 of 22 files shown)


support/lints/tracing_instrument/src/lib.rs
```diff
@@ -0,0 +1,104 @@
...
+    /// ```rust,ignore
+    /// #[bitwarden_logging::instrument]
+    /// fn derive_master_key(password: &str) {}
+    /// ```
+    pub TRACING_INSTRUMENT,
+    Allow,
+    "use the fully-qualified `#[bitwarden_logging::instrument]` instead of `#[tracing::instrument]` or bare `#[instrument]`"
+}
+
+impl EarlyLintPass for TracingInstrument {
```
dani-garcia: If you change this into a `LateLintPass`, you'll have type information and you'll be able to match against the `instrument` macro from the `tracing` crate specifically, which should avoid the need to require the fully qualified `bitwarden_logging::instrument`. It should also work if someone were to import the macro with a different name. Something like this:

```rust
impl<'tcx> LateLintPass<'tcx> for TracingInstrument {
    fn check_item(&mut self, cx: &LateContext<'tcx>, item: &'tcx Item<'tcx>) {
        check_span(cx, item.span);
    }
    fn check_impl_item(&mut self, cx: &LateContext<'tcx>, item: &'tcx ImplItem<'tcx>) {
        check_span(cx, item.span);
    }
    fn check_trait_item(&mut self, cx: &LateContext<'tcx>, item: &'tcx TraitItem<'tcx>) {
        check_span(cx, item.span);
    }
}

fn check_span(cx: &LateContext<'_>, span: Span) {
    for expn in span.macro_backtrace() {
        let ExpnKind::Macro(MacroKind::Attr, _) = expn.kind else { continue; };
        let Some(def_id) = expn.macro_def_id else { continue; };
        if cx.tcx.crate_name(def_id.krate).as_str() != "tracing_attributes" { continue; }
        span_lint_and_help(...);
        return;
    }
}
```

Also I'd recommend using `span_lint_and_sugg` instead of `span_lint_and_help`, as you can provide a machine applicable suggestion with that one.

coroiu: Great suggestion, thanks — done in 526d1af6c.

Switched to `LateLintPass` and now match by `cx.tcx.crate_name(def_id.krate) == "tracing_attributes"` walking the macro backtrace. Catches all the forms (`#[tracing::instrument]`, bare `#[instrument]` after `use tracing::instrument`, and aliased imports). The wrapper's internal `tracing::instrument` re-emission is filtered out by also checking for `bitwarden_logging_macro` in the same backtrace.

This also lets me drop the awkward "always fully qualify" convention from the docs — `use bitwarden_logging::instrument; #[instrument]` is now perfectly fine, and the lint distinguishes by macro identity rather than path text.

On `span_lint_and_sugg`: considered it but stuck with `span_lint_and_help` for now. A machine-applicable rewrite isn't quite safe here because users currently passing `skip(...)` would need their args translated to `fields(...)` opt-ins, which isn't mechanical. Open to revisiting with `Applicability::MaybeIncorrect` if you'd find that more useful than a help line.

crates/bitwarden-logging-macro/README.md
```diff
@@ -0,0 +1,42 @@
...
+- `#[tracing::instrument]`
+- `use tracing::instrument; #[instrument]`
+- `use tracing::instrument as foo; #[foo]`
+
+Expansions emitted by `bitwarden_logging::instrument` itself (which internally re-emits
+`tracing::instrument`) are filtered out.
+
+The lint currently defaults to `allow` so existing workspace call sites have time to migrate.
+Crates that have been swept can opt in with `#![warn(tracing_instrument)]` (or `deny`) at the
+crate root. The default will flip to `warn` once the workspace is clean.
```
Hinton: We should remember to update this.

Hinton (dismissed):

support/lints/tracing_instrument/src/lib.rs
```diff
@@ -0,0 +1,126 @@
...
+        let crate_name = cx.tcx.crate_name(def_id.krate);
+        match crate_name.as_str() {
+            TRACING_ATTRIBUTES_CRATE => {
+                // Only record the innermost `tracing::instrument` call site so the diagnostic
+                // points at the attribute the user actually wrote.
+                if tracing_call_site.is_none() {
+                    tracing_call_site = Some(expn.call_site);
+                }
+            }
+            WRAPPER_CRATE => {
```
dani-garcia: I wonder if rather than making an exception for the lint, we can just have the instrument macro emit `#[allow(tracing_instrument)]` on top of the tracing macro

dani-garcia (dismissed):
LGTM just a small comment.

Also CI is complaining that you have to run prettier on the readme hah

support/lints/tracing_instrument/src/lib.rs
coroiu: Alright, I've added suggestions where the conversion is obvious, and output a help text when it isn't 👍

support/lints/tracing_instrument/src/lib.rs
coroiu: This leaves the lint plugin clean, nice! Applied

dani-garcia (dismissed):

dani-garcia (approved):

quexten (dismissed):
TBH, I do not think this is the right approach or direction to take. If there are types we don't want logged, then we should implement Display/Debug for them. For things like passwords we could make a `SecretString` newtype that does not log. we could combine this with the dangerous crypto debug flag which *would* make all of these log.

That feels way more useful, because then I can flip a single flag to log all values (dangerously) as a dev at compile time, whereas if I change the macro to not log except opt in, I need to opt into every field.

This current approach will make it very unergonomic to get logging in all places, since now you need to edit all instruments, and with the above newtype approach, there would be no function where in KM code we would not instrument all variables. Effectively, this just makes writing the code for KM's purposes less ergonomic after the above newtype change, without any benefit.

quexten (changes requested):
> Migration of existing call sites is intentionally left for a follow-up.

This does not give KM which controls nearly all callsites of instrument which this affects any opportunity to comment on the approach.

Subsequently, I'll mark my review as request changes, to have a discussion about the approach, since you'll need KM approval either way to use this at the callsites.

support/lints/tracing_instrument/src/lib.rs
```diff
@@ -0,0 +1,176 @@
...
+    ///
+    /// Currently defaults to `Allow` to give existing call sites time to migrate without
+    /// breaking CI (the workspace runs `cargo dylint` with `-D warnings`). Crates that have
+    /// already been swept can opt in via `#![warn(tracing_instrument)]` (or `deny`) at the
+    /// crate root. The default will flip to `Warn` once the workspace is clean.
+    ///
+    /// ### Why is this bad?
+    ///
+    /// `tracing::instrument` records every argument that implements `Display`/`Debug` by
+    /// default. In a vault-handling SDK this is a foot-gun: forgetting `skip_all` on a
```
quexten, line 55: The footgun here is having loggable types, not having a macro that logs all params. The solution is not to ban logging, the solution is to make the types safe to log.

I also don't think this increases safety. We already log (and want to log) keys. It is reasonable to assume that we want to do the same with parameters such as password. These should be viewable in the log based on a compile time opt-in flag (dangerous crypto debug). As a reviewer it is hard to distinguish between these cases.

If we just consistently enforce a secret string type for secret information, then that is a way simpler rule to follow, while still getting us introspectability if we need it.

For example, currently a reviewer sees: "key, password" in a parameter list. Do they flag both? No they have to check whether each type is safe to log, and depending on that say "hey this field you opted into is bad". That is very easy to miss.

Instead, *ALL* secret values *MUST* be prevented from implementing a display that leaks information during prod builds. This is a way less error prone target to hit.

quexten: I'll note that during a local test branch, making this enforced by the type system already helped me find an instance that would not have been found by a reviewer -- salt is usually a value thats OK to log. However, a specific construction passes in the password as a salt. The type system caught this, so there is a second argument here, the type system will catch more cases of this than human review does.

coroiu:
@quexten anything you'd like me to change/add to this PR as a result of our discussion yesterday?

quexten:
@coroiu Sorry just did not loop around to resolving. As per the discussion yesterday, dismissing review.

quexten (approved):

PR #1140, opened 2026-05-26
[PM-24643] Add an easy to use formatting linting command to the sdk

coroiu:
## 📔 Objective

- Makes linting easier by using the same script for CI and local checks
- Makes linting faster in CI (17m -> 8m) by parallelizing the checks

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -27,7 +27,7 @@ jobs:
         include:
           - check: fmt
             needs_rust: true
-            needs_nightly: true
+            needs_rust_nightly: true
           - check: clippy
             needs_rust: true
           - check: sort
@@ -35,7 +35,7 @@ jobs:
             install: cargo install cargo-sort --version 2.0.2 --locked
           - check: udeps
             needs_rust: true
-            needs_nightly: true
+            needs_rust_nightly: true
             install: cargo install cargo-udeps --version 0.1.57 --locked
           - check: dylint
             needs_rust: true
@@ -64,7 +64,7 @@ jobs:
           echo "RUST_TOOLCHAIN=${RUST_TOOLCHAIN}" | tee -a "${GITHUB_OUTPUT}"
 
       - name: Set Rust nightly toolchain version
-        if: matrix.needs_nightly
+        if: matrix.needs_rust_nightly
         id: nightly-toolchain
         shell: bash
         run: |
@@ -79,7 +79,7 @@ jobs:
           components: clippy, rustfmt
 
       - name: Install Rust nightly
-        if: matrix.needs_nightly
+        if: matrix.needs_rust_nightly
         env:
           RUST_NIGHTLY_TOOLCHAIN: ${{ steps.nightly-toolchain.outputs.RUST_NIGHTLY_TOOLCHAIN }}
         run: |
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -12,6 +12,16 @@ env:
   CARGO_TERM_COLOR: always
 
 jobs:
+  check:
+    name: Check Style
+    runs-on: ubuntu-24.04
+    needs: style
+    if: always()
+    steps:
+      - name: Verify all style checks passed
+        if: ${{ needs.style.result != 'success' }}
+        run: exit 1
+
   style:
     name: ${{ matrix.check }}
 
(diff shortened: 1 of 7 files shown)


.github/workflows/lint.yml
```diff
@@ -13,127 +13,117 @@ env:
...
       contents: read
       security-events: write
 
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - check: fmt
+            needs_rust: true
+            needs_nightly: true
```
withinfocus: ⛏️ `needs_rust_nightly`

.github/workflows/lint.yml
```diff
@@ -13,127 +13,117 @@ env:
...
+          - check: clippy
+            needs_rust: true
+          - check: sort
+            needs_rust: true
+            install: cargo install cargo-sort --version 2.0.2 --locked
```
withinfocus, line 45: ❓ How are these versions monitored and maintained? Could be helpful to create a quick custom manager with Renovate to update them.

withinfocus (commented):
The AI changes are fine, but a couple things too.

Also, see https://github.com/bitwarden/template/pull/199 -- you don't have to use NPM as your linting gateway if you don't want to; the latest Git's hooks feature can be used for that and would be my preference for all new work.

scripts/lint.sh
Hinton, line 1: Windows support?

coroiu: Is that a thing we do?

coroiu: Also bash can be available on windows too

Hinton: Yes, all other projects runs on both windows and mac. Several of our developers use windows and we want to support external contributors. I don't think bash is generally available on windows unless you use WSL or git bash which I'm not sure we can assume?

What about just using a node script or powershell which is what we use in other places.

quexten: Just a note, we require devs to install powershell on linux/mac for `server` development, so requiring them to install bash on windows seems reasonable.

coroiu: Does `git` even work without bash? I think it's pretty safe to assume that bash exists on windows since it's a core component of git for windows. There are scaled down versions like `MinGit` that don't come with bash but `MinGit` is a lightweight, non-interactive distribution of Git for Windows, designed specifically for third-party applications like IDEs and graph visualizers, so I think we can safely scope that out. 

Unless support for Bash and Powershell is a general requirement we have documented somewhere that I've missed?

.github/workflows/lint.yml
coroiu: Renamed

.github/workflows/lint.yml
coroiu: Added to renovate

coroiu:
@withinfocus `npm` is mostly just used as a convenient way of running the bash script, and couldn't remove it even if we wanted to because we have javascript components that are linted using `prettier`. We could definitely change how git hooks are applied, but this PR deliberately doesn't touch the git hooks at all so I'd prefer to follow that up in a separate PR if we think it's something critical

withinfocus:
> this PR deliberately doesn't touch the git hooks at all so I'd prefer to follow that up in a separate PR if we think it's something critical

That's fine, just something to keep in mind as an improvement.

withinfocus (dismissed):

withinfocus:
You might need some admin help with merging this since you renamed a required check.

dereknance (dismissed):

scripts/lint.sh
Hinton: Doesn't `IDEs` imply we break commit flows in IDEs?

Powershell is a [documented](https://contributing.bitwarden.com/getting-started/tools/#local-environment) requirement, used in both server and browser extension, bash is not currently required on windows.

At a minimum we should test to ensure this works on windows.

coroiu:
@withinfocus Fixed without admin permissions, thanks for pointing me in the right direction!

scripts/lint.sh
addisonbeck: Thanks for the call out, we should definitely work to have all SDK tooling work across platforms. Considering bash is the current standard in the SDK I don't think we should block this work over it. I've added [this ticket](https://bitwarden.atlassian.net/browse/PM-38447) to the top of the Platform backlog for converting this script and others to powershell/ts/something else.

addisonbeck (approved):

withinfocus (approved):

PR #1141, opened 2026-05-26
[PM-38106] Implement streaming decryption for attachments

quexten:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-38106
Unblocking KM-side work for https://github.com/bitwarden/sdk-internal/pull/1093

## 📔 Objective

Adds support for streaming encryption and decryption of attachments on the crypto side.

There are a few goals which this PR balances:
- Low memory usage while keeping compatibility to the old attachment format
  - This presented limitations as will be clarified later
- Transparent forward compatibility with future, better encryption schemes
- Safe interface / no leaking of cryptographic implementation for consumers
- A standard interface that is easy to work with, within the ecosystem

The public interfaces for this are just standard `AsyncRead` / `AsyncWrite` implementations, making them simple to use, as they can be plugged into the standard ecosystem, such as tokio, and more importantly reqwest for HTTP requests without much custom code on the callers.

The interface is designed to be scalable to better encryption schemes. I have a local implementation of https://eprint.iacr.org/2015/189.pdf / https://docs.rs/aead/latest/aead/stream/index.html, which has much better properties, and validated that the interface scales to this and hides all cryptographic detail.

# AES256-CBC-HMAC
The AES256-CBC-HMAC implementation supports decryption and encryption of current attachments. Note: The existing format is particularly poorly suited to streamed decryption, because the mac is included as the header, not trailer. This means that we cannot emit any encrypted chunks before we have encrypted the full stream. Because of this, we need to buffer the encrypted chunks.

## Wasm Memory Hacks
Since WASM memory grows linearly, we need to introduce external memory via `Uint8array` that can be de-allocated again. We could implement this to cache to storage if required, but in this implementation this allocates in JS heap instead of expanding the JS memory. This makes it so that this at least does not affect the WASM memory and *can* be deallocated again.

# Future work - STREAM
Locally, I have verified that an implementation based on https://docs.rs/aead/latest/aead/stream/index.html works and is reasonable. The benefits are numerous:
- Emitted chunks are always integrity protected, even without reading the full stream
- Minimal memory usage since no full stream buffering is necessary
- Random access decryption possible
  - I added the traits for this in the PR but without an implementation, just to define the public API here.
- This will be the encryption scheme for the COSE typed keys - AES-GCM, CHACHA20-Poly1305

# Integration Test
Left out from this PR because it's unclear how vault will consume the API. I have a DM open with them.

### DIFF
--- a/crates/bitwarden-crypto/src/stream/aes256_cbc_hmac_legacy_stream.rs
+++ b/crates/bitwarden-crypto/src/stream/aes256_cbc_hmac_legacy_stream.rs
@@ -218,7 +218,7 @@ enum DecryptorState {
         key: Aes256CbcHmacKey,
     },
     Streaming {
-        decryptor: CbcDecryptor,
+        decryptor: Box<CbcDecryptor>,
         integrity_validator: HmacStreamValidator,
         expected_mac: Mac,
     },
@@ -241,7 +241,7 @@ impl DecryptorState {
         match self {
             Self::Uninitialized { key } => {
                 *self = Self::Streaming {
-                    decryptor: CbcDecryptor::new(&key.enc_key.0, &header.iv),
+                    decryptor: Box::new(CbcDecryptor::new(&key.enc_key.0, &header.iv)),
                     integrity_validator: HmacStreamValidator::new(&key.mac_key.0, &header.iv),
                     expected_mac: header.mac,
                 };
@@ -508,25 +508,28 @@ impl StreamingAes256CbcHmacEncryptor {
 
 impl StreamingEncryptor for StreamingAes256CbcHmacEncryptor {
     fn update(&mut self, plaintext_chunk: &[u8], last_block: bool) -> ChunkEncryptionResult {
-        let (encryptor, stream_validator, iv): (&mut Box<CbcEncryptor>, &mut HmacStreamValidator, &Iv) =
-            match &mut self.encryptor_state {
-                EncryptorState::Error | EncryptorState::Done => {
-                    return ChunkEncryptionResult::Error;
+        let (encryptor, stream_validator, iv): (
+            &mut Box<CbcEncryptor>,
+            &mut HmacStreamValidator,
+            &Iv,
+        ) = match &mut self.encryptor_state {
+            EncryptorState::Error | EncryptorState::Done => {
+                return ChunkEncryptionResult::Error;
+            }
+            EncryptorState::Emitting => {
+                if let Some(chunk) = self.ciphertext_buffer.emit_chunk() {
+                    return ChunkEncryptionResult::EncryptedChunk(chunk);
+                } else {
+                    self.encryptor_state = EncryptorState::Done;
+                    return ChunkEncryptionResult::FinalEncryptedChunk(Vec::new());
                 }
-                EncryptorState::Emitting => {
-                    if let Some(chunk) = self.ciphertext_buffer.emit_chunk() {
-                        return ChunkEncryptionResult::EncryptedChunk(chunk);
-                    } else {
-                        self.encryptor_state = EncryptorState::Done;
-                        return ChunkEncryptionResult::FinalEncryptedChunk(Vec::new());
-                    }
-                }
-                EncryptorState::Streaming {
-                    encryptor,
-                    integrity_validator,
-                    iv,
-                } => (encryptor, integrity_validator, iv),
-            };
+            }
+            EncryptorState::Streaming {
+                encryptor,
+                integrity_validator,
+                iv,
+            } => (encryptor, integrity_validator, iv),
+        };
 
         self.plaintext_buffer.extend_from_slice(plaintext_chunk);
 
--- a/crates/bitwarden-crypto/src/stream/aes256_cbc_hmac_legacy_stream.rs
+++ b/crates/bitwarden-crypto/src/stream/aes256_cbc_hmac_legacy_stream.rs
@@ -439,7 +439,7 @@ impl CiphertextBuffer {
 
     fn append(&mut self, data: &CbcCiphertextBlock) {
         self.inner
-            .copy_from_slice(self.size..self.size + data.len(), data)
+            .append(data)
             .expect("buffer should grow to fit");
         self.size += data.len();
     }
--- a/crates/bitwarden-crypto/src/stream/aes256_cbc_hmac_legacy_stream.rs
+++ b/crates/bitwarden-crypto/src/stream/aes256_cbc_hmac_legacy_stream.rs
@@ -1,0 +1,706 @@
+//! # AES-256-CBC-HMAC-Legacy-Stream
+//!
+//! Aes256CBC-HMAC-Legacy-Stream is a format for streaming encryption of attachments. It consists of
+//! a header, an AES-CBC stream, and a hmac over IV + CBC ciphertext. Because there is just one HMAC
+//! over the entire stream, it is not permissible to decrypt a partial portion of this stream, as
+//! the integrity of that portion cannot be guaranteed.
+//!
+//! ## Format
+//! The stream looks as follows:
+//! ```
+//! (KEY_E, KEY_A) = KEY
+//! IV | HMAC[KEY_A] (over IV + ciphertext) | AES-CBC[KEY_E]() ciphertext
+//! ```
+//!
+//! ## Limitations
+//!
+//! Because the HMAC is written to the start, encryption must be buffered in memory fully
+//! before emitting out as an encryption stream, since the HMAC can only be calculated
+//! after all ciphertext is produced. This is a limitation to the format.
+//!
+//! Further, the HMAC covers the entire stream, not chunks of it, so the entire stream
+//! must be fully read before the output is used. IMPORTANT: YOU MUST READ THE ENTIRE STREAM
+//! BEFORE USING THE DECRYPTED OUTPUT. This contract cannot be enforced by the interface and
+//! requires the correct usage of the caller.
+//!
+//! NOTE: The attachments, stored on the server contain a header idicating which format they are
+//! encrypted with. This header is *NOT* included in the stream processed by this file, and is
+//! expected to be stripped by the caller before passing the ciphertext to this code.
+//!
+//! Random access is not possible with this format, both because of the use of CBC chaining,
+//! and because of the single HMAC over the entire cipher stream.
+
+use aes::cipher::{BlockModeDecrypt, BlockModeEncrypt, KeyIvInit};
+use hkdf::HmacImpl;
+use hybrid_array::Array;
+use rand::Rng;
+use subtle::ConstantTimeEq;
+use zeroize::ZeroizeOnDrop;
+
+use super::{
+    ChunkDecryptionResult, ChunkEncryptionResult, StreamCreationError, StreamingDecryptor,
+    StreamingEncryptor,
+};
+use crate::{
+    Aes256CbcHmacKey, SymmetricCryptoKey, stream::large_memory_buffer::Buffer,
+    util::PbkdfSha256Hmac,
+};
+
+const AES256_CBC_BLOCK_SIZE: usize = 16;
+const AES256_CBC_IV_SIZE: usize = 16;
+const HMAC_SIZE: usize = 32;
+const HEADER_LENGTH: usize = AES256_CBC_IV_SIZE + HMAC_SIZE;
+const EMISSION_CHUNK_SIZE: usize = 1024 * 1024; // 1 MiB
+
+/// CBC IV
+type Iv = [u8; AES256_CBC_IV_SIZE];
+/// HMAC over IV + Ciphertext
+type Mac = [u8; HMAC_SIZE];
+/// Header is IV || HMAC
+type StreamHeaderBytes = [u8; HEADER_LENGTH];
+type CbcCiphertextBlock = [u8; AES256_CBC_BLOCK_SIZE];
+
+struct CbcDecryptor {
+    decryptor: cbc::Decryptor<aes::Aes256>,
+}
+
+impl CbcDecryptor {
+    fn new(key: &[u8; 32], iv: &Iv) -> Self {
+        Self {
+            decryptor: cbc::Decryptor::<aes::Aes256>::new(key.into(), iv.into()),
+        }
+    }
+
+    fn decrypt_block(&mut self, block: &CbcCiphertextBlock) -> CbcPlaintextBlock {
+        let mut block: Array<u8, _> = (*block).into();
+        self.decryptor.decrypt_block(&mut block);
+        CbcPlaintextBlock(
+            block
+                .as_slice()
+                .try_into()
+                .expect("block size checked by type"),
+        )
+    }
+}
+
+struct CbcEncryptor {
+    encryptor: cbc::Encryptor<aes::Aes256>,
+}
+
+impl CbcEncryptor {
+    fn new(key: &[u8; 32], iv: &Iv) -> Self {
+        Self {
+            encryptor: cbc::Encryptor::<aes::Aes256>::new(key.into(), iv.into()),
+        }
+    }
+
+    fn encrypt_block(&mut self, block: &mut CbcPlaintextBlock) -> CbcCiphertextBlock {
+        let mut block_array: Array<u8, _> = block.0.into();
+        self.encryptor.encrypt_block(&mut block_array);
+        block_array
+            .as_slice()
+            .try_into()
+            .expect("block size checked by type")
+    }
+}
+
+/// A higher level interface over the HMAC validation of the attachment ciphertext
+struct HmacStreamValidator {
+    hmac: PbkdfSha256Hmac,
+}
+
+impl HmacStreamValidator {
+    fn new(mac_key: &[u8; 32], iv: &Iv) -> Self {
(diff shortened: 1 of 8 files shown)


crates/bitwarden-crypto/src/stream/large_memory_buffer.rs
```diff
@@ -0,0 +1,170 @@
...
+}
+
+#[derive(Debug)]
+pub(crate) struct OutOfBoundsError;
+
+impl Buffer {
+    pub fn new() -> Self {
+        Self {
+            #[cfg(target_arch = "wasm32")]
+            inner: js_sys::Uint8Array::new_with_length(INITIAL_SIZE as u32).into(),
```
dani-garcia: We're using the `size: 0` below, so this `INITIAL_SIZE` allocation will always be ignored, as the first push will reallocate anyway, maybe we can set it to 0 too?

crates/bitwarden-crypto/src/stream/large_memory_buffer.rs
```diff
@@ -0,0 +1,170 @@
...
+        data: &[u8],
+    ) -> Result<(), OutOfBoundsError> {
+        if index.end > self.size {
+            self.grow_to_fit(index.end);
+        }
+
+        #[cfg(target_arch = "wasm32")]
+        {
+            let data_uint8_array = js_sys::Uint8Array::from(data);
+            self.inner.set(&data_uint8_array, index.start as u32);
```
dani-garcia: I believe if you do this you might get less allocations, haven't tested it though:

```suggestion
            self.inner
                .subarray(index.start as u32, index.end as u32)
                .copy_from(data);
```

crates/bitwarden-crypto/src/stream/large_memory_buffer.rs
```diff
@@ -0,0 +1,170 @@
...
+//!
+//! This is used for streaming encryption in the legacy format.
+
+/// Buffers are allocated and grown in multiples of this value.
+const BLOCK_SIZE: usize = 32 * 1024 * 1024;
+#[cfg(target_arch = "wasm32")]
+const INITIAL_SIZE: usize = 1024 * 1024;
+
+// Gives the size as a multiple of blocks that fits the requested size.
+fn round_up_to_block(size: usize) -> usize {
```
dani-garcia: Because we're going to be dealing with large files fairly often, I'd recommend switching the growth algorithm to instead allocate double the space on each move. This is what rust does with Vec internally already, and will reduce the number of allocations considerably, specially for large files.

```rust
// Option A: no floor, first grow sized exactly to first write, will require more reallocations/copies to get to a large size
fn next_size(current: usize, required: usize) -> usize {
    current.saturating_mul(2).max(required)
}

// Option B: 8 MiB floor, while smaller than the current block size, it will use more memory than option A, while reducing reallocations/copies 
const INITIAL_SIZE: usize = 8 * 1024 * 1024;

fn next_size(current: usize, required: usize) -> usize {
    current.saturating_mul(2).max(required).max(INITIAL_SIZE)
}
```
Some quick maths on the reallocations with all of them:
| Final size | Old (32 MiB blocks): grows / copies | New, no floor: grows / copies | New, 8 MiB floor: grows / copies |
|---|---|---|---|
| 100 KiB | 1 / 0 | 18 / 128 KiB | **1 / 0** |
| 500 KiB | 1 / 0 | 20 / 512 KiB | **1 / 0** |
| 5 MiB | 1 / 0 | 24 / 8 MiB | **1 / 0** |
| 50 MiB | 2 / 32 MiB | 27 / 64 MiB | **4 / 56 MiB** |
| 100 MiB | 4 / 192 MiB | 28 / 128 MiB | **5 / 120 MiB** |
| 500 MiB | 16 / 3.75 GiB | 30 / 512 MiB | **7 / 504 MiB** |

crates/bitwarden-crypto/src/stream/streaming_attachment_cipher.rs
```diff
@@ -0,0 +1,632 @@
...
+                Poll::Pending => return Poll::Pending,
+                Poll::Ready(Err(e)) => {
+                    this.state = StreamDecryptorState::Error;
+                    return Poll::Ready(Err(e));
+                }
+                Poll::Ready(Ok(())) => {
+                    let filled = scratch_buf.filled();
+                    if filled.is_empty() {
+                        this.inner_eof = true;
+                    } else if let Err(e) = this.feed_bytes_to_decryptor(filled) {
```
dani-garcia, line 254: I'm mildly worried about doing sync decryption work as part of the async future, though I can't think of a better way at the moment. Hopefully the stream chunks are small enough that decryption is going to be fast and not a big blocking problem here.

dani-garcia (commented):
Overall LGTM and the Buffer looks sound, just some small improvements on the Buffer allocation strategies

crates/bitwarden-crypto/src/stream/streaming_attachment_cipher.rs
quexten: What would be a size / timing? that is problematic?

dani-garcia: I think that's hard to quantify, the [rust docs](https://doc.rust-lang.org/std/future/trait.Future.html#runtime-characteristics) just say `quickly` which is not very specific: 

> An implementation of poll should strive to return quickly, and should not block. Returning quickly prevents unnecessarily clogging up threads or event loops. If it is known ahead of time that a call to poll may end up taking a while, the work should be offloaded to a thread pool (or something similar) to ensure that poll can return quickly.

My guess is that because we have a very low `READ_SCRATCH_SIZE`, any decryption for them is going to be in the microseconds range, which should be fine. I'd say as long as it doesn't get into high microseconds/low milliseconds range we're probably good, but we should have it in mind in case we notice any performance issues.

quexten: In that case I don't think there should be a problem here, worst case we can adjust the chunk size.

dani-garcia (dismissed):

crates/bitwarden-crypto/src/stream/streaming_attachment_cipher.rs
```diff
@@ -0,0 +1,663 @@
...
+}
+
+const READ_SCRATCH_SIZE: usize = 8 * 1024;
+
+// An enum representing the state of the streaming attachment decryptor. This is a state machine
+// for attachment stream parsing. As soon as the header bytes are parsed that discriminate the
+// encryption type, the state transitions to the appropriate streaming decryptor.
+enum StreamDecryptorState {
+    /// First byte of the wire has not yet been observed.
+    NeedDiscriminator { key: SymmetricCryptoKey },
```
mzieniukbw, line 62: ❓ Could this use key id instead ? This would require passing in key store context, but maybe it is the correct way to do this ?

crates/bitwarden-crypto/src/stream/streaming_attachment_cipher.rs
```diff
@@ -0,0 +1,663 @@
...
+    }
+
+    fn poll_flush(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<io::Result<()>> {
+        let this = self.get_mut();
+
+        // Drain whatever is pending (discriminator at startup, or nothing during steady state),
+        // then flush the inner writer.
+        if this.poll_drain_pending(cx).is_pending() {
+            return Poll::Pending;
+        }
```
mzieniukbw, line 436: ❓ Do we need to check for error ?
```suggestion
        }
        if matches!(this.state, StreamEncryptorState::Error) {
            return Poll::Ready(Err(io::Error::other(
                "streaming attachment: encryptor in error state",
            )));
        }
```

crates/bitwarden-crypto/src/stream/streaming_attachment_cipher.rs
```diff
@@ -0,0 +1,663 @@
...
+            this.state,
+            StreamEncryptorState::Finalized | StreamEncryptorState::Done
+        ) {
+            return Poll::Ready(Err(io::Error::other(
+                "streaming attachment: write after shutdown",
+            )));
+        }
+
+        // Make sure the discriminator (and any other queued bytes) are committed first.
+        if this.poll_drain_pending(cx).is_pending() {
```
mzieniukbw, line 390: ❓ Are we ok losing the original Err ? Should we propagate it OR add logging not to lose it - unless this is intentional not to expose any sensitive information about the stream ?

crates/bitwarden-crypto/src/stream/large_memory_buffer.rs
```diff
@@ -0,0 +1,151 @@
+//! A module for allocating large remote buffers not in WASM memory. WASM generally does not support
+//! resizing the heap after it has been allocated, so we need to allocate a buffer that can be
+//! released again.
+//!
+//! Allocations are made exponentially to minimize over-allocation while keeping re-allocations low.
```
mzieniukbw: ❓ Isn't this backwards ?
```suggestion
//! Allocations are made exponentially to maximize over-allocation to minimize re-allocations.
```

crates/bitwarden-crypto/src/stream/large_memory_buffer.rs
```diff
@@ -0,0 +1,151 @@
...
+}
+
+#[derive(Debug)]
+pub(crate) struct OutOfBoundsError;
+
+impl Buffer {
+    pub fn new() -> Self {
+        Self {
+            #[cfg(target_arch = "wasm32")]
+            inner: js_sys::Uint8Array::new_with_length(INITIAL_SIZE as u32).into(),
```
mzieniukbw: (Optional) Browsers seems to support resizeable arrays created in wasm. The array would need to be backed by `ArrayBuffer` and then can use [`https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/ArrayBuffer/resize](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/ArrayBuffer/resize) to resize.
Caveats:
- Fixed `maxByteLength` needs to be set (i.e. 1GB) https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/ArrayBuffer/maxByteLength
- Some older browser might not support it, would need to check `resizeable` property existence and check if it returns `true` - unless we commit to minimum browser versions (not sure if we already do with checking wasm support ?)

crates/bitwarden-crypto/src/stream/large_memory_buffer.rs
```diff
@@ -0,0 +1,151 @@
...
+
+            #[cfg(not(target_arch = "wasm32"))]
+            inner: vec![0; INITIAL_SIZE],
+
+            size: INITIAL_SIZE,
+        }
+    }
+
+    pub fn index(&self, index: std::ops::Range<usize>) -> Result<Vec<u8>, OutOfBoundsError> {
+        if index.end > self.size {
```
mzieniukbw: 🎨  Let's check for `index.start > index.end` too. (The same for `copy_from_slice`)

crates/bitwarden-crypto/src/stream/large_memory_buffer.rs
```diff
@@ -0,0 +1,151 @@
...
+            Ok(self.inner[index].to_vec())
+        }
+    }
+
+    pub fn copy_from_slice(
+        &mut self,
+        index: std::ops::Range<usize>,
+        data: &[u8],
+    ) -> Result<(), OutOfBoundsError> {
+        if index.end > self.size {
```
mzieniukbw: 🎨  Let's check for `data.len() != index.end - index.start` too.

crates/bitwarden-crypto/src/stream/large_memory_buffer.rs
```diff
@@ -0,0 +1,151 @@
...
+                .to_vec())
+        }
+
+        #[cfg(not(target_arch = "wasm32"))]
+        {
+            Ok(self.inner[index].to_vec())
+        }
+    }
+
+    pub fn copy_from_slice(
```
mzieniukbw: 🤔  The access pattern of this function is basically append to the end.
Could we simplify to `append` function that only accepts `data: &[u8]` ?

crates/bitwarden-crypto/src/stream/large_memory_buffer.rs
```diff
@@ -0,0 +1,151 @@
...
+    inner: Vec<u8>,
+
+    size: usize,
+}
+
+#[derive(Debug)]
+pub(crate) struct OutOfBoundsError;
+
+impl Buffer {
+    pub fn new() -> Self {
```
mzieniukbw: The attachment size is known from cipher attachment metadata (i don't recall if it is decrypted or encrypted size, still a massive improvement for larger attachments) - so we could have optional initial size argument, which we could pass when constructing the `StreamingAttachmentEncryptor` / `StreamingAttachmentDecryptor`

crates/bitwarden-crypto/src/stream/aes256_cbc_hmac_legacy_stream.rs
```diff
@@ -0,0 +1,711 @@
...
+    /// whether the calculated HMAC matches the expected HMAC from the header.
+    fn validate_stream_end(&self, expected_mac: &Mac) -> bool {
+        // `HmacImpl::finalize` consumes self; clone so the validator can stay borrowed
+        // by the decryptor state machine and be dropped normally when the state transitions.
+        let calculated_mac = self.hmac.clone().finalize();
+        calculated_mac.as_slice().ct_eq(expected_mac).into()
+    }
+}
+
+#[derive(ZeroizeOnDrop, Clone)]
```
mzieniukbw: ❓ This is the only place that zeroize on drop. But there are plenty of places here where we hold the `key` or other sensitive data.

crates/bitwarden-crypto/src/stream/streaming_attachment_cipher.rs
quexten: No, the use of the owned symmetric key is deliberate for a few reasons:
- Streaming decryptor state can be held across await points.
  - This makes it incompatible with working on key ids / key contexts, because key contexts cannot be held across await points
- We never expose the key to the outside, the public api visible to users of our crates works on key id / context. This makes it okay, because we mostly want to provide a nice public API, whereas `bitwarden-crypto` structs *may* use an API with different patterns. This is consistent with what we do in e.g. `PasswordProtectedKeyEnvelope` / `DataEnvelope` where the public API reads from context and then delegates to a private API that works on an owned key
- We *NEVER* want the key to be switched mid stream. If we do not own the key, we cannot enforce this?

crates/bitwarden-crypto/src/stream/large_memory_buffer.rs
quexten: The original comment is correct? Over-allocation (allocating more memory than we need) is a bad thing, so we want to minimize it, given the constraint that re-allocations are expensive.

What this comment is saying is that we both want low amount of over allocation, and low re-allocations.

crates/bitwarden-crypto/src/stream/aes256_cbc_hmac_legacy_stream.rs
quexten: We didn't actually need zeroize because we have a zeroizing global allocator. I've removed it now.

crates/bitwarden-crypto/src/stream/large_memory_buffer.rs
quexten: Changed api to `append` making this not required

crates/bitwarden-crypto/src/stream/large_memory_buffer.rs
quexten: Changed api to `append`

crates/bitwarden-crypto/src/stream/streaming_attachment_cipher.rs
quexten: I've updated the error handling now. Yes, we do not want to give specifics about which part of decryption failed, but other errors (IO) should be propagated.

crates/bitwarden-crypto/src/stream/large_memory_buffer.rs
quexten: ~~The buffer is only used by the `Encryptor` not the `Decryptor`, so the decryptor would not benefit from this. For the encryptor, this would prevevent around 7 reallocations for aes (and around 500MiB? of data copy in memory). Given that this is a few msec, and given that we are replacing the encryptor at some (hopefully near point) with a different encryption scheme, are we good to just put a comment in case we notice performance here being an issue?~~

~~I think a few msec are not that worth it to make the API more complex here, given that this a multi second operation anyways.~~

crates/bitwarden-crypto/src/stream/large_memory_buffer.rs
quexten: https://github.com/bitwarden/clients/issues/17536#issuecomment-3577477361

I think we could safely use this api. I'll check.

crates/bitwarden-crypto/src/stream/large_memory_buffer.rs
quexten: Applied for the encryptor

mzieniukbw (approved):

PR #1143, opened 2026-05-27
[PM-29567] Pin binary cargo tools via cargo-run-bin

coroiu:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-29567

## 📔 Objective

Add a [workspace.metadata.bin] block to the root Cargo.toml listing every binary cargo tool used by CI and scripts/lint.sh. CI and developers bootstrap cargo-run-bin once and invoke any tool via `cargo bin <tool>`, so dev versions stay in sync with CI.

scripts/lint.sh switches from PATH lookups (with manual install hints) to `cargo bin <tool>` invocations; the lint matrix drops its per-check install steps in favor of one shared cargo-run-bin install. The other workflows (check-powerset, build-rust-crates, build-android, version-bump, rust-test) get the same treatment.

Renovate's regex managers now target the manifest plus the cargo-run-bin bootstrap pin, replacing the four per-tool workflow regex managers.

### Follow-up fixes to get the lint matrix green

Moving the tools under cargo-run-bin surfaced two lint failures and one efficiency gap, fixed here:

- **clippy**: the clippy step set `RUSTFLAGS: -D warnings` for the whole step, which leaked into the from-source builds of `clippy-sarif`/`sarif-fmt` that `cargo bin` triggers in that step — a warning in their transitive dependency `serde-sarif` then became a hard error. `-D warnings` is now passed on the clippy command line instead, so the deny level applies only to the linted crates.

- **dylint**: `cargo-dylint` invokes `dylint-link` as rustc's linker, found by name on PATH. Run through `cargo bin`, cargo-run-bin prepends a shim directory whose `dylint-link` shim re-runs `cargo bin dylint-link`, which fails from the directories rustc links in (support/lints and each dependency's source dir, none of which declare `[workspace.metadata.bin]`). The dylint check now invokes the `cargo-dylint` binary directly with the real `dylint-link` on an absolute PATH, bypassing the shims. It builds only the two tools dylint needs rather than `cargo bin --install`, which also tries to build `cross` (which doesn't compile on the pinned toolchain).

- **caching**: cargo-run-bin builds its tools into `.bin`, which wasn't cached, so every lint check rebuilt its tools from source each run. A dedicated `actions/cache` step now caches `.bin` per check, keyed on `Cargo.toml` + `rust-toolchain.toml` (the tools depend on their pinned versions and the toolchain, not on `Cargo.lock`) with restore-keys for graceful fallback when a version is bumped.

## 🚨 Breaking Changes

None.

### DIFF
--- a/README.md
+++ b/README.md
@@ -454,8 +454,17 @@ The command is implemented in [scripts/lint.sh](./scripts/lint.sh) and mirrors
 
 ### Installation
 
-The tools each check needs (and pinned versions) are installed in the lint workflow above. The
-underlying tools are:
+Binary cargo tools (cargo-sort, cargo-udeps, cargo-dylint, etc.) are pinned in the root `Cargo.toml`
+under `[workspace.metadata.bin]` and installed lazily by
+[`cargo-run-bin`](https://crates.io/crates/cargo-run-bin), so dev and CI versions stay in sync.
+Bootstrap once:
+
+```bash
+cargo install cargo-run-bin --locked
+```
+
+After that, `npm run lint` (or `scripts/lint.sh` directly) handles installation of any missing tool
+on first invocation. The underlying tools are:
 
 - Nightly [cargo fmt](https://github.com/rust-lang/rustfmt) and
   [cargo udeps](https://github.com/est31/cargo-udeps)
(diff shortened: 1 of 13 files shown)


README.md
```diff
@@ -454,8 +454,17 @@ The command is implemented in [scripts/lint.sh](./scripts/lint.sh) and mirrors
 
 ### Installation
 
-The tools each check needs (and pinned versions) are installed in the lint workflow above. The
-underlying tools are:
```
djsmith85, line 458: Thank you, this step especially was weird to look up from the lint workflow and last time I setup my machine I was unaware of the pinned versions, as complaints about missing tools, came up while I was building the SDK locally.

djsmith85 (approved):

PR #1146, opened 2026-05-28
Integration test

quexten:


crates/bitwarden-wasm-internal/integration-tests/tests/streaming-attachment.test.ts
```diff
@@ -0,0 +1,126 @@
+import {
```
quexten, line 1: Note: This is generated by an LLM and not yet reviewed by me. Quality may vary, but it proves it works.

PR #1148, opened 2026-05-28
[PM-31128] Add reinit_user_crypto for mobile

Thomas-Avery:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-31128

## 📔 Objective

Add functionality to `CryptoClient` to allow mobile applications that receive a new `accountCryptographicState` and `V2UpgradeToken` from a userKey rotation upgrade sync to be able to reinit SDK's cryptography state.


crates/bitwarden-core/src/client/internal.rs
```diff
@@ -292,6 +293,15 @@ impl InternalClient {
...
         // dangerous-crypto-debug feature enabled, so this is safe
         info!("Setting user key with ID {:?}", user_key_id);
+
+        // The key store should not already have any keys initialized
+        if ctx.has_symmetric_key(SymmetricKeySlotId::User)
+            || ctx.has_private_key(PrivateKeySlotId::UserPrivateKey)
+            || ctx.has_signing_key(SigningKeySlotId::UserSigningKey)
+        {
+            return Err(EncryptionSettingsError::CryptoInitialization);
+        }
```
Thomas-Avery, line 300: Moving this check upstream seem like the best approach. Open to other suggestions.

crates/bitwarden-core/src/key_management/crypto/reinit_user_crypto.rs
```diff
@@ -0,0 +1,457 @@
...
+/// cryptographic state with the provided one, and upgrades the active user key from V1 to V2.
+///
+/// When mobile implements the state bridge, this expects the v2_upgrade_token to be written to
+/// state and performs local user data key migration.
+///
+/// Intended for mobile clients with `accountCryptographicState` and `V2UpgradeToken` received in
+/// a sync for a V1 -> V2 encryption upgrade. This allows the client to apply the received account
+/// cryptographic state and update to reinitialize the SDK without tearing down and recreating the
+/// client.
+pub(in crate::key_management) async fn reinit_user_crypto(
```
quexten: `pub(super)` will be more robust for moves here. If we ever rename the module then `crate::key_management` won't work.

crates/bitwarden-core/src/key_management/crypto/reinit_user_crypto.rs
```diff
@@ -0,0 +1,457 @@
...
...QXI6kcdpJzPXSeU9o+NC68QDlOIrMVxKFeE7w7PvVmAaxEo0YwmuAzzKy9QpdlK0aab/xEi8V4iXj4hGepqAvHkXIQd+r3FNeiLfllkb61p6WTjr5urcmDQMR94/wYoilpG5OlybHdbhsYHvIzYoLrC7fzl630gcO6t4nM24vdB6Ymg9BVpEgKRAxSbE62Tqacxqnz9AcmgItb48NiR/He3n3ydGjPYuKk/ihZMgEwAEZvSlNxYONSbYrIGDtOY+8Nbt6KiH3l06wjZW8tcmFeVlWv+tWotnTY9IqlAfvNVTjtsobqtQnvsiDjdEVtNy/s2ci5TH+NdZluca2OVEr91Wayxh70kpM6ib4UGbfdmGgCo74gtKvKSJU0rTHakQ5L9JlaSDD5FamBRyI0qfL43Ad9qOUZ8DaffDCyuaVyuqk7cz9HwmEmvWU3VQ+5t06n/5kRDXttcw8w+3qClEEdGo1KeENcnXCB32dQe3tDTFpuAIMLqwXs6FhpawfZ5kPYvLPczGWaqftIs/RXJ/EltGc0ugw2dmTLpoQhCqrcKEBDoYVk0LDZKsnzitOGdi9mOWse7Se8798ib1UsHFUjGzISEt6upestxOeupSTOh0v4+AjXbDzRUyogHww3V+Bqg71bkcMxtB+WM+pn1XNbVTyl9NR040nhP7KEf6e9ruXAtmrBC2ah5cFEpLIot77VFZ9ilLuitSz+7T8n1yAh1IEG6xxXxninAZIzi2qGbH69O5RSpOJuJTv17zTLJQIIc781JwQ2TTwTGnx5wZLbffhCasowJKd2EVcyMJyhz6ru0PvXWJ4hUdkARJs3Xu8dus9a86N8Xk6aAPzBDqzYb1vyFIfBxP0oO8xFHgd30Cgmz8UrSE3qeWRrF8ftrI6xQnFjHBGWD/JWSvd6YMcQED0aVuQkuNW9ST/DzQThPzRfPUoiL10yAmV7Ytu4fR3x2sF0Yfi87YhHFuCMpV/DsqxmUizyiJuD938eRcH8hzR/VO53Qo3UIsqOLcyXtTv6THjSlTopQ+JOLOnHm1w8dzYbLN44OG44rRsbihMUQp+wUZ6bsI8rrOnm9WErzkbQFbrfAINdoCiNa6cimYIjvvnMTaFWNymqY1vZxGztQiMiHiHYwTfwHTXrb9j0uPM=|09J28iXv9oWzYtzK2LBT6Yht4IT4MijEkk0fwFdrVQ4="
```
quexten: Above we use consts, but here a magic string? We should use the same pattern.

crates/bitwarden-core/src/key_management/crypto/reinit_user_crypto.rs
```diff
@@ -0,0 +1,457 @@
...
+            .dangerous_get_symmetric_key(SymmetricKeySlotId::User)
+            .unwrap();
+        assert_eq!(user_key, expected_v2_key);
+    }
+
+    /// V2 wrapped state from the test vectors. Wrapped under the V2 test user
+    /// key, so it only decrypts cleanly when paired with that key.
+    fn test_vector_v2_account_state() -> WrappedAccountCryptographicState {
+        WrappedAccountCryptographicState::V2 {
+            private_key: TEST_VECTOR_PRIVATE_KEY_V2.parse().unwrap(),
```
quexten, line 212: A thought: We should probably move these to be utility functions behind a feature flag that can be re-used throughout all tests, instead of putting them into all files.

crates/bitwarden-uniffi/swift/integration-tests/Tests/IntegrationTests/Utils.swift
```diff
@@ -11,6 +11,47 @@ let PRIVATE_KEY =
 let MASTER_KEY_WRAPPED_USER_KEY =
     "2.u2HDQ/nH2J7f5tYHctZx6Q==|NnUKODz8TPycWJA5svexe1wJIz2VexvLbZh2RDfhj5VI3wP8ZkR0Vicvdv7oJRyLI1GyaZDBCf9CTBunRTYUk39DbZl42Rb+Xmzds02EQhc=|rwuo5wgqvTJf3rgwOUfabUyzqhguMYb3sGBjOYqjevc="
 
+// PBKDF2 600k, V1 wrapped account state.
+// Used by reinit tests that need a V1 client whose user key matches
+// the V1 user key used to mint the captured upgrade-token vectors below.
+let V1_ALIGNED_USER_ID = "060000fb-0922-4dd3-b170-6e15cb5df8c8"
```
quexten: What does aligned mean here? Can we use consistent naming with the other consts? `TEST_VECTOR_PRIVATE_USER_ID_V1`, TEST_VECTOR_KDF_ITERATIONS_V1` and so on?

crates/bitwarden-uniffi/swift/integration-tests/Tests/IntegrationTests/Utils.swift
```diff
@@ -81,3 +122,80 @@ func makeInitializedClient(stateBridge: InMemoryStateBridge) async throws -> Cli
...
+        accountCryptographicState: makeV2AccountCryptographicState(),
+        method: .decryptedKey(decryptedUserKey: TEST_VECTOR_USER_KEY_V2_B64),
+        upgradeToken: nil
+    )
+
+    try await client.crypto().initializeUserCrypto(req: req)
+    return client
+}
+
+/// V2 wrapped account state from the test vectors. Decrypts cleanly only when
```
quexten:
```suggestion
/// V2 wrapped account state from the test vectors. Decrypts only when
```

crates/bitwarden-core/src/key_management/crypto/reinit_user_crypto.rs
```diff
@@ -0,0 +1,457 @@
...
+            })?;
+    }
+
+    // Re-wrap the local user data key under the new user key (no-op
+    // when not applicable).
+    let user_id = client
+        .internal
+        .get_user_id()
+        .ok_or(ReinitUserCryptoError::LocalUserDataMigrationFailed)?;
+    migrate_local_user_data_key_for_user_key_upgrade(client, user_id)
```
quexten: We are migrating the local user data key but are not migrating the PIN key here, which will leave the PIN encrypted with the old key. Ideally we'd extract these migrations so that we cannot get them inconsistently, between init and reinit.

```
    PinLockSystem::on_unlock(&PinLockSystem::with_client(client)).await;
```

quexten (changes requested):
A few smaller nits, but more importantly the PIN migration is missing.

crates/bitwarden-uniffi/swift/integration-tests/Tests/IntegrationTests/Utils.swift
Thomas-Avery: Updated in https://github.com/bitwarden/sdk-internal/pull/1148/changes/9a65a2d7439556890a3567897ce4fb672748a9bf

crates/bitwarden-uniffi/swift/integration-tests/Tests/IntegrationTests/Utils.swift
Thomas-Avery: Fixed in https://github.com/bitwarden/sdk-internal/pull/1148/changes/9a65a2d7439556890a3567897ce4fb672748a9bf

crates/bitwarden-core/src/key_management/crypto/reinit_user_crypto.rs
Thomas-Avery: IMO `WrappedAccountCryptographicState` should not live in `bitwarden-core` and these should be test utilities built behind a feature flag in whatever crate `WrappedAccountCryptographicState` should belong to. 

I don't think it is reasonable to increase the scope of this PR to do that.

crates/bitwarden-core/src/key_management/crypto/reinit_user_crypto.rs
Thomas-Avery: Updated in https://github.com/bitwarden/sdk-internal/pull/1148/changes/9a65a2d7439556890a3567897ce4fb672748a9bf

crates/bitwarden-core/src/key_management/crypto/reinit_user_crypto.rs
Thomas-Avery: `crypto.rs` re-exports it to be used in the crypto client so `pub(super)` doesn't work. These are all using the old feature client structure and would be a non-issue in the new client structure prescribed in the contributing documentation. For now I made it `pub(crate)` to solve for the rename concerns.

crates/bitwarden-core/src/key_management/crypto/reinit_user_crypto.rs
Thomas-Avery: I consolidated them into a shared `on_unlock_handler`. 

Something that I would like to hear your opinion on. Looking at this, all our local migrations need and assume the `v2_upgrade_token` is set in state. Should we change `reinit_user_crypto` to also require the upgrade token set in state instead of passed in via the request? I would think having one source of truth would be desirable.

coroiu (commented):
Ping me when this is ready for merge :)

crates/bitwarden-core/src/key_management/crypto/reinit_user_crypto.rs
quexten: > Something that I would like to hear your opinion on. Looking at this, all our local migrations need and assume the v2_upgrade_token is set in state. Should we change reinit_user_crypto to also require the upgrade token set in state instead of passed in via the request? I would think having one source of truth would be desirable.

IMO, in the long term:
- All public functions should have only legitimate variables (that vary) passed in, never state
  - I.e "unlock({pin, pin envelope, cryptographic state})" -> BAD
  - "unlock(pin)" -> GOOD
 
State should be KM internal concern, not public API.

Subsequently, we should not take the v2 token as public API if we do not have to, we also should not take in the account cryptographic state as public API if we not have to.

In the long term, the vision for how this would work is:
- SDK receives "on sync" handler, which gets triggered by a sync notification, and passes in a response model
  - KM has a sync handler registered, that takes the data and sets the data to state
  - The KM handler dispatches a call to reinit crypto, and reinit crypto is not a public API

*However* in this case we give the reinit API precisely because the above described desired state does not yet exist. Further, it is easy for mobile to forget to set the pre-conditions (setting the tokens, etc). So for this API, given that it is the bridge until we implement the above sdk consolidation, we should make it as safe for mobile to not be able to mis-use it. Subsequently, having the parameters required in the parameter set seems safer that accessing through state, as in the latter case mobile could forget to set the state, even though it is a worse public API.

quexten: I do want to simplify the unlock APIs as much as possible however, though, it is possible to mis-use it. Maybe we should make a ticket to implement a simplified version of "set data to state after sync in SDK" so that we can take that risk / burden off of mobile.

quexten (dismissed):

crates/bitwarden-core/src/key_management/crypto/reinit_user_crypto.rs
Thomas-Avery: @quexten  Going to recap here to make sure we are both on the same page.

For the local user data key migration and the PIN `on_unlock`, both require the getting the upgrade token from state. 

https://github.com/bitwarden/sdk-internal/blob/7bca9fcaccf17e6ae3e2dd4999bb62ffce84ad87/crates/bitwarden-core/src/key_management/local_user_data_key_state.rs#L57

https://github.com/bitwarden/sdk-internal/blob/7bca9fcaccf17e6ae3e2dd4999bb62ffce84ad87/crates/bitwarden-core/src/key_management/pin_lock_system.rs#L190




So for a time mobile will pass in the upgrade token and not have the state bridge implemented and this will just upgrade the `userKey`? Then when mobile implements the state bridge we will need them to set the upgrade token into state and pass it in here for things to work. The other option is to have the upgrade token pass into `reinit_user_crypto` and us set the upgrade_token into state there. The local data migrations don't work without the upgrade token set to state.

Thomas-Avery: Okay I dug into this deeper and I think I've covered all of our concerns with https://github.com/bitwarden/sdk-internal/pull/1148/changes/c1620a276ca616e79ace7348802764c8f059b218. 

First the idea that that we can upgrade just the `userKey` without the state bridge is incorrect. This would corrupt the localUserDataKey with how we have things setup. To solve for this we can just guard the entire `reinit_user_crypto` to only run when a state bridge is registered.

The concern I had with different sources for the upgrade token and your concern with mobile having to set state prior to calling can be solved as follows. Have the request model require the upgrade token and have `reinit_user_crypto` responsible for writing that to state bridge state prior to calling local migrations.

quexten: > First the idea that that we can upgrade just the userKey without the state bridge is incorrect. This would corrupt the localUserDataKey with how we have things setup. To solve for this we can just guard the entire reinit_user_crypto to only run when a state bridge is registered.

Just to be clear, I was not saying we do not need the state bridge. I was saying that a multi-step API is more brittle than having one interface to provide to mobile.

quexten: > The concern I had with different sources for the upgrade token and your concern with mobile having to set state prior to calling can be solved as follows. Have the request model require the upgrade token and have reinit_user_crypto responsible for writing that to state bridge state prior to calling local migrations.

I agree with this approach.

quexten (dismissed):

quexten (approved):

coroiu (approved):

PR #1153, opened 2026-06-01
Add support for AccountCryptographicState on auth response

dani-garcia:
## 📔 Objective

Expose `wrapped_account_crypto_state: Option<WrappedAccountCryptographicState>` on `LoginSuccessResponse`, now that https://bitwarden.atlassian.net/browse/PM-30222 is complete.

Supporting changes:
- New `LoginResponseError` enum composing `MasterPasswordError` and `AccountKeysResponseParseError`, used as the `TryFrom<LoginSuccessApiResponse>::Error`. We could have probably used `LoginErrorApiResponse` directly here, but I decided to keep the API specific error and the model parsing error separate. 
- `LoginErrorApiResponse` collapses its three explicit `From` impls (`reqwest::Error`, `reqwest_middleware::Error`, `MasterPasswordError`) into a single blanket `impl<T: std::error::Error>` to also support the new `LoginResponseError`.

### DIFF
--- a/crates/bitwarden-auth/src/login/api/response/login_error_api_response.rs
+++ b/crates/bitwarden-auth/src/login/api/response/login_error_api_response.rs
@@ -1,4 +1,3 @@
-use bitwarden_core::key_management::MasterPasswordError;
 use serde::Deserialize;
 
 #[derive(Deserialize, PartialEq, Eq, Debug)]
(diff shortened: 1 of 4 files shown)


JaredSnider-Bitwarden (approved):
Thank you for fixing this for us and adding the extra, requested tests! LGTM!

PR #1156, opened 2026-06-02
Migrate call sites to bitwarden_logging::instrument

coroiu:
## 🎟️ Tracking

Follow-up to #1139, which added `#[bitwarden_logging::instrument]` and the
`tracing_instrument` dylint but intentionally left migration of existing call
sites for later.

## 📔 Objective

Migrates existing call sites off `tracing::instrument` / bare `#[instrument]`
onto the fully-qualified `#[bitwarden_logging::instrument]` wrapper (which
enforces `skip_all` by default), and flips the `tracing_instrument` lint
default from `allow` to `warn` so regressions are caught going forward.

Swept this PR:

- **bitwarden-core**
- **bitwarden-vault**
- **bitwarden-user-crypto-management**

### Migration rules applied

- `skip_all` sites: dropped `skip_all` (the wrapper enforces it); `err`,
  `name`, `level`, and existing `fields(...)` flow through unchanged.
- `skip(specific args)` sites: the macro skips *all* args, so the arguments
  that were previously logged by default are re-expressed as explicit
  `fields(name = ?expr)` opt-ins. Span fields are therefore unchanged; the
  only difference is that the logging is now explicit at the call site. The
  re-added fields are non-sensitive key-slot ids / KDF config (never key
  material).
- Removed the now-unused `use tracing::instrument;` imports and write the
  attribute fully-qualified so the `skip_all` safety property is visible at
  every call site.

### Lint default

Flipped `TRACING_INSTRUMENT` from `Allow` to `Warn`. **bitwarden-crypto** is
deliberately *not* migrated here and is opted out with a crate-level
`#![allow(unknown_lints, tracing_instrument)]`: its `dangerous-crypto-debug`
sites intentionally log key material, which has to be re-expressed as explicit
`fields(...)` opt-ins under the wrapper's `skip_all` default. That conversion
deserves its own focused, security-reviewed PR.

## ⏭️ Follow-up

- Migrate **bitwarden-crypto** (including the `dangerous-crypto-debug`
  dual-arm sites) and remove the crate-level allow.

## 🚨 Breaking Changes

None. Public APIs are unchanged; this only affects tracing span fields, which
remain identical to before.


coroiu:
Superseded by #1194 (rebased onto latest main, renamed branch `arch/migrate-instrument` → `coroiu/migrate-instrument`).

PR #1157, opened 2026-06-02
[PM-35796] Invite link Crypto Bundle Generation and Unsealing

BTreston:
## 🎟️ Tracking
https://bitwarden.atlassian.net/browse/PM-35796

## 📔 Objective
Create the admin console owned crate in the SDK + WASM bindings that call the new KM interface `InviteKeyBundle`. Also edits `InviteKeyBundle` to return an `EncString` instead of `B64`.

Wired up in the clients in https://github.com/bitwarden/clients/pull/21004


BTreston:
@claude review

coroiu (dismissed):
`bitwarden-organization-invite-link` seems very specific 🤷

crates/bitwarden-organization-invite-link/src/invite.rs
```diff
@@ -0,0 +1,201 @@
...
+///
+/// - `invite_key`: raw invite key encoded as base64Url. **MUST NOT be sent to the server.**
+/// - `sealed_invite_key_envelope`: invite key sealed with the org key, serialized as a Bitwarden
+///   EncString (`"2.iv|data|mac"`). Safe to send to the server.
+#[derive(Clone, Serialize, Deserialize)]
+#[cfg_attr(feature = "wasm", derive(Tsify), tsify(into_wasm_abi, from_wasm_abi))]
+#[serde(rename_all = "camelCase")]
+pub struct OrganizationInviteCryptoBundle {
+    /// Raw invite key as base64Url. CRITICAL: MUST NOT be sent to the server.
+    pub invite_key: String,
```
quexten:
```suggestion
    pub invite_key: SymmetricKey,
```

crates/bitwarden-organization-invite-link/src/invite.rs
```diff
@@ -0,0 +1,201 @@
...
+    /// Invite key sealed with the organization key, as a Bitwarden EncString (`"2.iv|data|mac"`).
+    pub sealed_invite_key_envelope: String,
```
quexten:
```suggestion
    pub sealed_invite_key_envelope: EncString,
```

Ideally the newtype should actually be wasm exposed. I'm not sure why we didn't in the first place. But we should always use as specific of a type as possible, even if they are newtypes around string.

On the WASM side you can still directly pass the generated type into functions expecting string, though using a newtype there is also preferred.

crates/bitwarden-organization-invite-link/src/invite.rs
```diff
@@ -0,0 +1,201 @@
...
+    UnsealingFailed(InviteKeyBundleError),
+}
+
+/// The cryptographic bundle for an organization member invite.
+///
+/// - `invite_key`: raw invite key encoded as base64Url. **MUST NOT be sent to the server.**
+/// - `sealed_invite_key_envelope`: invite key sealed with the org key, serialized as a Bitwarden
+///   EncString (`"2.iv|data|mac"`). Safe to send to the server.
+#[derive(Clone, Serialize, Deserialize)]
+#[cfg_attr(feature = "wasm", derive(Tsify), tsify(into_wasm_abi, from_wasm_abi))]
```
quexten: Do we not need uniffi? Is this just in web for now?

crates/bitwarden-organization-invite-link/src/invite.rs
```diff
@@ -0,0 +1,201 @@
...
+use bitwarden_crypto::{BitwardenLegacyKeyBytes, KeyStore, SymmetricCryptoKey, key_slot_ids};
+use bitwarden_error::bitwarden_error;
+use bitwarden_organization_crypto::{InviteKeyBundle, InviteKeyBundleError, InviteKeyEnvelope};
+use serde::{Deserialize, Serialize};
+use thiserror::Error;
+#[cfg(feature = "wasm")]
+use tsify::Tsify;
+
+/// Errors from generating an organization invite crypto bundle.
+#[allow(missing_docs)]
```
quexten:
```suggestion
```

Please add docs. The missing docs allow is only meant for cases where we did not yet provide docs.

crates/bitwarden-organization-invite-link/src/invite.rs
```diff
@@ -0,0 +1,201 @@
...
+    /// Raw invite key as base64Url. CRITICAL: MUST NOT be sent to the server.
+    pub invite_key: String,
+    /// Invite key sealed with the organization key, as a Bitwarden EncString (`"2.iv|data|mac"`).
+    pub sealed_invite_key_envelope: String,
+}
+
+impl std::fmt::Debug for OrganizationInviteCryptoBundle {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("OrganizationInviteCryptoBundle")
+            .field("invite_key", &"<REDACTED>")
```
quexten: If invite key were using the proper symmetrickey type, this would not be needed because symmetrickey already implemnets debug and prevents you from logging it. That's why using string here is dangerous / relies on you not logging it.

crates/bitwarden-organization-invite-link/src/invite.rs
```diff
@@ -0,0 +1,201 @@
...
+                "sealed_invite_key_envelope",
+                &self.sealed_invite_key_envelope,
+            )
+            .finish()
+    }
+}
+
+/// Generates a new [`OrganizationInviteCryptoBundle`] sealed with the provided organization key.
+///
+/// Each call produces a unique, non-deterministic invite key.
```
quexten:
```suggestion
/// Each call produces a unique invite key sampled from a secure cryptographic source.
```

crates/bitwarden-organization-invite-link/src/invite.rs
```diff
@@ -0,0 +1,201 @@
...
+///
+/// # Security
+/// The `invite_key` field MUST NOT be sent to the server.
+#[cfg_attr(feature = "wasm", wasm_bindgen::prelude::wasm_bindgen)]
+pub fn generate_organization_invite_crypto_bundle(
+    org_key: Vec<u8>,
```
quexten:
```suggestion
    org_key: &SymmetricKey,
```

Please do not use `Vec<u8>` or `String` to represent keys. We want to use the `SymmetricKey` type for this.

crates/bitwarden-organization-invite-link/src/invite.rs
```diff
@@ -0,0 +1,201 @@
...
+}
+
+/// Unseals a `sealedInviteKeyEnvelope` (produced by [`generate_organization_invite_crypto_bundle`])
+/// using the organization key, returning the raw invite key as a base64Url string.
+///
+/// The returned invite key is safe to embed in a URL fragment for distribution to invitees.
+#[cfg_attr(feature = "wasm", wasm_bindgen::prelude::wasm_bindgen)]
+pub fn unseal_organization_invite_key(
+    org_key: Vec<u8>,
+    sealed_invite_key_envelope: String,
```
quexten:
```suggestion
    org_key: SymmetricKey,
    sealed_invite_key_envelope: &EncString, // OR &InviteKeyEnvelope??
```

quexten (changes requested):

quexten:
@BTreston Looks mostly good, but we have a few times where we are using generic types (String, Vec<u8>) when we can use the newtypes (SymmetricKey, EncString). These are preferred because the provide better type safety, but also prevent dangerous logging of cryptographic material.

crates/bitwarden-organization-invite-link/src/invite.rs
BTreston: Invite link is only configurable in admin console, which I don't think is available in any other clients beside web.

crates/bitwarden-organization-invite-link/src/invite.rs
```diff
@@ -0,0 +1,215 @@
...
+ * The `inviteKey` field MUST NOT be sent to the server.
+ */
+export function generate_organization_invite_crypto_bundle(org_key: SymmetricKey): OrganizationInviteCryptoBundle;
+
+/**
+ * Unseals a sealed invite key envelope using the organization key,
+ * returning the raw invite key as a base64Url string.
+ */
+export function unseal_organization_invite_key(org_key: SymmetricKey, sealed_invite_key_envelope: InviteKeyEnvelope): InviteKeyData;
+"#;
```
BTreston: I'm not really a fan of this. It requires this block of manually defined function bindings to be manually updated in lock-step with changes to the actual rust definitions, which is easily missed and kinda feels like it misses the whole point of using `wasm_bindgen`... But I could not figure out how to get the generated bindings to have the correct `SymetricKey` type that exists in clients (would generate as `uint8Array`)

crates/bitwarden-organization-crypto/src/wasm.rs
```diff
@@ -1,20 +1,30 @@
...
 "#;
 
+#[wasm_bindgen::prelude::wasm_bindgen(typescript_custom_section)]
+const TS_INVITE_KEY_ENVELOPE: &'static str = r#"
+export type InviteKeyEnvelope = Tagged<string, "InviteKeyEnvelope">;
+"#;
+
+/// WASM bindings for organization cryptography operations.
+#[wasm_bindgen::prelude::wasm_bindgen]
+pub struct OrganizationCryptoWasm;
```
quexten: I'm a bit confused what this does?

crates/bitwarden-organization-invite-link/src/invite.rs
```diff
@@ -0,0 +1,215 @@
...
+
+/// WASM bindings for organization invite link cryptographic operations.
+///
+/// Exposes [`generate_organization_invite_crypto_bundle`] and
+/// [`unseal_organization_invite_key`] to JavaScript/TypeScript consumers.
+#[cfg(feature = "wasm")]
+#[wasm_bindgen::prelude::wasm_bindgen]
+pub struct OrganizationInviteLinkWasm;
```
quexten: @BTreston Please have a look at the guide here: https://contributing.bitwarden.com/architecture/sdk/adding-functionality. This should use a Client. An example implementation is here: https://github.com/bitwarden/sdk-internal/blob/fe65dc5432c08a92ec03c419d2a02a5bcdcd5cb4/crates/bitwarden-vault/src/folder/folder_client.rs#L17.

The functions you are writing should be implemented on this client.
This massively simplifies the manual TS you are writing.

crates/bitwarden-organization-invite-link/src/invite.rs
quexten: I agree, this is definitely not great. I think the first step is to make these be implementations on top of a struct (the client struct), instead of static functions. That may already fix the wasm bindgen issues you are facing. If not I'll take a look because we don't expect most teams to dig into wasm bindgen deeply.

crates/bitwarden-organization-invite-link/src/invite.rs
```diff
@@ -0,0 +1,215 @@
...
+}
+
+/// Generates a new [`OrganizationInviteCryptoBundle`] sealed with the provided organization key.
+///
+/// Each call produces a unique key sampled from a secure cryptographic source.
+///
+/// # Security
+/// The `invite_key` field MUST NOT be sent to the server.
+#[cfg_attr(feature = "wasm", wasm_bindgen::prelude::wasm_bindgen(skip_typescript))]
+pub fn generate_organization_invite_crypto_bundle(
```
quexten: Let's not pass in the symmetric crypto key at all. If you implement this as a client, then you have access to the keycontext. The key context can give you the symmetric key for any organization. So then this simplifies down to:

- Pass in org UUID
- Code here reads key from ctx for org uuid
- Encryption happens via KM API

This also prevents you from ever having to deal with keys, since you can just pass the ctx, key slot to the invitekeybundle::make function.

crates/bitwarden-organization-invite-link/src/invite.rs
```diff
@@ -0,0 +1,215 @@
...
+    })
+}
+
+/// Unseals a `sealedInviteKeyEnvelope` (produced by [`generate_organization_invite_crypto_bundle`])
+/// using the organization key, returning the raw invite key as [`InviteKeyData`].
+///
+/// The returned invite key is safe to embed in a URL fragment for distribution to invitees.
+#[cfg_attr(feature = "wasm", wasm_bindgen::prelude::wasm_bindgen(skip_typescript))]
+pub fn unseal_organization_invite_key(
+    org_key: SymmetricCryptoKey,
```
quexten: Same comment here. You don't need to / shouldn't need to pass in the key. It should just come from the key context that you can get from the client.

crates/bitwarden-organization-invite-link/src/invite.rs
```diff
@@ -0,0 +1,215 @@
...
+    }
+
+    #[signing]
+    enum LocalSigningKeySlotId {
+        #[local]
+        Local(LocalId),
+    }
+
+    LocalKeySlotIds => LocalSymmetricKeySlotId, LocalPrivateKeySlotId, LocalSigningKeySlotId;
+}
```
quexten:
```suggestion

```
not required if you use the ctx from the `Client`.

quexten (changes requested):

crates/bitwarden-organization-crypto/src/wasm.rs
```diff
@@ -1,20 +1,26 @@
...
 
 #[wasm_bindgen::prelude::wasm_bindgen(typescript_custom_section)]
-const TS_CUSTOM_TYPES: &'static str = r#"
+const TS_INVITE_KEY_DATA: &'static str = r#"
 export type InviteKeyData = Tagged<string, "InviteKeyData">;
 "#;
 
+#[wasm_bindgen::prelude::wasm_bindgen(typescript_custom_section)]
+const TS_INVITE_KEY_ENVELOPE: &'static str = r#"
+export type InviteKeyEnvelope = Tagged<string, "InviteKeyEnvelope">;
```
quexten: Looks like this is duplicated with the invite link client? We have two custom type sections? I suggest they should only be in the organization crypto crate.

crates/bitwarden-organization-invite-link/src/invite_link_client.rs
```diff
@@ -0,0 +1,214 @@
...
+#[wasm_bindgen::prelude::wasm_bindgen(typescript_custom_section)]
+const TS_INVITE_KEY_DATA: &'static str = r#"
+export type InviteKeyData = Tagged<string, "InviteKeyData">;
+"#;
+
+#[cfg(feature = "wasm")]
+#[wasm_bindgen::prelude::wasm_bindgen(typescript_custom_section)]
+const TS_INVITE_KEY_ENVELOPE: &'static str = r#"
+export type InviteKeyEnvelope = Tagged<string, "InviteKeyEnvelope">;
+"#;
```
quexten:
```suggestion
```

quexten (commented):
Only small stuff remaining, mostly looking very good :)

crates/bitwarden-organization-crypto/src/wasm.rs
quexten: Looks like this is the same as https://github.com/bitwarden/sdk-internal/pull/1157#discussion_r3376402809

quexten:
Also looks like there are build issues now that https://github.com/bitwarden/sdk-internal/pull/1165 merged

crates/bitwarden-organization-crypto/src/wasm.rs
BTreston: In my fairly limited knowledge, the trait impls didn't produce standalone symbols and the consts aren't marked with `#[used]`. These were dropped during the linking process. Adding this fixed that because it emits sysmbols that the linker didn't drop.

crates/bitwarden-organization-crypto/src/wasm.rs
```diff
@@ -1,20 +1,30 @@
...
 use std::str::FromStr;
 
 use wasm_bindgen::convert::{FromWasmAbi, IntoWasmAbi, OptionFromWasmAbi};
 
 use crate::{InviteKeyBundleError, InviteKeyData, InviteKeyEnvelope};
 
+/// WASM bindings for organization cryptography operations.
+#[wasm_bindgen::prelude::wasm_bindgen]
+pub struct OrganizationCryptoWasm;
+
```
quexten:
```suggestion
```

quexten (commented):
Seems to build without the struct here.

crates/bitwarden-organization-crypto/src/wasm.rs
BTreston: Interesting. I get errors in the generated bindings file on my machine without this. This breaks the consuming code in the /clients repo (for me locally)
```
bitwarden_wasm_internal.d.ts:3030:14 - error TS2304: Cannot find name 'InviteKeyData'.

3030   inviteKey: InviteKeyData;
                  ~~~~~~~~~~~~~

bitwarden_wasm_internal.d.ts:3034:28 - error TS2304: Cannot find name 'InviteKeyEnvelope'.

3034   sealedInviteKeyEnvelope: InviteKeyEnvelope;
                                ~~~~~~~~~~~~~~~~~


Found 2 errors in the same file, starting at: bitwarden_wasm_internal.d.ts:3030
```
Maybe there is something on my end that's messed up. Any thoughts?

BTreston: This struct isn't needed if these custom ts type defs live next to their respective rust structs. IMO this is also nice b/c colocating these makes the intent clearer. As far as I can tell, there needs to be some concrete implementer (some struct or even a fn possibly?) that must implement [wasn_bindgen] in order to append these bindings to the generated d.ts file. I checked the generated bindings from `main` locally and it looks like the tagged string types are also not present there (at least according to my local LSP).

quexten: Are you saying the types do show up when colocating the typedefs with the structs? If so, let's do that.

quexten (approved):

quexten:
Thanks @BTreston for taking the time to address the feedbaack. Changes look good from my side now.

eliykat (approved):

coroiu (approved):

PR #1165, opened 2026-06-04
Migrate SymmetricCryptoKey creation to use make()

dani-garcia:
## 🎟️ Tracking

https://github.com/bitwarden/sdk-internal/pull/1104#discussion_r3339382191

## 📔 Objective

From a review on #1104, KM brought up that there are multiple ways to create a `SymmetricCryptoKey`, and we should move all usages to a single unified one and deprecate the rest.

This PR does just that, and replaces any uses of `SymmetricCryptoKey::make_*()` by `SymmetricCryptoKey::make(SymmetricKeyAlgorithm)`. Note that I haven't replaced the usages in `bitwarden-crypto` and instead just left the functions as `pub(crate)`

Also removed the unused dep mentioned here: https://github.com/bitwarden/sdk-internal/pull/1104#discussion_r3349432261


harr1424 (dismissed):

quexten (dismissed):
Sweet!

coroiu (dismissed):
Thank you for the follow up!

quexten (approved):

coroiu (approved):

harr1424 (approved):

gbubemismith (approved):

PR #1167, opened 2026-06-04
Update rustcrypto crates

dani-garcia:
## 📔 Objective

A few of the rustcrypto crates we use got new stable releases, so this PR updates to them. I've also updated a couple of the other RCs as needed to avoid version conflicts with their dependencies.

This should also fix the CI failure seen in https://github.com/bitwarden/sdk-internal/pull/1165: https://github.com/bitwarden/sdk-internal/actions/runs/26959728825/job/79546483105?pr=1165


quexten (dismissed):

dani-garcia:
Seems like #1173 already took care of this, nice

PR #1168, opened 2026-06-05
[PM-38387] Introduce KeePass KDBX importer

harr1424:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-38387

## 📔 Objective

Implement logic within the SDK to support KDBX file imports across all clients. This PR also introduces a new crate: `bitwarden-importers` to contain all future import importer support within the SDK. 

[AppSec Review for new keepass crate](https://bitwarden.atlassian.net/browse/VULN-619)

[Companion clients PR](https://github.com/bitwarden/clients/pull/21052)


harr1424:
@claude  please review

harr1424:
@claude please give this a re-review, ensuring previous findings have been addressed, while also surfacing any new findings

harr1424:
@claude please give this a re-review

harr1424:
@claude please re-review, taking into consideration that data owned by the keepass crate cannot be zeroized, I believe that all owned copies of this data are now zeroized in the codepaths I control as a consumer of the keepass crate

dereknance (dismissed):

crates/bitwarden-exporters/Cargo.toml
```diff
@@ -16,7 +16,7 @@ license-file.workspace = true
 keywords.workspace = true
 
 [features]
-uniffi = ["dep:uniffi", "bitwarden-core/uniffi"]
+uniffi = ["dep:uniffi", "bitwarden-core/uniffi", "bitwarden-vault/uniffi"]
```
itsadrago: Why is the `bitwarden-vault/uniffi` addition necessary?

harr1424: This is left over from my original implementation (before 6ac13e46fbd5d9f29d6d7939e68ca4835ebb4271)  where `bitwarden-exporters` was used to contain logic for CXF/CXP and the KDBX importer. 

At first I thought this might be needed as we begin to expand CXF/CXP work, but I'm pretty confident this will never be used in `bitwarden-exporters`.

harr1424:
@Hinton what do you think of our approach here? We are extracting dedicated importers (so far only KDBX but additional are expected to follow shortly) into a new `bitwarden-importers` crate.  This leaves `bitwarden-exporters` containing only CXF/CXP as far as I can tell. 

During work on this PR, I was instructed to consider cognitive load and ensure that tools-owned crates are accurately neamed, importers should be located  in `bitwarden-importers`, exporters should be located in `bitwarden-exporters`, should our CXF/CXP logic then be extracted into a dedicated crate?

itsadrago (approved):

dereknance (approved):

PR #1176, opened 2026-06-08
[PM-34719] Add Send functionality to rust CLI

adudek-bw:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-34719

## 📔 Objective

Add Send functionality to Rust CLI

Not all functionality is there. Current functionality status:

  │       Subcommand              │                                    Status                                    │
  │ bw send <data> shortcut │ ✅ Wired (text sends)                                                        │
  │ send list                             │ ✅ Wired                                                                     │
  │ send template <object>  │ ✅ Wired (runs unauthenticated)                                              │
  │ send get <id>           │ ⚠️  Wired except --output for file sends (file-decrypt write-to-disk is TODO) │
  │ send create             │ ⚠️  Text wired; file create returns explicit "not yet implemented" error      │
  │ send edit               │ ⚠️  Wired for CLI-flag overrides; full-object encodedJson shape is TODO       │
  │ send remove-password    │ ✅ Wired                                                                     │
  │ send delete             │ ✅ Wired                                                                     │

## Breaking Change

Client fix: https://github.com/bitwarden/clients/pull/21175


crates/bw/src/tools/mod.rs
dani-garcia, line 1: We should add tools as codeowners to this directory, at the moment it's just platform:

https://github.com/bitwarden/sdk-internal/blob/fe65dc5432c08a92ec03c419d2a02a5bcdcd5cb4/.github/CODEOWNERS#L62

crates/bw/src/tools/send.rs
```diff
@@ -0,0 +1,791 @@
...
+};
+use chrono::{Duration, Utc};
+use color_eyre::eyre::{Context as _, eyre};
+use serde::Serialize;
+
+use super::{SendArgs, SendCommands};
+use crate::render::{CommandOutput, CommandResult};
+
+impl SendArgs {
+    pub async fn run(self, client: Option<PasswordManagerClient>) -> CommandResult {
```
dani-garcia: This should be changed to implement `BwCommand` to match with the new CLI command pattern:
```rust
impl BwCommand for SendArgs {
    type Client = AnyState;

    async fn run(self, AnyState { user, ..}: AnyState) -> CommandResult {
         // user is Option<PMClient> here
    }
}
```

crates/bw/src/tools/send.rs
```diff
@@ -0,0 +1,791 @@
...
+    Ok(CommandOutput::Object(Box::new(views)))
+}
+
+async fn get_send(
+    client: &PasswordManagerClient,
+    id: &str,
+    output: Option<String>,
+    text: bool,
+) -> CommandResult {
+    let send_id: SendId = id
```
dani-garcia: I believe you should be able to update the Get enum variant to contain a `SendId` ID rather than a `String` one, which should save you from having to parse it manually.  

Same for other variants with String IDs like edit, removepassword and delete.

Also we should be able to update `delete_in_days` to be a `u64` to avoid the string parsing too.

adudek-bw:
Still putting together the "to be implemented" tickets

crates/bitwarden-send/src/edit.rs
```diff
@@ -77,15 +77,37 @@ pub struct SendEditRequest {
...
     /// Authentication method for accessing this Send.
-    pub auth: SendAuthType,
+    ///
+    /// `Some(_)` overwrites the existing auth (including switching to `SendAuthType::None`
+    /// to remove all protection). `None` is a partial-update marker that tells the SDK to
+    /// preserve the existing Send's auth configuration as-is — the existing `password` hash,
+    /// email list, and `authType` discriminant are forwarded verbatim to the server. Use
+    /// `None` when the caller is editing unrelated fields and must not silently strip a
+    /// previously configured password or email-OTP gate.
+    pub auth: Option<SendAuthType>,
```
harr1424: The doc comment is really helpful, but I wonder if there is any other way to effect this behavior without using Rust's `Option` enum, where `None` typically indicates a lack or absence of value. Not a blocker for me, but I feel like this will likely trip us up in the future and cause some confusion.

harr1424 (dismissed):
The way `Option` is used here is a bit unusual but I can appreciate it solves a specific technical challenge very well. Thanks for documenting it clearly.

crates/bitwarden-send/src/edit.rs
Hinton: Second this, this seems very confusing. Can you not edit the auth type in the CLI? Are there problems passing in the existing value?

crates/bw/src/tools/mod.rs
```diff
@@ -213,13 +216,20 @@ pub struct SendArgs {
...
-        default_value = "7"
+        default_value_t = 7
     )]
-    pub delete_in_days: String,
+    pub delete_in_days: u64,
 
     #[arg(long, help = "Optional password to access this Send.")]
     pub password: Option<String>,
 
+    // TODO(PM-34719): Gate behind the SendEmailOTP feature flag once flag plumbing exists in `bw`.
```
Hinton: Pretty sure this won't ship until this flag is long gone.

crates/bw/src/tools/mod.rs
```diff
@@ -256,10 +266,17 @@ pub enum SendCommands {
...
-        #[arg(long, help = "Specify a file path to save a File-type Send to.")]
-        output: Option<String>,
+        // The legacy CLI uses `--output <path>` here, but the top-level `bw` already defines
+        // a global `-o, --output` flag for the rendered output format (JSON, etc.). Naming
+        // this `--output-file` avoids a clap runtime panic when both args resolve to the
+        // same long flag with different types. Legacy parity is tracked in PM-34719 follow-ups.
+        #[arg(
+            long = "output-file",
+            help = "File path to save a file-type Send's decrypted contents to."
+        )]
```
Hinton: I've proposed that we remove the long format of output from the global in https://github.com/bitwarden/sdk-internal/pull/1178 which would free it up. I would suggest that we deprecated output and prefer some other name but it will prevent the breaking change.

crates/bw/src/tools/send.rs
```diff
@@ -0,0 +1,1001 @@
...
+            let view = client.sends().remove_password(id).await?;
+            Ok(CommandOutput::Object(Box::new(view)))
+        }
+        SendCommands::Delete { id } => {
+            let client = require_login(client)?;
+            client.sends().delete(id).await?;
+            Ok("Send deleted.".into())
+        }
+    }
+}
```
Hinton: Someone from platform do please correct me if I'm wrong, but I believe all subcommannds should impl the BitCommand struct, and rely on that dispatch logic.

Alongside that you probably want to lift in the command arg structs into this file rather than mod.rs since they don't belong there if the impl methods are here. This would lead to something like:

```rust
#[derive(Subcommand, Clone, Debug)]
pub enum SendCommands {
    #[command(about = "Edit a Send.")]
    Edit(EditArgs)

    // Remaining ...
}

#[derive(Args, Clone)]
struct EditArgs {}

impl BwCommand for EditArgs {
    type Client = Unlocked;

    async fn run(self, _: Unlocked) -> CommandResult {
        ...
    }
}
```

crates/bw/src/tools/mod.rs
adudek-bw: Sweet, will hold until #1178 lands, then rebase and rename :)

crates/bitwarden-send/src/edit.rs
adudek-bw: Updated. For password auth, the encrypted hash isn't surfaced through SendView, so a caller doing auth: Set(view.auth_type) can't reconstruct a working SendAuthType::Password { password: <hash> }. The preserve signal has to live at the request boundary so the SDK can forward the stored encrypted form verbatim.

crates/bitwarden-send/src/edit.rs
```diff
@@ -51,6 +51,31 @@ pub enum EditSendError {
     },
 }
 
+/// Controls how `bw send edit` updates the auth on an existing Send.
+///
+/// The previous shape (`Option<SendAuthType>`) overloaded `Option` with a sentinel:
+/// `None` did *not* mean "no auth", it meant "preserve". This enum encodes the
+/// preserve-versus-overwrite distinction directly in the type so readers don't have
+/// to recover the convention from doc-comments.
```
harr1424: ❓ I am not certain the background context on this will be helpful as a long-lived doc comment, this may have been more appropriate as a comment scoped to the PR review instead

crates/bitwarden-send/src/edit.rs
harr1424, line 1: 🎨 A lot of the doc comments in this file are really heavy, remember that this is what will be shown when a developer hovers over the enum/struct/fn. I try to keep these concise and focused on what someone writing code needs to immediately know.

harr1424 (dismissed):

crates/bw/src/tools/send.rs
Hinton: The structs are still in the `mod.rs`, which seems ill suited as the implementations live in the different file. The `send.rs` is also pretty big and with the move of structs from mod would benefit from being broken up instead.

crates/bitwarden-send/src/edit.rs
```diff
@@ -77,15 +102,52 @@ pub struct SendEditRequest {
...
+}
+
+/// Snapshot of an existing Send's wire-format auth fields, used to round-trip the auth
+/// configuration through a partial edit without requiring the plaintext password.
+#[derive(Debug)]
+struct PreservedAuth {
+    password: Option<String>,
+    emails: Option<String>,
+    auth_type: crate::AuthType,
+}
```
Hinton: **question:** We don't really need this information right? When updating a send we do _not_ currently include the password in the request, we just send the `authType` and the server handles retaining it. It should be enough to just have `AuthType`?

crates/bw/src/tools/mod.rs
```diff
@@ -249,105 +258,163 @@ pub struct SendArgs {
...
-    RemovePassword { id: String },
+    #[arg(long, help = "Notes to add to the Send.")]
+    pub notes: Option<String>,
 
-    #[command(about = "Delete a Send.")]
-    Delete { id: String },
+    #[arg(long, help = "Optional password to access this Send.")]
+    pub password: Option<String>,
+
+    // TODO(PM-34719): The legacy CLI enforces `--password` and `--emails` as mutually
```
Hinton: **question:** PM-34719 is this PR right? Is there a ticket tracking this defect?

crates/bw/src/tools/send.rs
```diff
@@ -0,0 +1,1053 @@
...
+            )),
+            Some(SendCommands::Edit(args)) if args.encoded_json.is_some() => Err(eyre!(
+                "`encoded_json` input on `bw send edit` is not yet implemented (PM-34719)."
+            )),
+            // `--output-file` on `get` similarly fails before the auth check: silently
+            // emitting JSON to stdout while the requested file path goes uncreated would
+            // be a worse UX than an explicit "not implemented" error.
+            Some(SendCommands::Get(args)) if args.output_file.is_some() => Err(eyre!(
+                "`--output-file` on `bw send get` is not yet implemented (PM-34719: file-decrypt pipeline pending)."
+            )),
```
Hinton, line 296: Same here, what ticket is tracking the follow up?

crates/bw/src/tools/send.rs
```diff
@@ -0,0 +1,1053 @@
...
+        // `--output-file` should see the precise "not yet implemented" error rather than
+        // a generic auth message. See PM-34719 for the file-decrypt pipeline follow-up.
+        get_send(&user, self.id, self.text).await
+    }
+}
+
+impl BwCommand for SendReceiveArgs {
+    // The `bw receive` command handles incoming sends; `bw send receive` is the legacy
+    // alias and is out of scope for this ticket. See `Commands::Receive` in main.rs.
+    // No login is required to reach that error path.
```
Hinton: **question:** Why is it out of scope? Is this deprecated in the existing CLI, if so can we just remove it completely?

Hinton (commented):
Several todos which seems to not be tracked in jira.

crates/bw/src/tools/send.rs
adudek-bw: The ticket is here: https://bitwarden.atlassian.net/browse/PM-34718

crates/bw/src/tools/mod.rs
adudek-bw: fixed!

crates/bitwarden-send/src/edit.rs
```diff
@@ -228,6 +269,58 @@ mod tests {
...
     use crate::{AuthType, SendTextView, SendType, SendViewType};
 
+    /// `AuthEdit` is serialized with `#[serde(tag = "type")]`, and its `Set` variant
+    /// carries a `SendAuthType` that *also* uses `#[serde(tag = "type")]`. With a tuple
+    /// variant (`Set(SendAuthType)`) the inner type's tag would flatten into the outer
+    /// object and collide with the outer tag, producing `{"type":"set","type":"password",...}`
+    /// (invalid JSON; deserialize would fail with "duplicate field 'type'"). The struct
+    /// variant `Set { auth: SendAuthType }` wraps the inner type in a sub-object instead,
+    /// keeping the two `"type"` tags in separate scopes. This test pins the wire shape so
+    /// the regression can't reappear.
```
harr1424: Sometimes Claude will write itself notes, this reads that way. I would really try to rephrase this to be more concise if the function was `pub` but since it isn't declared `pub` I am not sure why this was formatted as a doc comment `///` and not a regular file comment `//`. 

Certainly not a blocker, just a nit.

crates/bitwarden-send/src/edit.rs
```diff
@@ -458,12 +555,303 @@ mod tests {
...
         assert!(matches!(result.unwrap_err(), EditSendError::Api(_)));
     }
+
+    /// Builds a key store / repository fixture pre-populated with a Send whose encrypted
+    /// row carries the provided `password_hash`, `emails`, and `auth_type`. We construct
+    /// the encrypted `Send` directly (rather than going through `encrypt(SendView)`)
+    /// because `SendView` deliberately doesn't surface the password hash — those fields
+    /// only live on the wire-format `Send` struct. Preserve-mode tests use this to assert
+    /// that the SDK does *not* forward those fields back to the server (the server
+    /// retains them based on `authType` alone).
```
harr1424: All of the content in this comment helped me to understand what is happening here, but again it seems strange doc comments were used for a function without `pub` visibility.

crates/bw/src/tools/send.rs
```diff
@@ -0,0 +1,1054 @@
...
+        run_create(&user, request, full_object).await
+    }
+}
+
+impl BwCommand for SendEditArgs {
+    type Client = LoggedIn;
+
+    async fn run(self, LoggedIn { user, .. }: LoggedIn) -> CommandResult {
+        // The `encoded_json` early-error gate lives in [`SendArgs::run`] above so it can
+        // fire *before* the `LoggedIn` typestate extractor — see PM-34719 for the
```
harr1424: I think [PM-34719](https://bitwarden.atlassian.net/browse/PM-34719) is scoped to the entire SDK CLI Send implementation, just checking did you create a different ticket scoped to this follow up work? Or should the comment be updated?

crates/bw/src/tools/send.rs
```diff
@@ -0,0 +1,1054 @@
...
+        } = self;
+
+        let send_id = itemid.ok_or_else(|| {
+            eyre!("--itemid is required (or provide it via encoded JSON, TODO PM-34719).")
+        })?;
+
+        // TODO(PM-34719): The CLI builds the edit request from the existing decrypted view
+        // plus CLI overrides. The legacy CLI lets the user supply the full object via
+        // encodedJson; we should support that path too. TODO(PM-34719): Enforce type
+        // immutability on edit.
```
harr1424: Are these still `TODO`s? I don't see any `mut` but I don't fully understand the background context.

crates/bw/src/tools/send.rs
```diff
@@ -0,0 +1,1054 @@
...
+async fn list_sends(client: &PasswordManagerClient) -> CommandResult {
+    let views = client.sends().list().await?;
+    Ok(CommandOutput::Object(Box::new(views)))
+}
+
+async fn get_send(client: &PasswordManagerClient, id: SendId, text: bool) -> CommandResult {
+    let view = client.sends().get(id).await?;
+
+    if text {
+        // TODO(PM-34719): Building a shareable access URL requires the server's web vault URL
```
harr1424: I think this is still an unresolved TODO

crates/bw/src/tools/send.rs
```diff
@@ -0,0 +1,1054 @@
...
+        delete_in_days,
+        max_access_count,
+        hidden,
+        name,
+        notes,
+        password,
+        emails,
+    } = inputs;
+
+    // TODO(PM-34719): Validate `delete_in_days` against the legacy CLI's allowed set (1, 2, 3,
```
harr1424: The function still isn't performing this validation

crates/bw/src/tools/send.rs
```diff
@@ -0,0 +1,1054 @@
...
+    // 7, 14, 30) instead of accepting any positive integer.
+    let deletion_date = compute_deletion_date(delete_in_days)?;
+
+    let view_type = match (file.as_deref(), text.as_deref()) {
+        (Some(path), None) => {
+            // TODO(PM-34719): File sends require a premium account; enforce that check before
```
harr1424: I will continue flagging these just in case, but I suspect you intended to scope these follow-ups  to other tickets?

crates/bw/src/tools/send.rs
```diff
@@ -0,0 +1,1054 @@
...
+        None => existing.deletion_date,
+    };
+
+    let view_type = match (existing.text, existing.file) {
+        (Some(t), None) => SendViewType::Text(SendTextView {
+            text: t.text,
+            hidden: if hidden { true } else { t.hidden },
+        }),
+        (None, Some(f)) => SendViewType::File(f),
+        // TODO(PM-34719): Sends should always carry exactly one of text/file; the API can in
```
harr1424: This appears resolved

crates/bw/src/tools/send.rs
```diff
@@ -0,0 +1,1054 @@
...
+}
+
+async fn run_create(
+    client: &PasswordManagerClient,
+    request: SendAddRequest,
+    full_object: bool,
+) -> CommandResult {
+    let is_file = matches!(request.view_type, SendViewType::File(_));
+
+    // TODO(PM-34719): For file sends, the create flow has to follow up with `upload_send_file`
```
harr1424: Please update this `TODO` to match the appropriate follow-up ticket

crates/bw/src/tools/send.rs
```diff
@@ -0,0 +1,1054 @@
...
+        ));
+    }
+
+    let view = client.sends().create(request).await?;
+
+    if full_object {
+        return Ok(CommandOutput::Object(Box::new(view)));
+    }
+
+    // TODO(PM-34719): The default output is the access URL. We don't have the web vault URL
```
harr1424: Please update this `TODO` to match the appropriate follow-up ticket

crates/bw/tests/send.rs
```diff
@@ -0,0 +1,478 @@
...
+//!
+//! Mocking the live `SendClient` requires a full HTTP server fixture and `BW_EMAIL`/`BW_PASSWORD`
+//! login round-trip; that's out of scope for the parse-layer wiring tests here. See
+//! `crates/bitwarden-send/src/{create,edit,delete,...}.rs` for unit tests that exercise the
+//! underlying `SendClient` methods directly.
+
+mod common;
+use common::bw;
+
+// The crate name `bw` is a binary; to access its types we have to compile it as a library too.
```
harr1424:
```suggestion
// The crate named `bw` is a binary; to access its types we have to compile it as a library too.
```

crates/bw/tests/send.rs
```diff
@@ -0,0 +1,478 @@
...
+    let stderr = String::from_utf8_lossy(&output.stderr);
+    assert!(
+        !stderr.starts_with("error: unexpected argument"),
+        "clap should accept --password on edit; got:\n{stderr}"
+    );
+}
+
+#[test]
+fn send_create_encoded_json_returns_not_implemented_error() {
+    // PM-34719: full-object JSON input is not yet implemented. Supplying it should fail
```
harr1424:
```suggestion
    // full-object JSON input is not yet implemented. Supplying it should fail
```

crates/bw/tests/send.rs
```diff
@@ -0,0 +1,478 @@
...
+    );
+    assert!(
+        !stderr.contains("not logged in"),
+        "encoded_json check should pre-empt the auth check; got:\n{stderr}"
+    );
+}
+
+#[test]
+fn send_get_output_file_returns_not_implemented_error() {
+    // PM-34719: --output-file for file Sends needs the decrypt-file pipeline, which isn't
```
harr1424:
```suggestion
    //  --output-file for file Sends needs the decrypt-file pipeline, which isn't
```

harr1424 (changes requested):
Blocker for me: I think the Jira task [PM-34719](https://bitwarden.atlassian.net/browse/PM-34719) has been broadly used to delegate follow-up work, but that specific task is also the task scoped to introducing this functionality. I think we should scope the remaining functionality to dedicated tasks so that we don't loose track of them.

As a nit, not a blocker, there are still a lot of doc comments that read as if they written by an LLM for an LLM, they are very verbose and read like the narration of an iterative development journey instead of documenting final state/functionality. Most of these doc comments don't need to be doc comments (`///`) and could be regular file comments (`//`).

crates/bw/src/tools/mod.rs
```diff
@@ -216,11 +234,16 @@ pub struct SendArgs {
         short = 'd',
         long = "deleteInDays",
         help = "The number of days in the future to set deletion date.",
-        default_value_t = 7
+        default_value_t = 7,
+        value_parser = parse_delete_in_days,
```
Hinton: **suggestion:** It _might_ be possible to use https://docs.rs/clap/latest/clap/builder/struct.PossibleValuesParser.html directly which would let you simplify this to

```suggestion
        value_parser = PossibleValuesParser::new(DELETE_IN_DAYS_ALLOWED),
```

I have not tested this though.

crates/bw/src/tools/send.rs
Hinton: The referenced ticket in code is 34719 though. Which is this PR and that issue will presumably be closed.

crates/bitwarden-send/src/edit.rs
Hinton: You can have doc comments which will provide value inside the same function. I. would argue that half of that comment conveys internal details of the function which belongs to regular comments.

crates/bw/src/tools/mod.rs
```diff
@@ -249,105 +281,167 @@ pub struct SendArgs {
...
-            short = 'd',
-            long = "deleteInDays",
-            help = "The number of days in the future to set deletion date.",
-            default_value = "7"
-        )]
-        delete_in_days: String,
+    // The legacy CLI uses `--output <path>` here, but the top-level `bw` already defines
+    // a global `-o, --output` flag for the rendered output format (JSON, etc.). Naming
+    // this `--output-file` avoids a clap runtime panic when both args resolve to the
+    // same long flag with different types. Legacy parity is tracked in PM-34719 follow-ups.
```
Hinton: **issue:** `output` is available now, can you alias this argument to `output`.

Hinton (commented):
Consider also resolving this older comment: https://github.com/bitwarden/sdk-internal/pull/1176#discussion_r3413305392

crates/bw/src/tools/send.rs
adudek-bw: updated :)

crates/bw/src/tools/send.rs
adudek-bw: Yeah, added some followup tickets

adudek-bw:
Jira followup tickets are now properly noted :)

crates/bw/src/tools/send.rs
adudek-bw: Ah, this is more about not changing between text and file type, the comment has been changed to be clearer.

harr1424 (dismissed):
Thank you for adding all of the tickets! ✅

crates/bw/src/tools/mod.rs
Hinton: Seems this doesn't work out of the box, but this does, which buys you better help message:
```txt
  -d, --deleteInDays <DELETE_IN_DAYS>
          The number of days in the future to set deletion date.

          [default: 7]
          [possible values: 1, 2, 3, 7, 14, 30]
```
```rust
/// Allowed values for `--deleteInDays`, matching the legacy CLI's enumerated set.
const DELETE_IN_DAYS_ALLOWED: &[&str] = &["1", "2", "3", "7", "14", "30"];

/// Clap value parser for `--deleteInDays`. Restricts input to the legacy-CLI allowed
/// set (also surfaced in `--help`) and maps it back to the `u64` field type.
fn delete_in_days_parser()
-> impl clap::builder::TypedValueParser<Value = u64> + Clone {
    PossibleValuesParser::new(DELETE_IN_DAYS_ALLOWED)
        .map(|s| s.parse::<u64>().expect("allowed values are valid u64"))
}
```

Hinton (commented):
Two outstanding comments, we don't _need_ to do them but a response to the negative would be nice:

- https://github.com/bitwarden/sdk-internal/pull/1176#discussion_r3413305392
- https://github.com/bitwarden/sdk-internal/pull/1176#discussion_r3434254628

Looks good once the conflicts are resolved.

crates/bw/src/tools/send.rs
adudek-bw: I'll lift SendArgs, SendCommands, and the eight Send*Args structs into send.rs

crates/bw/src/tools/mod.rs
adudek-bw: applied :)

harr1424 (dismissed):

Hinton (approved):

dani-garcia (approved):

harr1424 (approved):

PR #1181, opened 2026-06-09
[PM-37919] Compile bitwarden-ssh with ECDSA flag

neuronull:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-37919

## 📔 Objective

`clients` currently has:
- support in the SSH Agent v2 for ECDSA keys
- logic that prevents the import of ECDSA keys (at the vault level) unless ECDSA feature flag is enabled.

With especially the second item, it's now safe to enable ECDSA support in bitwarden-ssh.

## Testing done

Used a local build of the SDK to integrate into the Desktop client.

(see ticket for recordings)

With the FF turned off:
- attempt to import an ecdsa key of all curve types → it should fail with Toast message that the key type is unsupported

With the FF turned on:
- attempt to import an ecdsa key of all curve types → it should succeed
- use an imported ECDSA key in the vault to perform
    - list keys (ssh -L) operations
    - sign request (ssh -T github , or to another host e.g. docker)
    - agent forwarding

### DIFF
--- a/crates/bitwarden-ssh/Cargo.toml
+++ b/crates/bitwarden-ssh/Cargo.toml
@@ -16,8 +16,7 @@ license-file.workspace = true
 keywords.workspace = true
 
 [features]
-ecdsa-keys = [
-] # Allow ECDSA key generation and import at runtime (disabled by default, since SSH Agent does not support ECDSA keys yet)
+ecdsa-keys = [] # ECDSA key generation and import
 wasm = [
     "bitwarden-error/wasm",
     "dep:tsify",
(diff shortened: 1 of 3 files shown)


quexten (approved):
@neuronull Can we make a ticket for compile time flag cleanup? I think we can remove the ecdsa-keys flag entirely and just default it in?

neuronull:
> @neuronull Can we make a ticket for compile time flag cleanup? I think we can remove the ecdsa-keys flag entirely and just default it in?

https://bitwarden.atlassian.net/browse/PM-38832

dereknance (approved):

neuronull:
(removing needs-qa , that will only happen on the clients repo)

PR #1182, opened 2026-06-10
[PM-38816] New crypto lib in new importers crate for Keeper DI

itsadrago:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-38816

## 📔 Objective

Need to add a new importers crate that includes crypto functions that Keeper DI needs. We're moving the crypto code from the clients repo to the sdk-internal repo

## Dependency Review
https://bitwarden.atlassian.net/browse/VULN-650 (aes-gcm)


itsadrago:
@claude perform a code review

itsadrago:
@claude perform another code review and make sure previous findings have been addressed

coroiu (commented):
Dismissing review since PR is marked as Draft, let me know when it's ready by re-requesting a review! :)

itsadrago:
@claude perform another code review

itsadrago:
@claude perform another code review and make sure previous findings have been addressed

harr1424 (dismissed):

harr1424 (dismissed):

harr1424 (dismissed):

crates/bitwarden-importers/src/importers/keeper/crypto.rs
```diff
@@ -0,0 +1,708 @@
...
+//!
+//! Every function here must stay compatible with data produced by Keeper's clients; do not change
+//! the formats.
+
+use aes::cipher::{
+    BlockModeDecrypt, KeyIvInit,
+    block_padding::{NoPadding, Pkcs7},
+};
+use aes_gcm::{Aes256Gcm, KeyInit, Nonce, aead::Aead};
+use bitwarden_crypto::pbkdf2;
```
quexten: Actually, I'd recommend just using plain pbkdf2. Thereis no reason to create a dep on the bitwarden-crypto crate which really is meant for creating the password manager crypto, not implementing support for external crypto of other managers.

crates/bitwarden-importers/src/importers/keeper/crypto.rs
```diff
@@ -0,0 +1,708 @@
...
+    cipher
+        .decrypt(&nonce_arr, ciphertext)
+        .map_err(|_| KeeperCryptoError::Decryption)
+}
+
+/// Encrypt `data` with an RSA public key using PKCS#1 v1.5 padding.
+///
+/// `public_key` is a DER-encoded PKCS#1 `RSAPublicKey` (the format Keeper ships its server keys
+/// in).
+pub fn encrypt_rsa(data: &[u8], public_key: &[u8]) -> Result<Vec<u8>, KeeperCryptoError> {
```
quexten, line 198: A question. Is there a reason we are moving *only* the crypto instead of a high level operation to the SDK? this will put more work on mobile and give them more opportunity to get it wrong.

crates/bitwarden-uniffi/src/tools/mod.rs
```diff
@@ -106,4 +111,117 @@ impl ImporterClient {
...
+
+    /// Keeper "direct" importer cryptography. Returns a stateless client exposing Keeper's
+    /// wire-format crypto primitives.
+    pub fn keeper_crypto(&self) -> Arc<KeeperCryptoClient> {
+        Arc::new(KeeperCryptoClient(self.0.keeper_crypto()))
+    }
+}
+
+#[derive(uniffi::Object)]
+pub struct KeeperCryptoClient(pub(crate) bitwarden_importers::keeper::KeeperCryptoClient);
```
quexten: Two notes: `tool` seems like the wrong subdirectory. Shouldn't it be `tools`?

This is actually not the recommended way to create uniffi clients. Please hav a look here https://contributing.bitwarden.com/architecture/sdk/client-patterns

crates/bitwarden-uniffi/src/tools/mod.rs
```diff
@@ -106,4 +111,117 @@ impl ImporterClient {
...
+
+/// Keeper "direct" importer cryptography, exposed for the still-TypeScript Keeper access layer.
+///
+/// These primitives implement Keeper's competitor wire formats (unauthenticated AES-CBC,
+/// AES-GCM with a prepended nonce, RSA PKCS#1 v1.5, ECDH-P256 → AES-GCM, and the custom
+/// `encryptionParams` blob). They are not Bitwarden cryptography.
+#[uniffi::export]
+impl KeeperCryptoClient {
+    /// Generate a new 32-byte AES encryption key.
+    pub fn generate_encryption_key(&self) -> Vec<u8> {
```
quexten: I'm concerned about the low-level nature of this interface. I also think this could use stronger types instead of just byte arrays. It's incredibly easy to get this wrong for the consumer (mobile, tools for clients). This entire interface is essentially tech debt at creation.

I'll also note that the naming is unclear for a lot of these without extra documentation. E.g "generate encryption key"

quexten (changes requested):
Not my code ownership for most of this, but I find the ultra-low-level nature of this very concerning. Ideally, even if it is just the crypto in the sdk, not the API calls, you would document what kind of structured messages keeper has and pass those in instead of passing in just raw byte arrays to decrypt / encrypt.

The only blocking comment here is the one about PBKDF2.

Cargo.toml
mzieniukbw, line 1: ❓ Were the dependency versions upgrades necessary ? Would be safer to revert it if it is not needed and do the upgrade separately with appropriate QA testing.

crates/bitwarden-uniffi/src/tools/mod.rs
itsadrago: I've completed the rename. Am still looking at the other comment

crates/bitwarden-importers/src/importers/keeper/crypto.rs
itsadrago: The crypto file migration is one part of a larger effort to move all the Keeper DI business logic over from the clients repo. The idea was to migrate smaller chunks of code at a time so that it would be easier to review and understand. This crypto code won't ultimately be consumed directly by the various clients through uniffi or wasm, but rather by other code in the bitwarden-importers crate that will expose higher level interfaces

crates/bitwarden-uniffi/src/tools/mod.rs
itsadrago: I've removed the uniffi and wasm bindings. I will create new ones when more of the higher level code is migrated

crates/bitwarden-uniffi/src/tools/mod.rs
itsadrago: After giving it some thought, related to my previous comment, I made a mistake in adding the uniffi and wasm interfaces this early. They should be a part of another code migration effort that will expose higher level interfaces that will eventually consume crypto code

crates/bitwarden-importers/src/importers/keeper/crypto.rs
```diff
@@ -0,0 +1,706 @@
...
+pub fn base64_url_decode(text: &str) -> Result<Vec<u8>, KeeperCryptoError> {
+    let trimmed = text.trim_end_matches('=');
+    data_encoding::BASE64URL_NOPAD
+        .decode(trimmed.as_bytes())
+        .map_err(|_| KeeperCryptoError::InvalidData)
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
```
quexten, line 397: Suggestion: Can we please record test vectors from keeper and make sure they decrypt here? If we do not have test vectors we cannot be confident that our code is correct, and that changes are correct, and that we keep stable support for their format.

crates/bitwarden-importers/src/importers/keeper/crypto.rs
```diff
@@ -0,0 +1,706 @@
...
+/// Total length of a valid Keeper `encryptionParams` blob: version(1) + iterations(3) + salt(16) +
+/// iv(16) + two 32-byte key blocks(64).
+const ENCRYPTION_PARAMS_SIZE: usize = 1 + 3 + 16 + 16 + 64;
+
+/// Errors produced by the Keeper crypto layer.
+///
+/// Messages are intentionally coarse and never include key material, plaintext or ciphertext, so an
+/// error can be surfaced to a client or log without leaking secrets.
+#[derive(Debug, thiserror::Error)]
+pub enum KeeperCryptoError {
```
quexten, line 52: Is there a reason all structs and functions in this module are `pub` instead of `pub(crate)`/ `pub(super)`? For our codebase, the expectation is to be as restrictive as possible, and we do not want other crates to depend on functions such as the base64 decode / encode.

This comment applies to the entire file and all instances of pub.

crates/bitwarden-importers/src/importers/keeper/crypto.rs
```diff
@@ -0,0 +1,706 @@
...
+}
+
+/// Encrypt `data` with an RSA public key using PKCS#1 v1.5 padding.
+///
+/// `public_key` is a DER-encoded PKCS#1 `RSAPublicKey` (the format Keeper ships its server keys
+/// in).
+pub fn encrypt_rsa(data: &[u8], public_key: &[u8]) -> Result<Vec<u8>, KeeperCryptoError> {
+    let key =
+        RsaPublicKey::from_pkcs1_der(public_key).map_err(|_| KeeperCryptoError::InvalidKey)?;
+    let mut rng = rand::rng();
```
quexten, line 201: Please note, we will have to move this to use the new `bitwarden-random` crate when https://github.com/bitwarden/sdk-internal/pull/1223 merges.

crates/bitwarden-importers/src/importers/keeper/crypto.rs
```diff
@@ -0,0 +1,706 @@
...
+/// A freshly generated P-256 key pair: PKCS#8 DER private key and uncompressed SEC1 public key.
+pub struct EcKeyPair {
+    /// PKCS#8 DER-encoded private key.
+    pub private_key: Vec<u8>,
+    /// Uncompressed SEC1 public key (`0x04 || X || Y`, 65 bytes).
+    pub public_key: Vec<u8>,
+}
+
+/// Generate `length` cryptographically secure random bytes.
+pub fn get_random_bytes(length: usize) -> Vec<u8> {
```
quexten, line 103: Please do not use Vec<u8> everywhere. This makes it hard to read, and also hard for the compiler to prevent mistakes. Currently it's entirely possible to generate random bytes, and make an EcKeyPair with the bytes obtained from it. We should prevent this  on the compiler level. I suggest making NewTypes for all Vec<u8> instances.

These can be passed across the FFI boundary. Ref:
https://github.com/bitwarden/sdk-internal/blob/a4fae39c4b2c82e5226a0917d80063e5388f9e80/crates/bitwarden-ipc/src/crypto_provider/noise/transport_state.rs#L21 , https://github.com/bitwarden/sdk-internal/blob/a4fae39c4b2c82e5226a0917d80063e5388f9e80/crates/bitwarden-crypto/src/enc_string/symmetric.rs#L90

This comment applies to all instances of `Vec<u8>`.

quexten, line 103: I suggest not having a generic function such as `get_random_bytes` being public (even on the mod level). Why not just expose the `generate_encryption_key()`?

crates/bitwarden-importers/src/importers/keeper/crypto.rs
```diff
@@ -0,0 +1,706 @@
...
+    let mut buf = vec![0u8; length];
+    rand::rng().fill_bytes(&mut buf);
+    buf
+}
+
+/// Generate a new 32-byte AES encryption key.
+pub fn generate_encryption_key() -> Zeroizing<Vec<u8>> {
```
quexten, line 110: If you make a new type as suggested above, this should actually live in an impl for the newtype. i.e:

```rust
//please note i do not know if this naming is correct
struct KeeperAesV1EncryptionKey([u8; KEEPER_KEY_SIZE]);

impl KeeperAesV1EncryptionKey {
  fn make() {
    let mut buf = ...
    // rng
    KeeperAesV1EncryptionKey(buf)
  }
}
```

crates/bitwarden-importers/src/importers/keeper/crypto.rs
```diff
@@ -0,0 +1,706 @@
...
+    Zeroizing::new(get_random_bytes(KEEPER_KEY_SIZE))
+}
+
+/// Decrypt an "aes-v1" packet: AES-256-CBC with PKCS#7 padding and no authentication (no MAC).
+///
+/// The packet is `IV(16) || ciphertext`. Keeper aes-v1 is unauthenticated, so a successful decrypt
+/// does not prove integrity — callers must not rely on this for tamper detection.
+pub fn decrypt_aes_v1(data: &[u8], key: &[u8]) -> Result<Vec<u8>, KeeperCryptoError> {
```
quexten, line 118: This should take a keeper key newtype as suggested above. Also, we should move this into the impl. Do you know if the same key is reused for aes_v1 and aes_v2 for keeper? If so put them into the same impl for the newtype, otherwise have separate impls / newtypes.

quexten, line 118: Suggestion: Make data a newtype.

```rust
struct KeeperV1Packet { iv, ciphertext }

impl TryFrom<&[u8]> for KeeperV1Packet {
   // impl the packet parsing here
}
```

then, the decrypt function does *one* thing not parsing + decrypting.

```rust
impl KeeperV1Key {
   fn decrypt(packet: &KeeperV1Packet) {
    let decrypted = cbc::Decryptor::<aes::Aes256>::new_from_slices(self.0, iv)
        .map_err(|_| KeeperCryptoError::InvalidKey)?
        .decrypt_padded::<Pkcs7>(&mut buf)
        .map_err(|_| KeeperCryptoError::Decryption)?;
    let len = decrypted.len();
    buf.truncate(len);
    Ok(buf)
   }
}
```

Then, the FFI implementation should automatically call the try from for the keeper packet.

crates/bitwarden-importers/src/importers/keeper/crypto.rs
```diff
@@ -0,0 +1,706 @@
...
+
+/// Encrypt an "aes-v2" packet with an explicit nonce. Kept private and used only by
+/// [`encrypt_aes_v2`] (random nonce) and the known-answer tests, so the nonce-reuse hazard is never
+/// exposed to callers.
+fn encrypt_aes_v2_with_nonce(
+    data: &[u8],
+    key: &[u8],
+    nonce: &[u8],
+) -> Result<Vec<u8>, KeeperCryptoError> {
+    if key.len() != KEEPER_KEY_SIZE {
```
quexten, line 154: You don't need all of these guard checks if you use newtypes. Way easier to read, and easier to keep correct.

crates/bitwarden-importers/src/importers/keeper/crypto.rs
```diff
@@ -0,0 +1,706 @@
...
+        return Err(KeeperCryptoError::InvalidKey);
+    }
+    let nonce_bytes: [u8; AES_GCM_NONCE_SIZE] = nonce
```
quexten, line 157: Is there a reason to even have a function where the nonce can be passed in? Even if private, why have it at all?

crates/bitwarden-importers/src/importers/keeper/crypto.rs
```diff
@@ -0,0 +1,706 @@
...
+}
+
+/// Encrypt an "aes-v2" packet with an explicit nonce. Kept private and used only by
+/// [`encrypt_aes_v2`] (random nonce) and the known-answer tests, so the nonce-reuse hazard is never
+/// exposed to callers.
+fn encrypt_aes_v2_with_nonce(
+    data: &[u8],
+    key: &[u8],
+    nonce: &[u8],
+) -> Result<Vec<u8>, KeeperCryptoError> {
```
quexten, line 153: Suggestion, make the return type a KeeperV2Packet

crates/bitwarden-importers/src/importers/keeper/crypto.rs
```diff
@@ -0,0 +1,706 @@
...
+    cipher
+        .decrypt(&nonce_arr, ciphertext)
+        .map_err(|_| KeeperCryptoError::Decryption)
+}
+
+/// Encrypt `data` with an RSA public key using PKCS#1 v1.5 padding.
+///
+/// `public_key` is a DER-encoded PKCS#1 `RSAPublicKey` (the format Keeper ships its server keys
+/// in).
+pub fn encrypt_rsa(data: &[u8], public_key: &[u8]) -> Result<Vec<u8>, KeeperCryptoError> {
```
quexten, line 198: Do you know what data is encrypted by rsa in the keeper protocol? Is there a high level draft I can read?

crates/bitwarden-importers/src/importers/keeper/crypto.rs
```diff
@@ -0,0 +1,706 @@
...
+    Zeroizing::new(hash.into())
+}
+
+/// Encrypt `data` for an EC public key (Keeper's ECC scheme).
+///
+/// An ephemeral key pair is generated; the shared key is `SHA-256(ECDH(recipient, ephemeral))` and
+/// the payload is sealed with aes-v2. The output is `ephemeralPublic(65) || aes-v2 packet`.
+///
+/// `public_key` is an uncompressed SEC1 point (`0x04 || X || Y`, 65 bytes).
+pub fn encrypt_ec(data: &[u8], public_key: &[u8]) -> Result<Vec<u8>, KeeperCryptoError> {
```
quexten, line 244: Newtypes would be nice.

crates/bitwarden-importers/src/importers/keeper/crypto.rs
```diff
@@ -0,0 +1,706 @@
...
+        .to_vec();
+    let public_key = secret.public_key().to_sec1_point(false).as_bytes().to_vec();
+    Ok(EcKeyPair {
+        private_key,
+        public_key,
+    })
+}
+
+/// Derive Keeper's ECDH content-encryption-key: ECDH(P-256) shared X coordinate, then SHA-256.
+fn derive_ecdh_key(public: &PublicKey, secret: &SecretKey) -> Zeroizing<[u8; 32]> {
```
quexten, line 232: Is an ECDH key a specific key type? I.e keeperv1 / keeperv2 (I'm still not sure if these key types are actually separate in keepers protocol).

crates/bitwarden-importers/src/importers/keeper/crypto.rs
```diff
@@ -0,0 +1,706 @@
...
+/// the payload is sealed with aes-v2. The output is `ephemeralPublic(65) || aes-v2 packet`.
+///
+/// `public_key` is an uncompressed SEC1 point (`0x04 || X || Y`, 65 bytes).
+pub fn encrypt_ec(data: &[u8], public_key: &[u8]) -> Result<Vec<u8>, KeeperCryptoError> {
+    let recipient =
+        PublicKey::from_sec1_bytes(public_key).map_err(|_| KeeperCryptoError::InvalidKey)?;
+    let ephemeral = SecretKey::generate_from_rng(&mut rand::rng());
+    let encryption_key = derive_ecdh_key(&recipient, &ephemeral);
+    let encrypted = encrypt_aes_v2(data, encryption_key.as_slice())?;
+    let ephemeral_public = ephemeral.public_key().to_sec1_point(false);
```
quexten, line 250: We mix encoding, validation, etc. a lot here. Newtypes will help make this code readable and safe.

quexten (changes requested):
I reviewed this under the assumption that there is a plan to move higher levels into the SDK at later stages. The feedback applies to any of the SDK migration levels, and is generic crypto, rust, SDK code standards advice.

crates/bitwarden-importers/src/importers/keeper/crypto.rs
```diff
@@ -0,0 +1,706 @@
+//! Keeper "direct" importer cryptography.
+//!
+//! This is a byte-for-byte port of the Keeper access layer's `crypto.ts`. It implements
+//! **Keeper's** competitor wire formats, not Bitwarden's, so it deliberately does **not** live in
```
harr1424, line 4:
```suggestion
//! **Keeper's** formats, not Bitwarden's, so it deliberately does **not** live in
```

crates/bitwarden-importers/src/importers/keeper/crypto.rs
```diff
@@ -0,0 +1,706 @@
...
+    Pkcs1v15Encrypt, RsaPrivateKey, RsaPublicKey,
+    pkcs1::{DecodeRsaPrivateKey, DecodeRsaPublicKey},
+};
+use sha2::{Digest, Sha256};
+use subtle::ConstantTimeEq;
+use zeroize::Zeroizing;
+
+/// AES symmetric key size used by Keeper (256-bit).
+const KEEPER_KEY_SIZE: usize = 32;
+/// AES-CBC block size / IV size.
```
harr1424, line 35: ❓ What is IV an acronym for?

crates/bitwarden-importers/README.md
```diff
@@ -9,6 +9,19 @@ Exposes `client.importers()` ([`ImporterClient`]) with:
   encrypts the entries for the user's personal vault or a given organization, and submits them to
   the server's import endpoint. Returns per-type counts.
 
+## Keeper direct importer
+
+The Keeper "direct" importer logs into Keeper's API and decrypts the vault on-device. Its access
+layer is being ported from TypeScript (`clients` repo) into the [`keeper`] module incrementally
+(strangler-fig). The first piece, [`keeper::crypto`], implements **Keeper's** competitor wire
```
harr1424, line 16: ❓ What is the relationship between this plant and the incremental port?  (strangler-fig)

harr1424 (commented):
Minor questions only, no blockers for me. I've deferred approval pending another reviewer's requested changes.

crates/bitwarden-importers/src/importers/keeper/crypto.rs
quexten: Note: We use IV a lot in the crypto crate and it seems okay to use. https://en.wikipedia.org/wiki/Initialization_vector

crates/bitwarden-importers/README.md
itsadrago: This PR doesn't really take part in the strangler pattern. We won't start any client integration until we migrate the protobuf messages layer over

quexten (approved):

crates/bitwarden-importers/Cargo.toml
```diff
@@ -46,20 +47,24 @@ bitwarden-error = { workspace = true }
...
 bitwarden-vault = { workspace = true }
+cbc = { workspace = true }
 chrono = { workspace = true, features = ["std"] }
 crypto-bigint = { workspace = true }
 data-encoding = { workspace = true }
 hkdf = { workspace = true }
 hmac = { workspace = true }
 icu_normalizer = { workspace = true }
 keepass = { workspace = true }
+p256 = { version = "=0.14.0", features = ["alloc", "ecdh", "pkcs8"], default-features = false }
```
harr1424, line 58: ✅ https://github.com/bitwarden/sdk-internal/blob/d2bd7a3eb06e8ea52c84e437b4c14c18c5fbec1a/crates/bitwarden-ssh/Cargo.toml#L32

Cargo.toml
```diff
@@ -21,6 +21,7 @@ keywords = ["bitwarden"]
 [workspace.dependencies]
 
 # External crates that are expected to maintain a consistent version across all crates
+aes = { version = "0.9.0", features = ["zeroize"] }
```
harr1424, line 24: ✅  https://bitwarden.atlassian.net/browse/VULN-650

Cargo.toml
```diff
@@ -74,6 +75,7 @@ bitwarden-user-crypto-management = { path = "crates/bitwarden-user-crypto-manage
 bitwarden-uuid = { path = "crates/bitwarden-uuid", version = "=3.0.0" }
 bitwarden-uuid-macro = { path = "crates/bitwarden-uuid-macro", version = "=3.0.0" }
 bitwarden-vault = { path = "crates/bitwarden-vault", version = "=3.0.0" }
+cbc = { version = "0.2.0", features = ["alloc", "zeroize"] }
```
harr1424, line 78: ✅  https://bitwarden.atlassian.net/browse/VULN-652

Cargo.toml
```diff
@@ -95,6 +97,7 @@ js-sys = { version = ">=0.3.72, <0.4" }
 keepass = { version = ">=0.13.7, <0.14", default-features = false }
 mockall = { version = ">=0.13.1, <0.16" }
 password-rules-parser = ">=1.1.0, <2"
+pbkdf2 = { version = "0.13.0-rc.10", default-features = false }
```
harr1424, line 100: ✅ https://github.com/bitwarden/sdk-internal/blob/b19177ea563f0d755227d931034015eb51cd751d/crates/bitwarden-crypto/Cargo.toml#L65

harr1424 (approved):

dani-garcia (approved):

PR #1184, opened 2026-06-11
[PM-36046] Fix 1pux SSH key formatting during import

harr1424:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-36046

## 📔 Objective

A [comment](https://github.com/bitwarden/clients/issues/20432#issuecomment-4635602313) on https://github.com/bitwarden/clients/issues/20432 revealed that `1pux` files contain SSH keys in a format divergent from what [RFC 7468](https://datatracker.ietf.org/doc/html/rfc7468) expects: 

>Generators MUST wrap the base64-encoded lines so that each line consists of exactly 64 characters...

This PR adds re-try logic in the event that `parse_pkcs8_pem()` returns `SshKeyImportError::Parsing` and will re-format the SSH key to match the RFC before re-attempting the parse.


crates/bitwarden-ssh/src/import.rs
```diff
@@ -37,9 +37,25 @@ pub fn import_key(
 fn import_pkcs8_key(
     encoded_key: String,
     password: Option<String>,
+) -> Result<SshKeyView, SshKeyImportError> {
+    match parse_pkcs8_pem(&encoded_key, password.as_deref()) {
+        // Some exporters (e.g. 1Password's 1PUX) emit the base64 body on a single line, which the
```
quexten, line 42: Two thoughts on this:
- If 1Password is not compliant with the RFC spec for generators, then we should report this as a bug to them
- We should try to upstream this to the ssh-key crate and not maintain this ourselves (or, maintain it for an interim period until upstreamed)

crates/bitwarden-ssh/src/import.rs
```diff
@@ -228,6 +287,24 @@ mod tests {
         assert_eq!(result.unwrap_err(), SshKeyImportError::WrongPassword);
     }
 
+    /// 1Password's 1PUX export re-encodes Ed25519 keys as PKCS#8 (`BEGIN PRIVATE KEY`) with the
+    /// whole base64 body on a single line. The strict RFC 7468 parser (`pem-rfc7468`) rejects the
+    /// unwrapped body, surfacing as "Failed to parse key". Wrapped bodies (our other fixtures,
+    /// `ssh-keygen` output) hide this, so the import must also accept the unwrapped form.
+    /// https://github.com/bitwarden/clients/issues/20432
+    #[test]
+    fn import_key_ed25519_pkcs8_unencrypted_single_line() {
```
quexten, line 295: Instead of constructing a test file here dynamically, can we please instead just save a test file with a single line as `ed25519_pkcs8_1password_single_line_unencrypted`?

quexten (changes requested):

crates/bitwarden-ssh/src/import.rs
harr1424: Per the [RFC](https://datatracker.ietf.org/doc/html/rfc7468) I don't think 1Password is technically non-compliant: 

> Generators MUST wrap the base64-encoded lines so that each line
   consists of exactly 64 characters except for the final line, which
   will encode the remainder of the data (within the 64-character line
   boundary), and they MUST NOT emit extraneous whitespace.  Parsers MAY
   handle other line sizes.  

1Password is not generating the key, they are reformatting it, the `ssh-key` crate expects to parse the key as if it had been generated and not re-formatted. I personally don't think 1Password will change their `1pux` output format, according to the GitHub user that opened the issue ProtonPass seems to handle this, but I agree that adding flexibility to parsers used in the `ssh-key` crate would be very valuable, since the RFC grants more flexibility to parsers.

quexten (approved):

dereknance (approved):

PR #1186, opened 2026-06-11
[PM-34053] Swap to new get_key_rotation_data for user key rotation

Thomas-Avery:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-34053

## 📔 Objective

Requires https://github.com/bitwarden/server/pull/7772 to be merged prior. 

Use the new endpoint for fetching all the key data needed for a user key rotation.


crates/bitwarden-user-crypto-management/src/key_rotation/mod.rs
Thomas-Avery, line 27: It would be better to merge `get_untrusted_organization_public_keys` and `get_untrusted_emergency_access_public_keys` into one method instead of calling `get_key_rotation_data` twice but I'm leaving that out of scope for this PR.

crates/bitwarden-user-crypto-management/src/key_rotation/mod.rs
```diff
@@ -28,10 +28,10 @@ impl UserCryptoManagementClient {
         &self,
     ) -> Result<Vec<V1OrganizationMembership>, RotateUserKeysError> {
         let api_client = &self.client.internal.get_api_configurations().api_client;
```
quexten, line 30: Note (non-blocking): Given these are called together here: 

https://github.com/bitwarden/clients/blob/c66167f574fe6d2ad93c2afe219a6942a0718005/libs/user-crypto-management/src/user-key-rotation.service.ts#L125

It seems unfortunate to do a sync twice (once for EA, once for orgs). I don't know if the HTTP client caches the calls, but ideally we would provide a single public function that replaces the current functions that provides both.

crates/bitwarden-user-crypto-management/src/key_rotation/unlock.rs
```diff
@@ -32,6 +42,24 @@ pub struct V1EmergencyAccessMembership {
...
+        Ok(Self {
+            id: require!(ea.id),
+            grantee_id: require!(ea.grantee_id),
+            // The name can be null if a user does not set a name; fall back to the email and
+            // then to "Unknown" so we always have a non-empty display name.
+            name: ea
+                .grantee_name
+                .or(ea.grantee_email)
+                .unwrap_or_else(|| "Unknown".to_string()),
+            public_key: parse_public_key(&require!(ea.public_key))?,
```
quexten, line 58: Do we know if users exist for which the user is in EA as a grantee, and also they lost their public key? if the public key is null here, (or corrupted i suppose), then this would fail rotation.

(I'm not saying these exist, and it's a case worth caring about, I wanted to ask what you think).

crates/bitwarden-user-crypto-management/src/key_rotation/unlock.rs
```diff
@@ -42,6 +70,24 @@ pub struct V1OrganizationMembership {
...
 }
 
+impl TryFrom<OrganizationPasswordResetKeyDataResponseModel> for V1OrganizationMembership {
+    type Error = KeyRotationDataParseError;
+
+    fn try_from(o: OrganizationPasswordResetKeyDataResponseModel) -> Result<Self, Self::Error> {
+        Ok(Self {
+            organization_id: require!(o.organization_id),
+            name: require!(o.organization_name),
+            public_key: parse_public_key(&require!(o.organization_public_key))?,
```
quexten, line 80: It's possible for organizations to loose their public key (or it was for a good while). I wonder how we should handle that. At this point @eliykat  do you know if we regenerate the public key for organizations that lots it via a plan change?

quexten (approved):
Some non-blocking comments

crates/bitwarden-user-crypto-management/src/key_rotation/mod.rs
quexten: https://github.com/bitwarden/sdk-internal/pull/1186#discussion_r3416427940

Ah, seems you also noted this.

crates/bitwarden-user-crypto-management/src/key_rotation/unlock.rs
eliykat: Yes - we do this during plan change or account/settings update: https://github.com/search?q=repo%3Abitwarden%2Fclients%20backfill&type=code

At first, organizations did not have pub/priv keypairs, and this was a soft migration to create them for old orgs. It also works to help organizations recover from that bug if they lose them.

But this does mean that in both cases, it's possible (although very rare) that an org does not have a pub/priv keypair and you should handle that case.

crates/bitwarden-user-crypto-management/src/key_rotation/unlock.rs
Thomas-Avery: I don't know of any case that has happen mostly because of the limited use of EA.

From my context the null case seems unlikely as they users wouldn't be able to setup a emergency access if they have a NULL or corrupted private key. The corrupted private key seems like the more likely of the two if these things actually exist.

One option to guard against this case is to bring the entire population through private key regen.

crates/bitwarden-user-crypto-management/src/key_rotation/unlock.rs
Thomas-Avery: I think the appropriate handling for user key rotation is to error and not process which it does today and in this PR. 

I assume the user would need to be an admin level to be able to invoke the organization pub/priv backfill?

quexten: Yeah. The org needs to be fixed by the admin first. We'll have to do eager migrations, and I'll recommend doing this as part of v2 organization memberships.

PR #1188, opened 2026-06-12
[PM-33107] Blob encryption key rotation migration

shane-melton:
## 🎟️ Tracking

[PM-33107](https://bitwarden.atlassian.net/browse/PM-33107)

Stacked on top of #1122 ([PM-32695] Cipher blob encryption).

## 📔 Objective

Make key rotation upgrade every individual-vault cipher to the sealed-blob format. Key rotation always lands the account on the V2 security state, so all individual ciphers must end up blob-encrypted afterward. `reencrypt_ciphers` now handles three cases:

- **Already blob-encrypted (with a per-item key):** only re-wrap the per-item cipher key under the new user key; the sealed blob payload is left untouched (no decrypt/re-encrypt).
- **Legacy with a per-item key:** re-wrap the key under the new user key, decrypt, then re-seal as a blob.
- **Legacy keyless:** decrypt under the current key and re-seal as a blob (sealing generates a fresh per-item key).

Supporting changes:

- `decrypt_blob_cipher` now takes an explicit `wrapping_key` slot instead of always deriving it from `cipher.key_identifier()`. During rotation the CEK is wrapped under a `Local` slot holding the new user key, so the wrapping key can no longer be assumed to be the user/organization key.
- `EncryptMode::Blob` routes through `encrypt_blob_cipher_with_wrapping_key`, respecting the caller-supplied key slot so rotation can target the new-user-key slot.
- `EncryptMode` and `Cipher::is_blob_encrypted()` are made public for use by the rotation crate.
- New tests covering all three reencryption paths (keyless-legacy upgrade, keyed-legacy upgrade, and existing-blob rewrap-without-re-encrypt).

## 🚨 Breaking Changes

None for clients. `EncryptMode` and `Cipher::is_blob_encrypted()` become public, and `decrypt_blob_cipher` gains a `wrapping_key` parameter (crate-internal).

[PM-33107]: https://bitwarden.atlassian.net/browse/PM-33107?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[PM-32695]: https://bitwarden.atlassian.net/browse/PM-32695?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mzieniukbw (dismissed):

shane-melton:
Had to rebase after the parent branch was merged to `main` no other changes were made since last review.

nick-livefront (approved):

mzieniukbw (approved):

PR #1196, opened 2026-06-18
[PM-39232] Stabilize invite key format

quexten:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-39232

## 📔 Objective

The `Invite` previously directly was a `EncString` under the hood. This meant that it would not be compatible with the milestone 3 work, since milestone 3 requires a new field being added (the invite key encrypted organization key).

This PR does the following:
- Update the API and stabilize it
  - We have a test vector included here. This test vector commits to a version that will always be kept compatible / ensures we do not break compatibility in the future
- Simplify verbiage to make the API easier to understand
- Add integration tests in the wasm layer to make sure the API is stable across the wasm FFI boundary

## 🚨 Breaking Changes

This is not breaking from the public API perspective, however the crypto format is a breaking change. This is not yet used in prod.


crates/bitwarden-organization-crypto/src/invite_key_bundle.rs
```diff
@@ -99,34 +99,59 @@ impl Serialize for InviteKeyData {
...
 "#;
 
-/// Struct for holding the wrapped invite key data. Currently supports encstring
-/// but the inner type must remain private as it may be extended in the future.
-#[derive(Clone)]
-pub struct InviteKeyEnvelope(EncString);
+/// Cryptographic invite for an organization.
+#[derive(Debug, Clone)]
+pub struct Invite {
+    organization_key_wrapped_invite_key: EncString,
```
quexten, line 109: Notes on reasoning here:
- Not envelope because this struct is more complex / contains other things. An envelope seals a single thing, but an invite represent more complex operations
  - create_invite
  - enable auto confimation
Thus, `Invite` is an easier term to use here.

crates/bitwarden-organization-crypto/src/invite_key_bundle.rs
```diff
@@ -163,16 +188,49 @@ impl InviteKeyEnvelope {
                 .clone(),
         ))
     }
+
+    #[allow(unused)]
+    fn enable_confirmation<Ids: KeySlotIds>(
+        &mut self,
+        _organization_key: Ids::Symmetric,
```
quexten, line 195: Added here to make sure we commit to an API we want to provide / know what functions we want to implement. I feel this missing is what lead to the previous design.

.github/CODEOWNERS
```diff
@@ -54,6 +54,8 @@ crates/bitwarden-user-crypto-management/** @bitwarden/team-key-management-dev
 crates/bitwarden-vault/** @bitwarden/team-vault-dev
 
 # Team-owned folders in other crates (to be avoided if possible)
+crates/bitwarden-wasm-internal/integration-tests/tests/invite/** @bitwarden/team-admin-console-dev
+crates/bitwarden-wasm-internal/integration-tests/tests/unlock/** @bitwarden/team-key-management-dev
```
quexten, line 58: To the platform reviewer: You don't have to review the tests, I expect these are reviewed by the owning teams :)

eliykat (approved):

mzieniukbw (approved):

djsmith85 (approved):
Approving changes to platform-owned files

PR #1197, opened 2026-06-18
Create an option to send already hashed passwords through SDK

adudek-bw:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-38391

## 📔 Objective

This PR adds a new `SendAuthType::HashedPassword { key_b64: String }` variant whose `auth_data` returns the supplied `keyB64` verbatim — no second PBKDF2 pass. The wire field is `keyB64` (matching the established camelCase convention in the clients code). The existing `Password { password }` variant is unchanged, so internal callers that hold plaintext (e.g. `bw send --password <plaintext>`) keep working as-is.

## 🚨 Breaking Changes

Additive variant — existing callers are unaffected and existing JSON still deserializes. TypeScript consumers that do exhaustive switch (auth.type) without a default arm will see a new case to handle once the bindings regenerate; the automated TypeScript compatibility check on this PR will flag any concrete breaks.

### DIFF
--- a/crates/bitwarden-send/src/send.rs
+++ b/crates/bitwarden-send/src/send.rs
@@ -120,10 +120,21 @@ pub enum AuthType {
 pub enum SendAuthType {
     /// No authentication required
     None,
-    /// Password-based authentication
+    /// Password-based authentication. The SDK derives the wire-format `keyB64` via PBKDF2
+    /// over the send key.
     Password {
-        /// The password required to access the Send
+        /// The plaintext password the recipient will enter to access the Send.
         password: String,
+    },
+    /// Pre-derived password. The caller has already run PBKDF2 client-side and supplies the
+    /// resulting base64-encoded hash; the SDK forwards it verbatim. Use this when the
+    /// hashing happens outside the SDK (e.g. the legacy TypeScript clients that derive in
+    /// `SendService.encrypt`). For new code that holds a plaintext password, use
+    /// `Password { ... }` and let the SDK do the derivation.
+    HashedPassword {
+        /// Base64-encoded PBKDF2 output (`keyB64`) ready for the wire.
+        #[serde(rename = "keyB64")]
+        key_b64: String,
     },
     /// Email-based OTP authentication
     Emails {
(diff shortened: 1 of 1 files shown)


harr1424 (approved):

crates/bitwarden-send/src/send.rs
```diff
@@ -120,11 +120,22 @@ pub enum AuthType {
...
-        /// The password required to access the Send
+        /// The plaintext password the recipient will enter to access the Send.
         password: String,
     },
+    /// Pre-derived password. The caller has already run PBKDF2 client-side and supplies the
+    /// resulting base64-encoded hash; the SDK forwards it verbatim. Use this when the
+    /// hashing happens outside the SDK (e.g. the legacy TypeScript clients that derive in
+    /// `SendService.encrypt`). For new code that holds a plaintext password, use
+    /// `Password { ... }` and let the SDK do the derivation.
+    HashedPassword {
```
quexten, line 134: Just a question; Isn't it easier to move over clients to the Password API instead of introducing a new API that acts at a lower level?

Generally, we want to move from lower level to higher level APIs, not the other way around.

adudek-bw: Long-term, migrating clients to the higher-level Password API is the cleaner shape. The reason this PR went with HashedPassword instead is plumbing on the clients side: SendService.encrypt currently derives keyB64 and discards the plaintext before the Send reaches the SDK call boundary, so switching to plaintext-via-Password requires threading the plaintext back through every Send save call site. The clients workaround in bitwarden/clients#20170 (https://github.com/bitwarden/clients/pull/20170) routes Password saves to the legacy HTTP path entirely to avoid the SDK double-derivation, and shipping HashedPassword lets that workaround be removed without the broader client refactor.

quexten: Awesome. Non-blocking suggesiton for the future: When you add temporary transitions like this, it's helpful to link the ticket that resolves it and to clearly state it's temporary.

quexten (approved):

PR #1202, opened 2026-06-22
[PM-35105] Add hazmat module and shared COSE symmetric encryption

quexten:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-35105

## 📔 Objective

Stacked PR **2/4** (on top of #1201).

Introduces a `hazmat::symmetric_encryption` module (moving XChaCha20-Poly1305 into it and adding AES-256-GCM), and a shared `cose::symmetric` layer that seals/unseals COSE Encrypt/Encrypt0 bodies with either cipher, declaring the content-encryption algorithm in the protected header. Migrates the data, symmetric-key, and password-protected key envelopes to the shared encrypt/decrypt helpers.

The password envelope's COSE byte layout shifts, so the pin corruption test now flips a byte at the envelope midpoint.

Base: `km/high-entropy-secret` (#1201).


crates/bitwarden-crypto/src/safe/symmetric_key_envelope.rs
```diff
@@ -142,9 +150,14 @@ impl SymmetricKeyEnvelope {
                 SymmetricKeyEnvelopeError::Parsing("Invalid content format".to_string())
             })?;
 
-        // Decrypt the key bytes
-        let key_bytes = crate::cose::decrypt_cose(&self.cose_encrypt0, wrapping_key_inner)
-            .map_err(|_| SymmetricKeyEnvelopeError::WrongKey)?;
+        // Decrypt the key bytes. The envelope always declares the content-encryption algorithm, so
```
quexten: Remove comment

quexten: Remove comment

crates/bitwarden-crypto/src/safe/symmetric_key_envelope.rs
```diff
@@ -97,14 +100,19 @@ impl SymmetricKeyEnvelope {
             SafeObjectNamespace::SymmetricKeyEnvelope,
             namespace,
         );
-        protected_header.alg = Some(coset::Algorithm::PrivateUse(XCHACHA20_POLY1305));
         protected_header.key_id = wrapping_key.key_id.as_slice().into();
 
-        let cose_encrypt0 = crate::cose::encrypt_cose(
-            CoseEncrypt0Builder::new().protected(protected_header),
+        // The content-encryption algorithm (XChaCha20-Poly1305) is declared in the protected header
```
quexten: Remove comment

crates/bitwarden-crypto/src/hazmat/mod.rs
```diff
@@ -0,0 +1,6 @@
+//! # Hazardous materials
```
quexten: Replace
```
Hazmat module

Inside here live low level crypto implementations that are not safe to use by themselves and can be easily misused. This should never be used from another crate, and handled with extreme care.
```

crates/bitwarden-crypto/src/cose/mod.rs
```diff
@@ -168,16 +132,19 @@ pub(crate) fn encrypt_xchacha20_poly1305(
...
-            let ciphertext =
-                crate::xchacha20::encrypt_xchacha20_poly1305(&(*key.enc_key).into(), data, aad);
-            nonce = ciphertext.nonce();
-            ciphertext.encrypted_bytes().to_vec()
+            XChaCha20Poly1305::encrypt(&(*key.enc_key).into(), &nonce, data, aad)
+                .encrypted_bytes()
+                .to_vec()
         })
-        .unprotected(coset::HeaderBuilder::new().iv(nonce.to_vec()).build())
+        .unprotected(
```
quexten: Encrypt, decrypt xchacha should live in symmetric.rs. also the changes for string need to be applied to all symmetric algs.

crates/bitwarden-crypto/src/cose/symmetric.rs
```diff
@@ -0,0 +1,539 @@
+//! Content encryption of COSE `CoseEncrypt`/`CoseEncrypt0` messages by the hazmat symmetric
```
quexten: Don't ref hazmat ciphers here.

crates/bitwarden-crypto/src/cose/symmetric.rs
```diff
@@ -0,0 +1,539 @@
...
+//! ciphers.
+//!
+//! The [`CoseEncryptCipher`] trait adds `encrypt_cose`/`decrypt_cose` (multi-recipient
```
quexten: Simplify:
- This is the second layer of the three layer stack
  - Lowest level: Hazmat Primitive
  - Mid level: Cose
  - High level: Consumer (Safe, Encstring)

remove the other fluff.

crates/bitwarden-crypto/src/safe/password_protected_key_envelope.rs
```diff
@@ -116,39 +115,43 @@ impl PasswordProtectedKeyEnvelope {
...
-        cose_encrypt.unprotected.iv = nonce.into();
+        let builder = coset::CoseEncryptBuilder::new().add_recipient({
+            let mut recipient = coset::CoseRecipientBuilder::new()
+                .unprotected(kdf_settings.into())
+                .build();
+            recipient.protected.header.alg = Some(coset::Algorithm::PrivateUse(ALG_ARGON2ID13));
+            recipient
+        });
+
+        // The envelope key derived above is the content-encryption key. `encrypt_cose` declares the
```
quexten: Remove this comment

crates/bitwarden-crypto/Cargo.toml
```diff
@@ -32,6 +32,7 @@ test-utils = []
 
 [dependencies]
 aes = { version = "0.9.0", features = ["zeroize"] }
+aes-gcm = { version = "0.11.0-rc.4", features = ["zeroize"] }
```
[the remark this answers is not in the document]
quexten: In-review https://bitwarden.atlassian.net/browse/VULN-650

crates/bitwarden-crypto/src/cose/symmetric.rs
```diff
@@ -0,0 +1,800 @@
...
+    /// Authenticates and decrypts the ciphertext of `cose_encrypt0` under `cek`. Behaves like
+    /// [`decrypt_cose`](Self::decrypt_cose), but for a single-recipient message.
+    fn decrypt_cose0(cose_encrypt0: &CoseEncrypt0, cek: &Self::Key)
+    -> Result<Vec<u8>, CryptoError>;
+}
+
+impl CoseEncryptCipher for Aes256Gcm {
+    const COSE_ALGORITHM: Algorithm = Algorithm::Assigned(iana::Algorithm::A256GCM);
+
+    fn encrypt_cose(
```
eligrubb, line 298: 🤔 non-blocking nit: the `cose` and `cose0` variants of `encrypt_` and `decrypt_` are identical except for the `builder` and return types. Could there be a way to avoid repeating this boilerplate setup in multiple places? It also means that security-related comments like the one below on nonce reuse are missing in the `_cose0` case.

crates/bitwarden-crypto/src/safe/password_protected_key_envelope.rs
```diff
@@ -116,39 +115,39 @@ impl PasswordProtectedKeyEnvelope {
...
-            })
-            .build();
-        cose_encrypt.unprotected.iv = nonce.into();
+        let builder = coset::CoseEncryptBuilder::new().add_recipient({
+            let mut recipient = coset::CoseRecipientBuilder::new()
+                .unprotected(kdf_settings.into())
+                .build();
+            recipient.protected.header.alg = Some(coset::Algorithm::PrivateUse(ALG_ARGON2ID13));
+            recipient
+        });
```
eligrubb, line 141: ❓non-blocking question, something for us to maybe discuss in our next sync: Why does PasswordProtectedKeyEnvelope (and now SecretProtectedKeyEnvelope) require multiple recipients (aka use `CoseEncrypt` instead of `CoseEncrypt0`)? Additionally, I notice that we set the recipient's protected header algorithm to Argon2id but we don't do the same for the new SecretProtectedKeyEnvelope and HKDF.

eligrubb (dismissed):
Good stuff! Some non-blocking questions is all.

crates/bitwarden-crypto/src/safe/password_protected_key_envelope.rs
quexten: > [5.2](https://datatracker.ietf.org/doc/html/rfc8152#section-5.2).  Single Recipient Encrypted
> The COSE_Encrypt0 encrypted structure does not have the ability to
   specify recipients of the message.  The structure assumes that the
   recipient of the object will already know the identity of the key to
   be used in order to decrypt the message.  If a key needs to be
   identified to the recipient, the enveloped structure ought to be
   used.


This to me suggests that this is not the right type because we want a (single) recipient struct, that specifies the (single) kdf-based recipient.

> Additionally, I notice that we set the recipient's protected header algorithm to Argon2id but we don't do the same for the new SecretProtectedKeyEnvelope and HKDF.

I looked into the spec and reference implementations, and the specified HKDF recipient does not state whether the KDF setting needs to be protected or not. Reference implementations accept unprotected KDF settings.

I think this was missed on the password based kdf settings, because there is no standard for it. Given it's not specified either, I believe we are good to do either. Do you think it's worth it to put it into the protected header for consistency?

crates/bitwarden-crypto/src/cose/symmetric.rs
quexten: Good point. Given the return type is different this gets a bit tricky sadly. The only real difference seems to be the missing recipient array. We could make a macro, but that seems to overcomplicate things. Do you have a suggestion?

djsmith85 (approved):
Approving platform-owned changes

eligrubb (approved):

PR #1147, opened 2026-05-28
feat(core): introduce Client::gov_mode() with URL-host inference

addisonbeck:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-38266

## 📔 Objective

Expose Client::gov_mode() on bitwarden_core::Client so SDK consumers can detect Gov environments without rebuilding URL inference locally.

Later on PM-36520 we'll replaces URL inference with a some sort of /api/config gov_mode read. After that isGovMode on clients will be updated to call the SDK.

PR #1205, opened 2026-06-22
[PM-38851] crypto cipher suite

mzieniukbw:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-38851

## 📔 Objective

Crypto cipher suite api and client, to help choose the correct cipher suites and algorithms when in Gov mode (FedRAMP environment).
Currently only supports KDF.
Crypto crate SAFE primitives outside of scope for this PR.
Mocks the `gov_mode` implementation to always return false - will be fixed by #1147. Note to Platform: This PR can be merged independently, since all environments as of today are in non-gov mode, hence have zero impact with this PR.

### DIFF
--- a/crates/bitwarden-core/src/client/builder.rs
+++ b/crates/bitwarden-core/src/client/builder.rs
@@ -3,7 +3,7 @@ use std::sync::RwLock;
 use std::sync::{Arc, OnceLock};
 
 use bitwarden_api_base::new_http_client_builder;
-use bitwarden_crypto::KeyStore;
+use bitwarden_crypto::{CipherSuite, KeyStore};
 use bitwarden_state::registry::StateRegistry;
 use reqwest::header::{self, HeaderValue};
 
@@ -132,7 +132,7 @@ impl ClientBuilder {
             client: bw_http_client,
         };
 
-        Client {
+        let client = Client {
             internal: Arc::new(InternalClient {
                 user_id: OnceLock::new(),
                 token_handler: self.token_handler,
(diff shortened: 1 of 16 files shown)


crates/bitwarden-core/src/client/builder.rs
```diff
@@ -145,7 +145,16 @@ impl ClientBuilder {
...
             }),
-        }
+        };
+
+        // Configure the key store's cipher suite from the client's environment, so all crypto
+        // operations (e.g. the KDF for a new account) pick compliant algorithms.
+        client
+            .internal
+            .get_key_store()
+            .set_cipher_suite(CipherSuite::from_gov_mode(client.gov_mode()));
```
coroiu, line 155: Hmm this seems like a round-about way of setting this, since the `gov_mode` value comes from inside the client, and we are modifying something that also lives inside the client.

suggestion: move `set_cipher_suite()` to the client constructor instead, that way it'll be very obvious when and how that function is called

mzieniukbw: Cipher suite is crucial component of crypto crate. The only reason why we can't construct the KeyStore with the Cipher suite is because gov mode is part of Clients - even though it does not need to be.
Since we were going to guard cipher suite value to be KM feature flagged, which is currently located in clients, i am inclined to move it for now.
KM would want eventually, for the cryptography implementation not to be fragmented across clients and SDK, but lively solely in SDK - which would mean determining gov mode and feature flags, without having to build Client (part of building client).

coroiu: Ok, fair point, I like that design, but it's gonna require platform to rethink how we provide and where we determine `gov_mode()`

coroiu (approved):

crates/bitwarden-core/src/client/builder.rs
mzieniukbw: I have discussed this with KM team and we are not going add a feature flag for this at this time. Considering this, we want the `set_cipher_suite` to stay in SDK instead of being moved to client.

quexten (approved):

crates/bitwarden-core/src/client/builder.rs
coroiu: @mzieniukbw I just want to make sure we're not talking past each other on the word "client." You're contrasting "stay in the SDK" with "moved to clients," but when I said "client" in my original comment I meant the `bitwarden_core::Client` struct, not the clients repository.

So to be concrete: my suggestion was just to move the `set_cipher_suite()` call into Client's constructor (still entirely within the SDK), not to push anything out to the clients repo. Are we on the same page?

PR #1209, opened 2026-06-23
Remove ecdsa-keys compilation flag

neuronull:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-38832

## 📔 Objective

We built with this flag as the default in https://github.com/bitwarden/sdk-internal/pull/1181

, and now that the ssh agent in the clients repo supports the key type, there isn't a reason to have the compilation flag anymore, so this change is to remove it and cleanup.


quexten (approved):
Does not need my explicit approval, but from @bitwarden/wg-ssh-keys side looks great.

dereknance (approved):
I love the smell of deleted code in the morning.

PR #1215, opened 2026-06-24
fix(deps): raise hybrid-array and typenum floors

addisonbeck:
## 📔 Objective

The direct-maximum-versions CI check runs cargo update to test that all dependencies can resolve to their latest compatible versions. crypto-bigint 0.7.0-0.7.4 are all YANKED upstream, so cargo update resolves to 0.7.5. That version requires hybrid-array ^0.4.12 and typenum ^1.20, which bitwarden-crypto's existing floors excluded, producing a resolver conflict that blocked all PRs.

Raise the two floors to admit what crypto-bigint 0.7.5 and rsa 0.10.0-rc.18 require. The committed Cargo.lock intentionally holds crypto-bigint at 0.7.3 (a locked-yanked version that still builds) so that local and Aikido-restricted dev environments do not pull anything newer than 7 days old. Only the no-Aikido CI runner resolves to 0.7.5 via its own cargo update.

Verified locally: cargo check --all-features -D warnings passes. Both newly-pulled versions (hybrid-array 0.4.12 @ 2026-05-09, typenum 1.20.1 @ 2026-05-29) are >7 days old.

### DIFF
--- a/crates/bitwarden-crypto/Cargo.toml
+++ b/crates/bitwarden-crypto/Cargo.toml
@@ -49,7 +49,7 @@ ed25519-dalek = { workspace = true, features = ["rand_core", "zeroize"] }
 hex = { version = ">=0.4.0, <0.5" }
 hkdf = "0.13.0"
 hmac = { workspace = true, features = ["zeroize"] }
-hybrid-array = { version = "0.4.10", features = ["alloc", "serde", "zeroize"] }
+hybrid-array = { version = ">=0.4.12, <0.5.0", features = ["alloc", "serde", "zeroize"] }
 js-sys = { workspace = true, optional = true }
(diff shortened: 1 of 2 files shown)


crates/bitwarden-crypto/Cargo.toml
```diff
@@ -69,7 +69,7 @@ thiserror = { workspace = true }
 tokio = { workspace = true, features = ["io-util"] }
 tracing = { workspace = true }
 tsify = { workspace = true, optional = true }
-typenum = ">=1.18.0, <1.20.0"
+typenum = ">=1.20.0, <1.21.0"
```
quexten, line 72: @addisonbeck KM already addressed this in https://github.com/bitwarden/sdk-internal/pull/1212

PR #1218, opened 2026-06-26
[PM-39258] Switch bitwarden-ipc to use P-256 curve by default

mzieniukbw:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-39258

## 📔 Objective

Switch `bitwarden-ipc` crate to use FIPS compliant cipher suite by default.
Adds new (default) `Noise_NN_P256_AESGCM_SHA256` cipher suite (handshake using P-256 curve, transport using AES-GCM cipher and SHA-256 for hash).

❓ Do we want to remove other cipher suites (or at least leave one for non-gov environment, likely ChaChaPoly) ?

## 🚨 Breaking Changes

None expected.
Note: While older client won't be able to communicate with new client with this change though ipc, the only use of this implementation is shared unlock, which is not enabled in prod. Older clients will eventually catch up (upgrade) to new version that will have this SDK, hence become compatible again.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -688,7 +688,7 @@ dependencies = [
  "criterion",
  "ed25519-dalek",
  "hex",
- "hkdf",
+ "hkdf 0.13.0",
  "hmac 0.13.0",
  "hybrid-array",
  "js-sys",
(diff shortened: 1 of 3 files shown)


quexten (approved):
I think we can remove the others here or do that later. note: by introducing it and default enabling it at the same time we are introducing a breaking change at runtime. That's since it's not released yet, for future changes after release we should add but default later, which is non breaking.

djsmith85 (approved):

PR #1219, opened 2026-06-27
[PM-35785] [PM-35787] Add strongly typed data and aggregation for policies

eliykat:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-35785
https://bitwarden.atlassian.net/browse/PM-35787

## 📔 Objective

#964 and associated PRs added basic policy enforcement to the SDK: given a list of policies, it would tell you which should be enforced against the user.

This PR adds additional functionality to the domain:
- strongly typed data handling, so that consumers can access the `policyView.data` JSON blob without deserializing it themselves
- policy aggregation, so that consumers can understand how multiple policies must be combined and enforced against the user. (e.g. if a user is part of multiple organizations that all have the master password policy enabled, their MP must satisfy all policies simultaneously).

Overview of changes:
- new mod called `enforcement` to house all enforcement logic
- `filter.rs` has been largely moved to `enforcement::Policy` because it defines the `Policy` trait
- `data.rs` adds trait and logic for strongly typed data handling
- `aggregate.rs` adds trait and logic for policy aggregation

TODOs:
- this PR turned out to be quite large (~1k LOC) but I wanted to illustrate how everything interacted. If desired I can break it up per the separate Jira tickets (`data.rs` in one PR, `aggregate.rs` in the next)
- should `PolicyFilter.enforced` panic, or return results?
- the big one: how to return the `Data` types across the FFI boundary? I think we need a `PolicyType` enum that is used to wrap data; but we already have one that is a plain discriminator for raw policy structs, and I'm not sure how that interacts. Do we just have to have 2 separate enum definitions: one without strongly typed data and one with?


crates/bitwarden-policies/src/enforcement/aggregate.rs
```diff
@@ -0,0 +1,250 @@
...
+///
+/// Implementing this trait unlocks
+/// [`PolicyAggregateFilter::get_enforced_aggregate_policy`] for the policy.
+/// [`NoData`] policies get a trivial implementation for free.
+pub trait PolicyAggregate: PolicyData {
+    /// Combines multiple [`Data`](PolicyData::Data) values into a single value.
+    /// How those values are combined is determined by the policy. Will only be
+    /// called with a non-empty `items`.
+    fn aggregate(&self, items: Vec<Self::Data>) -> Self::Data;
+}
```
eliykat, line 45: Should this be implemented by the `Policy` definition, or by the `Self::Data` type? It relates more to the Data type than the policy generally.

Alternative example:

```rust
pub trait FoldableData {
    fn fold_closure(result: Self, next: Self) -> Self;
}

/// if you want aggregate functionality, you implement this _instead of_ `PolicyData` (rather than in addition to)
pub trait AggregatePolicy {
    type Data: Default + DeserializeOwned + FoldableData;
}
```

And then `fold_closure` can be called in the iterator chain which is nice:

```rust
            filtered
                .iter()
                .map(|p| self.get_data_or_default(p))
                .fold(Self::Data::default(), |result, d| result.fold_closure(d))
```

I'm not sure if this is better or just different. It may be less intuitive for the policy author to implement.

crates/bitwarden-policies/src/enforcement/policy.rs
```diff
@@ -0,0 +1,319 @@
...
+    /// silently bypassed.
+    ///
+    /// # Inputs
+    ///
+    /// `view.r#type` must equal [`self.policy_type()`](Policy::policy_type),
+    /// and when `context` is `Some`, `context.id` must equal
+    /// `view.organization_id`. Both invariants should be guaranteed by
+    /// upstream filtering and lookup.
+    ///
+    /// # Panics
```
quexten, line 76: On panics:
- If it is a public API, avoid panics.
- If it is a non-public API, you MAY panic, if you control all callsites
  - ⚠️ The farther the invariant enforcing code is from the code that relies on the invariant and panics, the greater the risk. Usually you don't want to have more than a handful of callsites or LoC between.

Currently it seems rather easy to get this to panic, so I'd re-design this so that it cannot happen.

eliykat:
Closed in favour of https://github.com/bitwarden/sdk-internal/pull/1364.

PR #1221, opened 2026-06-29
[PM-39294] Gate shared unlock handshakes on transport reachability

coroiu:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-39294

Companion clients PR: bitwarden/clients#21551

## 📔 Objective

Shared-unlock followers repeatedly attempt Noise crypto handshakes against a leader (e.g. the desktop app) that often isn't running. Each attempt hits the 2s handshake timeout and surfaces as error spam plus reconnect churn. This introduces a lightweight **reachability** layer so a follower only talks to a leader it knows is present.

Design (an alternative to #1213):

- **Transport-authoritative, ping fallback.** New `CommunicationBackend::reachability(endpoint) -> Reachable | Unreachable | Unknown` (default `Unknown`). A definite transport answer wins; only `Unknown` falls back to a plaintext ping/pong liveness window. So browser↔desktop (native port state is known) never pings, while web↔extension (`postMessage`, no synchronous signal) uses ping/pong.
- **Owned, RAII-scoped tracking.** Consumers call `ReachabilityTracker::track(endpoint)` *after* discovering the peer and hold a `ReachabilityHandle`; dropping the last handle stops that endpoint's ping loop. One loop per endpoint via a weak registry, with deterministic cleanup on `Drop`.
- **Consumer-side gating.** `IpcClient::send` stays a dumb pipe; the shared-unlock `Follower` checks `is_reachable()` before `start_sessions`/heartbeat.
- **Crypto stays unaware.** A `ControlSplitter` backend decorator peels reserved control-topic (`$bw.control.*`) frames off before any crypto receiver sees them and routes them to the tracker, so `NoiseCryptoProvider` is unchanged — no handshake-abort guard needed.

## 🚨 Breaking Changes

`CommunicationBackend` gains a `reachability()` method, but it has a default (`Unknown`) so existing backends are unaffected.

## ⏰ Reminders

- Validated: `bitwarden-ipc` (40 tests), `bitwarden-shared-unlock` (14 tests), and wasm integration tests (reachability round-trip, transport-authoritative, shared-unlock, biometrics).
- The `crypto-bigint`/`rsa` commit is a build-resolution fix (yanked crates.io 0.7.x); it can be reviewed/reverted independently.

### DIFF
--- a/crates/bitwarden-ipc/Cargo.toml
+++ b/crates/bitwarden-ipc/Cargo.toml
@@ -43,6 +43,7 @@ tsify = { workspace = true, optional = true }
 uuid = { workspace = true }
 wasm-bindgen = { workspace = true, optional = true }
 wasm-bindgen-futures = { workspace = true, optional = true }
+web-time = "1.1.0"
 zeroize = { workspace = true, features = ["zeroize_derive"] }
 
 [lints]
--- a/crates/bitwarden-ipc/Cargo.toml
+++ b/crates/bitwarden-ipc/Cargo.toml
@@ -43,7 +43,6 @@ tsify = { workspace = true, optional = true }
 uuid = { workspace = true }
 wasm-bindgen = { workspace = true, optional = true }
 wasm-bindgen-futures = { workspace = true, optional = true }
-web-time = "1.1.0"
 zeroize = { workspace = true, features = ["zeroize_derive"] }
 
 [lints]
--- a/crates/bitwarden-ipc/src/control_splitter.rs
+++ b/crates/bitwarden-ipc/src/control_splitter.rs
@@ -26,7 +26,11 @@ use crate::{
 /// Wraps a communication backend so reachability control frames bypass crypto. The receivers handed
 /// to the crypto layer silently drop control frames; a single dedicated handler task routes them to
 /// the tracker (recording liveness and answering pings).
-pub(crate) struct ControlSplitter<Com> {
+// Public (but hidden) only because it appears in `IpcClientImpl`'s public trait bounds: the client
+// hands this wrapper to the crypto provider so control frames never reach it. It is not part of the
+// supported API and should not be used directly.
+#[doc(hidden)]
+pub struct ControlSplitter<Com> {
     backend: Arc<Com>,
     tracker: Arc<ReachabilityTracker>,
 }
@@ -88,7 +92,8 @@ impl<Com: CommunicationBackend> CommunicationBackend for ControlSplitter<Com> {
 }
 
 /// Receiver wrapper that drops control-topic frames so the crypto layer never sees them.
-pub(crate) struct ControlSplitterReceiver<R> {
+#[doc(hidden)]
+pub struct ControlSplitterReceiver<R> {
     inner: R,
 }
 
--- a/crates/bitwarden-ipc/src/control_splitter.rs
+++ b/crates/bitwarden-ipc/src/control_splitter.rs
@@ -11,11 +11,15 @@
 
 use std::sync::Arc;
 
+use bitwarden_threading::cancellation_token::CancellationToken;
+use tokio::select;
+
 use crate::{
+    control::is_control_topic,
     endpoint::Endpoint,
     error::IpcErrorKind,
     message::{IncomingMessage, OutgoingMessage},
-    reachability::{ReachabilityTracker, is_control_topic},
+    reachability::ReachabilityTracker,
     traits::{CommunicationBackend, CommunicationBackendReceiver, Reachability},
 };
 
@@ -34,21 +38,25 @@ impl<Com: CommunicationBackend> ControlSplitter<Com> {
 
     /// Spawn the single task that consumes control frames from the raw transport and drives the
     /// tracker. Spawned once when the client starts, so there is exactly one auto-pong responder
-    /// regardless of how many crypto receivers exist.
-    pub(crate) fn spawn_control_handler(&self) {
+    /// regardless of how many crypto receivers exist. The task stops when `cancellation_token` is
+    /// cancelled (i.e. when the client stops), so it does not leak across a stop/restart.
+    pub(crate) fn spawn_control_handler(&self, cancellation_token: CancellationToken) {
         let backend = self.backend.clone();
         let tracker = self.tracker.clone();
         let future = async move {
             let receiver = backend.subscribe().await;
             loop {
-                match receiver.receive().await {
-                    Ok(message) if is_control_topic(message.topic.as_deref()) => {
-                        tracker.handle_inbound(message).await;
-                    }
-                    // Data frames are delivered to the crypto layer's own receivers; ignore them.
-                    Ok(_) => {}
-                    Err(error) if error.is_fatal() => break,
-                    Err(_) => {}
+                select! {
+                    _ = cancellation_token.cancelled() => break,
+                    received = receiver.receive() => match received {
+                        Ok(message) if is_control_topic(message.topic.as_deref()) => {
+                            tracker.handle_inbound(message).await;
+                        }
+                        // Data frames are delivered to the crypto layer's own receivers; ignore them.
+                        Ok(_) => {}
+                        Err(error) if error.is_fatal() => break,
+                        Err(_) => {}
+                    },
                 }
             }
         };
--- a/crates/bitwarden-ipc/src/control_splitter.rs
+++ b/crates/bitwarden-ipc/src/control_splitter.rs
@@ -3,72 +3,43 @@
 //!
 //! Reachability ping/pong frames travel over the raw transport (so they can measure liveness
 //! without a crypto session), but the crypto layer reads the same transport and would try to decode
-//! them as Noise frames, aborting in-flight handshakes. This wrapper peels reserved control-topic
-//! frames off before any crypto-facing receiver yields them, and routes them to the
-//! [`ReachabilityTracker`] from a single dedicated handler task. `send` and `reachability` pass
-//! straight through to the inner backend, so the crypto provider stays entirely unaware of
+//! them as Noise frames, aborting in-flight handshakes. This wrapper splits the inbound stream in
+//! two: the receivers handed to the crypto layer ([`CommunicationBackend::subscribe`]) drop control
+//! frames, while [`ControlSplitter::subscribe_control`] yields only the control messages. `send`
+//! and `reachability` pass straight through, so the crypto provider stays entirely unaware of
 //! reachability.
 
 use std::sync::Arc;
 
-use bitwarden_threading::cancellation_token::CancellationToken;
-use tokio::select;
-
 use crate::{
-    control::is_control_topic,
+    control::{ControlMessage, IncomingControlMessage, is_control_topic},
     endpoint::Endpoint,
-    error::IpcErrorKind,
     message::{IncomingMessage, OutgoingMessage},
-    reachability::ReachabilityTracker,
     traits::{CommunicationBackend, CommunicationBackendReceiver, Reachability},
 };
 
-/// Wraps a communication backend so reachability control frames bypass crypto. The receivers handed
-/// to the crypto layer silently drop control frames; a single dedicated handler task routes them to
-/// the tracker (recording liveness and answering pings).
+/// Wraps a communication backend so reachability control frames are separated from data frames. The
+/// crypto layer only ever sees data; the reachability tracker consumes the control stream.
 // Public (but hidden) only because it appears in `IpcClientImpl`'s public trait bounds: the client
 // hands this wrapper to the crypto provider so control frames never reach it. It is not part of the
 // supported API and should not be used directly.
 #[doc(hidden)]
 pub struct ControlSplitter<Com> {
     backend: Arc<Com>,
-    tracker: Arc<ReachabilityTracker>,
 }
 
 impl<Com: CommunicationBackend> ControlSplitter<Com> {
-    pub(crate) fn new(backend: Arc<Com>, tracker: Arc<ReachabilityTracker>) -> Self {
-        Self { backend, tracker }
+    pub(crate) fn new(backend: Arc<Com>) -> Self {
+        Self { backend }
     }
 
-    /// Spawn the single task that consumes control frames from the raw transport and drives the
-    /// tracker. Spawned once when the client starts, so there is exactly one auto-pong responder
-    /// regardless of how many crypto receivers exist. The task stops when `cancellation_token` is
-    /// cancelled (i.e. when the client stops), so it does not leak across a stop/restart.
-    pub(crate) fn spawn_control_handler(&self, cancellation_token: CancellationToken) {
-        let backend = self.backend.clone();
-        let tracker = self.tracker.clone();
-        let future = async move {
-            let receiver = backend.subscribe().await;
-            loop {
-                select! {
-                    _ = cancellation_token.cancelled() => break,
-                    received = receiver.receive() => match received {
-                        Ok(message) if is_control_topic(message.topic.as_deref()) => {
-                            tracker.handle_inbound(message).await;
-                        }
-                        // Data frames are delivered to the crypto layer's own receivers; ignore them.
-                        Ok(_) => {}
-                        Err(error) if error.is_fatal() => break,
-                        Err(_) => {}
-                    },
-                }
-            }
-        };
-
-        #[cfg(not(target_arch = "wasm32"))]
-        tokio::spawn(future);
-        #[cfg(target_arch = "wasm32")]
-        wasm_bindgen_futures::spawn_local(future);
+    /// Subscribe to the inbound control-plane stream. Mirrors [`CommunicationBackendReceiver`]:
+    /// call [`ControlReceiver::receive`] for the next control message. Data frames are filtered
+    /// out (they reach the crypto layer via [`CommunicationBackend::subscribe`]).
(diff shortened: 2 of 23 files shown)


crates/bitwarden-ipc/src/traits/communication_backend.rs
```diff
@@ -1,10 +1,31 @@
...
 
+/// A transport-native answer to "is this endpoint reachable right now?".
+///
+/// The transport answers authoritatively when it can (e.g. a native-messaging port is either
+/// connected or not). When it has no synchronous way to tell (e.g. a `postMessage` relay to a
+/// content script that may or may not exist), it returns [`Reachability::Unknown`] and reachability
+/// falls back to the ping/pong liveness window tracked by the `ReachabilityTracker`.
+#[derive(Serialize, Deserialize, Debug, Clone, Copy, PartialEq, Eq)]
+#[cfg_attr(feature = "wasm", derive(tsify::Tsify))]
+#[cfg_attr(feature = "wasm", tsify(into_wasm_abi, from_wasm_abi))]
```
coroiu: suggestion: these can and should be merged

crates/bitwarden-ipc/src/traits/communication_backend.rs
```diff
@@ -27,6 +48,18 @@ pub trait CommunicationBackend: Send + Sync + 'static {
         message: OutgoingMessage,
     ) -> impl std::future::Future<Output = Result<(), Self::SendError>> + Send + Sync;
 
+    /// Query the transport-native reachability of `endpoint`.
+    ///
+    /// The default implementation returns [`Reachability::Unknown`], which makes reachability fall
+    /// back entirely to ping/pong liveness. Backends that can answer authoritatively (e.g. by
+    /// inspecting a connected port) should override this.
+    fn reachability(
```
coroiu, line 65: suggestion: wouldn't `reachable` be a better name?

crates/bitwarden-ipc/src/wasm/ipc_client.rs
```diff
@@ -126,4 +128,52 @@ impl JsIpcClient {
...
+        JsReachabilityTracker {
+            tracker: self.client.reachability(),
+        }
+    }
+}
+
+/// JavaScript wrapper around the reachability tracker. See
+/// [`ReachabilityTracker`](crate::ReachabilityTracker).
+#[wasm_bindgen(js_name = ReachabilityTracker)]
+pub struct JsReachabilityTracker {
```
coroiu: issue: please put this in a separate file

crates/bitwarden-ipc/src/control_splitter.rs
```diff
@@ -0,0 +1,99 @@
...
+                    Ok(message) if is_control_topic(message.topic.as_deref()) => {
+                        tracker.handle_inbound(message).await;
+                    }
+                    // Data frames are delivered to the crypto layer's own receivers; ignore them.
+                    Ok(_) => {}
+                    Err(error) if error.is_fatal() => break,
+                    Err(_) => {}
+                }
+            }
+        };
```
coroiu: suggestion: Hmm, I was thinking about this differently. Could we instead split messages into two different streams and subscribe to them separately in `IpcClient`? It feels weird to put `ReachabilityTracker` in the control splitter. Maybe the `ControlSplitter` could expose separate functions `receive_control` and `send_control`?

crates/bitwarden-ipc/src/reachability.rs
```diff
@@ -0,0 +1,447 @@
...
+use web_time::Instant;
+
+use crate::{
+    endpoint::Endpoint,
+    message::{IncomingMessage, OutgoingMessage},
+    traits::{CommunicationBackend, Reachability},
+};
+
+/// Reserved topic namespace for transport control frames that bypass the crypto channel.
+pub(crate) const CONTROL_TOPIC_PREFIX: &str = "$bw.control.";
```
coroiu: suggestion: it doesn't feel like this (or `is_control_topic`) belongs here? Could we put in a separate `control` module?

crates/bitwarden-ipc/src/reachability.rs
```diff
@@ -0,0 +1,447 @@
...
+                .last_seen
+                .lock()
+                .expect("reachability last_seen mutex poisoned"),
+            ACTIVE_WINDOW,
+        )
+    }
+}
+
+/// Tracks reachability of endpoints on behalf of consumers. Owned by the `IpcClient`; reachability
+/// pings and the auto-pong responder run over the transport this tracker was built from.
```
coroiu: issue: this doesn't seem like a very good public documentation

crates/bitwarden-ipc/src/reachability.rs
```diff
@@ -0,0 +1,447 @@
...
+
+    /// Begin (or join) tracking `endpoint`, returning a handle. Calls for the same endpoint share a
+    /// single ping loop; the loop is spawned only when the first handle for an endpoint is created.
+    pub fn track(&self, endpoint: Endpoint) -> ReachabilityHandle {
+        let mut map = self
+            .registry
+            .lock()
+            .expect("reachability registry poisoned");
+        // Opportunistically drop dangling entries left by handles that have since been freed.
+        map.retain(|_, weak| weak.strong_count() > 0);
```
coroiu: suggestion: can we create a utility function `clean` (or a better name if you have one) so we don't duplicate this?

crates/bitwarden-ipc/src/reachability/tracker.rs
```diff
@@ -0,0 +1,447 @@
...
+            registry: Arc::downgrade(&self.registry),
+        });
+        map.insert(endpoint.clone(), Arc::downgrade(&inner));
+
+        spawn_ping_loop(
+            Arc::downgrade(&inner),
+            self.send_fn.clone(),
+            self.reach_fn.clone(),
+            endpoint,
+        );
```
coroiu: question/suggestion: wouldn't it be better if the loop logic lived in the tracker?

crates/bitwarden-ipc/src/reachability.rs
```diff
@@ -0,0 +1,447 @@
...
+            .registry
+            .lock()
+            .expect("reachability registry poisoned")
+            .get(&source)
+            .and_then(Weak::upgrade)
+        {
+            *inner
+                .last_seen
+                .lock()
+                .expect("reachability last_seen mutex poisoned") = Some(Instant::now());
```
coroiu: question: Is instant supported on wasm?

crates/bitwarden-ipc/src/reachability/tracker.rs
```diff
@@ -0,0 +1,447 @@
...
+                        INACTIVE_PING_INTERVAL
+                    }
+                }
+                // The transport answered authoritatively; no ping needed. Keep polling at the
+                // backoff cadence so a later transition to `Unknown` resumes pinging.
+                Reachability::Reachable | Reachability::Unreachable => INACTIVE_PING_INTERVAL,
+            };
+
+            drop(inner);
+            bitwarden_threading::time::sleep(interval).await;
```
coroiu: suggestion: I think I mentioned this above it feels like if we moved this into the handle we can create separate loops for each endpoint and get better sleep precision.

crates/bitwarden-ipc/src/reachability.rs
```diff
@@ -0,0 +1,447 @@
...
+
+        if message.topic.as_deref() == Some(CONTROL_PING_TOPIC) {
+            (self.send_fn)(OutgoingMessage {
+                payload: Vec::new(),
+                destination: source,
+                topic: Some(CONTROL_PONG_TOPIC.to_owned()),
+            })
+            .await;
+        }
+    }
```
coroiu: question/suggestion: what if rewrote this as a loop that subscribes to typed control messages instead of manually having to filter? We could create a `ControlMessage` enum-style message instead of sending raw messages like this, and then subscribe to those? I guess we'd still have to filter the enum, but still, thoughts?

coroiu (commented):
Self-review for 🤖

crates/bitwarden-ipc/src/traits/communication_backend.rs
coroiu: 🤖 Done — merged into a single `#[cfg_attr(feature = "wasm", derive(tsify::Tsify), tsify(into_wasm_abi, from_wasm_abi))]` (commit `79b25929`).

crates/bitwarden-ipc/src/traits/communication_backend.rs
coroiu: 🤖 I went with `reachability` because the method returns the tri-state `Reachability` enum rather than a bool — `reachable()` reads as a boolean predicate. Happy to rename (e.g. `query_reachability`) if you'd still prefer it; let me know.

crates/bitwarden-ipc/src/wasm/ipc_client.rs
coroiu: 🤖 Done — moved `JsReachabilityTracker`/`JsReachabilityHandle` into their own `src/wasm/reachability.rs` (commit `79b25929`).

crates/bitwarden-ipc/src/control_splitter.rs
coroiu: 🤖 Agreed — coupling the tracker into the splitter feels wrong. Proposed refactor (folding in your typed-message and loop-location comments below):

- `control` module gains a `ControlMessage { Ping, Pong }` enum with frame conversion helpers.
- `ControlSplitter` becomes a generic data/control demux: it exposes `subscribe_control() -> Receiver<(ControlMessage, Source)>` and `send_control(...)`, and no longer knows anything about reachability.
- `IpcClient` owns the tracker and, in `start()`, spawns a small loop feeding control messages from `subscribe_control()` into the tracker.
- The ping loop moves into the tracker.

Want me to take this on as a follow-up commit?

crates/bitwarden-ipc/src/reachability.rs
coroiu: 🤖 Done — moved the control topics and `is_control_topic` into a new `control` module (`src/control.rs`) in `79b25929`. That's also the natural home for the `ControlMessage` enum proposed on the `ControlSplitter` thread.

crates/bitwarden-ipc/src/reachability.rs
coroiu: 🤖 Done — rewrote it from a consumer's perspective (how to obtain it, `track`/`is_reachable`, and the resolution model) in `79b25929`.

crates/bitwarden-ipc/src/reachability.rs
coroiu: 🤖 Done — extracted `fn prune(registry)`, now used by both `track()` and `Drop` (`79b25929`).

crates/bitwarden-ipc/src/reachability/tracker.rs
coroiu: 🤖 Agreed — I'll move the loop into the tracker as part of the control-plane refactor (see the `ControlSplitter` thread). It's already one loop per endpoint, so this just improves cohesion.

crates/bitwarden-ipc/src/reachability.rs
coroiu: 🤖 Yes — this is `web_time::Instant`, not `std::time::Instant` (`std`'s panics on `wasm32-unknown-unknown`). `web-time` backs it with `performance.now()` on wasm and is a drop-in everywhere else.

crates/bitwarden-ipc/src/reachability/tracker.rs
coroiu: 🤖 Clarification: the loops are already per-endpoint — `track()` spawns one loop per tracked endpoint, each with its own sleep — so there's no shared-loop precision loss today. Moving the loop into the tracker (per your other comment) keeps that and just improves cohesion; I'll fold it into the refactor.

crates/bitwarden-ipc/src/reachability.rs
coroiu: 🤖 Like it. I'd add a `ControlMessage { Ping, Pong }` enum in the `control` module with frame-conversion helpers, and have `ControlSplitter` expose a typed control stream so the tracker subscribes to `ControlMessage`s instead of filtering raw frames. Same refactor I proposed on the `ControlSplitter` thread — I'll roll them together.

crates/bitwarden-ipc/src/control_splitter.rs
```diff
@@ -0,0 +1,99 @@
...
+
+        #[cfg(not(target_arch = "wasm32"))]
+        tokio::spawn(future);
+        #[cfg(target_arch = "wasm32")]
+        wasm_bindgen_futures::spawn_local(future);
+    }
+}
```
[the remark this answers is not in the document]
coroiu: 🤖 Fixed in `ffdd72b3` — the control handler now takes the client's `CancellationToken` and `select!`s on it, so it stops on shutdown instead of looping forever (and a restart no longer spawns a duplicate responder).

crates/bitwarden-ipc/Cargo.toml
```diff
@@ -43,6 +43,7 @@ tsify = { workspace = true, optional = true }
 uuid = { workspace = true }
 wasm-bindgen = { workspace = true, optional = true }
 wasm-bindgen-futures = { workspace = true, optional = true }
+web-time = "1.1.0"
```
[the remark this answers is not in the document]
coroiu: 🤖 Noted — `web-time` is the wasm-bindgen-ecosystem `Instant`/`SystemTime` shim, used here only to get a monotonic clock that doesn't panic on `wasm32-unknown-unknown`. Flagging for AppSec dependency approval per process.

crates/bitwarden-ipc/src/traits/communication_backend.rs
coroiu: Fair, good point. Let's keep it as-is

crates/bitwarden-ipc/src/control_splitter.rs
coroiu: Yes, go for it

crates/bitwarden-ipc/src/reachability.rs
coroiu: How about we just drop web_time completely and instead rely on thread sleeps? That would avoid having to add a whole new dependency. Alternatively we add our custom implementation using `js_sys` in `bitwarden_threading`, but I'd prefer to avoid that

crates/bitwarden-ipc/src/reachability/tracker.rs
coroiu: Ok, sounds good. I'll resolve this

crates/bitwarden-ipc/src/reachability.rs
coroiu: 🤖 Done (commit `8e2c6aa8`) — dropped `web-time`. Liveness is now clock-free: a reply from the endpoint sets `live` immediately, and the per-endpoint ping loop clears it after a cycle with no reply — so no clock/dependency is needed and it behaves identically on native and wasm.

Note: `web-time` is still pulled in transitively by `bitwarden-shared-unlock`'s leader, which used it before this PR. Out of scope here, but happy to tackle that separately if you want it gone from the workspace entirely.

crates/bitwarden-ipc/src/control_splitter.rs
coroiu: 🤖 Starting on it next. This latest push was the CI/lint fixes (the `private_interfaces` + rustdoc `-D warnings` errors) plus the web-time drop. I'll land the ControlSplitter/ControlMessage restructure — a typed control stream via `subscribe_control`, `IpcClient` wiring the tracker, and the loop moved into the tracker — as a separate follow-up commit so it stays reviewable on its own. Will ping here when it's up.

coroiu: 🤖 Done (commit `e7870e43`). The `ControlSplitter` is now a pure data/control demux — it no longer holds the tracker or spawns anything. It exposes `subscribe_control()` returning a `ControlReceiver` whose `receive()` yields a typed `ControlMessage` (mirroring `CommunicationBackendReceiver`), while the crypto-facing `subscribe()` still drops control frames. `IpcClient` owns the tracker and wires the two together via `tracker.start(control_receiver, cancellation_token)`. The tracker module is also split into `reachability/{mod,handle,tracker}.rs`.

crates/bitwarden-ipc/src/reachability/tracker.rs
coroiu: 🤖 Done in `e7870e43` — the inbound loop now lives in `ReachabilityTracker::start()` (consuming the control receiver and recording liveness / answering pings), so `handle_inbound` is private. The per-endpoint ping loops are unchanged (still one per tracked endpoint).

crates/bitwarden-ipc/src/reachability.rs
coroiu: 🤖 Done in `e7870e43` — added a `ControlMessage { Ping, Pong }` enum in the `control` module (with topic/frame conversion helpers). The control receiver yields these typed messages, so the tracker matches on the enum instead of filtering raw topic strings.

crates/bitwarden-ipc/src/reachability.rs
coroiu: That's ok, we'll leave it like this for now

PR #1223, opened 2026-06-30
[PM-39259] Implement bitwarden-random crate

quexten:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-39259
https://github.com/bitwarden/clients/pull/21576

## 📔 Objective

Adds the `bitwarden-random` crate and migrates callers to it. Currently, various crates directly use ThreadRng. ThreadRng does not use FIPS approved algorithms, and we want to use SysRng / OsRng instead. To make future changes easier, and to ease testing, instead of just migrating the callsites, we instead implement a new crate - `bitwarden-random` - that provides a struct that implements all necessary random traits.

By default it just maps through to SysRng. An optional feature for seeded testing is provided, that does not require breaking the public API of cryptographic functions by seeding a thread-local RNG. This is behind a compile time flag and should never be enabled in production.

Further, we add an FFI exposed client for making byte arrays and UUIDs for both wasm and swift. This is to consolidate the random implementations into just one.

### DIFF
--- a/crates/bitwarden-random/src/ffi.rs
+++ b/crates/bitwarden-random/src/ffi.rs
@@ -1,0 +1,163 @@
+//! Cross-platform (WASM / UniFFI) bindings for random-number generation.
+
+use bitwarden_error::bitwarden_error;
+use rand::{Rng, RngExt};
+use thiserror::Error;
+#[cfg(feature = "wasm")]
+use wasm_bindgen::prelude::*;
+
+use crate::rng;
+
+/// Error returned by [`SdkRandomNumberClient::gen_bytes`].
+#[bitwarden_error(flat)]
+#[derive(Debug, Error)]
+pub enum GenBytesError {
+    /// More than 1 KiB of randomness was requested.
+    #[error("gen_bytes() is limited to 1KiB; requested {requested} bytes")]
+    TooManyBytes {
+        /// The number of bytes that was requested.
+        requested: u32,
+    },
+}
+
+/// Error returned by [`SdkRandomNumberClient::gen_range`].
+#[bitwarden_error(flat)]
+#[derive(Debug, Error)]
+pub enum GenRangeError {
+    /// `min` was greater than `max`, so the range is empty.
+    #[error("Invalid range: min ({min}) must not exceed max ({max})")]
+    InvalidRange {
+        /// The requested lower bound.
+        min: u32,
+        /// The requested upper bound.
+        max: u32,
+    },
+}
+
+/// Client exposing random-number generation to cross-platform bindings.
+#[derive(Default)]
+#[cfg_attr(feature = "uniffi", derive(uniffi::Object))]
+#[cfg_attr(feature = "wasm", wasm_bindgen)]
+pub struct SdkRandomNumberClient;
+
+#[cfg_attr(feature = "wasm", wasm_bindgen)]
+#[cfg_attr(feature = "uniffi", uniffi::export)]
+impl SdkRandomNumberClient {
+    /// Construct a new client.
+    #[cfg_attr(feature = "wasm", wasm_bindgen(constructor))]
+    #[cfg_attr(feature = "uniffi", uniffi::constructor)]
+    pub fn new() -> Self {
+        // UniFFI does not support associated (static) functions, so the client is exposed as an
+        // object with a constructor and `&self` methods. The WASM binding mirrors that
+        // shape (constructor + instance methods) so a single impl block serves both.
+        SdkRandomNumberClient
+    }
+
+    /// Generate `len` cryptographically-secure random bytes.
+    ///
+    /// Returns [`GenBytesError::TooManyBytes`] if `len` exceeds 1 KiB. If you want over 1KiB of
+    /// randomness, please reconsider your design.
+    pub fn gen_bytes(&self, len: u32) -> Result<Vec<u8>, GenBytesError> {
+        if len > 1024 {
+            return Err(GenBytesError::TooManyBytes { requested: len });
+        }
+
+        let mut buf = vec![0u8; len as usize];
+        rng().fill_bytes(&mut buf);
+        Ok(buf)
+    }
+
+    /// Generate a random v4 UUID, sampled from a CRNG
+    ///
+    /// This has 122 random bits of entropy.
+    pub fn gen_uuid(&self) -> String {
+        let mut bytes = [0u8; 16];
+        rng().fill_bytes(&mut bytes);
+        uuid::Builder::from_random_bytes(bytes)
+            .with_variant(uuid::Variant::RFC4122)
+            .with_version(uuid::Version::Random)
+            .into_uuid()
+            .to_string()
+    }
+
+    /// Generate a cryptographically-secure random number in the range `[min, max]` (inclusive).
+    ///
+    /// Returns [`GenRangeError::InvalidRange`] if `min > max`.
+    pub fn gen_range(&self, min: u32, max: u32) -> Result<u32, GenRangeError> {
+        if min > max {
+            return Err(GenRangeError::InvalidRange { min, max });
+        }
+        Ok(rng().random_range(min..=max))
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use uuid::Uuid;
+
+    use super::*;
+
+    #[test]
+    fn gen_bytes_returns_requested_length() {
+        let client = SdkRandomNumberClient::new();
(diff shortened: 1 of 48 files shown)


crates/bitwarden-random/src/ffi.rs
theMickster, line 1: ❓ Two multi-agent code reviews (opus & sonnet) brought up the following finding (yes I also saw that the GH workflow code review notes and passed on the finding). 
What are your thoughts @quexten? 
Are we able to avoid the panics with additional defensive coding like Claude suggested below or not? Cool with me if our decision is noted and clear that we want these panics, but just starting a convo.

<details>
<summary>gen_bytes() and gen_range() panic on out-of-range input → uncatchable crash on iOS</summary>

<br/>`crates/bitwarden-random/src/ffi.rs:31-36` (and `:58-59`)
<br/>**Caught by:** Security & logic agent / Bug analysis agent / Code quality agent
<br/>**Models:** sonnet *(opus pipeline dismissed it as documented/bounded — overturned by the skeptical gate)*

`SdkRandomNumberClient` is exported as `#[cfg_attr(feature = "uniffi", uniffi::export)]` / `#[cfg_attr(feature = "wasm", wasm_bindgen)]`, with two infallible (non-`Result`) methods that panic on caller-supplied input:

- `gen_bytes(&self, len: u32) -> Vec<u8>` panics when `len > 1024` (ffi.rs:32-36).
- `gen_range(&self, min: u32, max: u32) -> u32` panics when `min > max` (ffi.rs:59, via `rng().random_range(min..=max)` on an empty range).

**Verified behavior (third, adversarial review gate read the pinned UniFFI 0.31.0 source).** UniFFI wraps every exported call in `catch_unwind`, so the panic does **not** unwind across the FFI boundary — it becomes a `CALL_PANIC` status with a serialized message. The foreign-side consequence is platform-dependent:

- **iOS / Swift:** because these are infallible exports, the generated wrapper uses `try!` in a non-`throws` function. The lifted `UniffiInternalError.rustPanic` hits that force-try and **traps — the app crashes, and the caller has no syntactic way to catch it.** This is the real defect.
- **Android / Kotlin:** the panic surfaces as a catchable `InternalException` (still crashes if unhandled, but the caller *can* catch it).
- **WASM:** a catchable JS exception.

This implicates **P06** (minimized impact / availability): on iOS a swapped `min`/`max` or an oversized `len` — plausible client programming mistakes, not adversary-controlled remote input — terminate the app. Every other fallible exported method in `bitwarden-uniffi` returns `Result<_>` (e.g. `initialize_user_crypto`, `derive_key_connector`); these two break that established convention.

**Suggested fix:** change the signatures to `Result<Vec<u8>, RandomError>` / `Result<u32, RandomError>` and return `Err(...)` instead of panicking — a normal catchable error on every platform, consistent with the crate's convention.

*Severity is Important, not Blocker: the inputs are caller-side programming mistakes (not remote/adversarial) and `gen_bytes` is capped at 1 KiB by design. The original sonnet finding's "process abort on all mobile callers / unwinds across FFI" was corrected by the gate to the iOS-specific, catch_unwind-mediated behavior above.*

</details>

theMickster (changes requested):
Will you please add information & instructions for Claude Code in both the repo's root `.claude/CLAUDE.md` file and into a new `CLAUDE.md` file in the new crate? It's important that we strive to document precisely what we want Claude Code to do or not do when it comes to the required usage of the new crate.

---

Multiple multi-agent reviews found this additional location where it appears that we  need to apply this change. I double-checked and didn't see it here in the PR. Should we also update that as well?

<details>
<summary>RNG centralization sweep misses HighEntropySecret::make key-generation path</summary>
<br/>`crates/bitwarden-crypto/src/safe/high_entropy_secret.rs:73`
<br/>**Caught by:** Code quality agent / Architecture agent / Security & logic agent
<br/>**Models:** both *(opus and sonnet independently)*

The PR routes all SDK randomness through `bitwarden_random::rng()`, migrating every other `rand::rng()` call site in `bitwarden-crypto`. The cmon-bro gate confirmed against the actual code: `HighEntropySecret::make()` (a public production function declared at line 67) still calls `rand::rng().fill_bytes(secret.as_mut_slice())` at **line 73**, the file has no hunk in this PR's diff, and a crate-wide grep shows all 18 other `rand::rng()` survivors map 1:1 to migration hunks — this is the **only** site missed.

This generates high-entropy secret keying material (an **EK** — Encryption Keys — site). It is **not** a production security weakness: both `rand::rng()` and `bitwarden_random::rng()` draw from a CSPRNG, so EK-1's 256-bit guarantee is unaffected. The concrete consequence is consistency / test-determinism: under the test-only `dangerous-seeded-rng-for-testing` feature, `bitwarden_random::rng()` consults the seeded thread-local stream while `rand::rng()` never does — so this one path stays non-deterministic while every migrated sibling becomes reproducible, defeating the centralization the new crate exists to provide.

**Suggested fix:** change line 73 to `bitwarden_random::rng().fill_bytes(secret.as_mut_slice())` (`bitwarden-crypto` gains the `bitwarden-random` dependency in this PR, so it is in scope).

</details>

---

Multiple multi-agent reviews pointed out that we could improve test coverage. Please give this suggestion a review and consider adding a test as applicable. 

<details>
<summary>New make_aes256_cbc_hmac_key() WASM binding ships without a test</summary>

<br/>`crates/bitwarden-wasm-internal/src/pure_crypto.rs:153-156`
<br/>**Caught by:** Code quality agent
<br/>**Models:** both *(opus: "lacks a test"; sonnet: "undocumented + untested")*

The diff adds a net-new public WASM-exported `make_aes256_cbc_hmac_key() -> SymmetricCryptoKey` with `#[wasm_bindgen(unchecked_return_type = "SymmetricKey")]` (confirmed against the diff; not pre-existing). It has no accompanying test. Each sibling key-generation export has a one-line test (`test_make_aes256_cbc_hmac_key`, `test_make_xchacha20_poly1305_key`), establishing a per-key-gen test convention this new binding breaks.

**Corrections applied by the cmon-bro gate** (the original findings' framing was imprecise):
- The cited tests at pure_crypto.rs:621/627 actually exercise the **different** functions `make_user_key_aes256_cbc_hmac` / `make_user_key_xchacha20_poly1305` (which return `Vec<u8>`), not the new `SymmetricCryptoKey`-returning method — the shared name prefix is coincidental. The *test-convention-break* claim still holds; the "these are its tests" wording did not.
- The "undocumented breaks a doc convention" angle is **overstated**: the two direct siblings (`make_user_key_*`) are themselves undocumented, so the new method breaks no immediate-sibling doc convention. The missing **test** is the actionable debt.
- Sharper point the original findings undersold: this is the one net-new public API in an otherwise mechanical RNG-migration PR, and it returns `SymmetricCryptoKey` directly across the WASM ABI via `unchecked_return_type` while every sibling returns `Vec<u8>` — an API-shape departure worth a reviewer's explicit sign-off.

**Suggested fix:** add a test mirroring the sibling key-gen bindings (assert a valid 64-byte AES256-CBC-HMAC key), and confirm the new API surface + return shape is intentional.


</details>

quexten:
@theMickster addressed your feedback.

mcamirault (dismissed):
Tools changes look good

dereknance (dismissed):
✅ Platform

quexten:
Apologies for the force push, the history was in a weird state, where already merged commits from other PR's were showing up on this PR.

theMickster (approved):
The changes I requested look good. TY!

SaintPatrck (approved):

mcamirault (approved):
Tools files haven't changed since last review

crates/bitwarden-random/tests/wasm.rs
```diff
@@ -0,0 +1,57 @@
+//! WASM integration tests for the [`SdkRandomNumberClient`] FFI, exercising the bindings as
```
eligrubb, line 1: 🤔 It would be good to add typescript integration tests to the `bitwarden_wasm_internal` integration tests for `SdkRandomNumberClient`. The PR adds tests for the Swift side, but it would add confidence to extend to the TS tests too, unless you think the `#[wasm_bindgen_test]` tests are sufficient.

eligrubb (approved):
Nice job! I just had one non-blocking testing suggestion.

crates/bitwarden-random/tests/wasm.rs
quexten: Since we have 4/5 approvals, I'll make a follow-up PR, unless the last review is blocking.

dereknance (approved):

PR #1225, opened 2026-06-30
[PM-26955] Password Salt fallbacks to normalized email

ike-kottlowski:
## 🎟️ Tracking

[PM-26955](https://bitwarden.atlassian.net/browse/PM-26955)

## 📔 Objective

We need to fall back to the normalized email when the salt comes back from the pre login request `null`. This matches the fall back that currently happens in the client.

## 🚨 Breaking Changes

[PM-26955]: https://bitwarden.atlassian.net/browse/PM-26955?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


crates/bitwarden-auth/src/login/login_via_password/password_prelogin_response.rs
```diff
@@ -23,10 +23,12 @@ pub struct PasswordPreloginResponse {
     pub salt: String,
 }
 
-impl TryFrom<PasswordPreloginResponseModel> for PasswordPreloginResponse {
+impl TryFrom<(PasswordPreloginResponseModel, &str)> for PasswordPreloginResponse {
```
ike-kottlowski, line 27: 🎨 : This tuple in the TryFrom isn't stylistically "correct" since TryFrom usually is treated more like a mapper. In this context it's used to keep the change set small as well as the implied temporary nature of the fallback as outlined in [JIRA](https://bitwarden.atlassian.net/browse/PM-28143)

crates/bitwarden-auth/src/login/login_via_password/password_prelogin.rs
```diff
@@ -39,7 +39,7 @@ impl LoginClient {
         &self,
         email: String,
     ) -> Result<PasswordPreloginResponse, PasswordPreloginError> {
-        let request_model = PasswordPreloginRequestModel::new(email);
+        let request_model = PasswordPreloginRequestModel::new(email.clone());
```
ike-kottlowski, line 42: 🎨 Added the `clone()` here since we needed to consume the email in the `try_from`

enmande (approved):

PR #1229, opened 2026-07-06
[PM-39992] Wipe IPC Crypto Session When Destination is Unreachable

quexten:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-39992
Related clients PR: https://github.com/bitwarden/clients/pull/21653

## 📔 Objective

Currently there is a few bugs around disconnects with the IPC layer. For the crypto layer, if one side disconnects, and then comes back with a fresh crypto state, the first message will be dropped. Subsequently, a re-establishment of the session will happen. We do not want the first message to be dropped because this already leads to some breakages of the intended use-cases. Namely, downstream in the clients layer, the connection confirms the connect worked correctly by starting a version request. Currently, this fails every time after a restart of the desktop app, because of the above issue.

There are a few ways to fix this. In this PR the fix chose is to propagate through the unreachability error throughout the SDK. The transport layer can elect to inform the SDK that the destination is unreachable. In this case, the SDK crypto layer will discard the crypto session. The next time the destination is available again, a new handshake will happen.

This also prevents stale crypto sessions from sticking around forever.

The PR provides a test case for the above scenario.

### DIFF
--- a/crates/bitwarden-ipc/src/crypto_provider/noise/crypto_provider.rs
+++ b/crates/bitwarden-ipc/src/crypto_provider/noise/crypto_provider.rs
@@ -30,12 +30,12 @@ pub enum NoiseCryptoProviderError {
     Timeout,
     /// The destination could not be reached (the underlying transport is not connected).
     TransportUnreachable,
-    /// Could not send via the underlying transport. `fatal` is derived from the underlying
-    /// backend error's [`IpcErrorKind`] classification.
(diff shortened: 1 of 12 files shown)


dani-garcia (approved):
Platform changes lgtm

eligrubb (approved):
Looks good! enum > bool, always.

PR #1230, opened 2026-07-06
[PM-40047] Refactor cipher blob serialization to JSON to drop redundant base64

shane-melton:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-40047

## 📔 Objective

Blob-encrypted ciphers stored their sealed payload as base64-encoded CBOR in `Cipher.data`. Because the inner fields (`wrapped_cek`, `envelope`) are already base64 strings, the outer base64 added ~33% overhead for no benefit. This switches the sealed container to JSON stored verbatim (no outer base64), cutting stored blob size ~24% with no change to the inner crypto envelope. Format detection now keys off a top-level `format_version` key instead of base64/CBOR parse-success. Test vectors regenerated; the now-unused `ciborium` dependency removed.

## 🚨 Breaking Changes

None for stored data — this is a pre-release feature, so only test vectors regenerate (no migration). Requires the paired server-side detection change (also PM-40047). No client change needed — `data` is an opaque pass-through, and detection ships via the SDK bump.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1401,7 +1401,6 @@ dependencies = [
  "bitwarden-test",
  "bitwarden-uuid",
  "chrono",
- "ciborium",
  "data-encoding",
  "futures",
  "hmac 0.13.0",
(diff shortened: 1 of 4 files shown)


gbubemismith (approved):
Changes look good

PR #1231, opened 2026-07-07
Improve claude files

Hinton:
## 🎟️ Tracking

N/A — maintenance of the repo's Claude Code agent instructions.

## 📔 Objective

Restructure the Claude Code agent docs so guidance loads where it's relevant instead of all at once.

- Slim `.claude/CLAUDE.md` down to a high-level overview: what the SDK is, the license zones
  (`crates/` OSS vs `bitwarden_license/` commercial), commands (build/test/lint, WASM suite,
  generated-code regeneration), the four-layer architecture, and references. Details that only
  matter when touching specific code have been moved out.
- Add path-scoped rule files under `.claude/rules/` that load automatically when matching files are
  touched:
  - `crypto.md` — `bitwarden-crypto` / `bitwarden-organization-crypto` (backward-compat, `safe`
    module, no hazmat/key material, `KeyStoreContext` rules, naming, constant-time compares).
  - `generated-api-crates.md` — `bitwarden-api-api` / `bitwarden-api-identity` (don't hand-edit;
    regenerate via `build-api.sh` / templates).
  - `bindings.md` — `bitwarden-uniffi` / `bitwarden-wasm-internal` (thin bindings only, build/publish
    layout).
  - `rust-conventions.md` — repo-wide `.rs` conventions (data-model suffixes, threading,
    cross-binding exports, security).

Net effect: the same guidance, but scoped so each rule surfaces only when working in the relevant
crate rather than bloating the always-on context.

## 🚨 Breaking Changes

None. This only touches `.claude/` agent instruction files — no code, API, or serialization changes.

### DIFF
--- a/.claude/CLAUDE.md
+++ b/.claude/CLAUDE.md
@@ -1,168 +1,86 @@
-# Bitwarden Internal SDK
+# CLAUDE.md
 
-Cross-platform Rust SDK implementing Bitwarden's core business logic.
+This file provides guidance to Claude Code (claude.ai/code) when working with code in this
+repository.
 
-**Rust Edition:** The SDK targets the
-[2024](https://doc.rust-lang.org/nightly/edition-guide/rust-2024/index.html) edition of Rust.
+## What this repo is
 
-**Crate documentation**: Before working in any crate, read available documentation: `CLAUDE.md` for
-critical rules, `README.md` for architecture, `examples/` for usage patterns, and `tests/` for
-integration tests. **Before making changes or reviewing code, review relevant examples and tests for
-the specific functionality you're modifying.**
+Bitwarden's internal cross-platform SDK: core business logic in Rust (edition 2024), consumed by web
+and desktop clients through WASM bindings and by mobile through UniFFI. Not for public use — the API
+is unstable and changes without warning. Two license zones: `crates/` (OSS) and `bitwarden_license/`
+(commercial: `bitwarden-sm`, `bitwarden-commercial-vault`, `bitwarden-rotation-daemon`).
 
-## Architecture Overview
+Path-scoped rules live in `.claude/rules/` and load automatically when you touch matching files
+(crypto crates, generated API crates, binding crates, repo-wide Rust conventions). Several crates
+also carry their own `CLAUDE.md` and substantive `README.md` (e.g. `bitwarden-state`,
+`bitwarden-exporters`, `bitwarden-importers`, `bitwarden-ipc`, `bitwarden-threading`) — read the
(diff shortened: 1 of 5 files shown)


.claude/CLAUDE.md
```diff
@@ -1,168 +1,86 @@
...
+   apps consume — one instance per user (`HashMap<UserId, PasswordManagerClient>`), lifecycle init →
+   lock/unlock → logout (drop). The current sub-client list is in `crates/bitwarden-pm/src/lib.rs`.
+4. **Bindings** — `bitwarden-uniffi` (Swift/Kotlin) and `bitwarden-wasm-internal`
+   (TypeScript/JavaScript). Thin bindings only — no business logic.
+
+## API breaking changes
+
+A cross-repo workflow flags breaking API changes on every PR, and TypeScript compilation is tested
+against the `clients` repo. If you merge a breaking change, prepare the corresponding `clients` PR
+immediately — unresolved breaks block other SDK integrations.
```
SaintPatrck: ❓ What is the actual expectation here? Claude doesn't see this after something merges in GitHub, so I'm curious _when_ this should be acted on. To me, it reads more like instruction that should live in an encapsulating workflow than included here.

.claude/CLAUDE.md
```diff
@@ -1,168 +1,86 @@
...
-- Underlying script: `scripts/lint.sh`.
-
-**WASM Testing:**
-
-- `cargo test --target wasm32-unknown-unknown --features wasm -p bitwarden-error -p bitwarden-threading -p bitwarden-uuid` -
-  WASM-specific tests
+# CLAUDE.md
+
+This file provides guidance to Claude Code (claude.ai/code) when working with code in this
+repository.
```
SaintPatrck: ⛏️ This reads like human-facing documentation. Claude already knows what `CLAUDE.md` files are for. Can we retain the previous H1 header and put the "What this repo is" body below it?

```suggestion
# Bitwarden Internal SDK

Bitwarden's internal cross-platform SDK...
```

SaintPatrck (commented):
Couple of minor things in `CLAUDE.md`. Also, Claude's warning looks legit. Unless the crate it calls out is being added in an open/upcoming PR, we should refrain from referencing it.

.claude/CLAUDE.md
Hinton: I've found it useful to point claude at cross-repo work, but I agree this does not provide value here.

SaintPatrck (approved):

SaintPatrck:
Can you update the PR description before this is merged?

PR #1235, opened 2026-07-09
[PM-39239] Surface shareable access URLs in bw send output

adudek-bw:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-39239

## 📔 Objective

**For Rust CLI**

bw send get --text and the default output of bw send create now emit a shareable access URL that a user can paste to a recipient, instead of the raw access_id.

Previously --text emitted only the raw access_id and the create default output omitted the URL entirely — both because bw had no access to the configured web-vault base URL.


crates/bw/src/tools/send.rs
```diff
@@ -711,13 +708,84 @@ async fn run_create(
...
-    Ok(view
+    // The default output is the shareable access URL — the primary artifact a caller wants to
+    // hand to a recipient. `--fullObject` opts back into the full JSON view.
+    let url = build_access_url(client, &view)?;
+    Ok(url.into())
+}
+
+/// Build the shareable Send access URL from a decrypted [`SendView`].
+///
+/// Format: `<web-vault>/#/send/<access_id>/<url_b64_key>`.
```
harr1424: ❓ I need to look at this more closely, but this look dangerous and might behave differently across environments. 

@itsadrago can you please take a look at this and determine if this will be verifiable before a production release? I think a testing strategy was established during [PM-39449](https://bitwarden.atlassian.net/browse/PM-39449).

[See also](https://github.com/bitwarden/send/blob/master/script.js)

crates/bw/src/tools/send.rs
```diff
@@ -711,13 +708,84 @@ async fn run_create(
...
+        .ok_or_else(|| eyre!("Send is missing an access id; cannot build a shareable URL."))?;
+    let key = view
+        .key
+        .as_deref()
+        .ok_or_else(|| eyre!("Send is missing a key; cannot build a shareable URL."))?;
+
+    let web_vault = web_vault_url(client);
+    let url_key = to_url_b64(key);
+
+    Ok(format!("{web_vault}/#/send/{access_id}/{url_key}"))
```
djsmith85, line 749: While the approach itself is likely okay, it does bare risk of being brittle and needed in multiple places. I've started a discussion with the rest of the platform team, to see if there's a way to provide that url in a similar manner as the TS-clients currently do https://github.com/bitwarden/clients/blob/main/libs/common/src/platform/services/default-environment.service.ts#L441

crates/bw/src/tools/send.rs
mcamirault: My main takeaway from 39449 was that the US production env - and *only* that env - has a vanity URL that redirects to the vault, but hitting the vault link directly works too. Since that is the case I think the links produced by this code will work as-is, we just won't get the vanity URL in US prod. It would still be good to get confirmation from @itsadrago, that was a messy ticket :D

harr1424 (dismissed):

crates/bw/src/tools/send.rs
itsadrago: @mcamirault is correct. The links should work as-is, but we lose the vanity URL for the US prod env. I believe the CLI client doesn't have an authoritative source for the correct sends hostname and domain to use and so we have to derive it. We need a configuration service in this repo similar to `DefaultEnvironmentService` in the clients repo, but that's probably out of the scope of this work

itsadrago: Apologies, I had assumed we were actually getting a vault URL, but since we're actually using the api URL the code in the PR won't work. We will need to park this work until we can pull an actual send URL from a configuration in the sdk

adudek-bw:
Updated code with comments but still in somewhat of a holding pattern until a potential followup is decided on. I'd be willing to merge since this is only for rust CLI

mcamirault (dismissed):

crates/bw/src/tools/send.rs
djsmith85: @adudek-bw I'm going to go ahead and approve as discussions are still ongoing. Currently there is no pattern/guidance from platform to expose/pass on those environment URIs.

djsmith85 (dismissed):

djsmith85 (approved):

PR #1236, opened 2026-07-09
[PM-40220] Add PAM Access Rules

Hinton:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-40220

## 📔 Objective

Add the PAM (Privileged Access Management) Access Rules domain to the SDK as a self-contained commercial crate: the request/view DTOs, the condition model, local validation, and a CRUD client over the server's access-rules API.

- **`AccessRulesClient`** with the five CRUD operations (`list`, `get`, `create`, `update`, `delete`) over an organization's access rules, backed by the generated `access_rules_api()`.
- **`AccessRuleView` / `AccessRuleAddEditRequest`** DTOs with conversions to/from the generated `bitwarden-api-api` models. A single add/edit request type is shared by create and update because the server derives the immutable IDs from the URL rather than the request body.
- **`AccessCondition`** tagged enum (`human_approval`, `ip_allowlist`) with an untagged `Unknown` catch-all that preserves condition kinds this SDK version doesn't model (e.g. the server's `time_of_day`) verbatim across list / enable-disable round trips. Documented limitation: extra fields on a *modeled* kind are dropped on a round trip.
- **Local validation** (`validate_request`) run before every create/update, covering name length (measured in UTF-16 code units, ≤ 256), lease- and extension-duration consistency, condition count (≤ 10), and CIDR syntax — so malformed input fails fast with a typed `AccessRuleValidationError` instead of a server round trip.
- **`is_valid_cidr`** — a standalone, deliberately strict CIDR check (our own implementation rather than a dependency). It is stricter than both the server and .NET's `IPNetwork.TryParse`: it rejects set host bits, leading-zero/octal and hex octets, partial addresses, IPv6 zone IDs, and IPv4-mapped IPv6, so client and server can never disagree about which network a rule matches.
- **Flat `AccessRuleError`** (`Validation`, `InvalidConditions`, `MissingField`, `Chrono`, `Api`).
- **Tests**: unit tests for the response↔view/request conversions, condition round-tripping, and `is_valid_cidr` edge cases, plus mocked-API tests covering all five client operations.

The crate's public types derive `Tsify` and the client and `is_valid_cidr` are annotated `#[wasm_bindgen]`, so the crate is WASM-ready. This PR does not itself wire PAM into the web client (`bitwarden-wasm-internal`) — that is left to a follow-up.


bitwarden_license/bitwarden-pam/src/access_rules/models.rs
```diff
@@ -0,0 +1,294 @@
...
+    fn try_from(response: AccessRuleResponseModel) -> Result<Self, Self::Error> {
+        let conditions = match response.conditions {
+            None => Vec::new(),
+            Some(serde_json::Value::Array(items)) => items
+                .into_iter()
+                .map(|item| {
+                    serde_json::from_value(item)
+                        .map_err(|e| AccessRuleError::InvalidConditions(e.to_string()))
+                })
+                .collect::<Result<Vec<_>, _>>()?,
```
dani-garcia: This can be simplified a bit if we use from_value on the array directly:
```suggestion
            Some(value @ serde_json::Value::Array(_)) => serde_json::from_value(value)
                .map_err(|e| AccessRuleError::InvalidConditions(e.to_string()))?,
```

dani-garcia (commented):
LGTM, just one small suggestion on the deserializing

kspearrin (approved):

PR #1237, opened 2026-07-09
[PM-40179] Add bitwarden-pam crate scaffolding

Hinton:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-40179

## 📔 Objective

Adds the scaffolding for a new commercial crate, `bitwarden_license/bitwarden-pam`, which will own the Privileged Access Management (PAM) domain:

- **Crate skeleton**: `bitwarden-pam` added to the workspace, depending only on `bitwarden-core` (plus optional `wasm-bindgen` behind the `wasm` feature).
- **Client struct**: an empty `PamClient` using `#[derive(FromClient)]`, exposed via the `PamClientExt` extension trait on `Client`. Dependency fields (e.g. `Arc<ApiConfigurations>`) will be added alongside the methods that need them.
- **Application wiring**: reachable through `client.commercial().pam()` on `PasswordManagerClient` (behind the `bitwarden-license` feature).

### DIFF
--- a/crates/bitwarden-pm/Cargo.toml
+++ b/crates/bitwarden-pm/Cargo.toml
@@ -39,7 +39,7 @@ wasm = [
     "bitwarden-generators/wasm",
     "bitwarden-importers/wasm",
     "bitwarden-organization-invite-link/wasm",
-    "bitwarden-pam/wasm",
+    "bitwarden-pam?/wasm",
     "bitwarden-policies/wasm",
     "bitwarden-send/wasm",
     "bitwarden-state/wasm",
--- a/crates/bitwarden-pm/Cargo.toml
+++ b/crates/bitwarden-pm/Cargo.toml
@@ -39,6 +39,7 @@ wasm = [
     "bitwarden-generators/wasm",
     "bitwarden-importers/wasm",
     "bitwarden-organization-invite-link/wasm",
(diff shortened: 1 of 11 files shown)


eliykat (dismissed):
Approving CODEOWNERS changes only.

crates/bitwarden-pm/Cargo.toml
```diff
@@ -39,6 +39,7 @@ wasm = [
     "bitwarden-generators/wasm",
     "bitwarden-importers/wasm",
     "bitwarden-organization-invite-link/wasm",
+    "bitwarden-pam/wasm",
```
dani-garcia: The way we're adding the pam crate wasm feature here would implicitly enable the crate when using the wasm feature. We need to use ? here to make sure the crate doesn't get implicitly added.
```suggestion
    "bitwarden-pam?/wasm",
```

`bitwarden-commercial-vault` is also doing the same thing, so we've been including it accidentally in OSS builds 🫤, if you can you should also switch it to use ?, otherwise I'll do it in a followup.

dani-garcia (changes requested):

dani-garcia (approved):

quexten (approved):
Re-approving codeowners.

PR #1238, opened 2026-07-10
[PM-40201] Derive missing SSH public key and fingerprint in the SDK

gbubemismith:
## 🎟️ Tracking
https://bitwarden.atlassian.net/browse/PM-40201

## 📔 Objective
Some SSH key ciphers are stored with a private key but a null `publicKey` and/or `keyFingerprint`, created through a since-closed client loophole. When one of these items is present, decryption fails and the user's vault breaks completely

## 🚨 Breaking Changes
1. Encrypted SshKey fields are now optional: public_key/fingerprint went EncString → Option<EncString>, making them nullable/optional in the mobile (uniffi) and TS (tsify) bindings; consumers reading or constructing the encrypted SSH key must handle absence.
2. bitwarden-ssh return type changed:  `import_key/import_pkcs8_der_key/generate_sshkey` now return `SshKeyData` instead of `SshKeyView`; Rust-only (doesn't cross FFI).

### DIFF
--- a/crates/bitwarden-ssh/src/lib.rs
+++ b/crates/bitwarden-ssh/src/lib.rs
@@ -9,7 +9,6 @@ pub mod generator;
 #[allow(missing_docs)]
 pub mod import;
 
-use bitwarden_vault::SshKeyView;
 use error::SshKeyExportError;
 use pkcs8::LineEnding;
 use ssh_key::{HashAlg, PrivateKey};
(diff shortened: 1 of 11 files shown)


gbubemismith:
@bitwarden/team-key-management-dev, on the `bitwarden-ssh` crate, I dropped its dependency on bitwarden-vault. The only reason for the edge was the return type. `SshKeyData` is just { private_key, public_key, fingerprint: String }, FFI wrappers convert it back to SshKeyView, so callers are unaffected.

nick-livefront (dismissed):
@gbubemismith Are the mobile teams aware of this change so they can integrate the breaking changes?

crates/bitwarden-ssh/src/lib.rs
```diff
@@ -9,20 +9,30 @@ pub mod generator;
...
 #[cfg(feature = "uniffi")]
 uniffi::setup_scaffolding!();
 
-fn ssh_private_key_to_view(value: PrivateKey) -> Result<SshKeyView, SshKeyExportError> {
+/// Decoded SSH key material returned by this crate's import/generate functions.
+#[derive(Debug)]
+pub struct SshKeyData {
+    /// SSH private key in unencrypted OpenSSH format.
+    pub private_key: String,
+    /// SSH public key according to RFC 4253.
```
quexten, line 24: Should this be rfc4716?

quexten: Oh, section 6.6   `[RFC4253](https://datatracker.ietf.org/doc/html/rfc4253#section-6.6).`

quexten (dismissed):
Changes look good, I don't think KM owns the bitwarden-ssh crate though. It feels more like a @bitwarden/team-tools-dev  thing because all it does is control generator, import.

gbubemismith:
> @gbubemismith Are the mobile teams aware of this change so they can integrate the breaking changes?

I'll notify them today. I already had a discussion with one of their engineers before I created the PR

gbubemismith:
> Changes look good, I don't think KM owns the bitwarden-ssh crate though. It feels more like a @bitwarden/team-tools-dev thing because all it does is control generator, import.

Oh okay. For some reason I always assumed it was under KM. Thanks for reviewing

harr1424:
@gbubemismith Tools Team is OK with becoming the sole owner of `bitwarden-ssh` in case you would like to modify the `CODEOWNERS` file as part of this PR.

harr1424 (approved):

nick-livefront (approved):

theMickster (approved):
Approving on behalf of tech leads for the codeowners changes.

djsmith85 (approved):

PR #1239, opened 2026-07-10
[PM-35785] Add strongly typed policy data - alternative implementation

eliykat:
## 📔 Objective

An alternative approach to https://github.com/bitwarden/sdk-internal/pull/1219 based on @quexten 's feedback.

The approach in `main`, extended in the previous PR, is to have a registry struct where policy rules can be registered and dynamically looked up at runtime. This was intended to allow other teams to write their own rules and register them in our engine. However, this dynamic runtime behavior was complex and used generics which don't travel over the FFI boundary (an unsolved problem for that approach).

This PR takes a step back and tries to reimplement this behavior in a more simple and idiomatic way. It has breaking changes and is missing tests, so it is NOT intended for merge as-is. It demonstrates the end state as if it replaced the current implementation entirely.

Notes:

- the registry and dynamic lookup are deleted entirely
- added a new `EnrichedPolicyType` enum, which also represents each `PolicyType`, but carries its own strongly-typed data. This is effectively the "registry", as it has internal match lookups for:
  - converting from a `PolicyType` + json blob to the enriched type
  - getting the `PolicyDefinition` enforcement rules
- naming tweak for clarity: the `Policy` trait is now `PolicyDefinition`, because it's metadata that 'defines' how a policy works (open to other naming, naming is hard etc.)
- `policy_definitions.rs` contains both the per-policy data structs and the `PolicyDefinition` trait implementations. I think it's nice that we can use the data struct to also implement the trait - it keeps per-policy configuration together. (It's this file that could be split up and distributed among other feature teams.)
- overall seems simpler and easier to understand to me.

Remaining challenges:
- the cross-team interface issue: this effectively abandons the idea of an interface for other teams - they have to edit the internals of our implementation to add their policy. Maybe this is worth the tradeoff, but then we still need to make it as easy and as simple as possible for them, minimizing what they have to edit. I think we can iterate on this with macros if needed to minimize the boilerplate in these match statements.
- callers will expect to get back a certain policy type. e.g. if they call `filter(policies, contexts, PolicyType.MasterPassword)`, they will expect back all `MasterPassword`-type policies with corresponding data; not policies containing enums of unknown values that they then have to filter/match on. I'm not sure how to reconcile this. It may be a type cast that native code has to do.

Todos:
- determine the best serde option for exposing the enriched enum in TS - maybe internal tagging.
- if this is the best way forward, it needs to be made non-breaking. It should be able to sit behind the current public interface and just map back to a `PolicyView`.
- add tests to verify no change in behavior.


crates/bitwarden-policies/src/policy_definitions.rs
```diff
@@ -0,0 +1,455 @@
...
+//! Each policy type is defined here with its associated data structure (if any)
+//! and its [`PolicyDefinition`] trait implementation.
+//! Organized by policy type numeric value.
+
+use std::str::FromStr;
+
+use bitwarden_api_api::models::MasterPasswordPolicyResponseModel;
+use bitwarden_organizations::OrganizationUserType;
+use bitwarden_send::SendType;
+use bitwarden_vault::UriMatchType;
```
eliykat, line 12: Unclear whether our feature crate can import other feature crates like this. Their crate will almost certainly depend on this crate (to enforce policies in their domain), so it would be a circular dep. Not sure how this could be avoided.

This also causes the uniffi issues.

eliykat:
Closed in favour of https://github.com/bitwarden/sdk-internal/pull/1364.

PR #1240, opened 2026-07-10
[PM-40213] Add license section to each crate

quexten:
Adds a license entry for each crate, so that cargo-deny can understand them https://bitwarden.atlassian.net/browse/PM-40213


Hinton:
The license is _EITHER_ GPL or Bitwarden SDK. Defining it as GPL is misleading and why we use a license file.

PR #1241, opened 2026-07-10
chore(deps): update crossbeam-epoch version

quexten:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-40215

## 📔 Objective

Update the crossbeam utils version because of https://rustsec.org/advisories/RUSTSEC-2026-0204


dani-garcia:
I don't think we need a patch for this, nothing is pinning this version currently. You should be able to do:

```
cargo update -p crossbeam-epoch --precise 0.9.20
```

That said you may need IT to allow it as that version is 4 days old only.

quexten:
Adding an override in clients instead.

PR #1245, opened 2026-07-10
chore(deps): add license for crates owned by @bitwarden/team-key-management-dev

quexten:
https://bitwarden.atlassian.net/browse/PM-40213

Cargo deny needs a proper license entry instead of a license file. We add "GPL-3.0-only OR LicenseRef-Bitwarden-SDK"

### DIFF
--- a/crates/bitwarden-crypto/Cargo.toml
+++ b/crates/bitwarden-crypto/Cargo.toml
@@ -11,7 +11,7 @@ rust-version.workspace = true
 readme.workspace = true
 homepage.workspace = true
 repository.workspace = true
-license-file.workspace = true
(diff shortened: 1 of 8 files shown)


mzieniukbw (changes requested):
We also own `bitwarden-api-key-connector` and `bitwarden-crypto-cipher-suite`, but i just noticed we have not added the latter one to CODEOWNERS file. 
Adding to `CODEOWNERS` file addressed by #1250

quexten:
@mzieniukbw those two were merged as part of the platform PR.

mzieniukbw (approved):

PR #1253, opened 2026-07-10
Ac/pm 38906 decrypt list with failures for collections

JaredScar:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-38906

## 📔 Objective

Utilize SDK for decryption and utilize decryptMany methods so only 1 SDK call is used


crates/bitwarden-crypto/src/store/mod.rs
```diff
@@ -476,4 +478,67 @@ pub(crate) mod tests {
...
+    /// `DefaultCollectionEncryptionService` TS feature flag): decrypting one item at a time via
+    /// [KeyStore::decrypt] (the "original"/sequential path, not parallelized) versus decrypting
+    /// the whole list in a single call to [KeyStore::decrypt_list_with_failures] (the "bulk" path,
+    /// parallelized internally via rayon).
+    ///
+    /// This is a `#[test]`, not a criterion benchmark, because it needs the crate's `#[cfg(test)]`
+    /// helper types (`Data`/`DataView`/`TestIds`/`TestSymmKey`), which aren't visible to an
+    /// external `benches/` binary. Run with `cargo test speed_test_decrypt -- --nocapture` to see
+    /// the timing/throughput/speedup numbers; the assertions only check correctness, not speed, so
+    /// this won't flake in CI on slower/noisier machines.
```
BTreston: Can we simplify this a bit? The meaning is a bit lost in the verbosity.

crates/bitwarden-vault/src/collection_client.rs
```diff
@@ -14,6 +14,24 @@ use wasm_bindgen::prelude::wasm_bindgen;
...
+/// This struct contains two vectors: `successes` and `failures`.
+/// `successes` contains the decrypted `CollectionView` objects,
+/// while `failures` contains the original `Collection` objects that failed to decrypt.
+#[cfg_attr(
+    feature = "wasm",
+    derive(Tsify, Serialize, Deserialize),
+    tsify(into_wasm_abi, from_wasm_abi)
+)]
+#[cfg_attr(feature = "uniffi", derive(uniffi::Record))]
+pub struct DecryptCollectionListResult {
```
BTreston, line 28: Yeah, this is what we want to use in the client PR too. We might be able to use the generated SDK type, no?

just connecting the dots from [this comment](https://github.com/bitwarden/clients/pull/21798#discussion_r3561795764)

crates/bitwarden-crypto/src/store/mod.rs
eliykat: I think we can just revert it - we don't need benchmarking tests here and not in another team's code.

quexten: +1. If you want benches, unit tests are not the right place to do it. Also, you are making a lot of work for yourself by implementing this manually rather than with criterion. i recommend just using criterion.

>  /// This is a `#[test]`, not a criterion benchmark, because it needs the crate's `#[cfg(test)]`
    /// helper types (`Data`/`DataView`/`TestIds`/`TestSymmKey`), which aren't visible to an
    /// external `benches/` binary. Run with `cargo test speed_test_decrypt -- --nocapture` to see
    /// the timing/throughput/speedup numbers; the assertions only check correctness, not speed, so
    /// this won't flake in CI on slower/noisier machines.


You do not need test ids / testsymmkey, you can make your own key ids with a macro like this:
https://github.com/bitwarden/sdk-internal/blob/0a188b85f5f0693e5c4e0412822c133885dff9e1/crates/bitwarden-crypto/examples/protect_key_with_secret.rs#L117-L139

quexten: LMK if you want help building the benchmarks using criterion.

dereknance (dismissed):
✅ Platform

crates/bitwarden-uniffi/src/vault/collections.rs
```diff
@@ -34,6 +34,17 @@ impl CollectionsClient {
         Ok(self.0.decrypt_list(collections)?)
     }
 
+    /// Decrypt collection list with failures
+    /// Returns both successfully decrypted collections and any that failed to decrypt
+    // Note that this function still needs to return a Result, as the parameter conversion can
+    // still fail
```
eliykat: I don't think this is right. If converting values across the FFI boundary fails, that's not going to get caught by the `Result`. And there's no other reason why this should return a `Result`, because `decrypt_list_with_failures` fully represents any error states within the existing `DecryptCollectionListResult` return type.

You should be able to return `DecryptCollectionListResult` directly.

crates/bitwarden-vault/src/collection_client.rs
```diff
@@ -190,6 +224,61 @@ mod tests {
...
+    #[tokio::test]
+    async fn test_decrypt_list_with_failures_mixed_results() {
+        let client = Client::init_test_account(test_bitwarden_com_account()).await;
+
+        let valid_collection = test_collection();
+        let mut invalid_collection = test_collection();
+        // No organization key exists in the test account's key store for this id, so
+        // decryption of this single item must fail without affecting the others.
+        invalid_collection.organization_id =
+            "00000000-0000-0000-0000-000000000000".parse().unwrap();
```
eliykat: Simpler and preferable:


```suggestion
        invalid_collection.organization_id = OrganizationId::new_v4();
```

eliykat (changes requested):
Looks good aside from the stray comment.

Please also add wasm integration tests (see examples in `crates/bitwarden-wasm-internal/integration-tests`). They do not need to be comprehensive, they just check the FFI boundary.

gbubemismith (dismissed):
Changes look good

dereknance (dismissed):

eliykat (dismissed):

eliykat (approved):

quexten (approved):
Sweet. Does AC have plans to load collections via repositories so they don't have to be passed in in the future? (@eliykat i know we talked state recently)

BTreston (approved):

gbubemismith (approved):

dereknance (approved):

PR #1257, opened 2026-07-14
feat(crypto): add key-encryption-key, content-encryption-key, key-hierarchy notes to safe

quexten:
Adds the `KeyEncryptionKey` and `ContentEncryptionKey` structs to the safe module, and documentation on how to contstruct features. Moving forward, content encryption is done with aes-gcm. The keys are only used once, which satisfies the nonce-reuse bound. Key-encryption-keys instead use xaes-gcm. They are used generally in a multi-user / multi-device scenario, and are re-used, which necessitates the full nonce range.

Documentation is added on how this maps to the future key-hierarchy of vault encryption, and of send encryption.


crates/bitwarden-crypto/src/safe/data_envelope.rs
```diff
@@ -95,6 +95,35 @@ impl DataEnvelope {
...
+    /// Seals a struct into an encrypted blob, using an existing symmetric key already present in
+    /// the key store as the content-encryption-key, instead of generating a fresh one.
+    ///
+    /// # Warning
+    /// This bypasses the guarantee that a `DataEnvelope`'s content-encryption-key is freshly
+    /// generated and used to seal exactly one envelope. The caller is responsible for ensuring
+    /// `cek` is not reused to seal more than one `DataEnvelope`. Only the AES-256-GCM algorithm is
+    /// supported for externally-provided keys. Prefer [`DataEnvelope::seal`] unless you have a
+    /// specific reason the content-encryption-key must be supplied externally.
+    #[doc(hidden)]
```
quexten: Made this doc(hidden) so that teams don't accidentally use it. Most teams should be using seal instead.

crates/bitwarden-crypto/src/safe/data_envelope.rs
```diff
@@ -95,6 +95,35 @@ impl DataEnvelope {
...
+        cek: Ids::Symmetric,
+        ctx: &mut crate::store::KeyStoreContext<Ids>,
+    ) -> Result<Self, DataEnvelopeError>
+    where
+        T: Serialize + SealableVersionedData,
+    {
+        let key = ctx
+            .get_symmetric_key(cek)
+            .map_err(|_| DataEnvelopeError::KeyStore)?;
+        let SymmetricCryptoKey::Aes256GcmKey(key) = key else {
```
quexten: Todo: I *think* we want to support xaes here, but I'm not sure. Maybe xaes should be reseverd as a KEK (key encryption key), and DEKs should always be aes. Thoughts @eligrubb , @MGibson1 ?

MGibson1: I like simplicity and would advocate for _only_ XAES.

however, functionally, I don't think a DEK would reach the level of reuse needed to wear out an AES key. That would require 2^32 updates to the contained data. I don't see that happening.

eligrubb: My vote would be to use XAES by default. I agree that aes is probably fine for planned DEK uses, but XAES is the safe default for an API where the user can dictate the key. Ideally, a user shouldn't need to know about nonce collision at all to use a SAFE primitive's APIs.

eligrubb: @quexten #1264 has merged, you should be able to use xaes for data envelope now.

mzieniukbw (dismissed):
Except for the XAES discussion, LGTM

crates/bitwarden-crypto/src/safe/key_hierarchy/README.md
```diff
@@ -0,0 +1,118 @@
...
+a high-entropy secret (PRF output, a key-connector secret, a random URL fragment, biometric-derived
+bytes) — never key material the SDK is responsible for.
+
+```text
+Outside the SDK boundary - only secrets cross it, never keys:
+
+    Password / PIN  (low entropy)
+        |  PasswordProtectedKeyEnvelope  (slow, memory-hard KDF)
+        |
+    High-entropy secret  (PRF, key-connector, URL fragment)
```
quexten, line 16: This means that biometrics *should* not store the key directly, but always a high entropy secret, that wraps a key. (This is how we do windows biometrics, but not yet linux, mac).

This also means that the CLI session key *should* not be a key, but a high entropy secret.

crates/bitwarden-crypto/src/safe/key_hierarchy/README.md
```diff
@@ -0,0 +1,118 @@
+# Key Hierarchy
+
+Password manager encryption will organize cryptographic material into a strict hierarchy. This is
```
quexten, line 3: @MGibson1, @eligrubb, @mzieniukbw; re the XAES vs AES-GCM discussion. My proposal is now to make KEKs always XAES, while CEKs(DEKs) are always aes-gcm. I've added comments to the respective files in this PR. Could you confirm that this makes sense? If so, then this will be the new public abstraction. Attachment keys will be replaced by CEKs (at some point). Cipher keys will be replaced by KEK. User key -> KEK. OrganizationSymmetricKey -> KEK. (Any I'm missing?)

quexten: As per https://bitwarden.slack.com/archives/C07HY17HUHE/p1784557052309929?thread_ts=1784540973.819349&cid=C07HY17HUHE and following, accepted.

crates/bitwarden-crypto/src/safe/key_hierarchy/README.md
```diff
@@ -0,0 +1,118 @@
...
+migrated towards. Only passwords and high-entropy secrets ever cross the SDK boundary. **Keys never
+leave the SDK.** Outside the boundary, a client supplies a low-entropy secret (a PIN or password) or
+a high-entropy secret (PRF output, a key-connector secret, a random URL fragment, biometric-derived
+bytes) — never key material the SDK is responsible for.
+
+```text
+Outside the SDK boundary - only secrets cross it, never keys:
+
+    Password / PIN  (low entropy)
+        |  PasswordProtectedKeyEnvelope  (slow, memory-hard KDF)
```
mzieniukbw: > memory-hard KDF

This implies we use Argon2Id, which is not the case for FIDO environment.

crates/bitwarden-crypto/src/safe/key_hierarchy/README.md
```diff
@@ -0,0 +1,118 @@
...
+        |  encrypts content:
+        |    - DataEnvelope
+        |    - AttachmentStream (n/i)
+        v
+    +-------------------------------------------------------------+
+    | Content                                                     |
+    +-------------------------------------------------------------+
+```
+
+`(n/i)` = not implemented.
```
mzieniukbw, line 39: ❓ Will we remember to update this ?

crates/bitwarden-crypto/src/safe/data_envelope.rs
```diff
@@ -71,8 +74,20 @@ impl DataEnvelope {
...
         T: Serialize + SealableVersionedData,
     {
-        let (envelope, cek) = Self::seal_ref(&data, T::NAMESPACE)?;
-        let cek_id = ctx.generate_symmetric_key();
+        // The content-encryption-key is a fresh, single-use content-encryption-key (CEK) stored in
+        // the context.
+        let cek_id = ContentEncryptionKey::make(ctx);
+        let mut cek = match ctx.get_symmetric_key(cek_id) {
+            Ok(SymmetricCryptoKey::Aes256GcmKey(key)) => key.clone(),
+            _ => return Err(DataEnvelopeError::KeyStore),
```
mzieniukbw, line 82: ❓ I assume we want this, to be defensive, that data envelope only accept AES-GCM ?
This is weird though, since we essentially assert `ContentEncryptionKey::make` at runtime, which a standard test could do.
It's not like we have some kind of limitation on data envelope, it's just a contract that we have written between outselves and ContentEncryptionKey. I do wonder what this assertion being here brings.
Aside, i saw `is_key_algorithm_valid` function, which technically does the same thing - but i understand we want to be explicit here.

crates/bitwarden-crypto/src/safe/key_hierarchy/README.md
quexten: Yeah, that's always a problem with docs. I think we will, but would you recommend just dropping the comment?

crates/bitwarden-crypto/src/safe/key_hierarchy/README.md
quexten: Updated to say memory-hard or compute-hard

crates/bitwarden-crypto/src/safe/data_envelope.rs
quexten: There is no limitation here, and we could add support for more key types, but then we would need to extend seal_ref to support the new key types. I tried doing this with CoseKeyView, but the problem is that the "disable_operation" calls are not supported on generic SymmetricCryptoKeys, just on the specific key type variants that are cose keys.

I think we'd need to refactor a bit to make this nicer to read, so I kept it out of this PR.

crates/bitwarden-crypto/src/safe/key_hierarchy/README.md
mzieniukbw: It's ok, just won't want it forgotten.

mzieniukbw (approved):

PR #1259, opened 2026-07-14
feat(invite-link): add invite link stubs for milestone 3; require private key for invite creation

quexten:
## 🎟️ Tracking

## 📔 Objective

Add the stubs for milestone 3 so that the teams are unblocked. Please work against these stubs, implementations will follow.


crates/bitwarden-organization-invite-link/src/invite_link_client.rs
```diff
@@ -85,6 +99,73 @@ impl InviteLinkClient {
...
+            &ctx,
+        )?)
+    }
+
+    /// Accepts an invite by unsealing the organization key from the invite and re-encrypting it to
+    /// the accepting user's own public key (retrieved from the key store), producing an
+    /// [`UnsignedSharedKey`].
+    // TODO(Milestone 3): unseal the organization key from the invite via the invite key bundle.
+    // Until the crypto crate supports it, a placeholder key is encapsulated in its place.
+    pub fn accept_and_confirm_invite(
```
JimmyVo16, line 154: Just to confirm my understanding: This method will be called when the invited user accepts the invitation link. The returned key from this method will be saved in the [OrganizationUser.Key](https://github.com/bitwarden/server/blob/main/src/Sql/dbo/Tables/OrganizationUser.sql#L6) column on the server. This is a good approach since we need to pass the `organization_key` back to the TS client.

I think the only part we're missing is the logic to handle `defaultCollectionName`. Will that be handled elsewhere?

crates/bitwarden-organization-invite-link/src/invite_link_client.rs
```diff
@@ -55,16 +60,25 @@ impl InviteLinkClient {
     /// The organization key is looked up from the client's key store via
     /// [`SymmetricKeySlotId::Organization`]; the caller does not need to provide it directly.
     ///
+    /// `wrapped_private_key` is the organization's private key wrapped with the organization key.
+    ///
     /// Each call produces a unique invite key sampled from a secure cryptographic source.
     ///
     /// # Security
     /// The returned `invite_key` MUST NOT be sent to the server.
     pub fn make_invite(
```
JimmyVo16, line 69: Is it possible to pass in a boolean to indicate whether we want to enable confirmation during invite creation? Product wants confirmation to be enabled by default, so this would save us from having to add that logic on the TS client.

quexten: Sure!

crates/bitwarden-organization-invite-link/src/invite_link_client.rs
quexten: Yes! I can add the collection name here.

crates/bitwarden-organization-invite-link/src/invite_link_client.rs
```diff
@@ -85,6 +99,73 @@ impl InviteLinkClient {
...
+    /// Whether invite confirmation is supported. Currently always `false`.
+    // TODO(Milestone 3): delegate to the invite key bundle once confirmation is implemented.
+    pub fn supports_confirmation(&self) -> bool {
+        false
+    }
+
+    /// Enables confirmation on the given `invite`, returning the updated invite.
+    // TODO(Milestone 3): implement confirmation via the invite key bundle. Currently a stub that
+    // returns the invite unchanged.
+    pub fn enable_confirmation(
```
JimmyVo16, line 112: Will operations like `enable_confirmation` and `disable_confirmation` be idempotent?

PR #1261, opened 2026-07-14
[PM-30183] feat: add Salt newtype to bitwarden-crypto

ike-kottlowski:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-30183

## 📔 Objective

Adds a `Salt` newtype to `bitwarden-crypto` for better safety. `Salt::new` cleans and validates input from untrusted sources (e.g. values crossing the FFI boundary) by trimming whitespace, lowercasing, and rejecting empty results; `From<&str>`/`From<String>` wrap trusted values as-is. Also wires up the corresponding UniFFI custom type conversion and a new `CryptoError::InvalidSalt` variant.

Note: this PR only includes the `Salt` type itself and its UniFFI binding. There is no implementation yet.


crates/bitwarden-crypto/src/uniffi_support.rs
```diff
@@ -67,10 +65,11 @@ uniffi::custom_type!(PasswordProtectedKeyEnvelope, String, {
     lower: |obj| obj.into(),
 });
 
-uniffi::custom_type!(SecretProtectedKeyEnvelope, String, {
-    remote,
-    try_lift: |val| convert_result(SecretProtectedKeyEnvelope::from_str(&val)),
-    lower: |obj| obj.into(),
+uniffi::custom_type!(Salt, String, {
+    try_lift: |val| {
+        convert_result(Salt::new(val.as_str()))
```
quexten, line 72: Instead of `new` I'd probably prefer `from_string`. We should also make it pub(crate) so that creating a new string is not possible in the general case. We want them to either `from_email` or `make()` (random, not yet used).

crates/bitwarden-crypto/src/keys/salt.rs
```diff
@@ -0,0 +1,108 @@
...
+    /// Returns the salt as a string slice, e.g. to pass into a KDF function.
+    pub fn as_str(&self) -> &str {
+        &self.0
+    }
+
+    /// Constructs a [`Salt`] from untrusted input, cleaning it by trimming whitespace and
+    /// lowercasing it, and rejecting the value if the cleaned result is empty. Use this
+    /// constructor for salts originating outside the SDK (e.g. values crossing the FFI
+    /// boundary).
+    pub fn new(value: &str) -> Result<Self, CryptoError> {
```
quexten, line 32: This should probably be from_email. I don't know that I agree with the approach of always converting it. I think we should only ever convert it if we are getting it from a source that we know could be an email. Otherwise we would have to make sure that the random format (whatever we choose there) does not get broken by the transformation. If we just base64'd random bytes, the transformation would currently break correctness.

crates/bitwarden-auth/src/login/login_via_password/password_prelogin_response.rs
```diff
@@ -23,7 +23,6 @@ pub struct PasswordPreloginResponse {
     pub salt: String,
 }
 
-// TODO: PM-28143: tuple is a work around for the time being until PM-28143 removes the fallback
 impl TryFrom<(PasswordPreloginResponseModel, &str)> for PasswordPreloginResponse {
```
quexten, line 27: `&str` can have multiple meanings here. Maybe, we should make an "email" newtype, where you are asserting to the typesystem that something is an email. Something like: 
````
struct Email(String);
impl Email {
  fn from(email: &str) {..}
}

impl TryFrom<(PasswordPreloginResponseModel, Email)> {...}
```

that way, you cannot accidentally tryfrom(PasswordPreloginRespnseModel, raw salt), where raw salt should *Not* be converted to lowercase / was read from state or API.

crates/bitwarden-auth/src/login/login_via_password/password_prelogin_response.rs
quexten, line 53: When does salt ever not unwrap?

crates/bitwarden-crypto/src/keys/salt.rs
```diff
@@ -0,0 +1,108 @@
...
+/// is to give the salt its own type, so that the compiler can catch mistakes such as passing a
+/// salt where a password or raw email string was expected, or the reverse. Because
+/// [`PartialEq`]/[`Eq`] are derived structurally, mixing up two [`Salt`]s (for example a stale
+/// salt with a fresh one) is still possible - this type only protects against confusing a
+/// [`Salt`] with a completely unrelated [`String`].
+#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(transparent)]
+#[cfg_attr(feature = "wasm", derive(Tsify))]
+#[cfg_attr(feature = "wasm", tsify(into_wasm_abi, from_wasm_abi))]
+pub struct Salt(#[cfg_attr(feature = "wasm", tsify(type = "string"))] String);
```
quexten, line 20: Do we need this on mobile?

Wes hould make the contract as such:
The string inside is always a valid salt. It shall NOT be processed further. You implement From<Email>, and From<string>. From<Email> to-lower and trims the email to get a salt. From<string> just takes the raw string under the assumption that it is a valid salt.

The salt returned by the server should always From<string> correctly and should already be lowered / trimmed.

quexten (changes requested):

crates/bitwarden-crypto/src/keys/salt.rs
ike-kottlowski: Agreed. Which is sort of the tight rope we're walking here. right now, the salt is always defaulted to `email.tolower.trim` https://github.com/bitwarden/server/blob/480d88a3840eec0df97d19fe0ca833a526ceba21/src/Core/Entities/User.cs#L119 

So, the ceremony around lowering and trimming is more about defense-in-depth. It does feel as though it is getting convoluted.

crates/bitwarden-auth/src/login/login_via_password/password_prelogin_response.rs
ike-kottlowski: I think I interpreted "unwrap" to be similar to `salt = response?.salt ?? email.lower.trim` 🤔

quexten: Ah, OK, on server it is optional too. I hadn't realized that https://github.com/bitwarden/server/blob/f3192aa58595af25f3c177890c49e07a99385706/src/Identity/Models/Response/Accounts/PasswordPreloginResponseModel.cs#L36. We should consider making it non-optional.

crates/bitwarden-crypto/src/keys/salt.rs
quexten: Right, but if we make it so that the future salt is just a random b64 string, then lowering it will break it. So we would need to choose an encoding that survives being lowered + trimmed...

I'd much rather have a clean hand-off point from known emails to known static salts. Does that not work?

crates/bitwarden-crypto/src/keys/salt.rs
```diff
@@ -0,0 +1,108 @@
+use serde::{Deserialize, Serialize};
+#[cfg(feature = "wasm")]
+use tsify::Tsify;
+
+use crate::CryptoError;
+
+/// The salt used together with a user's password in a Key Derivation Function (KDF, see [`Kdf`](
+/// crate::Kdf)) to derive a [`MasterKey`](crate::MasterKey).
```
quexten, line 8: Note (no action needed): Eventually, the master key will be removed entirely, and only be present briefly during derivation of the authentication data / unlock data.

crates/bitwarden-crypto/src/keys/salt.rs
```diff
@@ -0,0 +1,108 @@
...
+///
+/// This is a thin wrapper around a [`String`], not an encrypted or hashed value. Its only purpose
```
quexten, line 10: Could we add into the comment "The salt MUST always be lower-case."

crates/bitwarden-crypto/src/keys/salt.rs
```diff
@@ -0,0 +1,108 @@
...
+        if cleaned.is_empty() {
+            return Err(CryptoError::InvalidSalt);
+        }
+        Ok(Self(cleaned))
+    }
+}
+
+impl From<String> for Salt {
+    fn from(value: String) -> Self {
+        Self(value)
```
quexten, line 43: Should we lower+trim here too? 🤔

crates/bitwarden-crypto/src/keys/salt.rs
```diff
@@ -0,0 +1,108 @@
...
+    }
+}
+
+impl From<&str> for Salt {
+    fn from(value: &str) -> Self {
+        Self(value.to_owned())
```
quexten, line 49: Should we lower + trim here too?

crates/bitwarden-crypto/src/keys/salt.rs
```diff
@@ -0,0 +1,108 @@
...
+    #[test]
+    fn serializes_as_a_plain_string() {
+        let salt = Salt::from("salt-value");
+        let serialized = serde_json::to_string(&salt).unwrap();
+        assert_eq!(serialized, "\"salt-value\"");
+
+        let deserialized: Salt = serde_json::from_str(&serialized).unwrap();
+        assert_eq!(deserialized, salt);
+    }
+
```
quexten, line 92: Probably should add a test case for an email?

quexten (approved):
Some smaller comments, but approving in general. As per DM discussion, salts will always be lower-cased values. They are already returned as lower-cased values from the server, but from our perspective the risk of introducing upper-case salts at this point is too high, in case they pass through an interface which transforms them to lower-case.

If we add random salt generation in the future, the allowed alphabet *MUST* survive the to-lower+trim transformation.

PR #1263, opened 2026-07-15
refactor(crypto): implement AES256-CBC, AES256-CBC-HMAC256 hazmat primitives

quexten:
Implements aes256-cbc (as legacy compat for old user-keys), and AES256-CBC-HMAC-SHA256 in the hazmat module. This will replace the `aes.rs` file at the crate root. These implementations are close to being zero-copy. Only the plaintext buffer that is passed out is allocated.

https://bitwarden.atlassian.net/browse/PM-40756

### DIFF
--- a/crates/bitwarden-crypto/src/hazmat/symmetric_encryption/mod.rs
+++ b/crates/bitwarden-crypto/src/hazmat/symmetric_encryption/mod.rs
@@ -10,12 +10,37 @@
 //! envelope or data envelope) instead.
 
 use crate::CryptoError;
+#[allow(dead_code)]
+pub(crate) mod aes256_cbc;
+
+#[allow(dead_code)]
+#[derive(Debug, PartialEq, Eq)]
+pub(crate) enum SymmetricEncryptionError {
+    /// The input is malformed — wrong length, invalid padding, or otherwise not a valid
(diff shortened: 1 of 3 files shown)


crates/bitwarden-crypto/src/hazmat/symmetric_encryption/mod.rs
```diff
@@ -10,13 +10,38 @@
...
+    /// ciphertext for the cipher.
+    FormatWrong,
+    /// The integrity check (MAC / authentication tag) failed; the ciphertext, associated data,
+    /// nonce/IV, and key do not match.
+    IntegrityCheckFailed,
+}
+
+impl From<SymmetricEncryptionError> for CryptoError {
+    fn from(_: SymmetricEncryptionError) -> Self {
+        CryptoError::KeyDecrypt
```
quexten, line 29: Will update in the future. Please note, mac should *ALWAYS* be checked before the padding, otherwise we may allow a padding oracle.

quexten:
Despite the AEAD bits living on this stack, do not feel the need to wait for it. The refactor can merge even without the aead stuff.

eligrubb (approved):

PR #1264, opened 2026-07-15
[PM-38857] Add support for `XAES-256-GCM` to SymmetricCryptoKey, Enstring, and the KeyStoreContext

eligrubb:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-38857

## 📔 Objective

In order to support future initiatives, `XAES-256-GCM` is being added as a new encryption algorithm option. For initial support, this PR adds a new `XAes256GcmKey` to `SymmetricCryptoKey` and adds support for encrypting and decrypting EncStrings with this new key and algorithm type. XAES-256-GCM EncStrings use the COSE serialization, reusing the `EncString` type introduced with `XChaCha20Poly1305`, using the algorithm value in the COSE protected header to differentiate. Support is also added for `XAES-256-GCM` as a symmetric key algorithm in the `KeyStoreContext`. 

This PR does not include support for using `XAES-256-GCM` for user keys or as the wrapping key in SAFE primitives. That will be addressed in separate, follow-up, work.

This PR adds a new dependency, the `xaes-256-gcm` crate. This dependency has already been through the AppSec review process for new dependencies. See VULN-655 (https://bitwarden.atlassian.net/browse/VULN-655) for details on the approval.

## 🚨 Breaking Changes

Adds new `XAes256GcmKey` option to public `SymmetricCryptoKey` enum.


dani-garcia (dismissed):

crates/bitwarden-core/src/key_management/crypto/reinit_user_crypto.rs
```diff
@@ -121,6 +121,10 @@ pub(crate) async fn reinit_user_crypto(
                 error!("Unexpected AES-256-GCM user key during reinit_user_crypto");
                 return Err(ReinitUserCryptoError::CryptoInitialization);
             }
+            SymmetricKeyAlgorithm::XAes256Gcm => {
```
quexten: I think this should be the same as the xchacha20-poly1305 arm, not the aes arm, since a v2 user - now xaes - would have an xaes user key after upgrade and we want to skip?

crates/bitwarden-core/src/key_management/crypto_client.rs
```diff
@@ -344,6 +344,7 @@ impl CryptoClient {
                 Err(CryptoClientError::UpgradeTokenRequired)
             }
             (SymmetricKeyAlgorithm::Aes256Gcm, _) => Err(CryptoClientError::InvalidKeyType),
+            (SymmetricKeyAlgorithm::XAes256Gcm, _) => Err(CryptoClientError::InvalidKeyType),
```
quexten: This also should be the same as xchacha not aes, i think?

crates/bitwarden-crypto/src/cose/mod.rs
```diff
@@ -30,6 +31,12 @@ use crate::{
 /// to be able to randomly generate nonces, and to not have to worry about key wearout. Since
 /// the draft was never published as an RFC, we use a private-use value for the algorithm.
 pub(crate) const XCHACHA20_POLY1305: i64 = -70000;
+/// XAES-256-GCM (<https://c2sp.org/XAES-256-GCM>) extended-nonce AEAD. Given
```
quexten, line 34: Nice, I like that we have a link here given the lack of rfc.

crates/bitwarden-crypto/src/cose/symmetric.rs
```diff
@@ -546,6 +653,85 @@ pub(crate) fn decrypt_xchacha20_poly1305(
...
+    key: &XAes256GcmKey,
+    content_format: ContentFormat,
+) -> Result<CoseEncrypt0Bytes, CryptoError> {
+    let mut plaintext = plaintext.to_vec();
+    let mut protected_header: Header = HeaderBuilder::from(content_format)
+        .key_id(key.key_id.as_slice().to_vec())
+        .build();
+    protected_header.alg = Some(Algorithm::PrivateUse(XAES_256_GCM));
+
+    if should_pad_content(&content_format) {
```
quexten, line 681:
```suggestion
    // GCM as a stream cipher does not have a block size. We want to hide exact 
    // input plaintext length, and pad the plaintext size, if the input is a string.
    if should_pad_content(&content_format) {
```

crates/bitwarden-crypto/src/keys/symmetric_crypto_key.rs
```diff
@@ -237,8 +240,47 @@ impl PartialEq for Aes256GcmKey {
     }
 }
 
-/// A borrowed view over a symmetric key that is encoded as a COSE key and used as the
```
quexten, line 240: I think this comment was accidentally removed?

crates/bitwarden-crypto/src/keys/symmetric_crypto_key.rs
```diff
@@ -437,23 +499,24 @@ impl SymmetricCryptoKey {
...
 
-    /// Returns a [`CoseKeyView`] for the COSE-key symmetric variants (AES-256-GCM,
-    /// XChaCha20-Poly1305), or `None` for the legacy AES-CBC variants.
+    /// Returns a [`CoseKeyView`] for symmetric key variants accepted by safe constructs.
+    /// XAES-256-GCM and legacy AES-CBC variants return `None`.
     pub(crate) fn as_cose_key_view(&self) -> Option<CoseKeyView<'_>> {
         match self {
             Self::Aes256GcmKey(k) => Some(CoseKeyView::Aes256Gcm(k)),
             Self::XChaCha20Poly1305Key(k) => Some(CoseKeyView::XChaCha20Poly1305(k)),
+            Self::XAes256GcmKey(_) => None,
```
quexten: I think this should have a cose key view?

quexten (changes requested):
Mostly looks good, there are a few instances where the code makes xaes behave like aes, but I think it should behave like xchacha, given that it's replacing xchacha.

crates/bitwarden-core/src/key_management/crypto/reinit_user_crypto.rs
eligrubb: Good catch, switched to match behavior of the `XChaCha20Poly1305` branch.

crates/bitwarden-core/src/key_management/crypto_client.rs
eligrubb: Same here, switched to match behavior of `XChaCha20Poly1305` branch.

crates/bitwarden-crypto/src/keys/symmetric_crypto_key.rs
eligrubb: Good catch, comment has been restored.

crates/bitwarden-crypto/src/keys/symmetric_crypto_key.rs
eligrubb: Yes, you are right, I've added `Xaes256GcmKey` to `CoseKeyView`. I also added a check to `encrypt_cose` and `encrypt_cose0` to ensure that the declared algorithm in the protected header matches the algorithm provided by any `cek`. This is required by [RFC 9052 Section 7.1](https://www.rfc-editor.org/info/rfc9052#section-7.1): "the application MUST verify that [the algorithm in the protected header] matches the algorithm for which the key is being used. If the algorithms do not match, then this key object MUST NOT be used to perform the cryptographic operation.". 

I did this by replacing the optional `default_algorithm` parameter in `decrypt_cose` and `decrypt_cose0` with a new enum `CoseAlgorithmPolicy` that allows callers to optionally provide a default, a content encryption key algorithm, or require the header provide the algorithm. Let me know if the design is confusing or you have a different solution in mind.

crates/bitwarden-crypto/src/safe/data_envelope.rs
```diff
@@ -224,11 +226,14 @@ impl DataEnvelope {
...
-        // Decrypt the message. `decrypt_cose0` recovers the content-encryption algorithm from the
-        // protected header (AES-256-GCM for new envelopes, XChaCha20-Poly1305 for legacy ones), so
-        // no decryption fallback is needed.
-        let decrypted_message = decrypt_cose0(&msg, None, cek.key_bytes())
-            .map_err(|_| DataEnvelopeError::Decryption)?;
+        // Bind the protected content-encryption algorithm to the independently typed CEK before
+        // attempting decryption. DataEnvelope has no legacy format that omits the algorithm.
+        let decrypted_message = decrypt_cose0(
+            &msg,
+            CoseAlgorithmPolicy::Exactly(cek.algorithm()),
```
quexten, line 233: Totally makes sense, good catch on the alg mismatch. One alternative would be to pass in the cose key view, and have the cose key view track the alg too; that way we are not passing in two separate objects that depend on each other in a sense, but it's just one distinct item.

crates/bitwarden-crypto/src/safe/symmetric_key_envelope.rs
```diff
@@ -137,12 +139,11 @@ impl SymmetricKeyEnvelope {
...
 
-        // Decrypt the key bytes. `decrypt_cose0` also validates that the declared
-        // content-encryption algorithm matches the cipher before attempting decryption. The
-        // envelope always declares the algorithm, so no decryption fallback is needed.
+        // The wrapping key is independently typed as XChaCha20-Poly1305, so require the protected
+        // content-encryption algorithm to match it before attempting decryption.
         let key_bytes = decrypt_cose0(
             &self.cose_encrypt0,
-            None,
+            CoseAlgorithmPolicy::Exactly(CoseContentEncryptionAlgorithm::XChaCha20Poly1305),
```
quexten, line 146: Just confirming, this is handled in a follow-up ticket right? (Supporting xaes-gcm).

crates/bitwarden-core/src/key_management/crypto/reinit_user_crypto.rs
```diff
@@ -189,16 +189,32 @@ mod tests {
             .register_bridge(Box::new(InMemoryStateBridge::default()));
     }
 
+    /// Make an XAES-256-GCM user key until normal V2 account initialization supports it.
```
quexten, line 192: thought: (don't need to address this): We have two patterns, sometimes we generate keys on the fly, sometimes we load them from const strings. I'm wondering whether we should enforce a consistent pattern (const strings?)...

crates/bitwarden-crypto/src/cose/symmetric.rs
```diff
@@ -28,46 +29,70 @@ fn should_pad_content(format: &ContentFormat) -> bool {
...
     XChaCha20Poly1305,
 }
 
+/// Policy for resolving and validating a COSE message's content-encryption algorithm.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub(crate) enum CoseAlgorithmPolicy {
+    /// Require a supported algorithm in the protected header.
+    RequireProtectedHeaderAlgorithm,
+    /// Require the protected algorithm to exactly match the independently known CEK algorithm.
+    Exactly(CoseContentEncryptionAlgorithm),
```
quexten, line 50: I like that we are representing these behaviors as enums!

crates/bitwarden-crypto/src/cose/symmetric.rs
```diff
@@ -797,4 +1254,154 @@ mod tests {
...
+            .ciphertext(message.ciphertext.unwrap())
+            .build()
+            .to_vec()
+            .unwrap()
+            .into()
+    }
+
+    #[test]
+    fn test_xaes256_gcm_roundtrip_content_formats() {
+        for content_format in [
```
quexten, line 1277: Nice; for xchacha we seem to have 4 separate tests, this is more consise. I wonder if macro has macro-generated parameterized tests for cases like this...

quexten (approved):

dani-garcia (approved):

crates/bitwarden-crypto/src/safe/symmetric_key_envelope.rs
eligrubb: Yes, follow-up PR will adjust the default for `SymmetricKeyEnvelope`

crates/bitwarden-crypto/src/safe/data_envelope.rs
eligrubb: This would mean that all Safe primitives would need to probably be built around CoseKeyView, which actually sounds intriguing to me. Something for future tech debt cleanup.

crates/bitwarden-core/src/key_management/crypto/reinit_user_crypto.rs
eligrubb: That makes sense to me, I think that consistently using const strings is a good pattern. IMO a follow-up tech debt ticket cleaning up all KM owned tests to use const strings would be useful.

PR #1265, opened 2026-07-15
[PM-40277] Key rotation corrupts attachments, fido2 credentials

mzieniukbw:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-40277

## 📔 Objective

SDK key rotation will currently corrupt fido2 credentials, attachments. This is because the cipherview has encrypted fields on it, namely fido2 credentials, attachment keys, which just get passed through.

Adds `CipherView::generate_cipher_key_from`, which takes the source key id explicitly (`generate_cipher_key` now delegates to it with `User` key id slot), and has the keyless cipher rotation path pass the real current key so the FIDO2 credentials and attachments are re-wrapped correctly.
Included test coverage that covers all cipher key rotation with fido2 and attachments possibilities.

## 🚨 Breaking Changes

None expected.
Android breaking change failure is unrelated to this PR.


crates/bitwarden-vault/src/cipher/cipher.rs
```diff
@@ -892,14 +892,34 @@ impl Cipher {
     }
 }
 impl CipherView {
-    #[allow(missing_docs)]
+    /// Generates a fresh per-item cipher key and re-wraps the cipher's attachment and FIDO2
+    /// sub-keys under it. The existing sub-keys are assumed to be wrapped under
+    /// [`self.key_identifier()`](IdentifyKey::key_identifier).
     pub fn generate_cipher_key(
```
quexten: It's great that we have this behavior documented now, but could we name the function more clearly? 
`reencrypt_to_cipher_key_encryption` or similar?

crates/bitwarden-vault/src/cipher/cipher.rs
```diff
@@ -892,14 +892,34 @@ impl Cipher {
...
+    /// attachment and FIDO2 sub-keys using an explicitly supplied `source_key` rather than
+    /// deriving it from [`IdentifyKey::key_identifier`]. Use this when the sub-keys are wrapped
+    /// under a key other than the cipher's identifier — e.g. during key rotation, where they are
+    /// under the current user key rather than the [`SymmetricKeySlotId::User`] slot.
+    ///
+    /// * `source_key` - The key the current attachment/FIDO2 sub-keys are wrapped under. For a
+    ///   keyless cipher this is the current user (or organization) key.
+    /// * `wrapping_key` - The key the freshly generated cipher key will be wrapped under (during
+    ///   rotation, the new user key).
+    pub fn generate_cipher_key_from(
```
quexten: ```upgrade_to_cipher_key_encryption_with_external_key```

quexten (dismissed):
In my opinion, we should re-structure the vault item encryption. It still feels very error-prone and brittle, but that shouldn't be part of this work.

crates/bitwarden-vault/src/cipher/cipher.rs
gbubemismith: @mzieniukbw I agree with @quexten here, I think it makes with easier to follow

gbubemismith (dismissed):

quexten (dismissed):

gbubemismith (dismissed):

quexten (dismissed):

gbubemismith (dismissed):

quexten (approved):

gbubemismith (approved):

PR #1273, opened 2026-07-17
feat(crypto): add safe crypto based invites

quexten:
Re-implements invites based on safe primitives. The cryptographic design is:
Rewrite the invite based on safe primitives. The construction is:

```
InviteSecret->SecretProtectedKeyEnvelope->InviteKey->DataEnvelope -> InviteSecret
                                            ^     |             |--> OrgPubKeyPrint
                                            |     |
OrganizationKey -> EncString ---------------|     ->SymmetricKeyEnvelope->OrganizationKey
```

The organization key still has access to the Invite Secret (via the invite key). The Invite Secret has access to the pub key print in an authenticated fashion, and to the organization key (if present) for milestone 3 - autoconfirmation.

### DIFF
--- a/crates/bitwarden-crypto/examples/protect_key_with_secret.rs
+++ b/crates/bitwarden-crypto/examples/protect_key_with_secret.rs
@@ -7,10 +7,8 @@
 //! bytes) and uses a cheap KDF.
 
 use bitwarden_crypto::{
-    KeyStore, KeyStoreContext, SymmetricKeyAlgorithm, key_slot_ids,
-    safe::{
-        HighEntropySecret, SecretProtectedKeyEnvelope, SecretProtectedKeyEnvelopeError,
-        SecretProtectedKeyEnvelopeNamespace,
+    KeyStore, KeyStoreContext, SymmetricKeyAlgorithm, key_slot_ids, safe::{
+        ContentEncryptionKey, HighEntropySecret, SecretProtectedKeyEnvelope, SecretProtectedKeyEnvelopeError, SecretProtectedKeyEnvelopeNamespace,
     },
 };
 
@@ -28,7 +26,7 @@ fn main() {
 
     // Alice has some data protected with a symmetric key. She wants the symmetric key protected
     // with a high-entropy secret (here, 16 random bytes).
-    let data_key = ctx.make_symmetric_key(SymmetricKeyAlgorithm::Aes256CbcHmac);
+    let data_key = ContentEncryptionKey::make(&mut ctx); 
     let secret = HighEntropySecret::make(16).expect("16 bytes is a valid size");
 
     // Seal the key with the secret.
@@ -37,6 +35,7 @@ fn main() {
     let envelope = SecretProtectedKeyEnvelope::seal(
         data_key,
         &secret,
+        // The namespace must be replaced with an appropriate namespace for the use-case
         SecretProtectedKeyEnvelopeNamespace::OrganizationInvite,
         &ctx,
     )
@@ -66,16 +65,18 @@ fn main() {
         .reseal(
             &secret,
             &new_secret,
+            // The namespace must be replaced with an appropriate namespace for the use-case
             SecretProtectedKeyEnvelopeNamespace::OrganizationInvite,
         )
         .expect("The secret should be valid");
     disk.save("data_key_envelope", (&envelope).into());
 
     // Alice wants to change the protected key. This requires creating a new envelope
-    let data_key = ctx.make_symmetric_key(SymmetricKeyAlgorithm::Aes256CbcHmac);
+    let data_key = ContentEncryptionKey::make(&mut ctx);
     let envelope = SecretProtectedKeyEnvelope::seal(
         data_key,
         &new_secret,
+        // The namespace must be replaced with an appropriate namespace for the use-case
         SecretProtectedKeyEnvelopeNamespace::OrganizationInvite,
         &ctx,
     )
@@ -87,6 +88,7 @@ fn main() {
     assert!(matches!(
         envelope.unseal(
             &wrong_secret,
+            // The namespace must be replaced with an appropriate namespace for the use-case
             SecretProtectedKeyEnvelopeNamespace::OrganizationInvite,
             &mut ctx
         ),
--- a/crates/bitwarden-crypto/examples/protect_key_with_secret.rs
+++ b/crates/bitwarden-crypto/examples/protect_key_with_secret.rs
@@ -22,7 +22,9 @@ fn main() {
     // Alice wants to protect a key with a high-entropy secret.
     // For example to:
     // - Protect a send with a random URL fragment secret
-    // - Protect a key with another (derived) key
+    //   - Here, The secret key envelope wraps a KEK, the KEK wraps a CEK, and the CEK wraps the data.
+    // - Protect a key with a (derived) secret, external to the SDK
+    //   - Here, the secret key envelope wraps a the key
     // For this, the `SecretProtectedKeyEnvelope` is used.
     // (For low-entropy secrets such as a PIN, use the `PasswordProtectedKeyEnvelope` instead.)
 
--- a/crates/bitwarden-crypto/examples/protect_key_with_secret.rs
+++ b/crates/bitwarden-crypto/examples/protect_key_with_secret.rs
@@ -22,7 +22,8 @@ fn main() {
     // Alice wants to protect a key with a high-entropy secret.
     // For example to:
     // - Protect a send with a random URL fragment secret
-    //   - Here, The secret key envelope wraps a KEK, the KEK wraps a CEK, and the CEK wraps the data.
+    //   - Here, The secret key envelope wraps a KEK, the KEK wraps a CEK, and the CEK wraps the
+    //     data.
     // - Protect a key with a (derived) secret, external to the SDK
     //   - Here, the secret key envelope wraps a the key
     // For this, the `SecretProtectedKeyEnvelope` is used.
@@ -58,6 +59,7 @@ fn main() {
     let _unsealed_data_key = deserialized
         .unseal(
             &secret,
+            // The namespace must be replaced with an appropriate namespace for the use-case
             SecretProtectedKeyEnvelopeNamespace::OrganizationInvite,
             &mut ctx,
         )
--- a/crates/bitwarden-crypto/examples/protect_key_with_secret.rs
+++ b/crates/bitwarden-crypto/examples/protect_key_with_secret.rs
@@ -7,8 +7,10 @@
 //! bytes) and uses a cheap KDF.
 
 use bitwarden_crypto::{
-    KeyStore, KeyStoreContext, SymmetricKeyAlgorithm, key_slot_ids, safe::{
-        ContentEncryptionKey, HighEntropySecret, SecretProtectedKeyEnvelope, SecretProtectedKeyEnvelopeError, SecretProtectedKeyEnvelopeNamespace,
+    KeyStore, KeyStoreContext, SymmetricKeyAlgorithm, key_slot_ids,
+    safe::{
+        ContentEncryptionKey, HighEntropySecret, SecretProtectedKeyEnvelope,
+        SecretProtectedKeyEnvelopeError, SecretProtectedKeyEnvelopeNamespace,
     },
 };
 
@@ -26,7 +28,7 @@ fn main() {
 
     // Alice has some data protected with a symmetric key. She wants the symmetric key protected
     // with a high-entropy secret (here, 16 random bytes).
-    let data_key = ContentEncryptionKey::make(&mut ctx); 
+    let data_key = ContentEncryptionKey::make(&mut ctx);
     let secret = HighEntropySecret::make(16).expect("16 bytes is a valid size");
 
     // Seal the key with the secret.
--- a/crates/bitwarden-crypto/examples/protect_key_with_secret.rs
+++ b/crates/bitwarden-crypto/examples/protect_key_with_secret.rs
@@ -7,7 +7,7 @@
 //! bytes) and uses a cheap KDF.
 
 use bitwarden_crypto::{
-    KeyStore, KeyStoreContext, SymmetricKeyAlgorithm, key_slot_ids,
+    KeyStore, KeyStoreContext, key_slot_ids,
     safe::{
         ContentEncryptionKey, HighEntropySecret, SecretProtectedKeyEnvelope,
         SecretProtectedKeyEnvelopeError, SecretProtectedKeyEnvelopeNamespace,
--- a/crates/bitwarden-crypto/examples/protect_key_with_secret.rs
+++ b/crates/bitwarden-crypto/examples/protect_key_with_secret.rs
@@ -37,7 +37,7 @@ fn main() {
     let envelope = SecretProtectedKeyEnvelope::seal(
         data_key,
         &secret,
-        SecretProtectedKeyEnvelopeNamespace::ExampleUse,
+        SecretProtectedKeyEnvelopeNamespace::OrganizationInvite,
         &ctx,
     )
     .expect("Sealing should work");
@@ -55,7 +55,7 @@ fn main() {
     let _unsealed_data_key = deserialized
         .unseal(
             &secret,
-            SecretProtectedKeyEnvelopeNamespace::ExampleUse,
+            SecretProtectedKeyEnvelopeNamespace::OrganizationInvite,
             &mut ctx,
         )
         .expect("Unsealing should work");
@@ -66,7 +66,7 @@ fn main() {
         .reseal(
             &secret,
             &new_secret,
-            SecretProtectedKeyEnvelopeNamespace::ExampleUse,
+            SecretProtectedKeyEnvelopeNamespace::OrganizationInvite,
         )
         .expect("The secret should be valid");
     disk.save("data_key_envelope", (&envelope).into());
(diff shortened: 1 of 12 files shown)


crates/bitwarden-crypto/src/safe/symmetric_key_envelope.rs
```diff
@@ -48,10 +47,34 @@ pub enum SymmetricKeyEnvelopeError {
...
 /// A symmetric key protected by another symmetric key
+#[derive(Clone)]
 pub struct SymmetricKeyEnvelope {
     cose_encrypt0: coset::CoseEncrypt0,
 }
 
+/// Resolves the content-encryption algorithm, wrapping-key id, and raw wrapping-key bytes for a
+/// symmetric key used to wrap another key in a [`SymmetricKeyEnvelope`]. Both XChaCha20-Poly1305
+/// and AES-256-GCM keys are supported as wrapping keys.
+fn wrapping_key_material(
```
quexten: Most likely need to update this after  https://github.com/bitwarden/sdk-internal/pull/1264 merges.

crates/bitwarden-organization-crypto/src/invite.rs
```diff
@@ -0,0 +1,868 @@
...
...{"invite_key_wrapped_invite_data":"g1hHpQEDA3gjYXBwbGljYXRpb24veC5iaXR3YXJkZW4uY2Jvci1wYWRkZWQEULAvkABNEIqBWSbJbR5w/PU6AAE4gQI6AAE4gAKhBUzu06Qb8Rqlag8wxIVYyAOEYjj8vngz+6efiYs4aJqpGKMDNYgXM7QEYDH85RfElcQ3tkBm5lNsLWy04I2gFaJV6hGRN8fZSG9Br3Q11yUIkTna6wy112uCn/szWnrceGg+tTUKjErl9FmkN6yzoSskR1U10dqVeG7AJRjzkAgZFHol1JycPVxhuJdszrF/HpNEOKUbw94vA8kVRQV7+JWJKgeUki0gb9kHPpE+P2vxtM7YJU5uhQXukKVg2KXkNTX7BwXDZ1aofIkFHzSzMtRuRVrEUjAn","invite_secret_wrapped_invite_key":"hFgopQEDAxhlOgABFVxQsC+QAE0QioFZJsltHnD89ToAATiBBjoAATiAAaEFTH8P6s7widuCRd5gsVhQ8Qdi4roT2jBcAb9Qn6f4pJL/BKGvWTrN3D9wvK3bT/AOgEcoiP3gZPZ8qvCiFBgErQ+IudaHnPxw1/W2QXvvZGbkaa1vnqL8eIPuzWrGVVyBg0CiASkzWCAIfgngqsWpnHP3zFYH88LcFD13xgvwzmmN0nirJ6BUZPY=","invite_key_wrapped_organization_key":"g1hGpQEDA3giYXBwbGljYXRpb24veC5iaXR3YXJkZW4ubGVnYWN5LWtleQRQsC+QAE0QioFZJsltHnD89ToAATiBAzoAATiAAqEFTMKnS2tiogG0nXwKFVhQvhXuJ7yorAyslqDLYYjUblYqKryq4aNOS9nGpHkJd+39Uo8cYblbqkBs6x7degHq+bj80RQUoNGpWBLGOPNkIP6tqMnAxQy0NR6sIm7trkI=","organization_key_wrapped_invite_key":"2.xpuLCrOF/Ai7q+zGn5ftSQ==|AyGdogR5TKqIO9nbABdtb57uJ/G/6J5Cj4AajWjWppe6ZJSfqt65fAmgT954+Qgf10ah9L+34jW27yOVitwXMEq/+7b8FaAUDJoh89Yi36Q=|BK4onlURPXFoaUt48UoUeD1BhFREYHHs1tuY6MoR7lQ="}"#;
```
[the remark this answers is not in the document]
quexten: @AikidoSec ignore: This is a test vector!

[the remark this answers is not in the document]
quexten: @AikidoSec ignore: This is a test vector!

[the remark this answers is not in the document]
quexten: @AikidoSec ignore: This is a test vector!

crates/bitwarden-organization-crypto/src/invite.rs
```diff
@@ -0,0 +1,868 @@
...
+
+        // Generate the URL-fragment invite secret (goes in the invite link).
+        let mut bytes = Zeroizing::new([0u8; INVITE_SECRET_LEN]);
+        bitwarden_random::rng().fill_bytes(bytes.as_mut_slice());
+        let invite_secret = InviteSecret(bytes);
+
+        // Generate the AES-256-GCM invite key (the hub). It is used for exactly two AES-256-GCM
+        // messages below (the invite-data DataEnvelope and the org-key SymmetricKeyEnvelope), each
+        // with a fresh random nonce, so there is no nonce-reuse concern.
+        let invite_key = ctx.make_symmetric_key(SymmetricKeyAlgorithm::Aes256Gcm);
```
quexten: Todo: Replace after xaesgcm support merges.

crates/bitwarden-crypto/examples/protect_key_with_secret.rs
```diff
@@ -37,7 +37,7 @@ fn main() {
     let envelope = SecretProtectedKeyEnvelope::seal(
         data_key,
         &secret,
-        SecretProtectedKeyEnvelopeNamespace::ExampleUse,
+        SecretProtectedKeyEnvelopeNamespace::OrganizationInvite,
```
mzieniukbw, line 44: ❓ Could we use `SecretProtectedKeyEnvelopeNamespace::ExampleNamespace` ?
(Which is currently conditioned to`#[cfg(test)]`, but in practise examples are run by CI, so the are like tests and example code for others at the same time)

crates/bitwarden-organization-crypto/src/invite.rs
```diff
@@ -0,0 +1,868 @@
...
+const TS_INVITE: &'static str = r#"
+export type Invite = Tagged<string, "Invite">;
+"#;
+
+/// Cryptographic invite for an organization, built around an AES-256-GCM invite key that acts as
+/// the hub tying everything together. See the module-level docs for the overall diagram.
+#[derive(Clone)]
+pub struct Invite {
+    /// The `InviteData` (org public-key thumbprint + invite secret) sealed with the invite key.
+    invite_key_wrapped_invite_data: DataEnvelope,
```
mzieniukbw: ❓ This does not sound correct, the key **seals** the data. I understood wrapping is key encrypting another key
```suggestion
    invite_key_sealed_invite_data: DataEnvelope,
```

crates/bitwarden-organization-crypto/src/invite.rs
```diff
@@ -0,0 +1,868 @@
...
+    /// The invite key sealed with the high-entropy invite secret, letting an invitee (who holds
+    /// only the invite secret from the link) recover the invite key.
+    invite_secret_wrapped_invite_key: SecretProtectedKeyEnvelope,
```
mzieniukbw: ❓ Similarly here, we use term wrapped, which does not sound correct.
```suggestion
    invite_secret_sealed_invite_key: SecretProtectedKeyEnvelope,
```

crates/bitwarden-organization-crypto/src/invite.rs
```diff
@@ -0,0 +1,868 @@
...
+    /// The organization key sealed with the invite key, letting a redeeming invitee recover the
```
mzieniukbw, line 205: ❓ Wrapped sounds more correct, since both are keys.

crates/bitwarden-organization-crypto/src/invite.rs
```diff
@@ -0,0 +1,868 @@
...
+    /// confirm the invitee out of band.
+    invite_key_wrapped_organization_key: Option<SymmetricKeyEnvelope>,
+    /// The invite key wrapped with the organization key, letting anyone holding the organization
+    /// key recover the invite key (and thus the invite data).
+    organization_key_wrapped_invite_key: EncString,
+}
+
+/// Wire format for [`Invite`]. This is what's serialized by serde (as JSON).
+#[derive(Serialize, Deserialize)]
+struct InviteWire {
```
mzieniukbw: ❓ Why do we need this, if the Invite is identical and the InviteWire is just intermediate struct between String and Invite.

crates/bitwarden-organization-crypto/src/invite.rs
quexten: You are right, `safe` should consistently use `seal` instead of `wrap`, `encrypted`, `encapsulated`.

crates/bitwarden-organization-crypto/src/invite.rs
```diff
@@ -0,0 +1,849 @@
...
...XIhQaz6yh8eyA653m4DJEHV00JjdQ/jSfTEfyDf6LiflGntDZ7gSYlQMRxuS71yVP9Mgm79yFY4aIw703G/HUjvIEt4XlWCLzvaLDmegpDz9z4eQCVk8v8JM1fS7BsA","invite_key_sealed_invite_data_cek":"g1g+pgE6AAEReQMYZQRQ7raDdes1HnAgiuiZWXyS9ToAARVcUPw60HtcnAwO6kRKd7MnQz86AAE4gQM6AAE4gAKhBVgYvxSL4XPZLwEXHPnEWweqeeKjyH9r85/ZWE3rsZy1MLmzkFOCJa8a0o20b1P3IH6c5NYtT3v1a50+X0GmgDKiMf/omjhwRRQ7ua6wK0O2JBlhM4JIE8jmtsyMd46mFDbURK1kGRvl7Q==","invite_secret_sealed_invite_key":"hFgopQEDAxhlOgABFVxQ7raDdes1HnAgiuiZWXyS9ToAATiBBjoAATiAAaEFTJvCToVTNSe1uWbD5FhUf061kAAj1sMGWjJ8IWPV2e2xk5Z+ownZ1RqUXG2jv9h2vnsEFZ2yglvKaSCGTsOuvOXv0ESwQk6eUtFAaZjxV/Rajyuu6YgEa712Tfb9Jz13QLGngYNAogEpM1gg1MsuVokUsK8WkqggHRdJ4jvzFsbN/bP4g0l+j6f97qb2","invite_key_sealed_organization_key":"g1hKpQE6AAEReQN4ImFwcGxpY2F0aW9uL3guYml0d2FyZGVuLmxlZ2FjeS1rZXkEUO62g3XrNR5wIIromVl8kvU6AAE4gQM6AAE4gAKhBVgYoth3hg+yUTLXF4ksaeT42IWGKuTv27B4WFAKV8Z7uKGZNHOONGgZQLbQMozgYX9tseuet413M314W0sV3cBKZIRSEZfj3NvHU8tE/6b2oGxPQIPKP3Tyhl84zhI4uG+Mo5WKkvtdPonD0A==","organization_key_sealed_invite_key":"2.XjZXnAwXK/cXCmHNgCPQzw==|WEGa37JPdNWVnMPWlfinyvZdXpMW8kpTBypXPWf/abbG4/+6vp/WJQmVlkVEflEkuZwjKWSkMWJPcACGBoabeBHgzqpkSYf3kyQb7VhePHQ=|kZLDCLGm1nCJhlVuNahFiZOecy5tKG2fXGCNZZzbDjk="}"#;
```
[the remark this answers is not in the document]
quexten: @AikidoSec ignore: Test vector.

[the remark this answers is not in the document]
quexten: @AikidoSec ignore: Test vector.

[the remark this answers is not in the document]
quexten: @AikidoSec ignore: Test vector.

crates/bitwarden-crypto/examples/protect_key_with_secret.rs
quexten: Examples are not run by CI actually. Had to find this out recently :(
If we want an example, we should make it clear that it should not be used. Maybe an extra: "ExampleDoNotUseThisInProdcutionCode"?

quexten: I chose to not change this for now, and just add comments above the examples.

crates/bitwarden-organization-crypto/src/invite.rs
quexten: I'll consistently use seal. I think the mnemonic of Envelope + seal / unseal / reseal is very easy to understand. Publicly facing to other teams we should not use: Encrypt, wrap, encapsulate. Those should be KM internal words.

crates/bitwarden-crypto/examples/protect_key_with_secret.rs
```diff
@@ -28,7 +28,7 @@ fn main() {
 
     // Alice has some data protected with a symmetric key. She wants the symmetric key protected
     // with a high-entropy secret (here, 16 random bytes).
-    let data_key = ctx.make_symmetric_key(SymmetricKeyAlgorithm::Aes256CbcHmac);
+    let data_key = ContentEncryptionKey::make(&mut ctx);
```
mzieniukbw, line 34: ❓ The "data" implies we need to use CEK, but few sentences above it, there is an example of how you can use for protecting other key, which makes it KEK:
> Protect a key with another (derived) key

This might give wrong impression based on example that you use CEK, without much thinking.
Can we do something about this to prevent misgauges ?

crates/bitwarden-crypto/examples/protect_key_with_secret.rs
```diff
@@ -55,7 +56,7 @@ fn main() {
     let _unsealed_data_key = deserialized
         .unseal(
             &secret,
-            SecretProtectedKeyEnvelopeNamespace::ExampleUse,
+            SecretProtectedKeyEnvelopeNamespace::OrganizationInvite,
```
mzieniukbw, line 63: ❓ Do we want to add the same comment ?
> // The namespace must be replaced with an appropriate namespace for the use-case

crates/bitwarden-organization-crypto/src/invite.rs
```diff
@@ -0,0 +1,849 @@
...
+export type Invite = Tagged<string, "Invite">;
+"#;
+
+/// Cryptographic invite for an organization, built around an XAES-256-GCM invite key that acts as
+/// the hub tying everything together. See the module-level docs for the overall diagram.
+#[derive(Clone, Serialize, Deserialize)]
+pub struct Invite {
+    /// The `InviteData` (org public-key thumbprint + invite secret) sealed with its own
+    /// content-encryption key (CEK).
+    invite_key_sealed_invite_data: DataEnvelope,
```
mzieniukbw: This name is confusing, since this is data representation. It is not sealed by invite key and is technically `invite_data_cek_sealed_invite_data`. What you are describing here is the process to get to invite key. You could also simply say `sealed_invite_data`, since normally sealing DataEnvelope is always accompanied by CEK key. (as of obvious you need `invite_data_cek` to unseal it ?)

crates/bitwarden-organization-crypto/src/invite.rs
```diff
@@ -0,0 +1,849 @@
...
+        self.invite_key_sealed_organization_key = Some(envelope);
+        Ok(())
+    }
+
+    /// Disables confirmation
+    pub fn disable_confirmation(&mut self) {
+        self.invite_key_sealed_organization_key = None;
+    }
+
+    /// Generates a brand new invite around a fresh AES-256-GCM invite key, binding it to the
```
mzieniukbw: KEK is XAES now
```suggestion
    /// Generates a brand new invite around a fresh XAES-256-GCM invite key, binding it to the
```

crates/bitwarden-organization-crypto/src/invite.rs
```diff
@@ -0,0 +1,849 @@
...
+        // with; a different key makes `make_for_private_key` fail before building the invite.
+        let wrong_org_key = ctx.generate_symmetric_key();
+        assert!(matches!(
+            Invite::make_for_private_key(wrong_org_key, &wrapped_private_key, &mut ctx),
+            Err(InviteKeyBundleError::InvalidPrivateKey)
+        ));
+    }
+
+    #[test]
+    fn test_into_base64_url() {
```
mzieniukbw: I think this is better:
```suggestion
    fn test_invite_secret_into_base64_url() {
```

crates/bitwarden-organization-crypto/src/invite.rs
```diff
@@ -0,0 +1,849 @@
...
+    fn test_debug() {
+        let key_store = KeyStore::<TestIds>::default();
+        let mut ctx = key_store.context_mut();
+        let wrapped_private_key = load_test_vectors(&mut ctx);
+
+        let (invite_secret, invite) =
+            Invite::make_for_private_key(TestSymmKey::Organization, &wrapped_private_key, &mut ctx)
+                .unwrap();
+        // Exercises both the `InviteSecret` and `Invite` `Debug` impls.
+        println!("{invite_secret:?}");
```
mzieniukbw, line 819: ❓ How about we make it a normal test with `format!` ?

mzieniukbw (changes requested):

crates/bitwarden-crypto/examples/protect_key_with_secret.rs
quexten: Good point! Updated with more clear examples.

crates/bitwarden-organization-crypto/src/invite.rs
quexten: Nice catch. `sealed_invite_data` i'll go with this.

crates/bitwarden-organization-crypto/src/invite.rs
quexten: We could, but this is the standard set so far in e.g. the safe module. The one thing I'll note is that if we do that, then we need to seed the RNG, in the bitwarden-random module during the test, because otherwise the invite produced is different every time.

quexten: (i'll out of scope it for this PR if that's ok)

mzieniukbw (approved):

PR #1274, opened 2026-07-17
feat(invite-link): re-structure invite-link-client to use and expose new invite link api

quexten:

Re-structure the invite link clients to use the new cryptographic api for the invite, and to expose the new API. HTTP requests are moved into the SDK. Two are left open for AC to move as a follow-up once they are implemented.


crates/bitwarden-wasm-internal/integration-tests/tests/invite/invite-link.test.ts
```diff
@@ -1,64 +0,0 @@
-import { OrganizationId } from "@bitwarden/sdk-internal";
```
quexten, line 1: Can't mock API requests for TS currently so we have to remove the tests again.

crates/bitwarden-organization-invite-link/src/invite_link_client.rs
```diff
@@ -1,95 +1,276 @@
...
+
+    /// Accepts an organization invite for the current user, optionally enrolling into account
+    /// recovery (when `enroll_into_account_recovery` is set) and — when the invite supports
+    /// confirmation — self-confirming.
+    pub async fn accept_and_optionally_confirm(
+        &self,
+        organization_id: OrganizationId,
+        code: String,
+        invite_secret: InviteSecret,
+        default_collection_name: String,
```
JaredSnider-Bitwarden, line 146: @quexten , `default_collection_name` is required but it is only used for the optional confirmation flow (milestone 3)?

crates/bitwarden-organization-invite-link/src/invite_link_client.rs
```diff
@@ -1,95 +1,276 @@
...
-#[cfg_attr(feature = "wasm", derive(Tsify), tsify(into_wasm_abi, from_wasm_abi))]
-#[serde(rename_all = "camelCase")]
-pub struct OrganizationInviteCryptoBundle {
-    /// Raw invite key. CRITICAL: MUST NOT be sent to the server.
-    #[cfg_attr(feature = "wasm", tsify(type = "InviteKeyData"))]
-    pub invite_key: InviteKeyData,
-    /// Invite key sealed with the organization key. Safe to send to the server.
-    #[cfg_attr(feature = "wasm", tsify(type = "Invite"))]
-    pub invite: Invite,
+pub enum InviteLinkError {
```
JaredSnider-Bitwarden, line 30: Per designs, I'm supposed to map any known server errors into translated errors (see my work in https://github.com/bitwarden/clients/pull/21574/changes#diff-6ce7a97def70673e98645baafcbf899749afaefecfd745af08133767ebcb586f). This looks like it is going to give `Api` flavored errors with an internal message which I guess I can interrogate for now. 

Note: I did get approval from product to just show a single translated, generic error if need be so we can improve this error contract in a follow up.  I think future milestones actually introduce error codes from the server vs string parsing which also improves this as well (cc: @eliykat am I right on the error codes approach? I think I saw it somewhere for milestone 3?)

crates/bitwarden-organization-invite-link/src/invite_link_client.rs
eliykat: Even then it's optional - should be an `Option`.

crates/bitwarden-organization-invite-link/src/invite_link_client.rs
eliykat: Yes I believe milestone 3 has proper ProblemDetail responses with codes so you don't have to string match. This is what Jimmy is clarifying in the [Slack thread](https://bitwarden.slack.com/archives/C0BJBL6M3DH/p1784661204598029?thread_ts=1784580196.243959&cid=C0BJBL6M3DH) about this.

crates/bitwarden-organization-invite-link/src/invite_link_client.rs
quexten: I think I disagree actually. `default_collection_name` is passed in because the SDK does not have i18n support. It is not a variable function parameter but a dependency. Dependencies should not be injected on-demand.

In other words, whether or not the name is needed is functionality that's encapsulated by the invite client. The function is explicitly called "accept_and_optionally_confirm", and hides the details of whether confirmation happens from the caller. If we did make it optional then:
1. You would have to first check "supports confirmation" before every call to know whether to pass it in. This is just more cumbersome
2. This would make illegal states / usage representable that was not before. What happens when you pass in none, but it is a link that requires confirmation? I guess another error. But why even allow that?

I would recommend to for this milestone pass in `""` given that it is unused. For the next milestone, i would pass in the default collection name always.

crates/bitwarden-organization-invite-link/src/invite_link_client.rs
quexten: If AC wants to offer parsing of the API responses, please add it to the PR stack. I'm not sure I'm qualified to encode those rules, I' don't know what the struct looks like / what variants there are.

crates/bitwarden-organization-invite-link/src/invite_link_client.rs
eliykat: My comment was unclear - I meant it's optional in the business sense. A default user collection ("My Items") is only created if the relevant policy is enabled, so there are times when it's not required even for the confirmation flow.

That said, the server already does these checks, and it would simplify the interface if you supply it unconditionally and then the server can choose to use it or not - keeping these checks to the back-end only, and minimizing the knowledge the caller needs.

So I'm OK with your recommendation.

mzieniukbw (dismissed):

mzieniukbw (dismissed):

eliykat:
@quexten I've made a minimal set of changes to:
- replace the stubbed network calls with real implementations.
- align this with what the native code expects on the TS side. The main differences are:
	- create (and the new refresh function) need to return the whole invite link response so that we can populate the admin-facing form
	- get_invite_secret is not expected to be stateful, so I've changed that back to accepting the invite in the parameter - changing this would trigger too much refactoring on the TS side for now.

I have a more complete PR that moves the whole service layer into the SDK, including the other network calls (get, delete, etc), but those changes are too extensive for the timeframes we're working in. That means there is some temporary glue code on the TS side to make this work. However, I will polish and submit that for review as a follow-up ticket so that we can get the whole feature into the SDK.

Two outstanding TODOs:
- I will regenerate server bindings once https://github.com/bitwarden/server/pull/8058 is merged, which will fix a compilation error
- I am getting one error when testing, which appears to be related to the FFI boundary for `get_invite_secret`:

```
Unhandled error in angular RuntimeError: unreachable executed
    passStringToWasm0 bitwarden_wasm_internal_bg.js:8926
    get_invite_secret bitwarden_wasm_internal_bg.js:2611
    reconstructUrl default-organization-invite-link.service.ts:100
    map map.ts:58
```

Can you please assist?

Reproduction steps if required:
- run against client PR: https://github.com/bitwarden/clients/pull/22054
- `pm-32497-generate-invite-link` feature flag on
- navigate to web app -> Admin Console -> Members -> Invite Member -> By link tab, then enter any domain and click save

crates/bitwarden-wasm-internal/integration-tests/tests/invite/invite-link.test.ts
quexten: NVM, added tests.

crates/bitwarden-wasm-internal/integration-tests/tests/org-fixtures.ts
```diff
@@ -23,3 +23,26 @@ export const TEST_ORGANIZATION_ID =
...
...7GVsa+JanExuT/8W/BKgTZGgp9ptGz98i0hOVLSBbIkOABDNs3ONmxilICzf9jipZsGFjwWpBXJgjAiK5qDHnlYTe6/ASdjvywwbiJjnwr+","invite_key_sealed_invite_data_cek":"g1g+pgE6AAEReQMYZQRQeHT1pCPgaB9CYG+Rb8scAjoAARVcUOeB/QFwHUwsvdEq0eCp5Us6AAE4gQM6AAE4gAKhBVgYnr1DARLvnvtw1bsYXa2gueder+FZz2D4WE1gkmCx7GnYs+54fw+OeTJhutaOhz9gd7lwfRj5V0tGlP3yNdOUxLyyBDN/+L1qQzjbo3Utm17Wfl1ZrVbBZcFsQzue0/Zq0Lk8+/h+cA==","invite_secret_sealed_invite_key":"hFgopQEDAxhlOgABFVxQeHT1pCPgaB9CYG+Rb8scAjoAATiBBjoAATiAAaEFTB07pVNo3oYWy0o/XlhUQA/hxa3AsrA2OnLDQE8ffqZ/J29CXGv3YbPqOgIEXWPWOKEo9SZW+TbhBi6q3Wf0EewTdeWADozr80J79rJTxTfXKS2yNCmT+fjIjJcI0ov8HpnlgYNAogEpM1ggvtgywem/2QA8V1q/wgKtv6Rk1jLDpbZrbvE5+efMGkL2","invite_key_sealed_organization_key":"g1hKpQE6AAEReQN4ImFwcGxpY2F0aW9uL3guYml0d2FyZGVuLmxlZ2FjeS1rZXkEUHh09aQj4GgfQmBvkW/LHAI6AAE4gQM6AAE4gAKhBVgYQ5qCc8T8brREcCQFDzA2QhqjeTjoHtl2WFBuAdHlH509FDCxbqgjPUM56nEXVawZFQSpotSWXQsCYAz/MUce7LREwgkUZ5f3x9y4CpkvNTPkmEGTw675emwwpR05x6gJ0aL/UoyoF4bSpw==","organization_key_sealed_invite_key":"2.YnutjO6dOLfu/FzUXmdpaw==|nQnSQSO5EjB/sq6aMEK3bcEY9PQdLTBmelj1ftsMP7tiWUeoS6RTLKzg4kv8qggUgx+xh+XjBxULwdYJeV3k+pChTrnTRogClph2m3N9RtQ=|/dsbX2NfWpSVFPTHa5HQxBT7b9EdlwVlEgjJzyq3SvA="}' as unknown as Invite;
```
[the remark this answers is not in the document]
quexten: @AikidoSec ignore: Test vector

crates/bitwarden-wasm-internal/integration-tests/tests/org-fixtures.ts
```diff
@@ -23,3 +23,26 @@ export const TEST_ORGANIZATION_ID =
...
+
+export const TEST_INVITE_SECRET =
+  "oSpRu_AENo_n11TqiLIryrJKaEPbQVjbdd6r4reyqUU" as unknown as InviteSecret;
```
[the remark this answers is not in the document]
quexten: @AikidoSec ignore: Test vector

eliykat (approved):
Approving CODEOWNERS changes and @quexten 's initial changes to AC code. I've asked another AC engineer to review as well.

crates/bitwarden-organization-invite-link/src/organization_invite_link.rs
```diff
@@ -0,0 +1,49 @@
...
+        Ok(Self {
+            id: require!(response.id),
+            code: require!(response.code),
+            organization_id: OrganizationId::new(require!(response.organization_id)),
+            allowed_domains: response.allowed_domains.unwrap_or_default(),
+            invite: require!(response.invite).parse()?,
+            supports_confirmation: response.supports_confirmation.unwrap_or(false),
+            creation_date: require!(response.creation_date)
+                .parse()
+                .map_err(|_| MissingFieldError("creation_date"))?,
```
mzieniukbw, line 46: ❓ Isn't this more like malformed ?

dereknance (approved):
✅ Platform: `crates/bitwarden-wasm-internal/integration-tests/tests/`

mzieniukbw (approved):

crates/bitwarden-organization-invite-link/src/invite_link_client.rs
BTreston, line 1: 🎨❓ Do we think its worth also moving the delete functionality into the SDK for consistency? There are no crypto operations on it currently, so it doesn't have to. But it would be nice to have all the CRUD operations live in the SDK client. It would make invite link completely portable across clients if we so desired. Perhaps a follow up?

BTreston (approved):
This all seems good to me, one non-blocking question for @eliykat

crates/bitwarden-organization-invite-link/src/organization_invite_link.rs
quexten: @eliykat Would you mind taking a look? Since this is nit I merged since this is VFO work with a tight deadline. / it lives in your code.

PR #1276, opened 2026-07-17
[PM-39754] Switch SymmetricKeyEnvelope default algorithm

eligrubb:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-39754

## 📔 Objective

Switch SymmetricKeyEnvelope default algorithm to use XAES-256-GCM instead of XChaCha20Poly1305. This also changes the default algorithm used by SessionKey to XAES-256-GCM. 

## 🚨 Breaking Changes

Neither of these constructions are used in production today, there is no need to handle migration or backward compatibility. SessionKey is used by the in-progress `bw` cli rewrite, but API surface is unchanged, no breaking changes or compatibility issues should be introduced.

### DIFF
--- a/crates/bitwarden-crypto/examples/protect_key_with_key.rs
+++ b/crates/bitwarden-crypto/examples/protect_key_with_key.rs
@@ -2,7 +2,7 @@
 //! [`SymmetrickeyEnvelope`].
 
 use bitwarden_crypto::{
-    KeyStore, KeyStoreContext, key_slot_ids,
+    KeyStore, KeyStoreContext, SymmetricKeyAlgorithm, key_slot_ids,
     safe::{SymmetricKeyEnvelope, SymmetricKeyEnvelopeError, SymmetricKeyEnvelopeNamespace},
 };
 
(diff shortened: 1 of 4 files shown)


crates/bitwarden-crypto/examples/protect_key_with_key.rs
```diff
@@ -14,7 +14,7 @@ fn main() {
     // Alice has a vault key and wants to protect it with another key.
     // For example, this can be used to persist the vault key while keeping it encrypted at rest.
     let vault_key = ctx.generate_symmetric_key();
-    let wrapping_key = ctx.generate_symmetric_key();
```
eligrubb, line 17: `.generate_symmetric_key()` actually creates a `cbc_hmac` key, so this example currently panics on main when sealing is attempted, expecting an `XChaCha20Poly1305` key. 

`cargo run -p bitwarden-crypto --example protect_key_with_key` runs successfully with this branch. Is there a way that examples could be run along with other tests to catch this in the future?

quexten: I'm surprised we do not yet... Will look into this, examples should be verified in the CI.

quexten (approved):

crates/bitwarden-crypto/examples/protect_key_with_key.rs
quexten: https://bitwarden.slack.com/archives/C054ZQSBS49/p1784532471391819

PR #1279, opened 2026-07-20
[PM-40158] Make XAES-256-GCM the default for v2 user keys

eligrubb:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-40158

## 📔 Objective

In order to use FIPS-compliant algorithms, this PR changes the default user key encryption algorithm to use XAES-256-GCM instead of XChaCha20-Poly1305. User key algorithm is assigned when an account is created or upgraded.

This PR updates the default for:
- user key creation
  - `WrappedAccountCryptographicState::make` in `crates/bitwarden-core/src/key_management/account_cryptographic_state.rs`
- v1 -> v2 user key upgrade rotation
  -   `make_rotation_context` in `crates/bitwarden-user-crypto-management/src/key_rotation/rotation_context.rs`
- relevant encryption v2 user tests

## 🚨 Breaking Changes

Encryption V2 is not currently rolled out and there are no v2 users in production, we do not need to handle backward compatibility for this change. There are some dev and QA test accounts that will need to be updated or regenerated after this change.


dani-garcia (dismissed):
Platform changes lgtm

crates/bitwarden-core/src/client/internal.rs
```diff
@@ -280,10 +280,7 @@ impl InternalClient {
                     .unwrap_v2(user_key_id, &mut ctx)
                     .map_err(|_| EncryptionSettingsError::InvalidUpgradeToken)?
             }
-            (
-                SymmetricCryptoKey::XChaCha20Poly1305Key(_) | SymmetricCryptoKey::XAes256GcmKey(_),
-                Some(_),
-            ) => {
+            (SymmetricCryptoKey::XAes256GcmKey(_), Some(_)) => {
```
quexten, line 283: RIP existing xchacha accounts.

crates/bitwarden-core/src/key_management/crypto/reinit_user_crypto.rs
```diff
@@ -166,10 +169,10 @@ mod tests {
...
     const TEST_VECTOR_SIGNED_PUBLIC_KEY_V2: &str = "hFgepAEnAxg8BFAmkP0QgfdMVbIujX55W/yNOgABOH8BoFkBTqNpYWxnb3JpdGhtAG1jb250ZW50Rm9ybWF0AGlwdWJsaWNLZXlZASYwggEiMA0GCSqGSIb3DQEBAQUAA4IBDwAwggEKAoIBAQDP/7WM8nUepxoJ0qtM+azxcly+eZ31qUjjZTZcX/gYw1MzkoXWAjqyeFH/bdktq1lEUwegrxkIxKkY2SMtp0CvPnaV1x5O8E6FBSJbKWRlDg181rfEhgm5tc6aR4PJ827IvFVm9xk6Sj091P5DHZDEOsWLZc2jYjtpUV3X38I4gSR7HiYnR4DcwcWkoJ3FhtxMCwYgPz6RVH0vzhLUmm1mgbzH6IH8Pf9DjLTZSxBikVO7S9s9jzhiZbTeeAl3FbNLxfj9Qkj+NoSfms7jGVTlBwvSXgjJs/ktGkT1cR5QcBMpU4bt41+l73MN8pXapCih9Awf1W+RY7imxpYOMFJ3AgMBAAFYQMq/hT4wod2w8xyoM7D86ctuLNX4ZRo+jRHf2sZfaO7QsvonG/ZYuNKF5fq8wpxMRjfoMvnY2TTShbgzLrW8BA4=";
```
quexten, line 174: Note: We also switched ed25519 -> mldsa. We may want to update the test vector (but not necessarily here).

crates/bitwarden-crypto/src/keys/symmetric_crypto_key.rs
```diff
@@ -88,7 +88,7 @@ impl TryFrom<wasm_bindgen::JsValue> for SymmetricCryptoKey {
 pub enum SymmetricKeyAlgorithm {
     /// Used for V1 user keys and data encryption
     Aes256CbcHmac,
-    /// Used for V2 user keys and data envelopes
+    /// Used by legacy COSE-based envelopes
```
quexten: "Old PIN PasswordProtectedKeyEnvelopes"

quexten (dismissed):
Only a comment nit

crates/bitwarden-core/src/client/internal.rs
eligrubb: xchacha you are the weakest link.

eligrubb:
Fixed comment @quexten pointed out. No other changes since last review.

quexten (approved):

dani-garcia (approved):

PR #1280, opened 2026-07-21
fix(passkey): fido2 assertion overwrites creation date of passkey

quexten:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-40658

## 📔 Objective

Probably related to passkey corruption bugs. Updating creds on usage should not update the creation date. This is just misleading.

### DIFF
--- a/crates/bitwarden-fido/src/lib.rs
+++ b/crates/bitwarden-fido/src/lib.rs
@@ -172,7 +172,7 @@ pub fn fill_with_credential(
         user_name: view.user_name.clone(),
         user_display_name: view.user_display_name.clone(),
         discoverable: "true".to_owned(),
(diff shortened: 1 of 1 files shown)


crates/bitwarden-fido/src/lib.rs
```diff
@@ -172,7 +172,7 @@ pub fn fill_with_credential(
         user_name: view.user_name.clone(),
         user_display_name: view.user_display_name.clone(),
         discoverable: "true".to_owned(),
-        creation_date: chrono::offset::Utc::now(),
```
quexten, line 175: I just want to note that the rest of this code unconditionally setting values may lead to bugs in the future.

dani-garcia (approved):

PR #1286, opened 2026-07-21
docs(crypto): clarify encryptable contract

quexten:
Encryptables currently have mis-usages, which lead to easy bugs. For example, encrypt(decrypt(encrypt(K1), K1), K2) may lead to an item partially being encrypted by K1, and partially by K2. The next decrypt will fail. This is entirely unexpceted and a footgun. This PR adds clarification on the intent (contract) of the traits, and tags usages that are violating the contract.


eligrubb (approved):
Love it!

nikwithak (approved):

PR #1295, opened 2026-07-21
Auth/PM-40520 - Registration - Open Org Invite Seal / Unseal Support 

JaredSnider-Bitwarden:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-40520
Tech breakdown: https://github.com/bitwarden/tech-breakdowns/tree/main/auth/PM-39706-open-invite-registration-crossing

## 📔 Objective

To provide a simple API on the Auth registration client to seal and unseal the new open organization invite link data so it can be safely carried through the registration with email verification process.


crates/bitwarden-auth/src/registration/seal_open_org_invite_data.rs
```diff
@@ -0,0 +1,216 @@
...
+    open_org_invite_data::{RegistrationOpenOrgInviteData, RegistrationOpenOrgInviteDataV1},
+    registration_client::{RegistrationClient, RegistrationError},
+};
+
+/// The plaintext invite context that a registrant will consume on the verification-email tab to
+/// complete the open-organization-invite acceptance. All three fields are required.
+#[cfg_attr(feature = "wasm", derive(Tsify), tsify(into_wasm_abi, from_wasm_abi))]
+#[cfg_attr(feature = "uniffi", derive(uniffi::Record))]
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq)]
+pub struct OpenOrgInviteData {
```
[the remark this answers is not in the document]
JaredSnider-Bitwarden: Claude here: Good catch — verified that `LoginRequest` (and every send-access / login-via-password / webauthn-prf model) uses `#[serde(rename_all = "camelCase")]`; only `JitMasterPasswordRegistrationRequest` and its `post_keys_for_*` siblings are the outlier, and that's the pattern I originally mirrored. Since nothing has shipped yet, aligning `OpenOrgInviteData` on camelCase in e843bb4b2.

crates/bitwarden-auth/src/registration/seal_open_org_invite_data.rs
```diff
@@ -0,0 +1,216 @@
...
+    /// pre-sealed form.
+    pub invite_key: String,
+}
+
+/// The output of [`RegistrationClient::seal_open_org_invite_data`]. Both fields are opaque wire
+/// artifacts safe to transport at layer boundaries; both are needed to unseal.
+#[cfg_attr(feature = "wasm", derive(Tsify), tsify(into_wasm_abi, from_wasm_abi))]
+#[cfg_attr(feature = "uniffi", derive(uniffi::Record))]
+#[derive(Serialize, Deserialize, Debug, Clone)]
+pub struct SealedOpenOrgInvite {
```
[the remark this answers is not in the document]
JaredSnider-Bitwarden: Claude here: Same treatment as `OpenOrgInviteData` — camelCased in e843bb4b2 so `sealed_data` / `high_entropy_secret` become `sealedData` / `highEntropySecret` on the WASM boundary.

crates/bitwarden-auth/src/registration/open_org_invite_data.rs
```diff
@@ -0,0 +1,41 @@
...
+
+use bitwarden_crypto::safe::{DataEnvelopeNamespace, SealableData, SealableVersionedData};
+use serde::{Deserialize, Serialize};
+
+/// Version 1 wire schema of the open-invite registration payload.
+///
+/// Once this shape ships it cannot be broken; if the fields change, add a new `V2` struct and
+/// register both variants on [`RegistrationOpenOrgInviteData`].
+#[derive(Serialize, Deserialize, Debug, PartialEq)]
+pub(super) struct RegistrationOpenOrgInviteDataV1 {
```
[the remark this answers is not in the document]
JaredSnider-Bitwarden: Claude here: Agreed — this is the CBOR-tagged wire format inside the sealed data, so aligning it now is basically free (no sealed blobs exist in the wild yet). Applied `#[serde(rename_all = "camelCase")]` to `RegistrationOpenOrgInviteDataV1` in e843bb4b2.

crates/bitwarden-crypto/src/safe/high_entropy_secret.rs
```diff
@@ -81,6 +80,30 @@ impl HighEntropySecret {
...
+    /// by [`HighEntropySecret::to_base64`] (or by the equivalent WASM ABI marshaling).
+    ///
+    /// The caller must guarantee that the decoded bytes originated from a genuine high-entropy
+    /// source. This constructor does not — and cannot — validate entropy on its own; feeding
+    /// low-entropy input to the cheap KDF that consumes this type will not provide the
+    /// brute-force resistance the KDF assumes.
+    pub fn from_base64(encoded: &str) -> Result<Self, HighEntropySecretError> {
+        let bytes = B64::try_from(encoded).map_err(|_| HighEntropySecretError::Malformed)?;
+        Ok(Self::from_internal(bytes.as_bytes()))
+    }
```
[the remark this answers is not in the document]
JaredSnider-Bitwarden: Claude here: Real security concern — thanks. Fixed in 8509fbe08: `from_base64` now applies the same `MIN_SECRET_LENGTH` floor `make(...)` uses, plus two guard tests (below-minimum rejected as `TooShort`, at minimum accepted). The doc comment is updated to clarify this is a length floor, not an entropy check — a 32-byte fixed ASCII string still passes, as documented.

crates/bitwarden-wasm-internal/integration-tests/tests/registration/open-org-invite-registration.test.ts
```diff
@@ -0,0 +1,63 @@
+import { makePasswordManagerClient, makeStateBridge } from "../utils";
+
+const SAMPLE_INPUT = {
+  organization_id: "1bc9ac1e-f5aa-45f2-94bf-b181009709b8",
+  invite_link_code: "abcd1234efgh5678",
+  invite_key: "raw-invite-key-material-base64url",
+};
```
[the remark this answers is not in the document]
JaredSnider-Bitwarden: Claude here: Updated in e843bb4b2 alongside the Rust struct rename.

crates/bitwarden-wasm-internal/integration-tests/tests/registration/open-org-invite-registration.test.ts
```diff
@@ -0,0 +1,63 @@
...
+  it("unseal_open_org_invite_data recovers the plaintext invite context with fields intact", async () => {
+    const client = makePasswordManagerClient(makeStateBridge());
+    const registration = client.auth().registration();
+
+    const sealed = await registration.seal_open_org_invite_data(SAMPLE_INPUT);
+    const unsealed = await registration.unseal_open_org_invite_data(sealed);
+
+    expect(unsealed.organization_id).toEqual(SAMPLE_INPUT.organization_id);
+    expect(unsealed.invite_link_code).toEqual(SAMPLE_INPUT.invite_link_code);
+    expect(unsealed.invite_key).toEqual(SAMPLE_INPUT.invite_key);
```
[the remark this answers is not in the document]
JaredSnider-Bitwarden: Claude here: Updated in e843bb4b2 — assertions now read `.organizationId`, `.inviteLinkCode`, `.inviteKey`.

crates/bitwarden-wasm-internal/integration-tests/tests/registration/open-org-invite-registration.test.ts
```diff
@@ -0,0 +1,63 @@
...
+    const sealedStr = sealed.sealed_data as unknown as string;
+    expect(sealedStr).not.toEqual("");
+    const b64 = sealedStr.replace(/-/g, "+").replace(/_/g, "/");
+    // Pad back to a multiple of 4 for Buffer.from("base64").
+    const padded = b64 + "=".repeat((4 - (b64.length % 4)) % 4);
+    const reencoded = Buffer.from(padded, "base64")
+      .toString("base64")
+      .replace(/\+/g, "-")
+      .replace(/\//g, "_")
+      .replace(/=+$/, "");
```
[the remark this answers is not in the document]
JaredSnider-Bitwarden: Claude here: You're right — Node 16+ supports `Buffer.from(str, "base64url")` natively and CI is on Node 20. Replaced the manual `+`/`-`, `/`/`_`, and padding substitutions with the native encoding in e843bb4b2.

crates/bitwarden-crypto/src/safe/high_entropy_secret.rs
```diff
@@ -96,6 +119,9 @@ pub enum HighEntropySecretError {
     /// The provided secret is too short to be used as a high-entropy secret.
     #[error("Secret is too short")]
     TooShort,
+    /// The provided string could not be decoded as standardized base64.
+    #[error("Secret is not valid base64")]
+    Malformed,
```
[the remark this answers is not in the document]
JaredSnider-Bitwarden: Claude here: Respectfully declining this one. `bitwarden-crypto` in `sdk-internal` isn't a public semver-stable crate, and both suggested alternatives don't quite fit:

- `bitwarden_encoding::NotB64EncodedError` only expresses decode-failure. Once the `MIN_SECRET_LENGTH` floor from the sibling review comment gets applied (see 8509fbe08), `from_base64` also has to signal *too-short-after-decode* — `NotB64EncodedError` can't represent that case.
- A dedicated parse-error type would give `from_base64` a different error surface than `make(...)`, which pushes the two-call-site consumer story in the wrong direction.

Keeping `HighEntropySecretError::Malformed`; the enum stays intentionally shared across constructors.

crates/bitwarden-crypto/src/safe/secret_protected_key_envelope.rs
```diff
@@ -668,6 +674,20 @@ mod tests {
         149, 151, 217, 94, 84, 67, 50, 107, 103, 74, 88, 72, 246,
     ];
 
+    #[test]
+    fn test_registration_open_org_invite_namespace_maps_to_expected_discriminant() {
+        // The wire discriminant is load-bearing once shipped; a regression here would silently
+        // invalidate every envelope sealed by production callers.
+        assert_eq!(
+            SecretProtectedKeyEnvelopeNamespace::try_from(3i128).unwrap(),
```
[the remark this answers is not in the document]
JaredSnider-Bitwarden: Claude here: Respectfully declining. Verified locally: `cargo clippy -p bitwarden-crypto --lib --tests -- -D warnings` and `cargo clippy -p bitwarden-auth --lib --tests -- -D warnings` both pass cleanly against the 58+ existing `.unwrap()` calls in the `#[cfg(test)]` modules of `data_envelope.rs`, `secret_protected_key_envelope.rs`, and `high_entropy_secret.rs` — including the new tests. `unwrap_used = "deny"` is set in the workspace `Cargo.toml`, but clippy exempts test cfg by design, and the codebase's convention throughout is to use `.unwrap()` in tests (every existing test in these files does). No lint failure to fix; leaving as-is for consistency.

crates/bitwarden-crypto/src/safe/data_envelope.rs
```diff
@@ -761,6 +767,47 @@ mod tests {
         assert!(matches!(result, Err(DataEnvelopeError::InvalidNamespace)));
     }
 
+    #[test]
+    fn test_registration_open_org_invite_namespace_maps_to_expected_discriminant() {
+        // The wire discriminant is load-bearing once shipped; a regression here would silently
+        // invalidate every envelope sealed by production callers.
+        assert_eq!(
+            DataEnvelopeNamespace::try_from(2i128).unwrap(),
```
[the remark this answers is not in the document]
JaredSnider-Bitwarden: Claude here: Respectfully declining. Verified locally: `cargo clippy -p bitwarden-crypto --lib --tests -- -D warnings` and `cargo clippy -p bitwarden-auth --lib --tests -- -D warnings` both pass cleanly against the 58+ existing `.unwrap()` calls in the `#[cfg(test)]` modules of `data_envelope.rs`, `secret_protected_key_envelope.rs`, and `high_entropy_secret.rs` — including the new tests. `unwrap_used = "deny"` is set in the workspace `Cargo.toml`, but clippy exempts test cfg by design, and the codebase's convention throughout is to use `.unwrap()` in tests (every existing test in these files does). No lint failure to fix; leaving as-is for consistency.

crates/bitwarden-crypto/src/safe/data_envelope.rs
```diff
@@ -761,6 +767,47 @@ mod tests {
...
+    fn test_registration_open_org_invite_data_rejects_cross_namespace_unseal() {
+        // AC 1.iii analogue: an envelope sealed under the new production namespace must not
+        // unseal under any other `DataEnvelopeNamespace` (here, the sibling production
+        // `VaultItem`). Mirrors `test_namespace_cross_contamination_protection` for the new
+        // variant.
+        let data: TestData = TestDataV1 { field: 42 }.into();
+
+        let (envelope, cek) =
+            DataEnvelope::seal_ref(&data, DataEnvelopeNamespace::RegistrationOpenOrgInviteData)
+                .unwrap();
```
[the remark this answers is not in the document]
JaredSnider-Bitwarden: Claude here: Respectfully declining. Verified locally: `cargo clippy -p bitwarden-crypto --lib --tests -- -D warnings` and `cargo clippy -p bitwarden-auth --lib --tests -- -D warnings` both pass cleanly against the 58+ existing `.unwrap()` calls in the `#[cfg(test)]` modules of `data_envelope.rs`, `secret_protected_key_envelope.rs`, and `high_entropy_secret.rs` — including the new tests. `unwrap_used = "deny"` is set in the workspace `Cargo.toml`, but clippy exempts test cfg by design, and the codebase's convention throughout is to use `.unwrap()` in tests (every existing test in these files does). No lint failure to fix; leaving as-is for consistency.

crates/bitwarden-crypto/src/safe/data_envelope.rs
```diff
@@ -761,6 +767,47 @@ mod tests {
...
+            Err(DataEnvelopeError::InvalidNamespace)
+        ));
+
+        // Correct namespace still succeeds.
+        let round_tripped: TestData = unseal_with_cek(
+            &envelope,
+            DataEnvelopeNamespace::RegistrationOpenOrgInviteData,
+            &cek,
+        )
+        .unwrap();
```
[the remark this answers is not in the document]
JaredSnider-Bitwarden: Claude here: Respectfully declining. Verified locally: `cargo clippy -p bitwarden-crypto --lib --tests -- -D warnings` and `cargo clippy -p bitwarden-auth --lib --tests -- -D warnings` both pass cleanly against the 58+ existing `.unwrap()` calls in the `#[cfg(test)]` modules of `data_envelope.rs`, `secret_protected_key_envelope.rs`, and `high_entropy_secret.rs` — including the new tests. `unwrap_used = "deny"` is set in the workspace `Cargo.toml`, but clippy exempts test cfg by design, and the codebase's convention throughout is to use `.unwrap()` in tests (every existing test in these files does). No lint failure to fix; leaving as-is for consistency.

crates/bitwarden-crypto/src/safe/high_entropy_secret.rs
```diff
@@ -213,4 +239,23 @@ mod tests {
             cloned.as_bytes().expose_owned()
         );
     }
+
+    #[test]
+    fn test_base64_round_trip_preserves_bytes() {
+        let secret = HighEntropySecret::make(32).unwrap();
+        let encoded = secret.to_base64();
+        let decoded = HighEntropySecret::from_base64(&encoded).unwrap();
```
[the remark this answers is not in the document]
JaredSnider-Bitwarden: Claude here: Respectfully declining. Verified locally: `cargo clippy -p bitwarden-crypto --lib --tests -- -D warnings` and `cargo clippy -p bitwarden-auth --lib --tests -- -D warnings` both pass cleanly against the 58+ existing `.unwrap()` calls in the `#[cfg(test)]` modules of `data_envelope.rs`, `secret_protected_key_envelope.rs`, and `high_entropy_secret.rs` — including the new tests. `unwrap_used = "deny"` is set in the workspace `Cargo.toml`, but clippy exempts test cfg by design, and the codebase's convention throughout is to use `.unwrap()` in tests (every existing test in these files does). No lint failure to fix; leaving as-is for consistency.

crates/bitwarden-auth/src/registration/open_org_invite/unseal.rs
```diff
@@ -0,0 +1,182 @@
...
+        assert_eq!(unsealed, input);
+    }
+
+    #[test]
+    fn unseal_fails_with_wrong_high_entropy_secret() {
+        let client = Client::new(None);
+        let registration_client = RegistrationClient::new(client);
+
+        let mut sealed = seal(&registration_client, sample_input());
+        let unrelated = HighEntropySecret::make(32).unwrap().to_base64();
```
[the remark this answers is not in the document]
JaredSnider-Bitwarden: Claude here: Respectfully declining. Verified locally: `cargo clippy -p bitwarden-crypto --lib --tests -- -D warnings` and `cargo clippy -p bitwarden-auth --lib --tests -- -D warnings` both pass cleanly against the 58+ existing `.unwrap()` calls in the `#[cfg(test)]` modules of `data_envelope.rs`, `secret_protected_key_envelope.rs`, and `high_entropy_secret.rs` — including the new tests. `unwrap_used = "deny"` is set in the workspace `Cargo.toml`, but clippy exempts test cfg by design, and the codebase's convention throughout is to use `.unwrap()` in tests (every existing test in these files does). No lint failure to fix; leaving as-is for consistency.

crates/bitwarden-auth/src/registration/open_org_invite/seal.rs
```diff
@@ -0,0 +1,218 @@
...
+    })
+}
+
+/// Internal CBOR wire schema for `sealed_data`. Byte arrays are the direct serializations of the
+/// respective envelope types (CBOR handles them as `bstr` natively), avoiding the
+/// double-base64 overhead of nesting the envelopes' own string forms.
+#[derive(Serialize, Deserialize)]
+pub(super) struct SealedEnvelopePair {
+    #[serde(rename = "d")]
+    pub(super) data_envelope: Vec<u8>,
```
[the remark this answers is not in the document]
JaredSnider-Bitwarden: Claude here: Great catch — confirmed. `serde`'s default `Vec<u8>` handling makes ciborium emit a CBOR array of integers (major type 0, one item per byte), not a `bstr`, so the round-trip worked but the doc comment was wrong about the wire shape and the encoded blob was roughly 2x what it needed to be.

Fixed in d539696de by adding `#[serde(with = "serde_bytes")]` to both fields, and rewrote the doc comment to describe why the attribute is load-bearing so future readers don't strip it thinking it's redundant. The round-trip test already proves correctness end-to-end.

The server's `MaxLength` cap (PM-40521) was sized for the pre-fix 1202-char output with a 3.4x buffer — after the fix it'll be smaller, so more headroom, no cap adjustment needed.

crates/bitwarden-auth/src/registration/open_org_invite_data.rs
```diff
@@ -0,0 +1,42 @@
+//! Versioned sealable payload for the open-organization-invite registration crossing.
+//!
+//! The payload structs live in their own file so both [`super::seal_open_org_invite_data`] and
+//! [`super::unseal_open_org_invite_data`] can reference them. The versioned enum matches what
+//! `bitwarden_crypto::generate_versioned_sealable!` would emit, expanded manually so it can be
+//! `pub(super)` and accessible across the sibling modules.
```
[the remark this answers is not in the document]
JaredSnider-Bitwarden: Claude here: Fair point — the trait docs are explicit about that. The reason I went manual was that the macro emits `enum $enum_name` with no visibility keyword, so the generated enum is module-private, and I have two sibling modules (`seal_open_org_invite_data` / `unseal_open_org_invite_data`) that both need to reference it.

Went with your first suggestion in bd2af6977: added a `$vis:vis` matcher in front of the enum name in the macro. Existing callers keep working (empty vis still matches), and new callers can pass any visibility — so `open_org_invite_data.rs` now uses `generate_versioned_sealable!(pub(super) RegistrationOpenOrgInviteData, ...)` and the ~15 lines of hand-written boilerplate are gone. Doc comment on the macro was updated to describe the new prefix too.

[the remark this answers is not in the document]
JaredSnider-Bitwarden: Ended up going with the other option you flagged ("or relocating the versioned type") in 4545a48027af8708d9ec1a3d0a429b83ae99232a, and reverted the macro visibility change from bd2af6977 as part of the same commit.

The three flat sibling files are now a single `open_org_invite/` module:

1. `mod.rs` — invokes `generate_versioned_sealable!` with no visibility prefix and re-exports the public API.
1. `wire_v1.rs` — the versioned wire schema struct.
1. `seal.rs` — public request/response types + seal path + shared framing helpers.
1. `unseal.rs` — unseal path + framing split helper.

With `seal` and `unseal` as descendants of `open_org_invite` (instead of top-level siblings), the versioned enum is module-private by default and still visible to both operations. That drops the need for a `pub(super)` output on the macro entirely, so the stock invocation from the trait docs is the one being used now. Net effect for this file: no manual `SealableVersionedData` impl, no macro extension, and the layout is closer to `bitwarden-vault/src/cipher/blob/` which uses the same shape.

crates/bitwarden-crypto/src/safe/data_envelope.rs
```diff
@@ -487,6 +487,11 @@ macro_rules! generate_versioned_sealable {
 pub enum DataEnvelopeNamespace {
     /// The namespace for vault items ("ciphers")
     VaultItem = 1,
+    /// Namespace for the invite context payload sealed on registration-start and unsealed on
+    /// registration-finish so an anonymous, email-verified new user can complete an open
+    /// organization invite in a single flow. The sealed struct carries the organization id,
+    /// invite link code, and invite key.
+    RegistrationOpenOrgInviteData = 2,
```
quexten: Note: This will collide with the invite work and one of them will have a merge conflict to resolve.

crates/bitwarden-crypto/src/safe/high_entropy_secret.rs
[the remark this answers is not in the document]
quexten: Lol. Bot vs bot. Copilot is spewing nonsense at any rate, agree with claude.

crates/bitwarden-auth/src/registration/open_org_invite/unseal.rs
```diff
@@ -0,0 +1,182 @@
...
+    /// wrong secret, tampered blob, or cross-namespace substitution.
+    pub fn unseal_open_org_invite_data(
+        &self,
+        sealed: SealedOpenOrgInvite,
+    ) -> Result<OpenOrgInviteSealRequest, RegistrationError> {
+        internal_unseal_open_org_invite_data(&self.client, sealed)
+    }
+}
+
+fn internal_unseal_open_org_invite_data(
```
quexten: Please note this is no longer recommended styling in the SDK and considered an anti pattern: https://contributing.bitwarden.com/architecture/sdk/client-patterns#per-method-files-or-subdirectories

crates/bitwarden-crypto/src/safe/high_entropy_secret.rs
```diff
@@ -81,6 +80,35 @@ impl HighEntropySecret {
...
+
+    /// Encodes the high-entropy secret as a standardized base64 string. This is the same wire
+    /// format the WASM ABI marshaling already uses for `HighEntropySecret`, surfaced as Rust API
+    /// so callers embedding the secret in an opaque wire artifact (e.g. a serde-serialized
+    /// envelope crossing a trust boundary) can round-trip without touching raw bytes.
+    ///
+    /// The returned string carries the secret in cleartext. Callers assume responsibility for it
+    /// — logging it, persisting it, or transmitting it over an untrusted channel each open a
+    /// leak channel.
+    pub fn to_base64(&self) -> String {
```
quexten: Can we please use the same pattern as in "DateEnvelope"?
```
impl FromStr for DataEnvelope {
    type Err = NotB64EncodedError;

    fn from_str(s: &str) -> Result<Self, Self::Err> {
        let data = B64::try_from(s)?;
        Ok(Self::from(data.into_bytes()))
    }
}

impl From<DataEnvelope> for String {
    fn from(val: DataEnvelope) -> Self {
        let serialized: Vec<u8> = (&val).into();
        B64::from(serialized).to_string()
    }
}

impl<'de> Deserialize<'de> for DataEnvelope {
    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
    where
        D: serde::Deserializer<'de>,
    {
        deserializer.deserialize_str(FromStrVisitor::new())
    }
}

impl Serialize for DataEnvelope {
    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
    where
        S: serde::Serializer,
    {
        let serialized: Vec<u8> = self.into();
        serializer.serialize_str(&B64::from(serialized).to_string())
    }
}
```

(Just prompt claude to use the same patterns, it should work fine ;)

crates/bitwarden-auth/src/registration/open_org_invite/seal.rs
```diff
@@ -0,0 +1,222 @@
...
+/// [`RegistrationClient::unseal_open_org_invite_data`]. Named shape-neutrally rather than with a
+/// `Request`/`Response` suffix because either would misrepresent one of the two roles.
+#[cfg_attr(feature = "wasm", derive(Tsify), tsify(into_wasm_abi, from_wasm_abi))]
+#[cfg_attr(feature = "uniffi", derive(uniffi::Record))]
+#[derive(Serialize, Deserialize, Debug, Clone)]
+#[serde(rename_all = "camelCase")]
+pub struct SealedOpenOrgInvite {
+    /// Base64url-encoded CBOR wrapper containing the inner data envelope and outer key envelope.
+    /// Rides the verification email URL as an opaque blob.
+    pub sealed_data: String,
```
quexten: Instead of string, can we use the types directly? They impl serialize so then you get type safety. On TS you can convert to string if you want.

crates/bitwarden-auth/src/registration/open_org_invite/seal.rs
```diff
@@ -0,0 +1,222 @@
...
+fn internal_seal_open_org_invite_data(
+    _client: &Client,
+    input: OpenOrgInviteSealRequest,
+) -> Result<SealedOpenOrgInvite, RegistrationError> {
+    // The CEK is transient and never persists across calls. A per-call `KeyStore` keeps the
+    // key material scoped to this operation so nothing lingers in the caller's key store.
+    let key_store: KeyStore<KeySlotIds> = KeyStore::default();
+    let mut ctx = key_store.context_mut();
+
+    let high_entropy_secret = HighEntropySecret::make(32).map_err(|_| RegistrationError::Crypto)?;
```
quexten: Probably want to move the 32 to the top of the file as a const. I wonder if we should make a default function: "make()". And then make_with_bytes()", where the first just chooses for you? 🤔

crates/bitwarden-auth/src/registration/mod.rs
```diff
@@ -3,12 +3,14 @@
...
 //! the Bitwarden server
 
+mod open_org_invite;
 mod post_keys_for_jit_password_registration;
 mod post_keys_for_key_connector_registration;
 mod post_keys_for_tde_registration;
 mod post_keys_for_user_password_registration;
 mod registration_client;
 
+pub use open_org_invite::{OpenOrgInviteSealRequest, SealedOpenOrgInvite};
```
quexten: Can we make a crate-level "integration" test in the tests subdirectory? That should test the public api of open org invite and do seal + unseal as a rountrip.

We should also add test vectors. We have a add test vector skill in the repository you can use for this. It should be json encoded strings that get pinned into the code, so that we can ensure permanent format compatibility.

crates/bitwarden-auth/src/registration/open_org_invite/seal.rs
```diff
@@ -0,0 +1,222 @@
...
+    let sealed_data = combine_envelopes(&data_envelope, &key_envelope)?;
+
+    Ok(SealedOpenOrgInvite {
+        sealed_data,
+        high_entropy_secret: high_entropy_secret.to_base64(),
+    })
+}
+
+/// Internal CBOR wire schema for `sealed_data`. Each field is the direct serialization of the
+/// respective envelope type, forced to CBOR `bstr` (major type 2) via `serde_bytes`. Without
```
quexten: The public docs here need cleanup. We should also add comments for claude to not put internal docs on the public docs of a pub struct. That is just confusing.

Public docs should document public interface and usage, not internals.

quexten (changes requested):

crates/bitwarden-wasm-internal/integration-tests/tests/registration/open-org-invite-registration.test.ts
```diff
@@ -0,0 +1,54 @@
...
+  it("seal_open_org_invite_data returns a non-empty sealedData and paired highEntropySecret", async () => {
+    const client = makePasswordManagerClient(makeStateBridge());
+
+    const sealed = client.auth().registration().seal_open_org_invite_data(SAMPLE_INPUT);
+
+    expect(sealed.sealedData).not.toEqual("");
+    expect(sealed.highEntropySecret).not.toEqual("");
+  });
+
+  it("unseal_open_org_invite_data recovers the plaintext invite context with fields intact", () => {
```
quexten, line 19: Nice! Round-trip test here.

crates/bitwarden-auth/src/registration/open_org_invite/seal.rs
JaredSnider-Bitwarden: Cleaned up in 62e9925f50467102a662b976dd5ee18c245a6470. 

<details>
<summary>Claude Implementation notes </summary>

The public rustdoc on `OpenOrgInviteSealRequest`, `SealedOpenOrgInvite`, and both `seal_open_org_invite_data` / `unseal_open_org_invite_data` methods now sticks to interface and usage — dropped the base64url/CBOR encoding notes, the naming-rationale paragraph, the "see module-level docs" reference (the module isn't `pub` so those docs never render on the crate's rustdoc anyway), and the cross-namespace substitution phrasing on the unseal error surface.
</details>

crates/bitwarden-auth/src/registration/open_org_invite/unseal.rs
JaredSnider-Bitwarden: Thank you for that heads up. I'd just instructed claude to follow conventions in the registration client, and I'd not seen the new client patterns doc! Appreciate it! Fixed in f93070a354167b2554431c9cc33676987c96e440. 

<details>
<summary>Claude Implementation notes </summary>

The `internal_seal_open_org_invite_data` and `internal_unseal_open_org_invite_data` free-function passthroughs are gone — the bodies now live directly inside the `impl RegistrationClient` methods. The unused `_client: &Client` param went with them, so both files drop the `bitwarden_core::Client` top-level import (moved to the `#[cfg(test)] mod tests` block, where `Client::new(None)` is still needed).

The `combine_envelopes` / `split_envelopes` / `SealedEnvelopePair` helpers stay as internal module-scoped items — they're shared between the seal and unseal paths, not thin passthroughs.
</details>

crates/bitwarden-crypto/src/safe/high_entropy_secret.rs
JaredSnider-Bitwarden: Thank you! Done in ba652a22ecaecbaf44c3499f050232bfb3760476.

<details>
<summary>Claude Implementation notes — trait pattern + deviations</summary>

`HighEntropySecret` now follows the DataEnvelope trait pattern: `FromStr`, `From<HighEntropySecret> for String`, `Serialize`, `Deserialize` (via `FromStrVisitor`). The ad-hoc `to_base64` / `from_base64` methods are gone, and callers use `.parse::<HighEntropySecret>()` and `String::from(secret)` / `.into()`.

Two intentional deviations from your snippet:

1. `type Err = HighEntropySecretError` on the FromStr impl, not `NotB64EncodedError`. `HighEntropySecret` has two failure modes — malformed base64 AND `MIN_SECRET_LENGTH` (from 8509fbe08) — and `NotB64EncodedError` only expresses the first.
1. Serialize body reads `self.secret.as_slice()` directly rather than routing through an `impl From<&HighEntropySecret> for Vec<u8>` like DataEnvelope does. Adding that impl would leak the secret bytes out through a non-zeroizing `Vec<u8>` handle, so kept the direct read.

Also routed `FromWasmAbi::from_abi` through `FromStr::from_str` so the length check applies at the WASM boundary too — previously it skipped the check, which was inconsistent with the Rust API after 8509fbe08.

</details>

crates/bitwarden-auth/src/registration/open_org_invite/seal.rs
JaredSnider-Bitwarden: Ok, Claude and I took a stab at this with 8314bc98a0383096cbcc2258d6069b96df7f28a7. However, in order to keep the JSON wire unchanged, we had to build a `SealedEnvelopePair` wrapper?  I'm not sure if that fits your vision or if not having typed individual envelopes in TS will bite me in the future... But it simplifies my client and server work to just have an opaque string vs two strings? Do you see any issues with that? Aside, if you like that approach, maybe KM would want to upstream that `SealedEnvelopePair` for future use cases where consuming clients want to keep things opaque. 

<details>
<summary>Claude Implementation notes</summary>

`SealedOpenOrgInvite` now holds `sealed_data: SealedEnvelopePair` and `high_entropy_secret: HighEntropySecret`. `SealedEnvelopePair` is a new `pub` type wrapping typed `DataEnvelope` + `SecretProtectedKeyEnvelope`, with the same trait pattern as `DataEnvelope`: `FromStr`, `From<T> for String`, `Serialize`, `Deserialize` (via `FromStrVisitor`), plus WASM ABI mirroring `HighEntropySecret`.

Wire behavior:

1. JSON stays `{"sealedData": "<base64url>", "highEntropySecret": "<base64>"}` — both fields still serialize as strings via the types' `Serialize` impls, so the server-passthrough URL query param shape is preserved.
1. TS type system sees `SealedEnvelopePair` and `HighEntropySecret` as `Tagged<String, "TypeName">` branded strings — runtime strings, compile-time type safety. TS can't accidentally swap the two, and callers who just want to treat them as strings still can.

Simplifications: `combine_envelopes` / `split_envelopes` / the private `Vec<u8>`-based `SealedEnvelopePair` wrapper are gone. Seal constructs the typed pair directly; unseal consumes the typed fields directly.

Dropped `#[cfg_attr(feature = "uniffi", derive(uniffi::Record))]` from `OpenOrgInviteSealRequest` and `SealedOpenOrgInvite`: `HighEntropySecret` and `SealedEnvelopePair` lack uniffi custom-type impls, so this crossing is WASM-only until those land.

Test additions:

1. `seal_json_round_trip_unseal_recovers_original_fields` — end-to-end serde round-trip through the seal → JSON → JSON → unseal path.
1. `sealed_envelope_pair_parse_rejects_valid_cbor_with_bad_key_envelope_bytes` — exercises the parse-envelope failure path that base64url/CBOR/truncation tests miss.
1. `sealed_open_org_invite_json_wire_shape_is_stable` — pins the two-key camelCase JSON shape with both values as strings.

Also cleaned up public rustdoc during the refactor to keep `pub` items describing interface + usage only (dropped internal encoding/construction detail from field docs and now-noisy `impl` doc comments).

</details>

crates/bitwarden-auth/src/registration/open_org_invite/seal.rs
JaredSnider-Bitwarden: Ah, good call: I've done that with 5874d8425b292961f8c8bc0f21431581626a4363. 

The default idea would be a good one to avoid multiple places to tweak if we ever want to change default sizes. I didn't take the maybe as a requirement that I had to implement so I didn't for now (leaving to KM unless you think we need it).

crates/bitwarden-auth/src/registration/mod.rs
JaredSnider-Bitwarden: Great call! I didn't know about the test vectors skill. That's sweet! Thank you for showing me that!

3e556bb92b16b8ce17e9fdd198a976667c43cbbd.

<details>
<summary>Claude Implementation notes</summary>

New file `crates/bitwarden-auth/tests/registration_open_org_invite.rs` exercises the crossing through the crate's public API only (`Client::auth_new().registration()`), so anything a downstream consumer sees is exactly what these tests see.

Four tests:

1. `seal_unseal_round_trip_via_public_api` — happy-path round trip.
1. `seal_json_round_trip_unseal_via_public_api` — production-shaped: seal, marshal to JSON, deserialize back, unseal. Mirrors the SDK → server → verification-email URL → SDK flow.
1. `unseal_with_wrong_secret_returns_crypto_error_via_public_api` — locks the public error contract: mismatched `HighEntropySecret` surfaces as `RegistrationError::Crypto`, not a panic or a different variant.
1. `pinned_wire_vector_unseals_to_expected_plaintext` — captured via the `create-testvectors` skill workflow. The `TEST_VECTOR_SEALED_JSON` constant is the wire form of a seal against a known plaintext; the test deserializes, unseals, and asserts plaintext equality.

crates/bitwarden-crypto/src/safe/high_entropy_secret.rs
```diff
@@ -83,6 +85,54 @@ impl HighEntropySecret {
     }
 }
 
+/// Parses a `HighEntropySecret` from a standardized base64 string previously produced by
+/// [`String::from`]`(secret)` (or by the equivalent WASM ABI marshaling).
+///
```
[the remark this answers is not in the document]
JaredSnider-Bitwarden: Fixed with a8b6c651f.

crates/bitwarden-crypto/src/safe/high_entropy_secret.rs
```diff
@@ -83,6 +85,54 @@ impl HighEntropySecret {
...
+        if bytes.as_bytes().len() < MIN_SECRET_LENGTH {
+            return Err(HighEntropySecretError::TooShort);
+        }
+        Ok(Self::from_internal(bytes.as_bytes()))
+    }
+}
+
+/// Encodes a `HighEntropySecret` as a standardized base64 string. The returned string carries
+/// the secret in cleartext — logging it, persisting it, or transmitting it over an untrusted
+/// channel each open a leak channel.
```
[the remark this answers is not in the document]
JaredSnider-Bitwarden: Fixed with a8b6c651f.

crates/bitwarden-auth/src/registration/open_org_invite/seal.rs
```diff
@@ -0,0 +1,354 @@
...
+    /// The wire string could not be decoded.
+    #[error("Sealed envelope pair is malformed")]
+    Malformed,
+}
+
+// WASM ABI: `SealedEnvelopePair` marshals as its wire string, matching the JSON wire form.
+#[cfg(feature = "wasm")]
+#[wasm_bindgen::prelude::wasm_bindgen(typescript_custom_section)]
+const TS_CUSTOM_TYPES: &'static str = r#"
+export type SealedEnvelopePair = Tagged<String, "SealedEnvelopePair">;
```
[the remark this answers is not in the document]
JaredSnider-Bitwarden: Fixed with a8b6c651f. Also fixed the same issue in HighEntropySecret's declaration (self-caught sibling).

crates/bitwarden-auth/src/registration/open_org_invite/seal.rs
quexten: "The json wire unchanged", the sealedenvelopepair encodes to b64 cbor not json...?

You could do json, and in fact it would probably make the encoding more lean, because the data envelope and secretkeyenvelope encode to b64 strings, so currently there is a double b64 encoding going on.

quexten: > But it simplifies my client and server work to just have an opaque string vs two strings? Do you see any issues with that? 

Yes it is reasonable from my perspective.

> Aside, if you like that approach, maybe KM would want to upstream that SealedEnvelopePair for future use cases where consuming clients want to keep things opaque.

I don't think this would be a construct that is useful upstream, especially with the current encoding overhead. In most cases there is a more complex hierarchy than just a secret -> dataenvelope directly. Usually, there would be an intermediate KEK, that is shared to the secret, and wraps the data envelope CEK, (and possibly more).

quexten: > I'm not sure if that fits your vision or if not having typed individual envelopes in TS will bite me in the future...


I don't think you need the individual envelopes in TS for this use-case.

crates/bitwarden-auth/src/registration/open_org_invite/seal.rs
```diff
@@ -0,0 +1,349 @@
...
+        let data_bytes: Vec<u8> = (&val.data_envelope).into();
+        let key_bytes: Vec<u8> = (&val.key_envelope).into();
+        let wire = SealedEnvelopePairWire {
+            data_envelope: data_bytes,
+            key_envelope: key_bytes,
+        };
+        let mut buf = Vec::new();
+        ciborium::ser::into_writer(&wire, &mut buf)
+            .expect("CBOR encoding of two byte fields cannot fail");
+        B64Url::from(buf).to_string()
```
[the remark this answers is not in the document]
JaredSnider-Bitwarden: This follows existing patterns in BW crypto. 

Claude:

1. Sibling `SecretProtectedKeyEnvelope::From<&_> for Vec<u8>` uses the identical `.expect(...)` on the same shape — this is the established convention in `bitwarden-crypto::safe`.
2. The panic is unreachable: two `#[serde(with = "serde_bytes")]` fields encode as CBOR `bstr`, and `Vec<u8>` writer I/O is infallible. `ciborium` has no fallible path here.
3. Refactoring would break the string-wire-form trait pattern (`FromStr` + `From<T> for String` + `Serialize`/`Deserialize`) used across every safe type in the crate, for a code path that cannot fire.

crates/bitwarden-auth/src/registration/open_org_invite/unseal.rs
```diff
@@ -0,0 +1,192 @@
+//! Unseals a [`SealedOpenOrgInvite`] back into the plaintext [`OpenOrgInviteSealRequest`],
+//! provided the caller supplies the paired [`HighEntropySecret`] the seal path returned. This
+//! runs on the verification-email tab after registration-finish has logged the user in.
+//!
+//! Reverses the two-envelope construction documented in [`super::seal`]: outer key envelope →
+//! inner CEK, then inner data envelope → plaintext.
+//!
+//! Any failure — malformed wire, wrong secret, tampered blob, cross-namespace — surfaces as
+//! [`RegistrationError::Crypto`]. The SDK performs no post-decrypt equality check on the
+//! plaintext; the AES-GCM auth tag at each envelope layer is the substitution defense.
```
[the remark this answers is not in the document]
JaredSnider-Bitwarden: Fixed with 8f2f989a5. Also cleaned up the same accuracy issue in the public method doc while I was there.

dani-garcia (dismissed):

crates/bitwarden-auth/src/registration/open_org_invite/seal.rs
```diff
@@ -0,0 +1,354 @@
+//! Seals an open-organization-invite context into an opaque wire artifact plus a paired
+//! [`HighEntropySecret`]. The sealed data rides the verification email URL; the secret stays
+//! client-side. Both are needed to recover the original invite context via [`super::unseal`].
+//!
+//! Two-envelope construction mirrors the pattern in
+//! `bitwarden-organization-crypto/src/invite_key_bundle.rs`:
```
quexten: Remove paragraph

crates/bitwarden-auth/src/registration/open_org_invite/seal.rs
```diff
@@ -0,0 +1,354 @@
...
+//! - inner [`DataEnvelope`] seals the `RegistrationOpenOrgInviteData` payload with a freshly
+//!   generated content-encryption key (CEK);
+//! - outer [`SecretProtectedKeyEnvelope`] seals that CEK with a per-registration
+//!   [`HighEntropySecret`].
+//!
+//! Substitution defense at both layers is the AES-GCM auth tag / wrong-key check — the SDK
+//! performs no post-decrypt equality check on the plaintext, matching how
```
quexten: Remove paragraph, this is at a weirdly low level.

crates/bitwarden-auth/src/registration/open_org_invite/seal.rs
```diff
@@ -0,0 +1,354 @@
...
+#[cfg(feature = "wasm")]
+use tsify::Tsify;
+#[cfg(feature = "wasm")]
+use wasm_bindgen::prelude::*;
+
+use super::{RegistrationOpenOrgInviteData, data_v1::RegistrationOpenOrgInviteDataV1};
+use crate::registration::registration_client::{RegistrationClient, RegistrationError};
+
+/// Byte length of the per-registration [`HighEntropySecret`] the seal path generates. 32 bytes
+/// = 256 bits, well above [`HighEntropySecret`]'s minimum-length floor and matching the
```
quexten: Very wordy comment..

crates/bitwarden-auth/src/registration/open_org_invite/seal.rs
```diff
@@ -0,0 +1,354 @@
...
+/// standard random-secret size elsewhere in the SDK.
+pub(super) const OPEN_ORG_INVITE_SECRET_SIZE_BYTES: usize = 32;
+
+/// Input to [`RegistrationClient::seal_open_org_invite_data`]. All three fields are required.
+// Not `uniffi::Record`-derived: [`SealedOpenOrgInvite`] holds typed cryptographic fields
```
quexten: Please rewrite the documentation. This is very confusing currently. Just describe clearly what the struct is used for.

crates/bitwarden-auth/src/registration/open_org_invite/seal.rs
```diff
@@ -0,0 +1,354 @@
...
+#[cfg_attr(feature = "wasm", derive(Tsify), tsify(into_wasm_abi, from_wasm_abi))]
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq)]
+#[serde(rename_all = "camelCase")]
+pub struct OpenOrgInvite {
+    /// The organization the registrant is joining.
+    pub organization_id: String,
+    /// The public invite link code carried in the shared invite URL.
+    pub invite_link_code: String,
+    /// The invite key associated with the invite link.
+    pub invite_key: String,
```
quexten: Invite secret is the term used now.

crates/bitwarden-auth/src/registration/open_org_invite/seal.rs
```diff
@@ -0,0 +1,354 @@
...
+}
+
+/// Sealed open-organization-invite payload. Produced by
+/// [`RegistrationClient::seal_open_org_invite_data`] and consumed by
+/// [`RegistrationClient::unseal_open_org_invite_data`]. Both fields are required to unseal;
```
quexten: What is the struct meant for? What is the purpose? Can we please make the docs clear.

crates/bitwarden-auth/src/registration/open_org_invite/seal.rs
```diff
@@ -0,0 +1,354 @@
...
+
+impl<'de> Deserialize<'de> for SealedOpenOrgInviteData {
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where
+        D: serde::Deserializer<'de>,
+    {
+        deserializer.deserialize_str(FromStrVisitor::new())
+    }
+}
+
```
quexten: TBH, seal and unseal living in the same file would make more sense, then you could have a round-trip unit test.

crates/bitwarden-crypto/src/safe/high_entropy_secret.rs
```diff
@@ -83,6 +85,54 @@ impl HighEntropySecret {
     }
 }
 
+/// Parses a `HighEntropySecret` from a standardized base64 string previously produced by
+/// `String::from(secret)` (or by the equivalent WASM ABI marshaling).
+///
+/// The caller must guarantee that the decoded bytes originated from a genuine high-entropy
+/// source. This constructor does not — and cannot — validate entropy on its own; feeding
+/// low-entropy input to the cheap KDF that consumes this type will not provide the
```
quexten: This is over-documented

crates/bitwarden-crypto/src/safe/high_entropy_secret.rs
```diff
@@ -83,6 +85,54 @@ impl HighEntropySecret {
...
+        let bytes = B64::try_from(s).map_err(|_| HighEntropySecretError::Malformed)?;
+        if bytes.as_bytes().len() < MIN_SECRET_LENGTH {
+            return Err(HighEntropySecretError::TooShort);
+        }
+        Ok(Self::from_internal(bytes.as_bytes()))
+    }
+}
+
+/// Encodes a `HighEntropySecret` as a standardized base64 string. The returned string carries
+/// the secret in cleartext — logging it, persisting it, or transmitting it over an untrusted
```
quexten: Over-documented, this adds nothing.

crates/bitwarden-crypto/src/safe/secret_protected_key_envelope.rs
```diff
@@ -565,6 +565,11 @@ pub enum SecretProtectedKeyEnvelopeNamespace {
     /// Bitwarden Desktop biometric (Windows Hello) unlock. The high-entropy secret is a PRF derived
     /// from the Windows Hello signing credential, and the sealed key is the user key.
     DesktopBiometricUnlock = 2,
+    /// Namespace for the outer envelope of the open-organization-invite registration crossing.
+    /// The sealed key protects the inner `DataEnvelope`'s content-encryption key; the high-entropy
+    /// secret is generated per registration and stored client-side to be paired with the sealed
```
quexten: "Is generated per registration and stored client side" is enough.

quexten (changes requested):
There are some very wordy comments in here to the point that they distract rather than give useful information. They are noise.

crates/bitwarden-crypto/src/safe/secret_protected_key_envelope.rs
JaredSnider-Bitwarden: I've fixed this with 86e80b3510daa4f3f741ca03e5a09a1a179e9dc1

crates/bitwarden-crypto/src/safe/high_entropy_secret.rs
JaredSnider-Bitwarden: Removed with 3ccff9e62f6fd986e49a2aedd4ac26fd7cecf1ea

crates/bitwarden-crypto/src/safe/high_entropy_secret.rs
JaredSnider-Bitwarden: Removed with 3ccff9e62f6fd986e49a2aedd4ac26fd7cecf1ea

crates/bitwarden-auth/src/registration/open_org_invite/seal.rs
JaredSnider-Bitwarden: The decision to split along method lines came out of both my preference to keep files leaner for reducing cognitive load + the contributing docs [SDK best practices](https://contributing.bitwarden.com/architecture/sdk/client-patterns#file-organization) you showed me which said to split files once they get over 500 lines (which the seal and unseal total 550 today). The two methods are also neatly put into their own subfolder (which I cleared with @dani-garcia as well b/c the pattern doesn't fit the examples in the docs perfectly). 

The `unseal.rs` file has round trip unit tests `seal_unseal_round_trip_recovers_original_fields` and `seal_json_round_trip_unseal_recovers_original_fields` + I have round trip integration tests. Thus, I would prefer to keep this architecture as is.

crates/bitwarden-auth/src/registration/open_org_invite/seal.rs
JaredSnider-Bitwarden: Thank you for the heads up. Rename applied with ba7fbd5e4d086f70d28d76be60ebeb2ac0251f29

crates/bitwarden-auth/src/registration/open_org_invite/seal.rs
JaredSnider-Bitwarden: Reduced with 21a1c2606448b2a576bd1ba6cdeda13b03882f78

crates/bitwarden-auth/src/registration/open_org_invite/seal.rs
JaredSnider-Bitwarden: 44f21d79a245d49b40bc41b6e8470b6100073736

crates/bitwarden-auth/src/registration/open_org_invite/seal.rs
JaredSnider-Bitwarden: 9ed4201ebca24da92ded852eadbf3d357ecf6738

crates/bitwarden-wasm-internal/integration-tests/tests/registration/open-org-invite-registration.test.ts
```diff
@@ -0,0 +1,54 @@
+import { makePasswordManagerClient, makeStateBridge } from "../utils";
+
+const SAMPLE_INPUT = {
+  organizationId: "1bc9ac1e-f5aa-45f2-94bf-b181009709b8",
+  inviteLinkCode: "abcd1234efgh5678",
+  inviteKey: "raw-invite-key-material-base64url",
+};
```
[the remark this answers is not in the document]
JaredSnider-Bitwarden: Fixed with e7a936f6faeb87304e73a584e22c4d3712a71084

crates/bitwarden-auth/src/registration/open_org_invite/seal.rs
quexten: I don't agree that this reduces cognitive load in this case, but it's your code ownership, so do not consider my comment blocking.

Having seal and unseal is two components of the same logic, and really should be in the same file in my opinion. Feel free to dismiss this comment.

quexten: The much better way to split this is to move the logic of the struct implementations (SealedOpenOrgInviteDataWire), out, and have the clients be merely composing calls on the struct. That is, your `SealedOpenOrgInviteData` should live in a separate file, and have a `seal`, `unseal` function, and be tested there, along with test vectors and a round-trip.

The `wire` implementation could also be a separate file.

Then, your client impl only constructs SealedOpenOrgInviteData. 

Currently, your `unseal` file imports `seal`'s Structs.

quexten: That is, this should be structured as:
```
open_org_invite_data.rs
- Contains impl for seal + unseal
serialization.rs
- Contains From<OpenOrgInviteData>, wire struct, serialization tests
data_v1.rs
client.rs
contains both seal and unseal functions, but they do not actually implement the seal and unseal like is currently done, they just call to open_org_invite_data + serialize to wire
```

quexten: I'll consider how we can make this pattern clear in contributing docs, but primarily logic for structs should be living in a single file. The clients should compose calls to impl on structs, but splitting logic that is relevant only to one struct and moving it into clients to assemble the struct is breaking the encapsulation of the struct, and thus an anti-pattern.

Consider that you are assembling a bunch of calls, and then setting the fields of the struct directly, which usually would be considered an example of improper abstraction.

quexten (dismissed):
Approving KM owned code. Please conisder the advice on structure. While it is optional, I do feel that the current implementation is an anti-pattern, as we are splitting the implementation of a struct's logic, and moving it into clients across multiple files, when not necessary.

This would be greatly simplified by moving the logic onto impl's on the struct, then making the client calls much lighter (and merging the files for the client functions, since now they are tiny).

crates/bitwarden-auth/src/registration/open_org_invite/seal.rs
JaredSnider-Bitwarden: Fixed with https://github.com/bitwarden/sdk-internal/commit/44f21d79a (forgot to respond to this one).

crates/bitwarden-auth/src/registration/open_org_invite/seal.rs
JaredSnider-Bitwarden: Thank you for your patience and continued help in getting this into a better shape. I believe I've restructured this as you've requested with a40ca8b01a7d74dddd7c29db9a0f5db3252b61b5. 

I had to rename `crates/bitwarden-auth/src/registration/open_org_invite` to `crates/bitwarden-auth/src/registration/open_org_invite_crypto`, but now things are split neatly by concern (Claude summary):
- mod.rs              — module glue: versioned-enum macro + public reexports
- data_v1.rs          — v1 plaintext payload (the innermost thing sealed by DataEnvelope)
- open_org_invite.rs  — domain types + seal/unseal impls + round-trip tests
- serialization.rs    — wire encoding + WASM ABI + wire/parse tests
- client.rs           — FFI pair struct + thin RegistrationClient wrappers

crates/bitwarden-auth/src/registration/open_org_invite_crypto/serialization.rs
```diff
@@ -0,0 +1,204 @@
...
+    data_envelope: Vec<u8>,
+    #[serde(rename = "k", with = "serde_bytes")]
+    key_envelope: Vec<u8>,
+}
+
+/// Errors returned when parsing a [`SealedOpenOrgInviteData`] from its wire form.
+#[derive(Debug, Error)]
+pub enum SealedOpenOrgInviteDataError {
+    /// The wire string could not be decoded.
+    #[error("Sealed envelope pair is malformed")]
```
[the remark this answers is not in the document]
JaredSnider-Bitwarden: Fixed with https://github.com/bitwarden/sdk-internal/pull/1295/commits/b45ac31d2ba6c0ceeba63185764d2b489673ec89

.github/CODEOWNERS
```diff
@@ -1,40 +1,58 @@
...
 # 1. Full ownership by team-platform-dev, for cases where teams have not yet onboarded at all.
+
 # 2. Joint team ownership with team-platform-dev, as a team onboards into SDK development and close guidance is necessary.
+
 # 3. Full team ownership, once a team is comfortable with SDK development and has enough shared experience to ensure alignment with SDK architecture.
 
 # Platform is the default owner for all files
-* @bitwarden/team-platform-dev
+
+- @bitwarden/team-platform-dev
```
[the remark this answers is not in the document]
JaredSnider-Bitwarden: Should be fixed with https://github.com/bitwarden/sdk-internal/pull/1295/commits/abc181043e0ee4b1387f1b16dfb9b71398700b05

crates/bitwarden-auth/src/registration/open_org_invite_crypto/serialization.rs
```diff
@@ -0,0 +1,205 @@
...
+
+/// Intermediate shape used to (de)serialize [`SealedOpenOrgInviteData`] to and from its wire
+/// bytes. Each envelope is carried as raw bytes under a short field name for compactness.
+#[derive(Serialize, Deserialize)]
+struct SealedOpenOrgInviteDataWire {
+    // Without serde_bytes, Vec<u8> encodes as a CBOR array of integers (~2x the size).
+    #[serde(rename = "d", with = "serde_bytes")]
+    data_envelope: Vec<u8>,
+    #[serde(rename = "k", with = "serde_bytes")]
+    key_envelope: Vec<u8>,
```
quexten, line 23: Nit: Naming this "sealed_data_envelope_cek" would be clearer. It is the CEK of the data envelope.

quexten (approved):
Looks good. I have one nit on the naming in the serialization format. I would at least add a comment that the key envelope contains the CEK of the data envelope., along with a claude generated diagram, similar to what we have at the top of the invite crypto file, that describes which key protects what.

dani-garcia (approved):

enmande (approved):

crates/bitwarden-auth/src/registration/open_org_invite_crypto/serialization.rs
JaredSnider-Bitwarden: Will address in fast follow.

JaredSnider-Bitwarden: Actually, the related type is called  `SecretProtectedKeyEnvelope`?  I'm mirroring KM naming here. I've added an additional comment in the follow up work to clarify, but respectfully, I'm not seeing a benefit of the other name.   Sealed is already in the struct name `SealedOpenOrgInviteDataWire`. It's an envelope that holds a CEK which can just be abbreviated as key therefore `key_envelope` is accurate?

See https://github.com/bitwarden/sdk-internal/pull/1309/changes/f3eceaf9a47824d3e55d92d4b0dc37c2be4725e2 for new comments to try to make things more clear.

PR #1297, opened 2026-07-24
feat(kdf-change): port to sdk

quexten:
https://github.com/bitwarden/clients/pull/22032
https://bitwarden.atlassian.net/browse/PM-40984

Moves KDF-Change fully to the SDK, including API calls and state management.

Added TS tests to test the public API contract of the SDK.


theMickster (approved):
Approving `CODEOWNERS` change on-behalf of tech-leads.

crates/bitwarden-uniffi/src/crypto.rs
```diff
@@ -101,7 +101,9 @@ impl CryptoClient {
     /// Create the data necessary to update the user's kdf settings. The user's encryption key is
     /// re-encrypted for the password under the new kdf settings. This returns the new encrypted
     /// user key and the new password hash but does not update sdk state.
+    /// Note: This is deprecated. Please use the user-crypto-management client instead.
```
djsmith85, line 104: ❓ Looking at the current usage across all BW repos: https://github.com/search?q=org%3Abitwarden%20make_update_kdf&type=code, are you planning on removing this directly after `clients` has been migrated to use the user-crypto-management ?

djsmith85 (approved):

mzieniukbw (approved):

crates/bitwarden-uniffi/src/crypto.rs
quexten: We need to wait for mobile to migrate too. https://github.com/bitwarden/ios/blob/0866dc918b57f0fe818c6f2f27570ffc8bb97d52/BitwardenShared/Core/Auth/Services/ChangeKdfService.swift#L106. KM does not generally port mobile features, and it is up to mobile to move off of the deprecated function.

PR #1303, opened 2026-07-28
Tools/pm 39238/complete file send lifecycle

adudek-bw:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-39238

## 📔 Objective

Completes the file-send lifecycle for the Rust CLI bw send, which PR #1176 (PM-34719) stubbed out with "not yet implemented" errors.

## 🚨 Breaking Changes

Changes to the WASM/UniFFI binding surface, safe today (no live consumers — the clients repo routes file-send creation to the legacy service and does not call these):
- SendClient::create_file_send gains a file_buffer: Vec<u8> parameter; CreateFileSendResponse gains encrypted_file_buffer: Vec<u8>.
- SendClient::upload_send_file gains file_upload_type and upload_url parameters.

The cross-repo TypeScript compatibility check will confirm no client breakage.


crates/bitwarden-core/src/auth/jwt_token.rs
```diff
@@ -82,7 +85,42 @@ mod tests {
         assert_eq!(token.sub, "e25d37f3-b603-40de-84ba-af96012f5a42");
         assert_eq!(token.email.as_deref(), Some("test@bitwarden.com"));
         assert_eq!(token.organization.as_deref(), None);
+        // This fixture carries `"premium":false` in its payload.
```
dani-garcia: This comment doesn't seem useful, we should just remove it.

crates/bw/src/tools/send.rs
```diff
@@ -622,6 +628,13 @@ fn build_edit_request(
         (None, Some(f)) => SendViewType::File(f),
         // Sends should always carry exactly one of text/file; the API can in theory return
         // both. The legacy CLI prefers text in that case, which is what we do here.
+        //
+        // PM-39238 disambiguation finding (item #4): there is NO deviation from legacy to fix
+        // here. `get` returns the full [`SendView`] (both `text` and `file` fields preserved), so
+        // a caller reading a mixed-shape response loses nothing. `create` is built from the typed
+        // [`SendViewType`] enum and so is unambiguous by construction. The only place a "prefer
+        // text" choice is forced is `edit`, where a single variant must be reconstructed from the
+        // existing row — and that choice matches the legacy CLI (`SendView.text ?? SendView.file`).
```
dani-garcia: This comment is a bit confusing. It just seems to restate what the above comment says (`Sends should always carry exactly one of text/file; the API can in theory return both. The legacy CLI prefers text in that case, which is what we do here.`). We can probably just remove it.

dani-garcia (dismissed):
Looks pretty good, but some of the comments are pretty wordy and unneeded. It would be good to trim them down.

crates/bw/src/tools/send.rs
```diff
@@ -0,0 +1,1599 @@
...
+/// 3. The ciphertext is uploaded via `upload_send_file`, which dispatches to the Direct or Azure
+///    backend based on the `file_upload_type` from step 2.
+async fn run_create_file(
+    client: &PasswordManagerClient,
+    request: SendAddRequest,
+    path: &str,
+) -> color_eyre::eyre::Result<bitwarden_send::SendView> {
+    // Read the plaintext before creating the send so a read failure aborts before we register a
+    // send that would then have no content.
+    let bytes = std::fs::read(path).wrap_err_with(|| format!("Could not read file {path}"))?;
```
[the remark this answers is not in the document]
adudek-bw: @AikidoSec ignore: local CLI argument, not a trust-boundary crossing — the operating user already has filesystem access to any path they pass here.

harr1424 (dismissed):
This looks good, I agree on the above regarding some of these comments, though.

crates/bw/src/tools/send.rs
```diff
@@ -715,6 +732,73 @@ async fn run_create(
...
```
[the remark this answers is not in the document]
adudek-bw: @AikidoSec ignore: local CLI argument, not a trust-boundary crossing — the operating user already has filesystem access to any path they pass here.

crates/bw/src/tools/send.rs
```diff
@@ -293,7 +294,7 @@ impl BwCommand for SendArgs {
             // emitting JSON to stdout while the requested file path goes uncreated would
             // be a worse UX than an explicit "not implemented" error.
             Some(SendCommands::Get(args)) if args.output_path.is_some() => Err(eyre!(
-                "`--output` on `bw send get` is not yet implemented (tracked under PM-39238)."
+                "`--output` on `bw send get` is not yet implemented (tracked under PM-34718)."
```
harr1424: ❓ Will this output show in our production release? If so, does it make sense to reference an internal Jira ticket in the output itself?

crates/bw/src/tools/send.rs
harr1424: ❓ Will this output show in our production release? If so, does it make sense to reference an internal Jira ticket in the output itself?

harr1424: ❓ Will this output show in our production release? If so, does it make sense to reference an internal Jira ticket in the output itself?

harr1424 (commented):
The comments seem much more focused now! I found some error output which production users might find strange; should we consider referencing the internal Jira tickets using code comments as opposed to command output?

crates/bw/src/tools/send.rs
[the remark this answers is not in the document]
withinfocus: @adudek-bw yes, the user has access to the filesystem, but if the path contains "../../../" for example it could still write where they're not expecting it to be. This should still be fixed.

dani-garcia (dismissed):

harr1424 (approved):

dani-garcia (approved):

PR #1310, opened 2026-07-30
[PM-41226] Add access-request and lease sub-clients

Hinton:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-41226

## 📔 Objective

Extend the `bitwarden-pam` crate with the **request→activate** leasing surface,
copying the shape of the existing `access_rules` module (client + models + flat
error, with Tsify/`wasm_bindgen` bindings and `mockall` unit tests):

- **`access_requests`** — `AccessRequestsClient` on `PamClient::access_requests()`:
  - `activate(id) -> AccessLeaseView` — activate an approved request, minting the lease
  - `cancel(id)` — cancel the caller's own pending request
  - `get(id)` / `list_mine()` — read the caller's requests
  - `AccessRequestView`, `AccessRequestStatus`
- **`leases`** — `LeasesClient` on `PamClient::leases()`:
  - `list_active()` / `list_mine()` — read the caller's leases
  - `extend(lease_id, request)` — extend an active lease
  - `end(lease_id, request)` — end a lease early (re-lock)
  - `AccessLeaseView`, `AccessLeaseStatus`, `AccessLeaseExtensionRequest`, `AccessLeaseRevokeRequest`
- Shared flat `LeasingError`; new `AccessLeaseId` / `AccessRequestId` newtypes.

Built entirely on bindings already present in `bitwarden-api-api` on `main`
(`access_requests_api`, `leases_api`) — no binding changes in this PR.

**Deliberately not included:** request *creation* (`POST /leases/ciphers/{id}`)
and the per-cipher access-state / pre-check reads. Those depend on the
`cipher_lease_api` binding, which isn't generated into `bitwarden-api-api` yet
and, per repo convention, lands via the bindings-update workflow rather than a
feature PR. Until then, clients keep creating requests over HTTP and can derive
per-cipher state from `leases().list_active()` + `access_requests().list_mine()`.
Human-approval (`decide`) is out of scope for the automatic flow.

Verification: `cargo test -p bitwarden-pam --all-features` (74 pass, 9 new),
clippy/fmt/cargo-sort clean, wasm target compiles, and the generated commercial
`.d.ts` exposes `commercial().pam().access_requests()` / `.leases()` with the
expected view/DTO/status types.

## 🚨 Breaking Changes

None — purely additive to the commercial `bitwarden-pam` crate.

### DIFF
--- a/bitwarden_license/bitwarden-pam/src/access_requests/models.rs
+++ b/bitwarden_license/bitwarden-pam/src/access_requests/models.rs
@@ -1,0 +1,110 @@
+use bitwarden_api_api::models::{
+    AccessRequestDetailsResponseModel, AccessRequestStatus as ApiAccessRequestStatus,
+};
+use bitwarden_collections::collection::CollectionId;
+use bitwarden_core::{OrganizationId, UserId, require};
+use bitwarden_vault::CipherId;
+use chrono::{DateTime, Utc};
+use serde::{Deserialize, Serialize};
+#[cfg(feature = "wasm")]
+use tsify::Tsify;
+
+use crate::{AccessRequestId, AccessRuleId, error::LeasingError};
+
+/// The lifecycle state of an access request.
+///
+/// The automatic (no human approval) path moves `Pending -> Approved -> Activated`; the requester
+/// activates the approved request to mint a lease. `Denied`, `Canceled`, and `Expired` are terminal
+/// states in which no lease exists.
+#[derive(Serialize, Deserialize, Debug, Clone, Copy, PartialEq, Eq)]
+#[cfg_attr(feature = "wasm", derive(Tsify), tsify(into_wasm_abi, from_wasm_abi))]
+#[serde(rename_all = "snake_case")]
+pub enum AccessRequestStatus {
+    /// Awaiting a decision (or, on the automatic path, awaiting the server's auto-approval).
+    Pending,
+    /// Approved but not yet activated; the requester may activate it to mint a lease.
+    Approved,
+    /// Activated - a lease has been minted from this request.
+    Activated,
+    /// Denied by an approver; terminal.
+    Denied,
+    /// Cancelled by the requester before resolution; terminal.
+    Canceled,
+    /// Approved but lapsed before the requester activated it; terminal.
+    Expired,
+    /// A status value this SDK version does not recognize. Kept as a distinct variant so listing
+    /// requests never fails on a newer server's status.
+    Unknown,
+}
+
+impl From<ApiAccessRequestStatus> for AccessRequestStatus {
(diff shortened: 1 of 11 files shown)


bitwarden_license/bitwarden-pam/src/access_requests/models.rs
```diff
@@ -0,0 +1,350 @@
...
+    /// The human approver's email; `None` for an automatic decision, or when the user could not
+    /// be resolved.
+    pub email: Option<String>,
+    /// The optional note the approver left with the decision.
+    pub comment: Option<String>,
+    /// The decision's verdict.
+    pub verdict: AccessDecisionVerdict,
+    /// When the decision was recorded (UTC).
+    pub decided_at: DateTime<Utc>,
+}
```
patriksvensson, line 98: The comment say:

```rust
/// For an [`Automatic`](AccessDeciderKind::Automatic) decision, 
/// `id`, `name`, and `email` are `None`; for a [`Human` 
/// (AccessDeciderKind::Human) decision they carry the 
/// approver's identity,
```

I think it would be better to make this a discriminated union which is more idiomatic to Rust.

```rust
pub enum AccessRequestDecisionView {
  Automatic(...),
  Human(...),
}
```

bitwarden_license/bitwarden-pam/src/leases/client.rs
```diff
@@ -0,0 +1,225 @@
...
+            .collect()
+    }
+
+    /// Extends an active lease, returning the updated originating request.
+    pub async fn extend(
+        &self,
+        lease_id: AccessLeaseId,
+        request: AccessLeaseExtensionRequest,
+    ) -> Result<AccessRequestView, LeasingError> {
+        let response = self
```
patriksvensson, line 63: Should we validate the request here? If someone creates a request with `duration_seconds` set to `-1`, the request will fail.

patriksvensson (commented):
Nothing major, and I'm happy to approve it, but I think it is worth having a discussion about `AccessRequestDecisionView` before so.

bitwarden_license/bitwarden-pam/src/access_requests/models.rs
Hinton: Correct, it ended up being a mix where we keep the struct for the shared fields and move decider into it's own enum.

bitwarden_license/bitwarden-pam/src/leases/client.rs
Hinton: Changed to `NonZeroU32`.

patriksvensson (dismissed):
LGTM 👍

patriksvensson (approved):

PR #1312, opened 2026-07-31
feat: add crypto sync handler crate

quexten:
Adds a new sync handler crate for the cryptographic items. This will be called by the sync implementation on mobile / clients.

This will be used both for the upcoming user-id field work, and also for preventing downgrades by the server to old cryptographic formats.


crates/bitwarden-crypto-sync-handler/Cargo.toml
```diff
@@ -0,0 +1,48 @@
+[package]
+name = "bitwarden-crypto-sync-handler"
+description = """
+temporary sync handler will live in this crate until sync fully moves to the sdk. at that point this logic will be moved over.
```
mzieniukbw: This aligns well with other team's features and crypto segregation. Why not make it permanent crate ?

crates/bitwarden-pm/src/lib.rs
```diff
@@ -85,9 +89,10 @@ impl PasswordManagerClient {
     pub fn new_with_sync(settings: Option<bitwarden_core::ClientSettings>) -> Self {
         let client = Self::new(settings);
 
-        client
-            .sync()
-            .register_sync_handler(Arc::new(FolderSyncHandler::from_client(&client.0)));
+        let sync = client.sync();
+        sync.register_sync_handler(Arc::new(FolderSyncHandler::from_client(&client.0)));
+        #[cfg(not(target_arch = "wasm32"))]
+        sync.register_sync_handler(Arc::new(CryptoSyncHandler::new(client.0.clone())));
```
mzieniukbw, line 94: ❓ Shouldn't this be the very first thing that happens in sync ? If we just rotated some keys it usually goes in pair with other rotated data. It feels natural to be first.

crates/bitwarden-crypto-sync-handler/src/crypto_sync_handler.rs
```diff
@@ -0,0 +1,210 @@
...
+pub struct CryptoSyncUserDecryption {
+    /// Unlock data for accounts that have a master password.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    #[cfg_attr(feature = "wasm", tsify(optional))]
+    pub master_password_unlock: Option<MasterPasswordUnlockData>,
+    /// Token allowing unlock after a V1 to V2 upgrade, when one is outstanding.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    #[cfg_attr(feature = "wasm", tsify(optional))]
+    pub v2_upgrade_token: Option<V2UpgradeToken>,
+}
```
mzieniukbw, line 61: ❓ What about WebAuthn PRF decryption options ?

crates/bitwarden-crypto-sync-handler/src/crypto_sync_handler.rs
```diff
@@ -0,0 +1,210 @@
...
+    if !state_bridge.is_bridge_registered() {
+        return;
+    }
+
+    match user_decryption.master_password_unlock.as_ref() {
+        Some(master_password_unlock) => {
+            state_bridge
+                .set_masterpassword_unlock_data(master_password_unlock)
+                .await;
+            state_bridge.set_kdf(&master_password_unlock.kdf).await;
```
mzieniukbw: ❓ why do we need separate KDF state ? I thought the idea is to move all individual states into unlock data or authentication data.
If we have to keep it separate for backwards compatibility, could we do that in TS instead ?

crates/bitwarden-crypto-sync-handler/src/crypto_sync_handler.rs
```diff
@@ -0,0 +1,210 @@
...
+        response: &bitwarden_api_api::models::SyncResponseModel,
+    ) -> Result<(), bitwarden_sync::SyncHandlerError> {
+        let user_decryption = response.user_decryption.as_deref();
+
+        let data = CryptoSyncData {
+            user_decryption: user_decryption.map(|d| CryptoSyncUserDecryption {
+                master_password_unlock: d.master_password_unlock.as_deref().and_then(|m| {
+                    MasterPasswordUnlockData::try_from(m)
+                        .inspect_err(|e| {
+                            warn!(error = ?e, "Sync response carried unparseable master password unlock data; treating it as absent")
```
mzieniukbw: ❓ What is the rationale for not aborting this function or even return Err ?
Are we supposed to handle crypto errors gracefully ? With the current implementation, if master password was the only unlock option, we will clear it from state, at which point, user will see blank lock screen, no error. I don't think that's any good.

mzieniukbw (changes requested):

crates/bitwarden-crypto-sync-handler/src/crypto_sync_handler.rs
quexten: I don't think we need a seprate KDF state, but currently clients uses it. I suppose we can drop it in clients completely in the follow-up PR; I'll check if there are any blockers..

quexten: @mzieniukbw While I agree we do not need the KDF state, we currently still need it. There are users that cannot have a masterpassword unlock data because they are TDE users. For these users, we could say "take *any* kdf" when generating their unlock data and setting it when offboarding or force-enabling KDF.

However, these users - TDE users without a MP, and kc users may still use PIN unlock. And pin unlock still uses the old pinkey method on mobile (or at least exposes it via the SDK, I did not check whether it's actually still used). Further, there is the "make_update_password" function, which is used on mobile to update the password. This should be migrated off of because it does not use MP unlock data / authentication data.

I have made a ticket here https://bitwarden.atlassian.net/browse/PM-41408, let's just keep this a port without behavior changes.

crates/bitwarden-pm/src/lib.rs
quexten: Ok, moving to happen first.

crates/bitwarden-crypto-sync-handler/src/crypto_sync_handler.rs
quexten: Initially, this was part of the same PR that adds the key id handling, and it was not needed there. But given that this is split out into a separate PR, it's reasonable to port over all of KM's domain.

Please note I'll keep organizations out for now because it is not entirely clear what we want the storage interface to look like yet, when taking into account v1/v2 orgs. I'd like to keep these out of the state bridge until we know what we want the storage to look like.

crates/bitwarden-crypto-sync-handler/src/crypto_sync_handler.rs
quexten: I'm not entirely sure yet either. If a sync partially has written (i.e lets say a user does a v2 upgrade and has newly encrypted ciphers and keys), and the sync already wrote some of the data to state, then the rest of it fails, that seems like it would cause an issue because we have an inconsistent state.

But maybe, we should divide into two categories of data, critical and non critical. If a cipher is missing from the sync, that seems OK, but if keys are missing, this is an issue. So maybe the pattern should be: Return error and abort sync for critical data like crypto (possibly log out), and ignore / skip for non critical data? What do you think?

mzieniukbw: That make sense. On crypto error during sync, within the core crypto data, the only reasonably, recoverable way out is like you suggest, a log out.
Maybe the definition of recoverable crypto error is: any data, that would prevent unlock and is not recoverable by data recovery tool ?

quexten: Well, there are two cases in which this can happen: Crypto is busted, and server is out of date (self hosted or vaultwarden / other custom server). While we don't support the latter, it does happen evidently by the recent vaultwarden related CS issues.

quexten: It will error now if the data does not parse from the request model.

crates/bitwarden-pm/src/lib.rs
mzieniukbw, line 99: ❓ As Claude already pointed out, we register sync handler in newly built SyncClient, but we don't return it, hence loose all the registrations as soon as we get out of this function.
Do we want to address it now or in later PR stack ?

mzieniukbw (approved):

crates/bitwarden-pm/src/lib.rs
quexten: Right, but that is already the case in `main`. I'm guessing this is done by @bitwarden/team-platform-dev as part of the sdk cli work.

dani-garcia (approved):

willmartian (approved):

PR #1317, opened 2026-07-31
[PM-34718] Implement bw receive / bw send receive

adudek-bw:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-34718

## 📔 Objective

Adds SendAccessKey (bitwarden-send), the anonymous-decrypt primitive that derives a Send's symmetric key from the URL fragment without any account key store or logged-in user. Wires up bw receive and its bw send receive alias: URL/API/identity resolution, password resolution (--password/--passwordenv/--passwordfile with interactive fallback gated on BW_NOINTERACTION), the password and email-OTP send-access token negotiation flow, and file download/decrypt/save via a new --output flag (replacing the previous --obj path flag to match bw send get's existing convention). Adds --fullObject for dumping the decrypted Send as JSON. Leaves bw send get --output as an explicit not-yet-implemented error with an expanded comment, since no owner-scoped download endpoint exists server-side.


crates/bw/tests/receive.rs
```diff
@@ -0,0 +1,643 @@
...
+        vec![
+            "send".to_string(),
+            "receive".to_string(),
+            "--help".to_string(),
+        ],
+    ] {
+        let output = run_bw(args.clone(), vec![]).await;
+        assert!(output.status.success(), "`bw {args:?}` should succeed");
+        let stdout = stdout(&output);
+        for flag in ["--password", "--passwordenv", "--passwordfile", "--output"] {
```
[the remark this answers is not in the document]
adudek-bw: @AikidoSec ignore: false positive — "hunter2" is a placeholder test password used throughout this codebase's tests, not a real credential.

crates/bw/tests/receive.rs
```diff
@@ -0,0 +1,643 @@
...
+            "text": { "text": encrypt_string("password gated"), "hidden": false },
+        })))
+        .mount(&server)
+        .await;
+
+    let output = run_bw(
+        vec![
+            "receive".to_string(),
+            receive_url(&server),
+            "--password".to_string(),
```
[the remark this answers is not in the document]
adudek-bw: @AikidoSec ignore: false positive — "hunter2" is a placeholder test password used throughout this codebase's tests, not a real credential.

crates/bitwarden-send/src/access.rs
```diff
@@ -74,8 +84,208 @@ pub struct SendFileDownloadData {
...
 }
 
+// ===== Decrypted views of an anonymous send access =====
+
+/// Plaintext view of a [`SendAccessResponse`], produced by
+/// [`SendAccessKey::decrypt_response`].
+///
+/// Mirrors the legacy CLI's `SendAccessResponse` output shape (`apps/cli`) so that a
+/// JSON dump of a received send stays recognizable to existing scripts, with two
+/// additions the wire response already carries (`expirationDate`, `creatorIdentifier`).
```
harr1424: ⛏️ It's a personal problem, but I really detest the way Claude throws around "wire" as if it's immedietely obvious what this refers to. 
❓ What does "wire response" refer to here? Is it the encrypted Send? A SendView? 

Just a few changed words could really increase clairty here.

crates/bitwarden-send/src/access.rs
```diff
@@ -74,8 +84,208 @@ pub struct SendFileDownloadData {
...
+/// Symmetric key material for an anonymous Send access, derived entirely from the URL
+/// fragment — no account key store or logged-in user involved. This is the only Send flow
+/// whose key doesn't come from the user's key store, hence a standalone type rather than a
+/// [`bitwarden_crypto::KeyStoreContext`] slot.
+///
+/// The key is opaque by design: callers need to *use* it three ways (hash a password,
+/// decrypt a response, decrypt a downloaded blob) but never need to *see* it.
+pub struct SendAccessKey {
+    /// The raw fragment key. Retained because the send password hash is salted with the
+    /// *unstretched* key, not with [`Self::key`].
```
harr1424: ❓ What does "unstretched" mean here?

crates/bitwarden-send/src/access.rs
```diff
@@ -493,4 +703,337 @@ mod tests {
             GetFileDownloadDataError::Api(_)
         ));
     }
+
+    // ===== SendAccessKey =====
+
+    mod send_access_key {
```
harr1424, line 575: ⚠️ This seems to be lacking documentation per our [guidelines](https://contributing.bitwarden.com/contributing/code-style/rust#modules)

crates/bw/src/tools/receive.rs
harr1424, line 1: ⚠️  There is a vulnerability pending direction from product that is very relevant to this file: https://bitwarden.atlassian.net/browse/PM-40120

crates/bw/src/tools/receive.rs
```diff
@@ -0,0 +1,807 @@
...
+/// itself is read from `{api}`. Precedence:
+///
+/// 1. A known Bitwarden cloud host (exact match — see [`CLOUD_HOSTS`]).
+/// 2. The locally configured deployment (`bw config server`) when the link's origin matches it —
+///    explicit `api`/`identity` overrides win, otherwise the base is suffixed.
+/// 3. Otherwise `<origin>/api` + `<origin>/identity`, the single-domain self-host convention and
+///    the legacy CLI's final fallback.
+///
+/// Pure so the precedence can be unit-tested without a client or a config file on disk.
+fn resolve_urls(url: &Url, config: Option<&ConfigFile>) -> (String, String) {
```
harr1424, line 150: ⚠️ This would be a good place to implement whatever is decided on the mentioned vulnerability

crates/bw/src/tools/receive.rs
```diff
@@ -0,0 +1,807 @@
...
+}
+
+/// Negotiate a send-access token, prompting for whatever credential the server says the Send
+/// needs. Mirrors the legacy `attemptAccess`: ask with no credentials first and branch on the
+/// typed `send_access_error_type` the server returns.
+///
+/// Legacy wraps every token request in a 3-attempt retry loop (`getTokenWithRetry`) for a
+/// `{kind: "expired"}` case that only exists because browser/extension storage caches tokens
+/// between calls. `bw receive` is a single-shot process with no token cache — the token is minted
+/// and consumed inside this one invocation — so there is nothing to expire and no retry here.
```
harr1424: ❓ The legacy CLI absolutely does cache Send access tokens - this comment would mean a significant behavioral divergence and is [patently false](https://github.com/bitwarden/clients/blob/f97b15bc4c8a0aa5236eb969e98a718c0330898b/libs/common/src/auth/send-access/services/default-send-token.service.ts#L179).

crates/bw/src/tools/receive.rs
```diff
@@ -0,0 +1,807 @@
...
+/// The `send_access_error_type` of an `invalid_request` response, if that's what this is.
+fn invalid_request_type(err: &SendAccessTokenError) -> Option<&SendAccessTokenInvalidRequestError> {
+    match err {
+        SendAccessTokenError::Expected(SendAccessTokenApiErrorResponse::InvalidRequest {
+            send_access_error_type,
+            ..
+        }) => send_access_error_type.as_ref(),
+        _ => None,
+    }
+}
```
harr1424: 🎨 It took some time to figure out why this exists, it might be worth adding or re-wording the documentation. This could also apply to the helper function immediately below.

crates/bw/src/tools/send.rs
```diff
@@ -292,9 +281,22 @@ impl BwCommand for SendArgs {
...
+            //   - There is no owner-scoped download-URL endpoint server-side. Only the two
+            //     anonymous send-access endpoints exist, so the only way to make this work today is
+            //     to route the Send's *owner* through the same endpoint recipients use — which
+            //     increments the Send's `access_count` and burns one of the recipient's allowed
+            //     accesses just to let the owner look at their own file. That's a behavior change,
+            //     not a parity fix.
+            //   - It isn't a parity gap either: the legacy CLI never implemented this flag.
+            //     `SendGetCommand.run` declares `--output` and then never reads `options.output`.
+            // Implementing it needs a real owner-scoped server endpoint first; that requires its
+            // own ticket.
```
harr1424: ⚠️ I have a strong preference this is re-worded by a human - this reads like Claude interviewing a developer and is raising an alarm over a behavioral change. Why does the `--output` flag affect the `access_count`? None of this is very clear. 

EDIT I think I understand what this is saying now, why don't we just adhere to what the legacy CLI does?

crates/bitwarden-send/src/access.rs
adudek-bw: I think it means it wasn't run through the KDF, I'm going through and clarifying a lot of the claude comments lol

crates/bw/src/tools/receive.rs
harr1424: @adudek-bw was there any decision made regarding how this vulnerability applies to Sends using the SDK logic?

crates/bitwarden-send/src/access.rs
harr1424: I agree this [makes sense](https://en.wikipedia.org/wiki/Key_stretching) but since it would be nice to document this with certainty.

crates/bitwarden-send/src/access.rs
harr1424: Is this still lacking documentation?

crates/bw/src/tools/receive.rs
harr1424: ❓ Can you explain this? 
>`bw receive` is a single-shot process with no token cache — the token is minted nd consumed inside this one invocation — so there is nothing to expire and no retry here.

Is this an intentional behavioral change? My earlier comment in this thread linked to very different behavior in the legacy TS CLI.

crates/bw/src/tools/receive.rs
harr1424: ❓ Would it make sense to resolve this now? Instead of re-introducing this on the SDK path?

harr1424 (commented):
Some of the comments and open questions from the last review cycle still appear unaddressed.

crates/bw/src/tools/send.rs
harr1424: ❓ I am still not sure why we aren't matching the legacy TS CLI behavior for this command exactly. I don't think we should add this command which will return an error in all cases. 

⚠️ As a user I would really question why this was shipped to a production release if it isn't implemented. 

❓ How does the current TS CLI command work? Is there some reason blocking that same approach here?

crates/bw/src/tools/receive.rs
adudek-bw: resolve_urls here already requests the token from the server specified in the Send URL, so it looks like we're good to go :)

crates/bw/src/tools/receive.rs
adudek-bw: resolve_urls already implements the decided fix, updating the comments :)

crates/bw/src/tools/send.rs
adudek-bw: To clarify the scope first: this doesn't error in all cases, only --output does. Plain bw send get works normally, the error only fires when --output is passed (send.rs:286-288).

On why not match legacy exactly: legacy's send.program.ts declares --output as a flag, but SendGetCommand.run never reads it — so legacy's actual behavior is to silently accept the flag and do nothing with it. No file gets written, no error, no indication to the user that their request wasn't honored. Matching that here would mean the same silent no-op.

The reason it's not implemented outright: it needs an owner-scoped download endpoint that doesn't exist server-side yet. The only download route that exists is the anonymous send-access endpoint bw receive uses, which isn't reusable for an authenticated owner fetching their own Send.

Given the choice between silently dropping the request (matching legacy) or telling the user immediately that it's not supported yet, I went with the explicit error, it's a smaller surface change and it doesn't leave the user assuming a file was saved when it wasn't. Open to reconsidering if you'd rather match legacy's silent-no-op behavior for exact parity, but I think the explicit error is the safer default here.

dereknance (dismissed):

crates/bw/src/tools/receive.rs
harr1424: Thank you!

crates/bw/src/tools/receive.rs
harr1424: Was there any resolution on this? Dropping the caching is a behavioral change, I am happy to sign off on this change if it was intentional but my former two comments don't appear to be addressed.

crates/bitwarden-send/src/access.rs
harr1424: 🎨 Not a blocker but this doc comment was never updated for clarity

crates/bw/src/tools/send.rs
harr1424: Thanks for explaining, if the flag is truly a no-op, and has never worked in the existing TS CLI, then maybe it makes more send to drop it enitrely?

harr1424 (commented):
There is still one item I consider a blocker that has gone unaddressed across multiple review cycles, this is the loss of any caching mechanism. If this is deliberate and product has approved of the behavioral change I am comfortable approving this. 

Optionally, it may make sense to remove the `--output` flag for `bw send get` if this flag will only ever result in an error.

adudek-bw:
> There is still one item I consider a blocker that has gone unaddressed across multiple review cycles, this is the loss of any caching mechanism. If this is deliberate and product has approved of the behavioral change I am comfortable approving this.
> 
> Optionally, it may make sense to remove the `--output` flag for `bw send get` if this flag will only ever result in an error.

Ah okay, I totally misunderstood this, I'll reimplement the caching because I don't think there was a sign off on that. As for the output, I think that's a good point! That can be removed.

crates/bw/src/tools/receive.rs
```diff
@@ -119,25 +159,41 @@ fn parse_send_url(url: &Url) -> Result<(String, String)> {
...
+/// correctness alone requires — [`send_access_token_cache`]'s `(resolved_host, send_id)` key
+/// already prevents a token from being looked up under a different host than the one it was
+/// minted against — but it keeps this CLI from ever writing a persistent, on-disk credential for
+/// an arbitrary, untrusted Send-link host, not just from misusing one it already wrote.
 ///
 /// Pure so the precedence can be unit-tested without a client or a config file on disk.
-fn resolve_urls(url: &Url, config: Option<&ConfigFile>) -> (String, String) {
+///
+/// See the module-level PM-40120 note: this unconditional host resolution — always the link's
+/// own host, never a separately configured environment — is the decided fix, not a gap.
```
harr1424:
```suggestion
/// own host, never a separately configured environment.
```

🎨 I try to read these before they end up in a PR - there is probably more that could be done here for our future selves trying to quickly understand this code.

crates/bw/src/tools/send_access_token_cache.rs
harr1424, line 1: 👍 In general these seem like a solid improvement over the existing legacy cache mechanism. Just a note that my [initial approach](https://github.com/bitwarden/clients/pull/22089) did something similar. As described in [this comment](https://bitwarden.atlassian.net/browse/PM-40120?focusedCommentId=168055) that approach didn't fully mitigate the vulnerability, and was perceived as overly complicated. 

As long as tokens cannot cross domain we are good.

crates/bw/src/tools/send.rs
harr1424, line 1: ❓ Was all of the code removed in this file related to removing the `--output` flag?

harr1424 (dismissed):

crates/bw/src/tools/send.rs
adudek-bw: Not all, some was deduplicating the ReceiveArg

harr1424 (approved):

dereknance (approved):

PR #1347, opened 2026-08-03
Downgrade some packages that are too new

dani-garcia:
## 📔 Objective
 
The recent lockfile PR #1346 updated some packages to versions that are too new for aikido, causing local builds to fail. This PR just ran `cargo update` to downgrade the packages to the latest available versions.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -378,12 +378,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "72b3254f16251a8381aa12e40e3c4d2f0199f8c6508fbecb9d91f575e0fbb8c6"
 
 [[package]]
-name = "base64"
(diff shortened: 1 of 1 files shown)


quexten (approved):
We should probably either add aikido to the CI of lockfile updates, or test the updates locally to prevent this in the future.

dani-garcia:
@quexten The strange thing is that renovate also has a 7 day cooldown, so there might be some disagreement on what 7 days means between renovate and aikido:
https://github.com/bitwarden/renovate-config/blob/59308866ffea61e01212409040cd5b51b8ecbe12/non-pinned.json#L17

quexten:
@dani-garcia Then we should probably add a grace period in rennovate and set it *larger than* akido (8, 9, 10 days).

PR #1353, opened 2026-08-05
[PM-41073] Remove nested EncStrings on decrypted CipherViews

nikwithak:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-41073

## 📔 Objective

When decrypting a `Cipher` into a `CipherView`, we currently leave the following fields in an encrypted state as an `EncString`, for later decryption. This breaks the contract of the `Decryptable` trait, which should leave us with a fully decrypted object. This PR updates those implementations and types to be fully decrypted and exposed when in a decrypted `CipherView` state.

Currently in draft, as additional changes may be needed if any issues arise while working on the client PRs.

## 🚨 Breaking Changes

Yes - this has breaking changes with the SDK, by changing the contracts with the decrypted types. Changes will be needed in all clients currently using the SDK to consume the new approach.

Clients will need to update their usage of Fido2 credentials and TOTP tokens, and can remove the extra step of decrypting them on access, and use the values, which are now decrypted fully, directly. Any clients directly using the cipher key will also need to update that use, as it is no longer wrapped by the user/org key inside a  `CipherView`.

I'll update this with links to those PRs once they are ready.

Clients PR (DRAFT / WIP):  https://github.com/bitwarden/clients/pull/22245


nikwithak:
Draft clients PR Here: https://github.com/bitwarden/clients/pull/22245 (still WIP)

crates/bitwarden-vault/src/cipher/cipher.rs
```diff
@@ -539,8 +538,7 @@ pub struct CipherListView {
     pub folder_id: Option<FolderId>,
     pub collection_ids: Vec<CollectionId>,
 
-    /// Temporary, required to support calculating TOTP from CipherListView.
-    pub key: Option<EncString>,
+    pub key: Option<String>,
```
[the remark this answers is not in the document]
quexten: Accepted from KM side.

crates/bitwarden-vault/src/totp.rs
```diff
@@ -730,6 +730,18 @@ mod tests {
 
     #[test]
     fn test_generate_totp_cipher_view() {
+        let key = SymmetricCryptoKey::try_from("w2LO+nwV4oxwswVYCxlOfRUseXfvU03VzvKQHrqeklPgiMZrspUe6sOBToCnDn9Ay0tuCBn8ykVVRb7PWhub2Q==".to_string()).unwrap();
+        let key_store = create_test_crypto_with_user_key(key);
+
+        // Decrypt the TOTP URI which was previously stored as an EncString in LoginListView.totp
+        let enc_totp: EncString = "2.hqdioUAc81FsKQmO1XuLQg==|oDRdsJrQjoFu9NrFVy8tcJBAFKBx95gHaXZnWdXbKpsxWnOr2sKipIG43pKKUFuq|3gKZMiboceIB5SLVOULKg2iuyu6xzos22dfJbvx0EHk=".parse().unwrap();
+        let totp_str: String = {
```
quexten: Since this is a static string, why not just replace this logic by the totp secret? Don't need to dynamically decrypt a static secret every time.

crates/bitwarden-vault/src/totp.rs
```diff
@@ -86,12 +86,12 @@ pub fn generate_totp(key: String, time: Option<DateTime<Utc>>) -> Result<TotpRes
 ///
 /// See [generate_totp] for more information.
 pub fn generate_totp_cipher_view(
-    ctx: &mut KeyStoreContext<KeySlotIds>,
+    _ctx: &mut KeyStoreContext<KeySlotIds>,
```
quexten: Can we drop the ctx? Or are we trying to not impact consumers of the interface for now?

crates/bitwarden-exporters/src/models.rs
```diff
@@ -95,7 +95,7 @@ impl From<PasswordHistoryView> for crate::PasswordHistory {
 /// Convert a `LoginView` into a `crate::Login`.
 fn from_login(
     view: &CipherView,
-    key_store: &KeyStore<KeySlotIds>,
```
quexten, line 98: Can we drop this or are we trying to not impact consumers for now?

crates/bitwarden-vault/src/cipher/cipher.rs
```diff
@@ -629,35 +627,32 @@ pub struct ListOrganizationCiphersResult {
...
-// ciphertext (`key`, the cipher content-encryption key wrapped under the decrypting key) and copies
-// it through unchanged (`key: cipher_view.key` below) instead of re-wrapping it under `key`. As a
-// result decrypt(K) -> encrypt(K1) -> decrypt(K1) does NOT round-trip.
 impl CipherView {
+    pub(crate) fn load_cipher_key_slot(
+        &self,
+        ctx: &mut KeyStoreContext<KeySlotIds>,
+        wrapping_key: SymmetricKeySlotId,
+    ) -> Result<SymmetricKeySlotId, CryptoError> {
+        match &self.key {
```
quexten, line 639: Hopefully can simplify with the follow-up of enabling cipher keys by default.

crates/bitwarden-vault/src/cipher/cipher.rs
```diff
@@ -629,35 +627,32 @@ pub struct ListOrganizationCiphersResult {
...
+            Some(b64) => attachment::load_symmetric_key_slot(ctx, b64),
```
quexten: Seems weird to use the attachment-owned function here? 🤔

quexten: I'd just use ctx methods directly.

crates/bitwarden-vault/src/cipher/cipher.rs
```diff
@@ -667,9 +662,11 @@ impl CipherView {
             organization_id: cipher_view.organization_id,
             folder_id: cipher_view.folder_id,
             collection_ids: cipher_view.collection_ids,
-            // ⚠️ pass-through of wrapped key-bound ciphertext — see the contract-violation note
-            // above.
-            key: cipher_view.key,
+            key: cipher_view
+                .key
+                .as_ref()
+                .map(|_| ctx.wrap_symmetric_key(key, ciphers_key))
```
quexten, line 663: I just want to make sure I understand this logic correctly.
- `key` is either user key or org key
- `ciphers_key` is either user-key, org-key, or cipher-key (if cipherview.key is present)
- then, the `as_ref().map()` does the following
  - if cipher_view.key is not present (i.e when ciphers-key is a user-key or org key) map to none, becaus as_ref is none
  - if cipher_view.key is present, then wrap the ciphers_key (which must be the cipher-key) with the user/org key

The logic seems a bit complex / brittle still, but we can fix that as part of the default enablement of cipher keys.

crates/bitwarden-vault/src/cipher/cipher.rs
```diff
@@ -908,17 +908,14 @@ impl CipherView {
...
-        let old_unwrapping_key = self.key_identifier();
-        let old_ciphers_key = Cipher::decrypt_cipher_key(ctx, old_unwrapping_key, &self.key)?;
-
         let new_key = ctx.generate_symmetric_key();
 
-        self.reencrypt_attachment_keys(ctx, old_ciphers_key, new_key)?;
-        self.reencrypt_fido2_credentials(ctx, old_ciphers_key, new_key)?;
-
-        self.key = Some(ctx.wrap_symmetric_key(wrapping_key, new_key)?);
+        // Attachment keys are stored as raw base64 on AttachmentView; re-wrapping under the
```
quexten: Do we need this comment if attachment keys are no longer handled here?

crates/bitwarden-vault/src/cipher/cipher.rs
```diff
@@ -908,17 +908,14 @@ impl CipherView {
...
+        // new cipher key happens inside CompositeEncryptable at encrypt time.
+        #[allow(deprecated)]
+        let raw = ctx.dangerous_get_symmetric_key(new_key)?.clone();
+        self.key = Some(raw.to_base64().to_string());
```
quexten: I think you can just put SymmetricKey instead of string on the FFI struct. It should convert fine, and we made the FFI constructs. That would save you a bunch of conversion.

crates/bitwarden-vault/src/cipher/cipher.rs
```diff
@@ -936,138 +933,73 @@ impl CipherView {
...
     pub fn reencrypt_cipher_keys(
         &mut self,
-        ctx: &mut KeyStoreContext<KeySlotIds>,
-        new_wrapping_key: SymmetricKeySlotId,
+        _new_wrapping_key: SymmetricKeySlotId,
     ) -> Result<(), CipherError> {
-        let old_key = self.key_identifier();
-
         // If any attachment is missing a key we can't reencrypt the attachment keys
         if self.attachments.iter().flatten().any(|a| a.key.is_none()) {
```
quexten, line 952: Probably want to clean this up at some point, because this does not (in opposition to the naming) actually re-encrypt anything.

crates/bitwarden-vault/src/cipher/login.rs
```diff
@@ -128,9 +128,11 @@ pub struct Fido2CredentialView {
     pub key_type: String,
     pub key_algorithm: String,
     pub key_curve: String,
-    // This value doesn't need to be returned to the client
-    // so we keep it encrypted until we need it
-    pub key_value: EncString,
+    /// The raw EC private key material for this passkey credential.
```
quexten: Replace `EC` with `raw PKCS8 Private key`. It is not guaranteed that this is an elliptic curve key, despite that being currently the case. If we switch to post-quantum passkeys in the future, it would be an ML-DSA private key seed.

crates/bitwarden-vault/src/cipher/login.rs
```diff
@@ -128,9 +128,11 @@ pub struct Fido2CredentialView {
...
+    /// Previously kept as `EncString` until explicitly requested; it is now decrypted eagerly as
```
quexten: This paragraph is probably not needed.

quexten (commented):
Will continue review later, but posting the first set of comments. Good work so far, this will massively simplify the architecture.

crates/bitwarden-vault/src/cipher/cipher_client/edit.rs
```diff
@@ -72,7 +72,7 @@ pub struct CipherEditRequest {
     pub revision_date: DateTime<Utc>,
     pub archived_date: Option<DateTime<Utc>>,
     pub attachments: Vec<AttachmentView>,
-    pub key: Option<EncString>,
+    pub key: Option<String>,
```
quexten, line 69: do we want to consider just using symmetrickey here instead of string? would save you a bunch of conversion work.

crates/bitwarden-vault/src/cipher/blob/encryption.rs
```diff
@@ -36,15 +36,15 @@ impl From<BlobEncryptionError> for CryptoError {
...
-/// Seals a `CipherView` into an opaque blob string, using `wrapping_key` as
-/// the outer key that protects the cipher's wrapped CEK.
+/// Seals a `CipherView` into an opaque blob string under the given `cipher_key` slot.
+/// The caller is responsible for loading the key slot before calling (e.g. via
+/// `CipherView::load_cipher_key_slot`); this avoids allocating a duplicate slot.
 fn seal_cipher(
     view: &CipherView,
     ctx: &mut KeyStoreContext<KeySlotIds>,
-    wrapping_key: SymmetricKeySlotId,
+    cipher_key: SymmetricKeySlotId,
```
quexten: Is cipher-key here guaranteed to be the cipher key, or is it possible for the user-key / org key to be used here?

crates/bitwarden-vault/src/cipher/attachment.rs
```diff
@@ -75,46 +75,7 @@ pub struct AttachmentView {
...
-    /// fully migrated to the SDK.
-    ///
-    /// **Ticket**: <https://bitwarden.atlassian.net/browse/PM-23005>
-    ///
-    /// Do not rely on this field for long-term use.
-    #[cfg(feature = "wasm")]
-    pub decrypted_key: Option<String>,
-}
-
-impl AttachmentView {
```
quexten, line 93: Nice! So much easier!

crates/bitwarden-vault/src/cipher/attachment.rs
```diff
@@ -75,46 +75,7 @@ pub struct AttachmentView {
...
-        ctx: &mut KeyStoreContext<KeySlotIds>,
-        old_key: SymmetricKeySlotId,
-        new_key: SymmetricKeySlotId,
-    ) -> Result<(), CryptoError> {
-        for attachment in attachment_views {
-            attachment.reencrypt_key(ctx, old_key, new_key)?;
-        }
-        Ok(())
-    }
+    pub key: Option<String>,
```
quexten: SymmetricKey?

crates/bitwarden-vault/src/cipher/attachment.rs
```diff
@@ -198,36 +161,33 @@ fn size_name(size: usize) -> String {
     format!("{} {}", size_round, units[unit])
 }
 
+/// Parses a base64-encoded raw symmetric key and loads it into an ephemeral key-store slot.
+/// Used by both the attachment encrypt and decrypt paths.
+pub(super) fn load_symmetric_key_slot(
+    ctx: &mut KeyStoreContext<KeySlotIds>,
+    b64: &str,
+) -> Result<SymmetricKeySlotId, CryptoError> {
+    let raw = SymmetricCryptoKey::try_from(b64.to_owned())?;
```
quexten: If you use symmetrickey as teh FFI struct, you don't need this function since you already have a symmetrickey, you can just call ctx_add_local_symmetric_key

crates/bitwarden-vault/src/cipher/attachment.rs
```diff
@@ -336,9 +286,7 @@ pub(crate) fn decrypt_attachments_with_failures(
                     size: attachment.size.clone(),
                     size_name: attachment.size_name.clone(),
                     file_name: None,
-                    key: attachment.key.clone(),
-                    #[cfg(feature = "wasm")]
-                    decrypted_key: None,
+                    key: None,
```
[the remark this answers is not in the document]
quexten: Seems reasonable. This is a failed decryption and we cannot reasonable place a key here. Ideally we would have an enum of different struct representations (failed, not failed) but this is OK too.

quexten (commented):
Okay, review complete from my end.

crates/bitwarden-exporters/src/models.rs
shane-melton: I lean towards dropping it, but it does admittedly increase the scope/diff by going further up the chain.

crates/bitwarden-vault/src/totp.rs
nikwithak: [a5a508c4751448557f52de0c2dbba5006225067d](https://github.com/bitwarden/sdk-internal/pull/1353/changes/a5a508c4751448557f52de0c2dbba5006225067d)

crates/bitwarden-vault/src/totp.rs
nikwithak: I was trying not to impact consumers, but given that we are already introducing breaking changes anyway, I'll go ahead and clean up the tech debt all in one go. 

[a4eaf60beef2e0280c75f4be1616752107b7c6b4](https://github.com/bitwarden/sdk-internal/pull/1353/changes/a4eaf60beef2e0280c75f4be1616752107b7c6b4)

crates/bitwarden-exporters/src/models.rs
nikwithak: [a4eaf60beef2e0280c75f4be1616752107b7c6b4](https://github.com/bitwarden/sdk-internal/pull/1353/changes/a4eaf60beef2e0280c75f4be1616752107b7c6b4)

crates/bitwarden-vault/src/cipher/cipher.rs
nikwithak: [68872e892f81529aaf1c794535b5ca6185f79b40](https://github.com/bitwarden/sdk-internal/pull/1353/changes/68872e892f81529aaf1c794535b5ca6185f79b40)

shane-melton: Agreed, using the attachment-owned function adds confusion here, we should just use the ctx method.

crates/bitwarden-vault/src/cipher/cipher.rs
nikwithak: Yes that is how it's working - agreed that it's a little bit brittle. The presence of `cipher_view.key` decides whether we wrap the key (since at that point `ciphers_key` represents the `cipher_key`

It does this determintion via the `map` call, but I could change to an `if cipher_view.key.is_some() {..} else {None}` or a `match` if that would make it clearer

crates/bitwarden-vault/src/cipher/cipher.rs
nikwithak: [58d71166818e8ca5665938ee30407ea577f76ee1](https://github.com/bitwarden/sdk-internal/pull/1353/changes/58d71166818e8ca5665938ee30407ea577f76ee1)

crates/bitwarden-vault/src/cipher/login.rs
nikwithak: [b7af535df43910006bea9dfa895caff6dce82002](https://github.com/bitwarden/sdk-internal/pull/1353/changes/b7af535df43910006bea9dfa895caff6dce82002)

crates/bitwarden-vault/src/cipher/login.rs
nikwithak: [b7af535df43910006bea9dfa895caff6dce82002](https://github.com/bitwarden/sdk-internal/pull/1353/changes/b7af535df43910006bea9dfa895caff6dce82002)

crates/bitwarden-vault/src/cipher/blob/encryption.rs
nikwithak: In theory it should be the cipher key, since all ciphers *should* be using a cipher key at this point, but it isn't a guarantee since the caller can pass in any key. Will change back to wrapping_key.

crates/bitwarden-vault/src/cipher/cipher.rs
```diff
@@ -936,138 +934,63 @@ impl CipherView {
...
-    ) -> Result<(), CryptoError> {
-        if let Some(attachments) = &mut self.attachments {
-            AttachmentView::reencrypt_keys(attachments, ctx, old_key, new_key)?;
-        }
-        Ok(())
+    #[allow(missing_docs)]
+    #[deprecated(
+        note = "Use `get_fido2_credentials` instead - Fido2Credentials are no longer encrypted in `CipherView`"
+    )]
+    pub fn decrypt_fido2_credentials(&self) -> Vec<Fido2CredentialView> {
```
shane-melton: :art: I think we should be able to just remove this. We're the only consumers of it internally.

crates/bitwarden-vault/src/cipher/cipher.rs
```diff
@@ -936,138 +934,63 @@ impl CipherView {
...
-    ) -> Result<(), CryptoError> {
-        if let Some(login) = self.login.as_mut() {
-            login.reencrypt_fido2_credentials(ctx, old_key, new_key)?;
-        }
-        Ok(())
+            .cloned()
+            .unwrap_or_default()
     }
 
     /// Moves the cipher to an organization by re-encrypting the cipher keys with the organization
```
shane-melton: :pick: Comment can be updated, we're no longer re-encrypting here now.

crates/bitwarden-vault/src/cipher/login.rs
```diff
@@ -128,9 +128,10 @@ pub struct Fido2CredentialView {
     pub key_type: String,
     pub key_algorithm: String,
     pub key_curve: String,
-    // This value doesn't need to be returned to the client
-    // so we keep it encrypted until we need it
-    pub key_value: EncString,
+    /// The raw private key material for this passkey credential.
+    /// Callers that receive `Fido2CredentialView` over a binding boundary should
+    /// treat this field with the same care as any private key material.
+    pub key_value: String,
```
shane-melton, line 134: 💭  Hmm, this is makes `Fido2CredentialFullView` obsolete. We should look at deprecating/removing that view. Maybe not in this PR if it increases the scope too much (IDE is showing around ~44 usages that would need to be updated)

crates/bitwarden-vault/src/cipher/attachment.rs
shane-melton: I'm open to the stronger type, though when I was experimenting with it for my own understanding I was getting Unffi binding errors.

shane-melton (commented):
Partial review that I'll pick up tomorrow, but things are looking good!

crates/bitwarden-vault/src/cipher/attachment.rs
quexten: Okay, let's skip this for now then, though we (km) should fix it at some point.

crates/bitwarden-vault/src/cipher/login.rs
nikwithak: Good callout - agreed, let's remove it as a follow-up task!

crates/bitwarden-vault/src/cipher/attachment.rs
nikwithak: I was able to get it working with the `SymmetricCryptoKey` type, by registering it in `crates/bitwarden-vault/src/uniffi_support.rs`

nikwithak:
@quexten @shane-melton I believe I've addressed all the comments now - ready for another look!

crates/bitwarden-vault/src/cipher/attachment.rs
```diff
@@ -330,15 +262,18 @@ pub(crate) fn decrypt_attachments_with_failures(
             Ok(decrypted) => successes.push(decrypted),
             Err(e) => {
                 tracing::warn!(attachment_id = ?attachment.id, error = %e, "Failed to decrypt attachment");
+                let recovered_key = attachment.key.as_ref().and_then(|attachment_key| {
```
shane-melton, line 265: :+1: Nice, we're still attempting to decrypt the key in case the filename fails above.

crates/bitwarden-vault/src/cipher/blob/encryption.rs
```diff
@@ -36,28 +36,28 @@ impl From<BlobEncryptionError> for CryptoError {
...
+/// `CipherView::load_cipher_key_slot`); this avoids allocating a duplicate slot.
 fn seal_cipher(
     view: &CipherView,
     ctx: &mut KeyStoreContext<KeySlotIds>,
     wrapping_key: SymmetricKeySlotId,
 ) -> Result<String, BlobEncryptionError> {
-    let cipher_key = Cipher::decrypt_cipher_key(ctx, wrapping_key, &view.key)?;
-    let blob = CipherBlobLatest::from_cipher_view(view, ctx, cipher_key)?;
-    seal_blob_content(blob, cipher_key, ctx)
+    let blob = CipherBlobLatest::from_cipher_view(view)?;
```
shane-melton, line 47: :+1: Glad this is simplified even further and has one fewer level of key shuffling.

crates/bitwarden-vault/src/cipher/blob/encryption.rs
```diff
@@ -36,28 +36,28 @@ impl From<BlobEncryptionError> for CryptoError {
...
 
-/// Seals a `CipherView` into an opaque blob string, using `wrapping_key` as
-/// the outer key that protects the cipher's wrapped CEK.
+/// Seals a `CipherView` into an opaque blob string under the given `wrapping_key` slot.
+/// The caller is responsible for loading the key slot before calling (e.g. via
+/// `CipherView::load_cipher_key_slot`); this avoids allocating a duplicate slot.
 fn seal_cipher(
     view: &CipherView,
     ctx: &mut KeyStoreContext<KeySlotIds>,
     wrapping_key: SymmetricKeySlotId,
```
shane-melton, line 45: :question: Thoughts on re-naming this `cipher_key` to avoid the ambiguity with the `wrapping_key` name below which is referring to the User/Org key?

crates/bitwarden-vault/src/cipher/blob/encryption.rs
```diff
@@ -36,28 +36,28 @@ impl From<BlobEncryptionError> for CryptoError {
...
 }
 
-/// Seals a constructed `CipherBlobLatest` under `cipher_key`, returning the
+/// Seals a constructed `CipherBlobLatest` under `wrapping_key`, returning the
 /// opaque string form. Shared by all `CipherBlobLatest` producers so they
 /// don't each re-implement the versioned-enum wrap + COSE seal + base64 chain.
 fn seal_blob_content(
     blob: CipherBlobLatest,
-    cipher_key: SymmetricKeySlotId,
+    wrapping_key: SymmetricKeySlotId,
```
shane-melton, line 56: :question: Same here, this should always be a cipher key and renaming so could potentially help avoid confusion?

crates/bitwarden-vault/src/cipher/blob/encryption.rs
```diff
@@ -213,14 +222,14 @@ pub(crate) fn decrypt_blob_cipher(
         password_history: None,
     };
 
-    blob.apply_to_cipher_view(&mut view, ctx, cipher_key)?;
+    blob.apply_to_cipher_view(&mut view)?;
```
shane-melton, line 225: :+1: So much nicer not to have to pass a context and key to what was supposed to be a simple mapper. :tada:

crates/bitwarden-vault/src/cipher/cipher_client/edit.rs
shane-melton: :+1: I think this would be beneficial, ust like you were able to do for the `AttachmentView.key`

crates/bitwarden-vault/src/cipher/cipher.rs
```diff
@@ -533,8 +533,7 @@ pub struct CipherListView {
     pub folder_id: Option<FolderId>,
     pub collection_ids: Vec<CollectionId>,
 
-    /// Temporary, required to support calculating TOTP from CipherListView.
-    pub key: Option<EncString>,
+    pub key: Option<String>,
```
shane-melton, line 536: :question: Could we just remove this? Or does it increase the scope? We never encrypt a `CipherListView` so it should never need the key.

shane-melton (commented):
Nice work! I think we're pretty much there, only a few final comments/questions.

quexten (dismissed):
One note: It may be beneficial to split the cipher-key out. Specifically, if the initial theory for the passkey corruption is correct, then this would currently - temporarily - make it worse until the KM side of the key-id tracking is done. 

Previously, the encrypted cipher key would fail to decrypt with the wrong user key, and an update was not possible. Now, because the cipher-key is not encrypted at all, thee cipher would be encrypted with the cipher key present for the wrong user-key, so the corruption type would expand to all passkey assertions that happen with a wrong user-key in sdk, rather than just the subset of assertions with a wrong user-key for ciphers without a cipher-key.

Once the KM work is done, this will be prevented either way.

crates/bitwarden-vault/src/cipher/blob/encryption.rs
nikwithak: I had renamed it to `cipher_key` originally, and changed it back based on this comment: https://github.com/bitwarden/sdk-internal/pull/1353#discussion_r3726072508

In actuality, though, it should always be the cipher key at this point. Happy to rename it back again

crates/bitwarden-vault/src/cipher/cipher.rs
nikwithak: Good callout - since we no longer need it to decrypt other values, I think we can remove it entirely.

crates/bitwarden-vault/src/cipher/blob/encryption.rs
shane-melton: Ah, I see, I'm good to keep it as is for now and I can rename in my removal of the cipher key feature flag when we _can_ guarantee it is always a `cipher_key`.

nikwithak: Too late: [fea9886d58232c52815e302a71dd5d488aa39be5](https://github.com/bitwarden/sdk-internal/pull/1353/changes/fea9886d58232c52815e302a71dd5d488aa39be5)

crates/bitwarden-vault/src/cipher/cipher.rs
nikwithak: [70b27930b40a292c4de8cb43de2eaf0bc8e5ba04](https://github.com/bitwarden/sdk-internal/pull/1353/changes/70b27930b40a292c4de8cb43de2eaf0bc8e5ba04)

crates/bitwarden-vault/src/cipher/blob/encryption.rs
shane-melton: That works too haha

shane-melton:
Looks like the linter/clippy aren't happy

shane-melton:
Still a few more unhappy CI jobs (if you're already on it, please ignore me, I was just watching the branch for my stacked PR)

shane-melton (dismissed):
🚀

dani-garcia (dismissed):

review-summary.md
```diff
@@ -0,0 +1,45 @@
+## 🤖 Bitwarden Claude Code Review
```
dani-garcia: You left the claude review file here, we should remove this.

shane-melton:
@nikwithak Can we pull in the latest main to address the conflicts? (Also we have that left over `review-summary.md` we should clean up)

crates/bitwarden-vault/src/cipher/cipher_client/mod.rs
```diff
@@ -801,11 +771,9 @@ mod tests {
             .unwrap()
             .clone();
 
-        // Ensure attachment key is still the same since it's protected by the cipher key
-        assert_eq!(
-            attachment.clone().key.as_ref().unwrap().to_string(),
-            attachment_view.key.as_ref().unwrap().to_string()
-        );
+        // The attachment key (raw bytes) is unchanged; it's re-wrapped under the cipher key at
```
quexten, line 774: Actually, you can add bitwarden-random with the dangerous-seeded-rng-for-testing feature in your dev dependencies. That way, you can seed the rng:

https://github.com/bitwarden/sdk-internal/blob/e91ca8971d8e22081432c8bd39621b8cb41f404e/crates/bitwarden-random/src/seeded.rs#L14

and the encryption result will be deterministic / you can assert against it after recording it once.

quexten (dismissed):

harr1424 (dismissed):
Tools owned changes look good ✅

nikwithak:
This needs another round of ✔️ after resolving merge conflicts.

quexten (approved):
I merged main into this since integration tests were failing.

dani-garcia (approved):

crates/bitwarden-vault/src/cipher/login.rs
```diff
@@ -488,12 +485,11 @@ impl Decryptable<KeySlotIds, SymmetricKeySlotId, LoginView> for Login {
...
-            // is a decrypted DTO, but `fido2_credentials` are copied through still
-            // encrypted (`self.fido2_credentials.clone()`) rather than decrypted,
-            // because `LoginView` stores them as the encrypted `Vec<Fido2Credential>`.
-            // Consumers must decrypt each credential separately.
-            fido2_credentials: self.fido2_credentials.clone(),
+            fido2_credentials: self
+                .fido2_credentials
+                .as_ref()
+                .map(|c| c.decrypt(ctx, key))
+                .transpose()?,
```
shane-melton: ❓ This causes `Login` to fail when only a fido2 credential fails, correct? Does that conflict with the "lenient" decryption this flow represents?

e.g. If a fido2 credential fails, it will bubble up to [lenient_decrypt_cipher_view](https://github.com/bitwarden/sdk-internal/blob/0e1c0a69f34ab8868b13054446d75c66934e1f8b/crates/bitwarden-vault/src/cipher/cipher.rs#L753) and then be swallowed by `.ok().flatten()`. Which effectively drops the entire login view for a single failed item within.

shane-melton (commented):
One last minute question

harr1424 (approved):

crates/bitwarden-vault/src/cipher/login.rs
shane-melton: To be addressed in: https://github.com/bitwarden/sdk-internal/pull/1518

shane-melton (approved):

PR #1360, opened 2026-08-06
Auth / SealedOpenOrgInviteData - Fix TS2552 in WASM bindings by colocating custom section

JaredSnider-Bitwarden:
## 🎟️ Tracking

n/a

## 📔 Objective

**Claude Summary**

Local WASM debug builds were failing with `TS2552: Cannot find name 'SealedOpenOrgInviteData'` in the generated `.d.ts`.

**Root cause.** The `#[wasm_bindgen(typescript_custom_section)]` declaring the `SealedOpenOrgInviteData` TS alias lived in `serialization.rs`, which contains only trait impls — no non-generic function reachable from a `#[wasm_bindgen]` export. Debug builds use many codegen units, so `wasm-ld` treats the unit as unreachable and drops it, taking the section's `#[used] static` with it. `wasm-bindgen` never sees the alias, so the generated `.d.ts` references a type it never declares.

**Why CI missed it.** The workspace `[profile.release]` sets `codegen-units = 1`, which merges every crate symbol into a single unit and eliminates the per-CU DCE that was dropping the section. CI runs `./build.sh -r` (release), so the bug was invisible in CI while breaking every local debug build.

**Fix.** Moved the `typescript_custom_section` block next to the `SealedOpenOrgInviteData` struct in `open_org_invite.rs`. The non-generic methods `SealedOpenOrgInviteData::seal`/`unseal` are called from `RegistrationClient::seal_open_org_invite_data` — a `#[wasm_bindgen]` export — so their codegen unit stays reachable in both profiles. Added a brief note to `.claude/CLAUDE.md` documenting the rule and the debug-vs-release trap so future contributors don't re-introduce the same shape of bug.

**Verified.** All four `build.sh` combinations pass (debug/release × OSS/commercial); `SealedOpenOrgInviteData` is declared and the reference resolves in every generated `.d.ts`. Also empirically isolated the mechanism: reverting the section to `serialization.rs` reproduces the debug failure; setting only `codegen-units = 1` on `[profile.dev]` (LTO still off) makes the section survive even in that broken placement — confirming `codegen-units = 1` alone is the mask, not LTO.

### DIFF
--- a/.claude/CLAUDE.md
+++ b/.claude/CLAUDE.md
@@ -47,6 +47,12 @@ Generated code:
   or the "Update API Bindings" GitHub workflow.
 - WASM npm packages (`@bitwarden/sdk-internal`, `@bitwarden/commercial-sdk-internal`) build via
   `crates/bitwarden-wasm-internal/build.sh` (`-r` release, `-b` commercial).
+- WASM TS aliases via `#[wasm_bindgen(typescript_custom_section)]` only reliably survive `wasm-ld`
+  in debug builds when colocated in the same source file as an externally-referenced struct/fn.
+  Debug uses many codegen units and can drop the whole unit containing the section; release masks
+  this via `codegen-units = 1` + `lto = true`. CI runs release (`-r`) — CI-green does not prove the
+  local debug build passes; run `build.sh` without `-r` when adding one. For types with no
+  downstream Rust consumer, declare the alias in `bitwarden-wasm-internal/src/custom_types.rs`.
 
 ## Architecture
 
--- a/.claude/CLAUDE.md
+++ b/.claude/CLAUDE.md
@@ -47,12 +47,11 @@ Generated code:
   or the "Update API Bindings" GitHub workflow.
 - WASM npm packages (`@bitwarden/sdk-internal`, `@bitwarden/commercial-sdk-internal`) build via
   `crates/bitwarden-wasm-internal/build.sh` (`-r` release, `-b` commercial).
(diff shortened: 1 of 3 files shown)


.claude/CLAUDE.md
```diff
@@ -47,6 +47,12 @@ Generated code:
   or the "Update API Bindings" GitHub workflow.
 - WASM npm packages (`@bitwarden/sdk-internal`, `@bitwarden/commercial-sdk-internal`) build via
   `crates/bitwarden-wasm-internal/build.sh` (`-r` release, `-b` commercial).
+- WASM TS aliases via `#[wasm_bindgen(typescript_custom_section)]` only reliably survive `wasm-ld`
```
dani-garcia: I think the trigger that makes it survive is specifically a non-generic function reachable from a wasm_bindgen export, I don't think structs survive into the object stage. Also LTO isn't involved here in my experience, it's just tthe codegen units.

dani-garcia (dismissed):
LGTM, just added some clarification  on the comment

.claude/CLAUDE.md
JaredSnider-Bitwarden: Adjusted with https://github.com/bitwarden/sdk-internal/pull/1360/changes/6102f2bb69bf4d1adbf544e01eb1e40cf5ec02d3

enmande (approved):

dani-garcia (approved):

SaintPatrck (approved):

PR #1361, opened 2026-08-07
Validate wasm debug builds

dani-garcia:
## 📔 Objective

Due to some linker shenanigans, a `typescript_custom_section` definition that is on a module without other exported functions fails to appear on the final TS code when building in debug mode. This is due to the fact that the linker doesn't load the module object file if it thinks it's not needed. The reason it only happens in debug mode is that release is using codegen-units = 1, and so everything gets compiled into the same object file. 

This PR adds a debug CI check to ensure we don't break it in the future. We also run the tests on it to ensure nothing is broken internally.

Requires #1360 to build. 

For reference on how to fix this weird linker issue, there are two options:
- Move the definition to a different module that has exported functions that are used elsewhere, so the module gets loaded
- Add dummy function that is globally accessible in that module to make sure the module gets loaded. The easiest way is to use extern C and no_mangle:
```rust
const TS_CUSTOM_TYPES: &'static str = r#"
export type SealedOpenOrgInviteData = Tagged<string, "SealedOpenOrgInviteData">;
"#;

pub extern "C" fn __dummy_sealed_open_org_invite_data() {}
```


quexten (commented):
Question: Do you think it makes sense to run the wasm integration tests against the debug builds too to test the FFI layer? Same with wasm2js builds (until we finally drop them)?

PR #1364, opened 2026-08-08
[PM-35785] Strongly typed policy enforcement

eliykat:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-35785

## 📔 Objective

Iterate on the policies crate to provide a better policy enforcement interface. This is done in 2 ways:
- represent the outcome of a policy enforcement decision through a dedicated return value, `EnforcedPolicy`, rather than simply filtering `PolicyView`s out of an array (which is what `filter_by_type` does today)
- handle per-policy data deserialization (some policies carry additional configuration data)

This PR is fairly large, but the commits are ordered to try to separate interesting vs. uninteresting changes. The commits, in order:

1. add comprehensive tests for the `filter_by_type` interface before refactoring - to ensure we don't cause a regression in existing consumers.
2. delete the runtime builder pattern, which overcomplicated the design. Replace it with a simple match statement on the `PolicyType` enum, which looks up the corresponding `Policy` implementation. Add implementations for each policy we have today.
3. add the new policy enforcement API - **this is the new business logic and (I suggest) where most of the review should focus**.
4. test cleanup.
5. (PR review feedback after this point)

See the expanded README for an overview of the crate and how to add and consume policies.

One design note: there are separate client interfaces for native rust vs. FFI consumers. This is to accommodate the per-policy data. As the native rust interface shows, we can resolve this at compile time (using generics) for consumers, and I want to keep that as the first-class interface rather than forcing everyone to use the enum and have redundant match statements everywhere.

The next step after this PR would be to introduce state so that consumers are not passing in the policyViews and organization contexts. The new interfaces are not intended to be used yet, as that will be a breaking change.

## 🚨 Breaking Changes

- `clients` - missing an `asUuid` call. Will be fixed by https://github.com/bitwarden/clients/pull/22392 after merge. EDIT: I missed some test code - see https://github.com/bitwarden/clients/pull/22443 as well.


djsmith85 (changes requested):
Platform-owned changes are limited to the added integration tests.

Please extend the CODEOWNERS file with this line
`crates/bitwarden-wasm-internal/integration-tests/tests/policies/** @bitwarden/team-admin-console-dev`

crates/bitwarden-policies/src/models.rs
```diff
@@ -54,3 +59,140 @@ pub struct OrganizationUserPolicyContext {
...
+            Some(raw) => serde_json::from_str(raw).unwrap_or_else(|e| {
+                tracing::warn!(
+                    policy_type = ?policy.policy_type(),
+                    organization_id = %view.organization_id,
+                    "Failed to parse policy data, falling back to default: {e}"
+                );
+                Default::default()
+            }),
+            None => Default::default(),
+        };
```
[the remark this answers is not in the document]
eliykat: I still need to go back and fill in all the data structs for existing policies (in follow-up PR). Once I do that, the end state should be that all policies with data have a `Data` type defined, and therefore no policy with `Data = ()` would receive data to parse. Then, it seems both rare and reasonable to log a warning here ("I received data when the policy doesn't expect it"). Open to feedback though.

@claude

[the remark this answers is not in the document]
eliykat: This case has been covered by using the debug level instead with a more accurate error message.

djsmith85 (approved):
Approving for platform, as there's no longer any changes platform owns.

@sven-bitwarden Please ensure to also review `crates/bitwarden-wasm-internal/integration-tests/tests/policies/policy-client.test.ts` as the CODEOWNERS changes are only enforced after this has merged.

crates/bitwarden-wasm-internal/integration-tests/tests/policies/policy-client.test.ts
```diff
@@ -0,0 +1,260 @@
...
+    id,
+    role: options.role ?? OrganizationUserType.User,
+    status: options.status ?? OrganizationUserStatusType.Confirmed,
+    enabled: options.enabled ?? true,
+    usePolicies: options.usePolicies ?? true,
+    isProviderUser: options.isProviderUser ?? false,
+  };
+}
+
+describe("PolicyClient", () => {
```
quexten, line 65: Nice! Testing the FFI boundary!

crates/bitwarden-policies/src/policy_client.rs
```diff
@@ -38,34 +25,123 @@ impl Default for PolicyClient {
...
+/// Native rust callers should use the non-erased interfaces instead.
 #[cfg_attr(feature = "wasm", wasm_bindgen)]
 impl PolicyClient {
-    /// Filter policies of the given type for the current user.
+    /// Evaluate enforcement of the given policy type across all organizations,
+    /// returning type-erased decisions for the FFI boundary.
     ///
-    /// Untyped FFI path: native/WASM callers pass a runtime `policy_type` integer.
-    /// Delegates to the registry, falling back to default rules for unknown types.
+    /// Not yet intended for consumer use: exposed across the FFI boundary for
```
quexten, line 41: I wonder whether we should have a separate feature we compile the wasm build with that exposes these (and other test only) methods, but that is not enabled in prod builds.

crates/bitwarden-policies/src/policy.rs
```diff
@@ -0,0 +1,564 @@
...
+    /// The strongly-typed data for this policy. The [`Default`] value is
+    /// the fall-back whenever the policy is not enforced or the raw data could
+    /// not be parsed.
+    type Data: Default + DeserializeOwned;
+
+    /// Returns the organization roles that are exempt from this policy.
+    ///
+    /// Defaults to [`Owner`](OrganizationUserType::Owner) and
+    /// [`Admin`](OrganizationUserType::Admin).
+    fn exempt_roles(&self) -> &[OrganizationUserType] {
```
quexten, line 41: I would argue that having a default makes this harder to reason about. It's a general problem with OOP in TS where subclasses make it very hard to reason about the full behavior. We usually do not have that problem in rust because we don't have inheritance.

IMO each type should explicitly state the excempt roles.

(However feel free to disagree and ignore).

quexten (approved):
> One design note: there are separate client interfaces for native rust vs. FFI consumers. This is to accommodate the per-policy data. As the native rust interface shows, we can resolve this at compile time (using generics) for consumers, and I want to keep that as the first-class interface rather than forcing everyone to use the enum and have redundant match statements everywhere.

IMO I'm not sure whether that's worth the complexity of maintaining two different APIs. It's easy enough to just match with a default catch arm. In rust this should be expected / reasonable to work with.

Anyways, interested to see how this feels when it is more widely used by other consumers, maybe we'll double down on it, or revise.

crates/bitwarden-policies/src/policy.rs
eliykat: I think this is a good observation, particularly for exempt roles which are frequently overridden. I will consider this for a follow-up.

sven-bitwarden (approved):

PR #1365, opened 2026-08-09
[PM-39455] Defer account recovery rotation to organizatio on a V1 to V2 user key rotation

mzieniukbw:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-39455

## 📔 Objective

When forced user key rotation creates a V2 upgrade token, the client sends no new account recovery key. It sends one entry per enrolled organization with an empty key, so the server keeps the key the organization already holds. Later, organization admins update that key from the token.
This prevents showing a trust prompt to the user and prevents sharing it's new user key with server (who provides public key of the organization, hence could be swapped by malicious server).

Requires https://github.com/bitwarden/server/pull/8174, since at the moment server will reject a request with null account recovery key (also known as reset password key).

Note for reviewers:
- Should the v2 upgrade token creation (and request creation) still live in `unlock.rs` module ? It is no longer just unlock.
- ~~The org trust dialog shows first in clients, before the SDK key rotation. To my understanding this change is part of https://bitwarden.atlassian.net/browse/PM-30160 (there are mobile tickets too), but as a result all clients will need to know whether user is part of org and whether that org member have account recover enabled for that user. This is already known as part of sync / profile response from server. This fragments many clients, but could instead be lifted into SDK.~~ This is already part of SDK and in use by clients.

## 🚨 Breaking Changes

None expected, V2 key rotation is behind a feature flag in clients, none exists in prod at this moment.
The https://github.com/bitwarden/server/pull/8174 introduces a change to forced user key rotation request, which changes the `ResetPasswordWithOrgIdRequestModel` into `OrganizationUserAccountRecoveryRequestModel` (identical in payload). A compilation error is expected once the server change is merged and automated api models are updated in SDK, but since it's trivial to fix (just change the struct), it will be addressed immediately / as part of this PR later.

### DIFF
--- a/crates/bitwarden-user-crypto-management/src/key_rotation/rotate_user_keys.rs
+++ b/crates/bitwarden-user-crypto-management/src/key_rotation/rotate_user_keys.rs
@@ -55,6 +55,8 @@ pub enum UpgradeTokenAction {
 pub struct RotateUserKeysRequest {
     pub key_rotation_method: KeyRotationMethod,
     pub trusted_emergency_access_public_keys: Vec<PublicKey>,
+    /// Organization public keys the user confirmed as trusted. Only needed when the rotation
+    /// shares the new user key for account recovery.
     pub trusted_organization_public_keys: Vec<PublicKey>,
     pub upgrade_token_action: UpgradeTokenAction,
 }
@@ -155,6 +157,7 @@ async fn internal_rotate_user_keys(
             &sync,
             request.trusted_organization_public_keys.as_slice(),
             request.trusted_emergency_access_public_keys.as_slice(),
+            request.upgrade_token_action.clone(),
             &mut ctx,
         )?;
 
@@ -208,7 +211,7 @@ async fn internal_rotate_user_keys(
             },
             rotation_context.current_user_key_id,
             rotation_context.new_user_key_id,
-            request.upgrade_token_action,
+            rotation_context.creates_v2_upgrade_token,
             &mut ctx,
         )
         .map_err(|_| RotateUserKeysError::Crypto)?;
@@ -284,7 +287,9 @@ mod tests {
     use chrono::DateTime;
 
     use super::*;
-    use crate::key_rotation::partial_rotateable_keyset::PartialRotateableKeyset;
+    use crate::key_rotation::{
+        partial_rotateable_keyset::PartialRotateableKeyset, unlock::V1OrganizationMembership,
+    };
 
     fn make_state_bridge() -> StateBridgeClient {
         let client = Client::new(None);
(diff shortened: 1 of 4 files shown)


quexten:
> The org trust dialog shows first in clients, before the SDK key rotation. To my understanding this change is part of https://bitwarden.atlassian.net/browse/PM-30160 (there are mobile tickets too), but as a result all clients will need to know whether user is part of org and whether that org member have account recover enabled for that user. 

Isn't this already in the SDK? Via   `get_untrusted_memberships(): Promise<UntrustedMembershipsResponse>;`?

crates/bitwarden-user-crypto-management/src/key_rotation/rotate_user_keys.rs
```diff
@@ -326,6 +331,26 @@ mod tests {
         (store, sync)
     }
 
+    /// Builds a V1 account that is enrolled in account recovery for a single organization.
+    fn make_test_key_store_and_synced_data_with_organization()
```
quexten, line 335: Not relevant for this PR, but I have some BEEEP work that will hopefully make these kinds of tests easier / cleaner. We should sync up on it, because I think you'd be interested too.

quexten: Essentially, a good amount of full account test vectors of all account versions + helper methods to read and do tests for sets of these would make this much easier and increase the coverage.

crates/bitwarden-user-crypto-management/src/key_rotation/rotation_context.rs
```diff
@@ -4,13 +4,41 @@ use tracing::{debug, info, warn};
...
+        return false;
+    }
+
+    matches!(
+        (
+            ctx.get_symmetric_key_algorithm(current_user_key_id),
+            ctx.get_symmetric_key_algorithm(new_user_key_id),
+        ),
+        (
+            Ok(SymmetricKeyAlgorithm::Aes256CbcHmac),
```
quexten, line 36: Just noting this feels brittle. What if we ever change the v2 agorithm? (not blocking)

crates/bitwarden-user-crypto-management/src/key_rotation/rotation_context.rs
```diff
@@ -87,11 +109,41 @@ pub(super) fn make_rotation_context(
...
+    );
+
+    let v1_organization_memberships = if creates_v2_upgrade_token {
+        // Organization admins update account recovery from the token. The user is never asked to
+        // trust these keys.
+        info!(
+            "Skipping the trust check for {} organization public key(s), the rotation does not use them",
+            sync.organization_memberships.len()
+        );
+        sync.organization_memberships.clone()
```
quexten, line 126: Do we need to include them at all? Could we just send an empty array? I'm a bit concerned that we may end up having bugs where somehow these are passed in as trusted, and then actually re-encrypted to / it feels somewhat brittle.

quexten (approved):

crates/bitwarden-user-crypto-management/src/key_rotation/rotation_context.rs
mzieniukbw: I have a follow up PR, i will address it there. I am guessing something along the lines of `is not v1` would be more future proof.

crates/bitwarden-user-crypto-management/src/key_rotation/rotation_context.rs
mzieniukbw: Good idea. I will have a follow up PR to address this.

PR #1377, opened 2026-08-12
[PM-39979] Item-type Send backend integration work

mcamirault:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-39979

## 📔 Objective

This PR adds support for creating and editing Item-type Sends in the SDK, which will be used for the integration work starting with [this clients PR](https://github.com/bitwarden/clients/pull/22351)


crates/bitwarden-send/src/send.rs
```diff
@@ -250,7 +285,22 @@ impl CompositeEncryptable<KeySlotIds, SymmetricKeySlotId, SendApiModels> for Sen
...
+            SendViewType::Item(i) => {
+                let encrypted = i.encrypt_composite(ctx, key)?;
+                let serialized_cipher =
+                    serde_json::to_string(&encrypted.data).unwrap_or("{}".to_string());
+                Ok((
+                    bitwarden_api_api::models::SendType::Item,
+                    None,
+                    None,
+                    Some(Box::new(bitwarden_api_api::models::SendDataModel {
+                        encryption_version: Some(SendEncryptionType::V1.into()),
```
harr1424, line 299: 🎨 It might be nice to define the encryption type being used as a `const` at the top of the file. This would make it simpler to change down the road.

crates/bw/src/tools/send.rs
```diff
@@ -687,6 +687,8 @@ impl From<SendJsonInput> for SendView {
             r#type: input.r#type,
             file: input.file,
             text: input.text,
+            // TODO - Use the `data` field when implementing item-type Send support
```
harr1424, line 690: ❓ Is this still a TODO?

harr1424 (approved):

crates/bw/src/tools/send.rs
mcamirault: I'm not sure if creating these item-type Sends via the CLI is a desired feature, but if it is I think it'll be a follow-on to the current work.

PR #1378, opened 2026-08-12
tests: add test vectors

quexten:
Adds structured test vectors for different account versions. We will add loading functionality later. The aim here is to have all actions (pin, unlock, key rotation) tested for all version types, to prevent regressions such as "a cipher key cipher breaks upon key rotation" which we've recently had.

Ticket: https://bitwarden.atlassian.net/browse/PM-41875

There will be follow-up PR's adding support for these in the integration tests. We can also add rust facilities to make using these in unit tests easier.

### DIFF
--- a/.aikido
+++ b/.aikido
@@ -1,0 +1,3 @@
+exclude:
+  paths:
+    - test-vectors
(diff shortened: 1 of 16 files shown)


coroiu (commented):
Approval won't work across draft -> ready for review transition, put PR in ready for review and I'll look at it

PR #1381, opened 2026-08-13
feat: add trust log spec

quexten:
Adds an empty trust log crate, plus specs for the trust log, and how it is used with organizations.

### DIFF
--- a/crates/bitwarden-trust-log/trust-log.md
+++ b/crates/bitwarden-trust-log/trust-log.md
@@ -399,6 +399,38 @@ is out of scope for this version of the document.
 
 ---
 
+## Split view attack
+
+The hash chain makes a log internally consistent, but internal consistency says nothing about
+_which_ chain a reader was shown. The server is the sole distributor of logs, so it can maintain two
+divergent chains for the same actor and serve one to Alice and the other to Bob. Both views validate
+under [Chain validation](#chain-validation).
+
+```mermaid
+flowchart LR
+  L0["seq 0"] --> L1["seq 1"] --> L2["seq 2"]
+  L2 --> A3["seq 3: add_key K_attacker"]
+  L2 --> B3["seq 3: revoke_key K_old"]
+  A3 --> AV["View served to Bob"]
+  B3 --> BV["View served to Alice"]
+```
+
+The consequence is that a key revocation can be made effective for the account holder while never
+becoming effective for the relying parties encrypting to that account, and vice versa: an
+attacker-controlled key can be accepted by everyone except the account holder, who would notice it.
+Each party's view is self-consistent, so no local check fires. The `pin` mechanism
+([Log state](#log-state)) narrows the window between parties who exchange pins, but does not close
+it: the server chooses which view each party pins against.
+
+Detecting a split view requires comparing views across parties — gossip between clients, a
+transparency log with an independent witness set, or a signed-tree-head auditing protocol as in
+[RFC9162] and [CONIKS]. **Resolving the split view attack is out of scope for this document.**
+
+We will revisit this later; most likely using a tlog that holds a merkle tree that in each epoch
+commits to all latest head hashes.
+
+---
+
 ## Chain validation
 
 A verifier processing a trust log must, in sequence order:
--- a/crates/bitwarden-trust-log/trust-log.md
+++ b/crates/bitwarden-trust-log/trust-log.md
@@ -1,0 +1,450 @@
+# Trust log design
+
+**Audience:** engineers implementing or reviewing the trust log in clients, the SDK, or the server.
+
+**Status:** draft — under discussion inside Bitwarden.
+
+**Owner:** Key-management team
+
+## Overview
+
+The trust log is a tamper-resistant, append-only, signed log of security-critical actions, kept per
+actor — a user, an organization, or a provider — and served by an untrusted server. Each log entry
+is a **link**, split into a small signed **outer link** and an **inner message**. The chain of outer
+links provides completeness: no action can be inserted, removed, or reordered without detection.
+
+This set of logs acts as the key-transparency system for Bitwarden. Additionally, it allows us to
+cryptographically record and attest to permissions (policies around data transfer, etc).
+
+This will serve as the foundation to:
+
+- Organization Encryption V2
+- Organization Cryptographically Enforced Policies
+  - Automatic Key Connector Enrollment
+  - Automatic Data Ownership Transfer
+  - Automatic Retroactive Account Recovery Enrollment
+  - Automatic Activation of Auto-Confirmation
+    - Safer Autoconfirmation
+- Emergency Access V2
+- (Maybe) Device Keys
+
+With the key-transparency effective, this will prevent most fingerprints while preventing the server
+from injecting public keys.
+
+## Introduction
+
+The trust log is a tamper-resistant, append-only data structure that records actions. Each actor — a
+user, an organization, or a provider — has one trust log each. The goal is to record actions
+concerning that actor in a way that provably came from the actor, that can be revoked (by appending
+an action stating the revocation), and that the server can neither inject into, modify, nor omit
+from.
+
+The trust log is intended to become the central root from which clients determine:
+
+- the currently valid keys of other users, organizations, and providers;
+- their own currently valid keys;
+- the organizations and providers they agreed to be part of;
+- the security-critical permissions an organization holds over a user;
+- the devices currently active for a user.
+
+Each entry in the log is a **link**, and every link has two parts:
+
+- an **outer link** — small, always visible to any authorized reader of the log, and **the thing
+  that is signed**;
+- an **inner message** — the full payload, whose readability is gated by the outer link's declared
+  visibility.
+
+The chain of outer links gives _completeness_: the account holder can prove no action was inserted,
+removed, or reordered.
+
+### Goals
+
+| #   | Goal                                                                                                                                                   |
+| --- | ------------------------------------------------------------------------------------------------------------------------------------------------------ |
+| G1  | An account holder can detect any modification, reordering, or omission of their own history.                                                           |
+| G2  | A relying party can determine which keys an account controlled _at a given point in its history_, without trusting the server.                         |
+| G3  | The server cannot suppress a security-critical action (key revocation, policy change)                                                                  |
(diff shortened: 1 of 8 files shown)


crates/bitwarden-trust-log/Spec.md
```diff
@@ -0,0 +1,1251 @@
...
+       | public    |    | public    |    | semi_pub  |    | public    |
+       +-----+-----+    +-----+-----+    +-----+-----+    +-----+-----+
+             |                |                |                |
+             v                v                v                v
+       +-----------+    +-----------+    +===========+    +-----------+
+ inner | AddKeyBody|    | AddKeyBody|    #  WITHHELD  #   | RemoveKey |
+       +-----------+    +-----------+    +===========+    +-----------+
+
+  <---   previous_outer_link_hash;  all-zero at genesis
+  The outer chain is served WHOLE to every authorized reader, always (§4.1):
```
mzieniukbw, line 257: ~~❓ I am worried about performance if the chain becomes large. Is there some kind of optimisation possible (without risking the truncation or modification) on the client side ? I.e: we keep a shallow copy of only relevant logs sequences of a particular trust log (i.e. if user rotated their keys and we want to grant emergency access to a given user, we only need to know about recent sequences)~~
I answered myself with section 7

crates/bitwarden-trust-log/Spec.md
```diff
@@ -0,0 +1,1251 @@
...
+
+ClientMetadata = {
+  client_type:    enum { web, browser_extension, desktop, mobile, cli, server },
+  client_version: string,
+  device_id:      bytes[16] | null,
+}
+
+GlobalTrustLogRef = {              // all three null, or all three set; never mixed
+  merkle_root:  bytes[32] | null,
+  epoch:        uint64 | null,     // sequence number of that link in the server's log
```
mzieniukbw, line 304: ❓ shouldn't this be `sequence_number` ?

crates/bitwarden-trust-log/Spec.md
```diff
@@ -0,0 +1,1251 @@
...
+can be deployed before the server's log is.
+
+Three rules make that concession safe, and a verifier MUST enforce all three:
+
+1. **All or nothing.** The three fields are either all null or all set. A `global_trust_log` with
+   some fields null and others set MUST be rejected.
+2. **Witnessing latches on, per log.** Once any link in a log carries a non-null `global_trust_log`,
+   every later link in that log MUST carry one too. A log cannot go back to being unwitnessed, so
+   the server cannot induce a client to drop the reference from links it appends later.
+3. **A null `global_trust_log` is not a witness.** Links carrying one are protected by the hash
```
mzieniukbw, line 322: ❓ Why we wouldn't always want to require `global_trust_log` ? If the entire idea is that we can't trust the server then it certainly it would make sense to always have a reference to it, at a given point of time ?

crates/bitwarden-trust-log/Spec.md
```diff
@@ -0,0 +1,1251 @@
...
+
+## 4. Visibility
+
+```
+Visibility = enum { public, semi_public, private }
+```
+
+| Value         | Who may read the inner message                                                                                                            |
+| ------------- | ----------------------------------------------------------------------------------------------------------------------------------------- |
+| `public`      | Anyone who may read the outer chain.                                                                                                      |
```
mzieniukbw, line 352: inner chain

crates/bitwarden-trust-log/Spec.md
```diff
@@ -0,0 +1,1251 @@
...
+| `semi_public` | The account holder, and principals sharing the relevant scope (e.g. members/admins of the organization named in the body and the server). |
```
mzieniukbw, line 353: How about "privileged" ?

crates/bitwarden-trust-log/Spec.md
```diff
@@ -0,0 +1,1251 @@
...
+be encrypted when stored on the server, so that the server cannot read the body.
+
+### 4.1. Preventing Omission by the Server
+
+The outer chain is served **in full, to every authorized reader, regardless of visibility**.
+Visibility gates inner messages only. Since sequence numbers are contiguous and each link commits to
+the previous link's hash, a missing link in the interior of the chain is immediately visible (the
+_Sequence_ and _Outer chain_ checks, §8). A missing link at the _tail_ is covered by §9.
+
+If the server serves the outer link but refuses an inner message the reader is entitled to — this is
```
mzieniukbw: entitles to ? reject ?

crates/bitwarden-trust-log/Spec.md
```diff
@@ -0,0 +1,1251 @@
...
+  type:                  enum { kem, public_key_encryption, signature },
+  algorithm:             enum { ml-dsa, rsa },     // parameter sets unspecified
+  public_key:            bytes,                    // the public / verifying key itself
+  thumbprint:            bytes,                    // over public_key
+}
+
+RemoveKeyBody = {
+  key_id:      bytes[16],      // MUST refer to a key added earlier in this log
+  type:        enum { kem, public_key_encryption, signature },
+  reason:      enum { rotation, compromise },
```
mzieniukbw, line 418: 1. ❓ Should we distinguish manual vs automatic rotation ?
2. ❓ How would we distinguish compromise from key rotation ?

crates/bitwarden-trust-log/Spec.md
```diff
@@ -0,0 +1,1251 @@
...
+whether that is the right trade against interaction-graph leakage is not settled ⚠️. The `private`
+value is declared by no type and is currently headroom — an unused enum value invites misuse, so it
+SHOULD either gain a type or be removed.
+
+### 5.1. Bodies
+
+```
+AddKeyBody = {
+  key_id:                bytes[16],
+  type:                  enum { kem, public_key_encryption, signature },
```
mzieniukbw, line 409: Why not user symmetric keys ?

crates/bitwarden-trust-log/Spec.md
```diff
@@ -0,0 +1,1251 @@
...
+Each entry in the log is a **link**, and every link has two parts:
+
+- an **outer link** — small, always visible to any authorized reader of the log, and **the thing
+  that is signed**;
+- an **inner message** — the full payload, whose readability is gated by the outer link's declared
+  visibility.
+
+The chain of outer links gives _completeness_: the account holder can prove no action was inserted,
+removed, or reordered. The **server's own trust log** (§9) — which on a fixed schedule records the
+latest state of every other trust log, and which every link back-references — gives _freshness_: it
```
mzieniukbw, line 103: ~~❓ If server provides it's own trust log to the party that asks for it (account holder), then malicious server could build one one the fly (per request), since server likely knows everything about the account holder.
Unless server have it's own signing key that the account holder can verify when validating each entry that references server ?~~
Answer is in section 9.1

crates/bitwarden-trust-log/Spec.md
```diff
@@ -0,0 +1,1251 @@
...
+may be treated as consented to by this user. Once a counterparty appears in `known_users`, a
+different key for it MUST be justified by an `add_key` at a later position in that counterparty's
+own log, which is what makes a swapped recipient key detectable.
+
+### 6.3. Log Application Rules
+
+| Message type                     | Effect on state                                                                                                                                                                         |
+| -------------------------------- | --------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `add_key`                        | Insert into this trust log's key set (`trust_log_keys`). Genesis establishes the first signature key.                                                                                   |
+| `remove_key`                     | Set `revoked_at_sequence` + `revocation_reason` on that key of `trust_log_keys`. Never delete.                                                                                          |
```
mzieniukbw, line 652: I wonder about this message type. It's not like you remove the key, you replace it. So it's always remove key followed by add key (or actually the other way around?). Maybe this should be "replace_key", "revoke_key" or "invalidate_key" ?

quexten: Agreed, replacing.

crates/bitwarden-trust-log/Spec.md
quexten: Yeah; Initially it was a separate construct from trust log, but now that it's the same it should be seqno. 

I'll note that i have out of scoped the global trust log for the initial version and will move it to follow-up.

crates/bitwarden-trust-log/Spec.md
quexten: I suppose we could record them into the trust log. I'm not yet sure about the purpose for symmetric keys 🤔

crates/bitwarden-trust-log/Spec.md
quexten: This specifies just the format. I don't think we have a way to get this data currently / it would go unused, but I do think it's useful to have support for speciyfing why you revoked a key.

@MGibson1 has previously frequently said "we can't know for rotations why someone does them, so we shouldn't re-trust keys", but if we track the reason, we can distinguish those cases.

crates/bitwarden-trust-log/Spec.md
quexten: I think this is correct. "Who may read the inner message?" "Anyone who can read the outer chain". Each chain link references an inner message. There is no inner chain 🤔

crates/bitwarden-trust-log/Spec.md
quexten: Makes sense.

dani-garcia (approved):
Platform changes lgtm!

coltonhurst (approved):
Approving just for the `.github/CODEOWNERS` change 👍

crates/bitwarden-trust-log/trust-log.md
```diff
@@ -0,0 +1,482 @@
...
+SignedOuterLink = {
+  outer:      OuterLink,
+  signature:  bytes,     // over the canonical encoding of `outer`
+}
+```
+
+### Genesis link
+
+The link at `sequence_number = 0` must be `add_key` carrying a key of `type = signature`, must be
+self-signed by that key, and must carry `previous_outer_link_hash` of all zeroes. The body carries
```
iinuwa: > must be self-signed by that key

Optimization suggestion: if we can avoid the self-signature, which is not conveying trust since trust is assumed, that can save a lot of overhead for smaller logs using PQC algorithms. I think that we could just skip signing and check the fingerprint only.

PR #1384, opened 2026-08-14
PAM: approver surface and shared domain logic

patriksvensson:
## 🎟️ Tracking

Consumed by bitwarden/clients#22510, which deletes ~730 lines of hand-rolled wire parsing against this.

## 📔 Objective

Two commits, both in `bitwarden_license/bitwarden-pam`.

### 1. `ApprovalsClient` — the approver surface

`PamClient::approvals()` with `list_inbox`, `list_history` and `decide`, wrapping `AccessRequestsApi::{get_inbox, get_history, decide}` (bindings already on `main` since #1254, so no bindings-update workflow).

This was the one documented gap. The web client reached those three routes over raw HTTP because the SDK exposed only requester-scoped operations; its own module docs called that "SDK work — not a fourth raw-HTTP route".

A separate client rather than extra methods on `AccessRequestsClient`, whose doc contract is explicitly requester-scoped.

- `AccessRequestView` gains `expired_at`. The server does not track it in v1 (always null today), which the doc comment records.
- `AccessDecisionVerdict::Unknown` is a read-side spelling for a verdict a newer server returned, never something to submit — `TryFrom` rejects it with the new `LeasingError::UnsubmittableVerdict` and the request never reaches the API.

### 2. Shared domain logic

Three rules the web client owned that every other platform would otherwise reimplement:

- **`CipherAccessStateView::badge_state`** — the spec's badge precedence (active lease → ready to activate → pending → resting), applied once at conversion. `unavailable` and `expired` are deliberately not modelled: the state response is caller-scoped, so there is no data to derive them from.
- **`AccessLeaseView::ended_by_holder`** — `AccessLeaseStatus::Revoked` covers both a holder ending their own lease and an operator revoking it. The web client told them apart by scanning the request's decision log for a human `deny` whose decider matched the requester; the lease already carries `revoked_by_user_id`, so that heuristic was never necessary.
- **`AccessRequestCreateRequest::validate`** — the server's two window checks and its 24h cap, run by `request()` before POSTing, following `AccessRulesClient::create`. A request with neither a duration nor a window is left alone: the server decides which path applies, and a local rule there would reject requests the server accepts.

These are computed fields and an internal validation hook rather than methods on the views, because Tsify types cross the wasm boundary as plain objects — a method on a view is unreachable from TypeScript. (Verified: no Tsify type in this repo carries a `wasm_bindgen` impl block.)

### Deliberately excluded

- **The requester/inbox actionability predicates.** They need a clock, so they cannot be computed fields, and per the above they cannot be view methods either. They stay in TypeScript.
- **The clients' server-error catalog**, which substring-matches 13 English server messages. Moving it would centralise the same fragility rather than fix it; the real fix is a server error code.

## ⚙️ Code changes

- `src/approvals/{mod,client,models}.rs` (new)
- `src/access_requests/validate.rs` (new)
- `src/access_requests/{models,client,mod}.rs`, `src/leases/models.rs`, `src/error.rs`, `src/lib.rs`, `src/pam_client.rs`

## 📋 Verification

- `cargo test -p bitwarden-pam --all-features` — 125 passed
- `cargo check -p bitwarden-pam --all-features --all-targets`, `npm run lint -- --only clippy`, `cargo doc --no-deps`, `npm run lint:fix -- --only fmt` — all clean
- Built the WASM bindings for both packages and confirmed the generated `.d.ts` exposes `approvals(): ApprovalsClient`, `badgeState`, `endedByHolder`, `expiredAt` and the `UnsubmittableVerdict` variant; bitwarden/clients#22510 typechecks against it.


Hinton:
# PAM UAT — what should move into the SDK

Analysis of [bitwarden/clients#22430](https://github.com/bitwarden/clients/pull/22430)
("PAM UAT", 135 files / +14.1k, base `pam/vault-gating-seams`) against
[bitwarden/sdk-internal#1384](https://github.com/bitwarden/sdk-internal/pull/1384)
(same branch name `pam/uat`, base `pam/cipher-partial-data`, currently an empty
placeholder commit).

Question asked: which parts of the clients PR **can** and **should** move into
`bitwarden_license/bitwarden-pam`. No code written yet — this doc is the proposal.

Worktrees used: `worktrees/pam-uat/{clients,sdk-internal}`.

## Verdict at a glance

| # | Candidate | Move? | Cost | Blocked on |
|---|---|---|---|---|
| 1 | Approver surface: inbox / history / decide | **Yes — do it first** | ~600 LoC Rust + tests | nothing |
| 2 | Pure domain logic (badge state, filters, window math, decision-log helpers) | Yes, second PR | ~400 LoC Rust + tests | nothing |
| 3 | Server error catalog (`classifyRequestAccessError`) | Yes, but see caveat | small | ideally a server error-code change |
| 4 | i18n key selection, duration presets, Angular/rxjs plumbing, HTML input formatting | **No** | — | — |
| 5 | `partial` / `partialData` type bridges in `libs/common` | Already SDK work, done but unshipped | — | SDK bump (breaking) |

## 1. Approver surface — the one documented gap

`bitwarden_license/bit-web/src/app/pam/approvals/approval-api.service.ts` is a
deliberately-narrow raw-HTTP seam for three routes:

- `GET  /access-requests/inbox`
- `GET  /access-requests/history`
- `POST /access-requests/{id}/decision`

The clients module's own `CLAUDE.md` states the rule and the exception:

> Every PAM call goes through the Rust SDK (`client.commercial().pam()`), never HTTP.
> […] **Do not widen it.** […] If another PAM capability turns out to be missing from
> the SDK, that is SDK work — not a fourth raw-HTTP route.

So this isn't a judgement call — the clients PR already labels it SDK work and binds it
behind an abstraction so "the eventual swap is a provider change in `provide-pam.ts` and
nothing else."

### Why it's cheap now

The generated bindings already exist on `main` (landed in #1254, `Update API bindings to
c2d97d5`), on the `AccessRequestsApi` trait:

```rust
async fn decide(&self, id: Uuid, m: AccessDecisionRequestModel)
    -> Result<AccessRequestDetailsResponseModel, Error>;   // POST /access-requests/{id}/decision
async fn get_history(&self) -> Result<AccessRequestDetailsResponseModelListResponseModel, Error>;
async fn get_inbox(&self)   -> Result<AccessRequestDetailsResponseModelListResponseModel, Error>;
```

`AccessRequestDetailsResponseModel` is the **same** model `AccessRequestView` already
converts from for `list_mine`/`get`, and it already carries every field the clients
hand-rolled response reads — including `expired_at`. `AccessDecisionVerdict` is a
generated int-serializing enum (`0 = deny`, `1 = approve`), which is exactly what the
clients `WIRE_VERDICT` map reimplements by hand.

Nothing needs the bindings-update workflow. This is a pure feature PR in
`bitwarden-pam`.

### Shape (decided)

A **new `ApprovalsClient`** on `PamClient::approvals()`, not extra methods on
`AccessRequestsClient`. Rationale: `AccessRequestsClient`'s doc contract is explicitly
requester-scoped ("Client for a requester's PAM access requests"), and the clients module
already splits `approvals/` from `access-requests/` along the same line. Mirrors the
existing `access_rules` / `access_requests` / `leases` module layout.

```rust
// bitwarden_license/bitwarden-pam/src/approvals/{mod,client,models}.rs
impl ApprovalsClient {
    /// GET /access-requests/inbox — requests awaiting the caller's decision.
    /// Server-scoped by Manage permission on the request's collection and already
    /// filtered to pending, so an empty list is normal for a non-approver.
    pub async fn list_inbox(&self) -> Result<Vec<AccessRequestView>, LeasingError>;

    /// GET /access-requests/history — decided requests for collections the caller manages.
    pub async fn list_history(&self) -> Result<Vec<AccessRequestView>, LeasingError>;

    /// POST /access-requests/{id}/decision
    /// Only `status`, `resolved_at`, and the decision just recorded are guaranteed
    /// populated — callers merge onto the row they hold rather than replacing it.
    pub async fn decide(&self, id: AccessRequestId, request: AccessDecisionRequest)
        -> Result<AccessRequestView, LeasingError>;
}

pub struct AccessDecisionRequest {
    pub verdict: AccessDecisionVerdict,   // reuse the existing read-side enum
    pub comment: Option<String>,
}
```

`AccessDecisionVerdict::Unknown` must be rejected on the write path (it's a read-side
spelling for a verdict this SDK doesn't recognise, never something to submit) — either a
new `LeasingError` variant or a `TryFrom`. The clients class throws for the same reason.

Reuses `LeasingError` — approvals are the other half of the same request lifecycle, same
argument the existing shared error type already makes.

### One additive model change

`AccessRequestView` is missing `expired_at`, which the clients response class adds as an
extra field and `isActionableInboxRequest` depends on (a request the server marked lapsed
while pending must drop out of the inbox). Add:

```rust
/// When the server marked the request lapsed while still pending (UTC); None otherwise.
pub expired_at: Option<DateTime<Utc>>,
```

Additive and already on the API model — no server or bindings change.

### What this retires in clients

| File | LoC |
|---|---|
| `approvals/approval-api.service.ts` | 47 |
| `approvals/default-approval-api.service.ts` + spec | 189 |
| `approvals/access-decision.request.ts` | 38 |
| `approvals/responses/access-request.response.ts` + spec | 452 |
| **total** | **~726** |

That deleted code is entirely PascalCase wire parsing, status-spelling normalisation
(`cancelled`→`canceled`, lease `cancelled`→`revoked`), verdict int↔string translation, and
branded-id widening — all of which the SDK's `TryFrom<AccessRequestDetailsResponseModel>`
already does once, correctly, for every client. `ApproverInboxService` keeps its shape;
only the injected service changes.

Both response files carry comments anticipating exactly this ("When these three routes
move into the SDK, this class is deleted and nothing downstream changes").

## 2. Pure domain logic

Precedent is established: `bitwarden-pam` already owns `access_rules::validate_request`
and `is_valid_cidr` with a wasm-exposed typed error enum
(`AccessRuleValidationError`), and the clients CIDR editor "delegates to the SDK's
`is_valid_cidr`". These are the same kind of thing.

Candidates, all framework-free today (they live under `helpers/` precisely because the
module keeps that layer Angular-free):

| Clients source | What it decides | Note |
|---|---|---|
| `access-state-badge/access-badge-state.ts` → `cipherAccessBadgeState` | Badge precedence: active lease → ready → pending → privileged | The spec's precedence rule; one copy for web/desktop/browser/mobile |
| `helpers/actionable-requests.ts` | Whether a request still needs the requester's attention (nav badge count) | Encodes the server's activate-window rule |
| `approvals/inbox-request-filter.ts` | Whether an inbox row can still produce usable access | Needs `expired_at` (see §1) |
| `helpers/requested-window.ts` | `lease_not_after − lease_not_before` | Trivial but shared by 4 call sites |
| `helpers/find-human-decision.ts` | Unwrap `decider: Automatic \| Human` | Pure `AccessDecider` accessor — belongs next to the type |
| `helpers/can-approve.ts` | The no-self-approval rule | 3 lines, but it's a policy rule the server also enforces |
| `helpers/request-access-window.ts` → `requestWindowProblem`, `MAX_REQUEST_ACCESS_WINDOW_SECONDS` | Mirrors the server's two window checks (end > start, span ≤ 24h) | Direct analogue of `validate_request`; the 86 400 cap is a server constant duplicated in TS |
| `access-requests/my-access-row.ts` → `historyDisplayStatus`/`endedByHolder`, `extensionsByLeaseId` | Ended-by-holder vs revoked-by-operator; folding extensions onto their parent lease | Move the **derivation**, not the i18n keys — see below |

### `historyDisplayStatus` deserves care, not a straight port

The clients version derives "the holder ended their own lease" by scanning the decision
log for a human `deny` whose decider id equals the requester's — because
`AccessLeaseStatus` collapses self-end and operator-revoke into `revoked`.

But `AccessLeaseView` already carries `revoked_by_user_id`, and `MyAccessService` already
loads every lease (`leases().list_mine()` returns all leases, active or not). So the
heuristic is avoidable with data already on hand. **If this moves to the SDK, the function
should take the lease and compare `revoked_by_user_id == requester_id`** rather than
reproducing the decision-log scan. Splitting the derivation (SDK) from the i18n key
selection (clients) makes that clean.

## 3. The server error catalog

`helpers/request-access-error.ts` hardcodes 13 server error **sentences** to detect the
three "you already have this" reconciliation cases and the field-level validation
failures. Its own comment:

> Reproduced here rather than imported because the strings cross the wire as prose: the
> SDK surfaces a server 400 as a `LeasingError` with `variant: "Api"` and the server's
> message on `.message`, with no machine-readable code to switch on. When the server grows
> a code, this catalog is the single place to retire.

Moving the catalog to the SDK means it retires in one place for every client instead of
once per platform, and `LeasingError` can grow real variants
(`AlreadyActive`, `AlreadyApproved`, `AlreadyPending`, `ReasonRequired`, …) so consumers
switch on a variant rather than `String::contains`.

Caveat, stated plainly: substring-matching English server prose inside the SDK is not
better engineering than doing it in TypeScript — it just centralises the same fragility.
The real fix is a server error code. **Recommendation:** raise the server change, and
either (a) hold this item until codes exist, or (b) move it now behind clearly-labelled
`LeasingError` variants so the swap to codes is internal to the SDK. Do not move it
silently as if it were clean.

## 4. Should not move

- **i18n key selection** — `statusLabelKey`, `durationLabel`, `elapsedLabel`,
  `accessRuleSummaryKeys`, `resolveResolver`'s key choice. Key names are the clients'
  message catalog; the SDK has no business naming them.
- **`formatRemaining`, `elapsedLabel` output, `pickDurationUnit`** — locale/presentation.
  `Intl` is the right tool and it lives in the browser.
- **Duration presets** (`helpers/lease-window.utils.ts`) — each entry pairs a number with
  an i18n key, so the whole table is clients-shaped. `snapToNearestDuration` is generic
  but not worth a crate boundary on its own.
- **`composeRequestWindow`, `toDateInputValue`, `toTimeInputValue`, `defaultRequestWindow`**
  — `<input type="date">`/`<input type="time">` string formats and local-midnight clamping.
  Web-form concerns. (The *validation* half of that file is §2.)
- **`hasApprovalPrivileges$`, `canViewApprovalsGuard`** — read the clients
  `Organization.canManageAccessRules` model through rxjs.
- **All Angular/rxjs services** — `MyAccessService`, `ApproverInboxService`,
  `DefaultAccessEventService`, `DefaultAccessRefreshService`, `DefaultPamNavBadgeService`,
  `AccessRequestCancelService`, `DefaultVaultRowAccessActionsService`,
  `PamGatedCipherReloaderService`. Optimistic patching, `concatMap` ordering,
  `shareReplay({refCount})` lifetimes, DI seams — platform composition, correctly placed.
- **`AccessNameResolverService`** — resolves names from the local decrypted vault.
- **`accessRuleSummaryKeys` / `rulesGoverningCollection`** — the first is i18n keys; the
  second is a two-line filter. Low value.
- **`NotificationType.RefreshApproverInbox/RefreshAccessRequest` (28/29)** — clients
  platform notification enum.

## 5. Adjacent findings

**The `partial` / `partialData` type bridges are not "move to SDK" — they're "ship the SDK
work that's already done."** `libs/common/src/vault/models/domain/cipher.ts` and
`view/cipher.view.ts` declare local intersection types
(`SdkCipher & { partialData?: string }`, `SdkCipherView & { partial?: boolean }`) because
the Rust side exists only on the PAM branch. Confirmed present in this worktree:
`crates/bitwarden-vault/src/cipher/cipher.rs:335` (`partial_data`), `:489` and `:586`
(`partial` on `CipherView` and `CipherListView`).

The clients comment records why the bump isn't just a version change: SDK commit `99ffb6ef`
("Use state bridge instead of platform state") dropped `UserKeyState` and
`EphemeralPinEnvelopeState`, which `libs/common/src/key-management` still imports. Those
call sites migrate in the same change. Worth tracking as its own piece of work — it is the
gate on deleting both bridges.

**`GET /leases/history` is bound but unexposed.** `LeasesApi::get_history` exists in
`bitwarden-api-api`; `LeasesClient` exposes `list_active`, `list_mine`, `extend`, `end`
only. No clients surface needs it today — noting it so it isn't mistaken for a gap later.

**`unavailable` badge state is unreachable.** `AccessBadgeState` declares an
`unavailable` kind (item held by another user) that `cipherAccessBadgeState` never
produces, because `CipherAccessStateView` is caller-scoped and never reports someone
else's lease. Producing it needs a server response-model change plus an SDK field — real
SDK+server work, not a move. Already flagged as a tracked follow-up in the clients code.

## Recommended sequencing

1. **`ApprovalsClient` + `expired_at`** (this PR, #1384). Self-contained, no dependencies,
   retires ~726 lines of clients code and the module's one documented architectural
   exception. Follow the existing `access_requests` pattern: client + models + `mockall`
   unit tests, Tsify/`wasm_bindgen` bindings, verify
   `commercial().pam().approvals()` in the generated `.d.ts`.
2. **Clients follow-up**: swap `provide-pam.ts` to the SDK, delete the four approvals HTTP
   files, extend `AccessRequestSdkService`'s sibling contract or add an
   `ApprovalSdkService`.
3. **Pure domain logic** (§2) as a second SDK PR, with `historyDisplayStatus` reworked to
   use `revoked_by_user_id`.
4. **Server error codes**, then §3.

## Decisions taken

- New `ApprovalsClient` on `PamClient::approvals()` — not extra methods on
  `AccessRequestsClient` (keeps that client's requester-scoped contract intact).
- Analysis first; no Rust written under #1384 until this is agreed.

bitwarden_license/bitwarden-pam/src/access_requests/models.rs
```diff
@@ -187,6 +187,10 @@ pub struct AccessRequestView {
     pub submitted_at: DateTime<Utc>,
     /// When the request was approved, denied, or cancelled (UTC); None while pending.
     pub resolved_at: Option<DateTime<Utc>>,
+    /// When an approved request lapsed unactivated (UTC); None otherwise. The server does not
+    /// track this in v1, so it is always None today - its absence does not prove the request
+    /// is still live.
+    pub expired_at: Option<DateTime<Utc>>,
```
Hinton: This is missing from server.

bitwarden_license/bitwarden-pam/src/access_requests/client.rs
```diff
@@ -57,13 +57,15 @@ impl AccessRequestsClient {
...
-    /// Opens an access request for a cipher. Run [`pre_check`](Self::pre_check) first to know
-    /// which fields [`AccessRequestCreateRequest`] expects.
+    /// Validates and opens an access request for a cipher. Run [`pre_check`](Self::pre_check)
+    /// first to know which fields [`AccessRequestCreateRequest`] expects.
     pub async fn request(
         &self,
         cipher_id: CipherId,
         request: AccessRequestCreateRequest,
     ) -> Result<AccessRequestResultView, LeasingError> {
+        request.validate()?;
```
Hinton: I think this validation should happen as part of converting `AccessRequestCreateRequest` into whatever the api expects. That way this can't be circumvented.

bitwarden_license/bitwarden-pam/src/error.rs
```diff
@@ -2,22 +2,32 @@ use bitwarden_core::{ApiError, MissingFieldError};
 use bitwarden_error::bitwarden_error;
 use thiserror::Error;
 
+use crate::access_requests::AccessRequestWindowError;
+
 /// Errors returned from the PAM leasing clients
-/// ([`AccessRequestsClient`](crate::AccessRequestsClient)
-/// and [`LeasesClient`](crate::LeasesClient)).
+/// ([`AccessRequestsClient`](crate::AccessRequestsClient),
+/// [`LeasesClient`](crate::LeasesClient), and [`ApprovalsClient`](crate::ApprovalsClient)).
```
Hinton: Seems that we should split this error up to better represent the error states between the different clients. It's probably an even better idea to not re-use the same error across a single clients if they differ noteworthy between functions.

bitwarden_license/bitwarden-pam/src/leases/models.rs
```diff
@@ -74,24 +74,36 @@ pub struct AccessLeaseView {
     pub revoked_at: Option<DateTime<Utc>>,
     /// The user who revoked the lease; None unless it was revoked early.
     pub revoked_by_user_id: Option<UserId>,
+    /// Whether the holder ended their own lease, as opposed to an operator revoking it. `false`
+    /// for a lease that was never revoked. This exists because
+    /// [`AccessLeaseStatus::Revoked`] alone can't tell the two apart - it covers both a
+    /// self-service end and an operator-initiated revoke - so callers no longer need to infer it
+    /// by scanning the originating request's decision log for a human `deny` whose decider id
+    /// matches the requester.
+    pub ended_by_holder: bool,
```
Hinton: Should we use some enum here to indicate the state of an access lease better?

PR #1388, opened 2026-08-15
feat(crypto): add sync handling and state write for user key id

quexten:
https://bitwarden.atlassian.net/browse/PM-40814

Syncs the key id to disk via the crypto handler, and adds a backfill endpoint. The backfill endpoint will be called in a encrypted migrator (rather than on sync).


crates/bitwarden-user-crypto-management/src/key_id_backfill.rs
```diff
@@ -0,0 +1,290 @@
...
+    /// The API call recording the key id failed.
+    #[error("API call failed during user key id backfill")]
+    Api,
+}
+
+#[cfg_attr(feature = "wasm", wasm_bindgen)]
+#[cfg_attr(feature = "uniffi", uniffi::export(async_runtime = "tokio"))]
+impl UserCryptoManagementClient {
+    /// Returns whether the server is missing the id of the user's current user key.
+    pub async fn needs_backfill(&self) -> Result<bool, KeyIdBackfillError> {
```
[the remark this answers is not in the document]
quexten: Renamed.

crates/bitwarden-user-crypto-management/src/key_id_backfill.rs
```diff
@@ -0,0 +1,301 @@
...
+            }))
+            .await
+            .map_err(|e| {
+                error!("Failed to post the user key id: {e:?}");
+                KeyIdBackfillError::Api
+            })?;
+
+        // Written only after the server accepted it, so state never claims an id the server lacks.
+        let state_bridge = self.client.km_state_bridge();
+        if state_bridge.is_bridge_registered() {
```
mzieniukbw: This behaviour is inconsistent with `user_key_id_needs_backfill`, one errors and another one is permissive about it (no error).
Since most of the state bridge does not error when state bridge is not registered, should we do the same in `user_key_id_needs_backfill` ?

crates/bitwarden-user-crypto-management/src/key_id_backfill.rs
```diff
@@ -0,0 +1,301 @@
...
+        }
+
+        Ok(())
+    }
+}
+
+impl UserCryptoManagementClient {
+    /// Reads the key id of the user key currently in the key store.
+    ///
+    /// `None` means the key has no id, which is the case for V1 user keys.
```
mzieniukbw: ❓ I saw other PR that added key ids to V1 user keys. Isn't that comment inaccurate ?

quexten: Currently this is true but yes you are right with the other PR it is no longer true. I'll remove the two comments about this.

crates/bitwarden-user-crypto-management/src/key_id_backfill.rs
quexten: I think we can actually just error. iOS, Android, `clients` all now implement state bridge, so we can remove these checks soon.

crates/bitwarden-vault/src/cipher/cipher.rs
```diff
@@ -377,7 +391,7 @@ impl TryFrom<Cipher> for CipherRequestModel {
     /// Structural mapping from an encrypted [`Cipher`] to the API's expected
     /// [`CipherRequestModel`]. No crypto — all encryption happened upstream in
     /// `CipherView::encrypt_composite`. Callers are responsible for setting
-    /// `encrypted_for` after the conversion.
+    /// `encrypted_for` and `encrypted_by_key_id` after the conversion.
```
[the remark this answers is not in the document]
quexten: Will be handled in a follow-up PR.

mzieniukbw (approved):

dani-garcia (approved):

PR #1392, opened 2026-08-19
fix(biometrics): fix biometric unlock over ipc with concurrent status request

quexten:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-42191

## 📔 Objective

There is a concurrency issue where the typed request subscription breaks on parsing the wrong request type. In this case `unlock`, `status` are concurrent / breaking *only when the browser ui is open*, so only when using the popout browser not the popup which closes during bio unlock.

Adds concurrency tests proving that this works correctly now.

### DIFF
--- a/crates/bitwarden-ipc/src/ipc_client.rs
+++ b/crates/bitwarden-ipc/src/ipc_client.rs
@@ -921,6 +921,23 @@ mod tests {
             }
         }
 
+        /// A second request type whose response is a unit enum, which serde serializes to a bare
+        /// JSON string. Deserializing it as [`TestResponse`] fails, which is what makes it a
+        /// faithful stand-in for the real `GetBiometricsStatus` / `UnlockBiometrics` pair.
+        #[derive(Debug, Clone, PartialEq, Deserialize, Serialize)]
+        struct OtherRequest;
+
+        #[derive(Debug, Clone, PartialEq, Deserialize, Serialize)]
+        enum OtherResponse {
+            Available,
+        }
+
+        impl RpcRequest for OtherRequest {
+            type Response = OtherResponse;
+
+            const NAME: &str = "OtherRequest";
+        }
+
         #[tokio::test]
         async fn request_sends_message_and_returns_response() {
             let crypto_provider = NoEncryptionCryptoProvider;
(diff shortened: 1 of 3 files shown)


coroiu (approved):

quexten:
Broken test unrelated and fixed in https://github.com/bitwarden/sdk-internal/pull/1391.

quexten:
@djsmith85 was dismissing @coroiu 's approving review intentional?
(And apologies, I had removed your review request after @coroiu approved but it seems github ate this change).

djsmith85:
> @djsmith85 was dismissing @coroiu 's approving review intentional? (And apologies, I had removed your review request after @coroiu approved but it seems github ate this change).

Yes, I think because @coroiu's approval was during draft, it didn't satisfy as also approving for "team-platform-dev", hence it also requesting me. I pinged him for re-review to also approve for platform.

coroiu (commented):
Lol, didn't think it would request from someone else

crates/bitwarden-ipc/src/ipc_client.rs
```diff
@@ -979,6 +996,107 @@ mod tests {
...
+                    client
+                        .request::<OtherRequest>(
+                            OtherRequest,
+                            Endpoint::BrowserBackground { id: HostId::Own },
+                            None,
+                        )
+                        .await
+                })
+            };
+            tokio::time::sleep(Duration::from_millis(100)).await;
```
mzieniukbw, line 1038: nit: We could bring this delay down to few milliseconds and increase reliability if we would instead in a loop check for length of `communication_provider.outgoing().await` to be 2  (otherwise sleep for few ms and loop again).
(optional, since all tests do the sleep with ~100ms delay and might not be able to use this in the same way)

crates/bitwarden-ipc/src/ipc_client_ext.rs
```diff
@@ -112,10 +112,20 @@ pub trait IpcClientExt: IpcClient {
...
-                    .map_err(RequestError::Receive)?;
+                {
+                    Ok(received) => received,
+                    // Every RPC response shares a single payload type name, and therefore a single
+                    // topic, so this subscription also receives the responses belonging to other
+                    // requests that are in flight concurrently. Those do not necessarily
+                    // deserialize into this request's response type, so a typing error here is
+                    // expected: skip the message and keep waiting for our own response rather than
+                    // failing over a message that was never addressed to us.
+                    Err(TypedReceiveError::Typing(_)) => continue,
```
mzieniukbw, line 126: I honestly think entire typed response subscription concept is broken.
It would would be way better if request message, that expects reply, would contain a dedicated topic for replies (we already support this to my knowledge). The receiver would then send the response message onto the provided topic, instead of current catch-all topic for responses.
Example change, that does not require loop or handling typed Err case:
```
  let response_topic = format!("RpcResponseMessage:{request_id}");

  let mut response_subscription = IpcClientTypedSubscription::<                                                                               
      IncomingRpcResponseMessage<Request::Response>,                                                                                          
  >::new(self.subscribe(Some(response_topic.clone())).await?);

  let request_payload = RpcRequestMessage {                                                                                                   
      request,                                                                                                                                
      request_id,                                                                                                                             
      request_type: Request::NAME.to_owned(),                                                                                                 
      response_topic,                                                                                                                         
  }; 
```

mzieniukbw (approved):
Approved, but i don't think this is a long term solution.

crates/bitwarden-ipc/src/ipc_client_ext.rs
quexten: Replied in DM

PR #1393, opened 2026-08-19
[PM-34041] Implement status command for the CLI

djsmith85:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-34041

### DIFF
--- a/crates/bw/src/platform/config.rs
+++ b/crates/bw/src/platform/config.rs
@@ -56,10 +56,7 @@ impl BwCommand for ConfigCommand {
     async fn run(self, state: AnyState) -> CommandResult {
         // If we're not provided any values, then this is a request to get the current server URL.
         if self.is_get() {
-            let server = read_config_json()?
-                .and_then(|c| c.server)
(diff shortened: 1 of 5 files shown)


crates/bw/src/platform/config.rs
```diff
@@ -123,10 +120,22 @@ pub struct ConfigFile {
...
 
+/// The base server URL saved by `bw config server <url>`, or [`DEFAULT_SERVER_URL`] when unset.
+///
+/// Shared by `bw config server` (GET) and `bw status` so the two cannot report different servers.
+pub(crate) fn resolve_server_url() -> Result<String> {
+    Ok(read_config_json()?
+        .and_then(|c| c.server)
+        .unwrap_or_else(|| DEFAULT_SERVER_URL.into()))
+}
+
```
djsmith85: Exposes `resolve_server_url` from the config command to retrieve the configured server urls. This might need to be moved into some kind of other service. Some equivalent of the `environment-service` we have on `clients` 🤷 ? This can be done as a follow-up and doesn't necessarily need to land here.

PR #1396, opened 2026-08-20
feat(importers): add encryption context to import cxf

quexten:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-42247

## 📔 Objective

Expose an encryption context with the user id and user key id used to encrypt the vault item.

## 🚨 Breaking Changes

`import_cxf` now returns an encryption context. This should be wired through to place the user id and key id onto the HTTP request used to post it to the server.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -831,6 +831,7 @@ dependencies = [
  "serde",
  "serde_json",
  "thiserror 2.0.19",
+ "tokio",
  "tsify",
  "uniffi",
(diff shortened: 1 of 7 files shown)


quexten:
Tests fixed in https://github.com/bitwarden/sdk-internal/pull/1391

harr1424 (approved):

dereknance (approved):

PR #1398, opened 2026-08-20
test(shared-unlock): add a test harness and tests for the shared unlock protocol

quexten:
## 🎟️ Tracking
https://bitwarden.atlassian.net/browse/PM-43320

## 📔 Objective

Adds a comprehensive test suite for the shared unlock protocol. It adds a test harness, that (somewhat) accurately simulates real client behavior, then runs a few scenarios and topologies.

Where possible we do not want to assert specific whitebox internals, but blackbox behavior observable from the outside. The tests are aimed to be composed of high level functions.

To the reviewers, please let me know if there is anything about the tests that can be better. I'm trying to establish a better pattern for testing here.

> [!NOTE]
> I'm not sure if the event log tracking is worth the complexity it introduces. We do gain a few tests from it, but in general just asserting the outputs should be enough. Keeping them for now but may remove them later.


.vscode/settings.json
```diff
@@ -35,5 +35,41 @@
...
+  "rust-analyzer.runnables.extraEnv": {
+    "RUST_LOG": "bitwarden_shared_unlock=debug,shared_unlock_harness=debug"
+  },
+  // Every workspace feature except `wasm`. Enabling `wasm` on a native target makes
+  // rust-analyzer analyze (and run tests against) the wasm-bindgen code paths, which do not
+  // compile/run off wasm32. Keep this list in sync with:
+  //   cargo metadata --no-deps --format-version 1 | jq -r '.packages[].features | keys[]' | sort -u
+  "rust-analyzer.cargo.features": [
+    "bitwarden-license",
+    "bitwarden-test",
```
quexten, line 54: We can move these out to a separate PR; i'll leave that up to any reviewer. Feel free to veto them being included here.

.vscode/settings.json
```diff
@@ -35,5 +35,41 @@
...
+  // (`--show-output`) already prints a passing test's captured output, which is where these lines
+  // land. A failing assertion prints the captured log on its own, with or without this.
+  "rust-analyzer.runnables.extraEnv": {
+    "RUST_LOG": "bitwarden_shared_unlock=debug,shared_unlock_harness=debug"
+  },
+  // Every workspace feature except `wasm`. Enabling `wasm` on a native target makes
+  // rust-analyzer analyze (and run tests against) the wasm-bindgen code paths, which do not
+  // compile/run off wasm32. Keep this list in sync with:
+  //   cargo metadata --no-deps --format-version 1 | jq -r '.packages[].features | keys[]' | sort -u
+  "rust-analyzer.cargo.features": [
```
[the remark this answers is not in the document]
quexten: This was meant for humans only. Sorry bots have no veto power here.

dereknance (dismissed):
✅ Platform

.vscode/settings.json
```diff
@@ -35,5 +35,41 @@
...
+    "internal-test-utils",
+    "mockall",
+    "no-memory-hardening",
+    "performance-tracing",
+    "secrets",
+    "snow",
+    "test-fixtures",
+    "test-support",
+    "test-utils",
+    "uniffi"
```
mzieniukbw: Why not exclude uniffi as well? It is not needed, except when building for mobile, wasted cpu cycles in most cases.

.vscode/settings.json
mzieniukbw: Does this need an announcement in engineering group for awareness ?
I also wonder, wouldn't that degrade coding experience  in VSCode ?

crates/bitwarden-shared-unlock/tests/harness/logs.rs
```diff
@@ -0,0 +1,318 @@
...
+    })
+}
+
+/// Installs the capture subscriber if it is not already installed. Every topology calls this.
+pub(crate) fn install() {
+    let _ = state();
+}
+
+/// Milliseconds since the capture buffer was installed, so assertions can bound a window.
+pub(crate) fn now_ms() -> u128 {
```
mzieniukbw, line 176: nit: This probably don't belong in this module.

crates/bitwarden-shared-unlock/tests/scenarios/client_quirks.rs
```diff
@@ -0,0 +1,118 @@
+//! Behaviours real clients exhibit that the plain protocol does not model.
+
+use crate::prelude::*;
+
+/// A follower replaying an applied lock as a manual lock changes nothing
+///
+/// ```text
+/// follower 🔓 --> 🔒 --replay as manual lock--> stays 🔒
```
mzieniukbw: These two icons are near identical. Maybe user different ones or just use text ?

crates/bitwarden-shared-unlock/tests/scenarios/client_quirks.rs
```diff
@@ -0,0 +1,118 @@
...
+        "the browser's process reload to complete",
+        CONVERGE_TIMEOUT,
+        |event| {
+            event.kind() == Some(kind::RELOAD)
+                && event.field("detail") == Some("process reload complete")
+        },
+    )
+    .await;
+    // 3. Give it several ticks and assert it reloaded exactly once.
+    bitwarden_threading::time::sleep(fast_timing().sync_interval * 6).await;
```
mzieniukbw: Is this magic number have any special meaning, except the "for safety" ? I wonder if there is any better way to do this, so we don't end up with flaky tests ?

crates/bitwarden-shared-unlock/tests/scenarios/v_shaped_topology.rs
mzieniukbw, line 1: how about a test(s), where browser goes offline (initially), does not share the new state, but when it comes online it syncs to other web ?
I guess you could do the same for tree topology.

crates/bitwarden-shared-unlock/Cargo.toml
```diff
@@ -50,7 +51,10 @@ js-sys = { workspace = true }
 [dev-dependencies]
 bitwarden-encoding = { workspace = true }
 bitwarden-ipc = { workspace = true, features = ["test-support"] }
+# Self dependency so the integration tests are compiled against the crate with `test-support` on.
+bitwarden-shared-unlock = { path = ".", features = ["test-support"] }
```
mzieniukbw, line 55: I don't understand this.

quexten: Some functions such as `create_for_test` are not meant to be public API. They are gated by the `test-support` feature. This means that when you compile the tests, by default the compiler says "function doesn't exist" because by default the tests are compiled with the crate's default features, not all features, and test-support is not a default feature. `[dev-dependencies]` are included for tests (but not prod builds f.e). So we just add the own crate but with the `[test-support]` dep so that we have the feature available during the tests for that crate.

quexten: LMK if that makes sense. I didn't find a better way to do this in rust + cargo :(

.vscode/settings.json
quexten: Hmm; maybe it's better to just split out these changes to a different PR...

crates/bitwarden-shared-unlock/tests/scenarios/v_shaped_topology.rs
quexten: We had a similar test in leader_unavailable, but i added a few more for this scenario now. Essentially, the leader_unavailable now has:

1. follower locks / unlocks while offline
2. leader locks / unlocks while offline
   (I'll note that leader and follower, after the rewrite mean much less, but it's still useful / the integration tests shouldn't have to know about that)

Then, I added similar cases to the v-shaped and tree topology.

crates/bitwarden-shared-unlock/tests/harness/logs.rs
quexten: Do you mean the entire logs concept or just this function? 🤔

crates/bitwarden-shared-unlock/tests/scenarios/client_quirks.rs
quexten: I'm a bit hesitant to add specific hooks for integration tests. I just bumped the time waaay up now to multiple orders of magnitude, so maybe that's fine 🤔

crates/bitwarden-shared-unlock/src/lib.rs
```diff
@@ -133,7 +133,7 @@ pub const SYNC_INTERVAL: std::time::Duration = std::time::Duration::from_secs(5)
 /// leader.
 pub const VAULT_TIMEOUT_GRACE_PERIOD: std::time::Duration = std::time::Duration::from_secs(2);
 /// How long a peer may go without syncing before it is pruned and no longer synced to.
-pub const PEER_STALE_AFTER: std::time::Duration = std::time::Duration::from_secs(30);
+pub const PEER_STALE_AFTER: std::time::Duration = std::time::Duration::from_secs(10);
```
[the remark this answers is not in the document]
quexten: Given that the protocol changed, it seems reasonable that if within 10s there is no response, we can drop it. I guess we can move the change out if needed 🤔

mzieniukbw (approved):

PR #1403, opened 2026-08-20
[PM-40985] Introduce managed settings crates

djsmith85:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-40985

## 📔 Objective

This PR introduces 2 new crates aimed to define types and a client to consume client configuration forced by an operating system's Unified Endpoint Management (UEM/MDM) channel. 

These new crates are not wired up to `bitwarden-core/ClientBuilder` or `bitwarden_pm::PasswordManagerClient` yet.
In follow up PRs I'll be adding support for wasm and uniffi and include the plumbing which will cause a breaking change for TS clients and mobile clients

### Changes
- **Create bitwarden-managed-settings-types crate** - https://github.com/bitwarden/sdk-internal/commit/a192bf94a2f938de7b0d5860ddeb60d07ded8925
- **Create bitwarden-managed-settings crate** - https://github.com/bitwarden/sdk-internal/commit/3dc009b6321f81e82951c2dc969a0db5137e44da

### DIFF
--- a/crates/bitwarden-managed-settings-types/README.md
+++ b/crates/bitwarden-managed-settings-types/README.md
@@ -1,0 +1,25 @@
+# Bitwarden Managed Settings Types
+
+Contains the shared data types for managed settings — configuration an organization's IT
+administrators enforce through the host operating system's Unified Endpoint Management (UEM)
+channel.
+
+[`ManagementProfile`] is a point-in-time snapshot of that configuration: a schema
+[`version`](ManagementProfile::version), an [`updated_at`](ManagementProfile::updated_at) timestamp
+recording when the host last refreshed it, and a [`settings`](ManagementProfile::settings) map of
+dotted keys (e.g. `"environment.base"`) to JSON-encoded value strings. Values are stored as plain
+`String` rather than `serde_json::Value` because `String` has a UniFFI representation; callers
+decode on demand.
+
+Read a profile with [`ManagementProfile::is_managed`] to test whether a key is administrator-forced,
+[`ManagementProfile::get`] for the raw JSON-encoded string, or [`ManagementProfile::get_as`] to
+decode into a concrete type — which fails with [`ManagedSettingsError::Decode`] when the stored
+value does not match the requested shape. [`ManagementProfile::empty`] builds a profile that manages
+nothing.
+
+This crate exists separately from `bitwarden-managed-settings` to break a dependency cycle:
+`bitwarden-core` must name [`ManagementProfile`] to hold the shared profile cell, while the
+higher-level `ManagedSettingsClient` handle depends on `bitwarden-core`.
+
+Managed settings are administrator-supplied client configuration. They are not Vault Data, involve
+no cryptography, and carry no key material.
--- a/crates/bitwarden-managed-settings/README.md
+++ b/crates/bitwarden-managed-settings/README.md
@@ -1,0 +1,27 @@
+# Bitwarden Managed Settings
+
+Provides [`ManagedSettingsClient`], the SDK's read handle onto the settings an organization's IT
+administrators enforce through the host operating system's Unified Endpoint Management (UEM)
+channel.
+
+The host application constructs a [`ManagedSettingsClient`] at startup with
+[`ManagedSettingsClient::new`], acquires a [`ManagementProfile`] from the host platform, and pushes
+it in with [`ManagedSettingsClient::update_profile`]. Passing `None` clears the profile. Clones of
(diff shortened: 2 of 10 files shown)


crates/bitwarden-managed-settings/README.md
```diff
@@ -0,0 +1,27 @@
...
+decode it into a concrete type with [`ManagementProfile::get_as`] on the profile returned by
+[`ManagedSettingsClient::current_profile`]. Guard writes to the corresponding state with
+[`ManagedSettingsClient::is_managed`], only writing when the key is not administrator-forced.
+
+The first profile push is asynchronous on every platform, so a consumer may not observe a managed
+setting immediately after startup.
+
+A managed setting overrides user and global state and built-in defaults. It carries no automatic
+precedence over an enterprise policy — where a policy and a managed setting both bear on one
+effective setting, the consuming feature is responsible for resolving the conflict.
```
dereknance, line 23: ⛏️ a little rewording helps clarify intent here I think.

```suggestion
Generally, a managed setting overrides user state, global state, and any defaults. When a policy
and a managed setting both affect a given setting, the consuming feature is responsible
for resolving the conflict.
```

crates/bitwarden-managed-settings-types/README.md
```diff
@@ -0,0 +1,25 @@
...
+
+Read a profile with [`ManagementProfile::is_managed`] to test whether a key is administrator-forced,
+[`ManagementProfile::get`] for the raw JSON-encoded string, or [`ManagementProfile::get_as`] to
+decode into a concrete type — which fails with [`ManagedSettingsError::Decode`] when the stored
+value does not match the requested shape. [`ManagementProfile::empty`] builds a profile that manages
+nothing.
+
+This crate exists separately from `bitwarden-managed-settings` to break a dependency cycle:
+`bitwarden-core` must name [`ManagementProfile`] to hold the shared profile cell, while the
+higher-level `ManagedSettingsClient` handle depends on `bitwarden-core`.
```
dereknance, line 22: ⛏️ Some suggestions that are just my personal taste.


```suggestion
Contains the shared data types for managed settings, which is configuration an organization's IT
administrators enforce through the host operating system's Unified Endpoint Management (UEM)
channel.

[`ManagementProfile`] is a point-in-time snapshot of that configuration containing a schema
[`version`](ManagementProfile::version), an [`updated_at`](ManagementProfile::updated_at) timestamp
recording when the host last refreshed it, and a [`settings`](ManagementProfile::settings) map of
dotted keys (e.g. `"environment.base"`) to JSON-encoded value strings. Values are stored as plain
`String` rather than `serde_json::Value` because `String` has a UniFFI representation.

Read a profile with [`ManagementProfile::is_managed`] to test whether a key is administrator-forced,
[`ManagementProfile::get`] for the raw JSON-encoded string, or [`ManagementProfile::get_as`] to
decode into a concrete type. Decoding fails with [`ManagedSettingsError::Decode`] when the stored
value does not match the requested shape. [`ManagementProfile::empty`] builds a profile that manages
nothing.

This crate exists separately from `bitwarden-managed-settings` to break a dependency cycle where
`bitwarden-core` must name [`ManagementProfile`] to hold the shared profile cell, while the
higher-level `ManagedSettingsClient` handle depends on `bitwarden-core`.
```

dereknance (approved):
These changes look great. I only have a couple non-blocking ⛏️ on READMEs.

crates/bitwarden-managed-settings-types/README.md
dereknance: For some reason, the diff on that suggestion doesn't show the precise changes.

- `managed settings — configuration an organization's IT` ➡️ `managed settings, which is configuration an organization's IT`
- `that configuration: a schema` ➡️ `that configuration containing a schema`
- `UniFFI representation; callers decode on demand.` ➡️ `UniFFI representation.`
- `a concrete type — which fails with` ➡️ `a concrete type. Decoding fails with`
- `dependency cycle: bitwarden-core` ➡️ `dependency cycle where bitwarden-core`

crates/bitwarden-managed-settings/README.md
djsmith85: Changes applied with e95422f604313a7c8079c6b8bf2248461d0b34d9

crates/bitwarden-managed-settings-types/README.md
djsmith85: Changes applied with 449164095e7220b779c216cee8dc42000289bc33

PR #1409, opened 2026-08-25
feat(ipc): route RPC responses on per-request dedicated topics [Option A]

coroiu:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-42445

> **One of two alternative designs for PM-42445.** This is **Option A (per-request dedicated response topic)**. The sibling PR #1410 implements **Option B (deferred-body envelope)** as a non-breaking alternative. They are mutually exclusive — pick one, close the other.

## 📔 Objective

Redesign the RPC request-response correlation. Today every response is published on a single shared `RpcResponseMessage` topic, and the client subscribes to that catch-all and loops, filtering by `request_id`. Because the typed subscription deserializes the body *before* `request_id` is inspected, a concurrent response of a different type either force-fits into the wrong `Response` type or (with the #1392 skip-fix) gets silently skipped — which also swallows a genuinely malformed response to *this* request and turns it into an indefinite wait.

This PR gives each request that expects a reply a **dedicated response topic** (`"RpcResponseMessage:{request_id}"`). The requester subscribes only to that topic before sending, and the handler echoes the response onto it. Because the topic is unique per request:

- the manual receive loop and the typed-`Err` handling disappear from the request path;
- concurrent requests of different types can no longer see each other's responses (fixes the #1392 biometric-unlock concurrency bug); and
- a malformed response to this request surfaces as an error immediately instead of hanging.

The RPC layer stays cleanly layered on top of IPC: the topic is just an opaque routing string that IPC never interprets.

### Changes
- `RpcRequestMessage::new(request)` mints the request id and dedicated response topic and sets `request_type` from the request's `NAME`; `RpcRequestPayload::response_topic()` accessor added.
- `request()` builds the message via `RpcRequestMessage::new`, subscribes to its `response_topic`, sends, and deserializes the single response directly (no loop, no `request_id` filter).
- `handle_rpc_request` publishes the response on the request's `response_topic`.
- RPC message types no longer implement `PayloadTypeName`. Both send paths build `OutgoingMessage` directly with an explicit topic — `RPC_REQUEST_PAYLOAD_TYPE_NAME` for requests (the fixed dispatch topic the receiver matches on) and the per-request topic for responses. The RPC layer no longer goes through `TypedOutgoingMessage`; that machinery remains only for general typed pub/sub (`send_typed`, `subscribe_typed`, `LeaderMessage`/`FollowerMessage`).
- Tests updated to route responses per-topic; new `malformed_response_surfaces_error_instead_of_hanging` regression test.

## ✅ Verification

- `cargo test -p bitwarden-ipc`: 31 passed.
- WASM integration tests (`bitwarden-wasm-internal/integration-tests`): 70 passed / 11 suites, including `unlock/biometrics-ipc.test.ts`, which drives `IpcClient.request()` and the registered handlers end-to-end over IPC.

## 🚨 Breaking Changes

Yes, at the wire level: `RpcRequestMessage` gains a required `response_topic` field and responses move from the shared `RpcResponseMessage` topic to a per-request topic. Requester and handler must run matching SDK versions — mixed old/new peers (either direction) hang rather than error, and because `discover` (the version-skew detection request) rides the same path, it hangs too.

**In practice this has no impact.** The RPC request/response API has no released consumers: it is entirely behind feature flags and has never shipped, so there is no older peer to skew against. The break can land now for free. A backward-compat migration (keep the shared topic and `request_id` filter for one release, make `response_topic` optional) would only be warranted if there were deployed consumers — there are none.

Note this is a runtime wire break, not a TypeScript API break, so the automated SDK compatibility check (which only compiles clients) does not flag it.

### DIFF
--- a/crates/bitwarden-ipc/src/ipc_client.rs
+++ b/crates/bitwarden-ipc/src/ipc_client.rs
@@ -11,10 +11,7 @@ use crate::{
         AlreadyRunningError, ErrorKind, IpcErrorKind, ReceiveError, SendError, SubscribeError,
         TypedReceiveError,
     },
-    message::{
-        IncomingMessage, OutgoingMessage, PayloadTypeName, TypedIncomingMessage,
-        TypedOutgoingMessage,
-    },
+    message::{IncomingMessage, OutgoingMessage, PayloadTypeName, TypedIncomingMessage},
     rpc::{
         exec::{handler::ErasedRpcHandler, handler_registry::RpcHandlerRegistry},
         request_message::{RPC_REQUEST_PAYLOAD_TYPE_NAME, RpcRequestPayload},
@@ -308,14 +305,15 @@ fn handle_rpc_request<Crypto, Com, Ses>(
                 result: response,
             };
 
-            let outgoing = TypedOutgoingMessage {
-                payload: response_message,
-                destination: incoming_message.source.into(),
-            }
-            .try_into()
-            .map_err(|e: serde_utils::SerializeError| HandleError::Serialize(e.to_string()))?;
+            // Publish the response on the dedicated topic the requester is subscribed to.
+            let payload = serde_utils::to_vec(&response_message)
+                .map_err(|e: serde_utils::SerializeError| HandleError::Serialize(e.to_string()))?;
 
-            Ok(outgoing)
+            Ok(OutgoingMessage {
+                payload,
+                destination: incoming_message.source.into(),
+                topic: Some(request.response_topic().to_owned()),
+            })
         }
 
         match handle(incoming_message, &inner.handlers).await {
@@ -985,9 +983,9 @@ mod tests {
                     tab_id: 9001,
                     document_id: "doc-1".to_string(),
                 },
-                topic: Some(
-                    IncomingRpcResponseMessage::<TestRequest>::PAYLOAD_TYPE_NAME.to_owned(),
-                ),
+                // The requester subscribes to the dedicated topic it minted, so the response must
+                // be published there.
+                topic: Some(outgoing_request.response_topic.clone()),
             };
             communication_provider.push_incoming(simulated_response);
 
@@ -996,11 +994,10 @@ mod tests {
             assert_eq!(result.unwrap().result, 3);
         }
 
-        /// A response belonging to one in-flight request must not disturb another. Both requests
-        /// subscribe to the same topic (every RPC response shares one payload type name), so each
-        /// one sees the other's response; correlating on `request_id` before deserializing the body
-        /// is what keeps `OtherResponse::Available` -- a bare JSON string -- from being force-fit
-        /// into `TestResponse` and failing the unrelated request.
+        /// Two requests in flight at once must each receive their own response. Each request mints
+        /// its own response topic, so `TestRequest` and `OtherRequest` never see each other's
+        /// responses even though their response types differ (`OtherResponse::Available` is a bare
+        /// JSON string that would not deserialize into `TestResponse`).
         #[tokio::test]
         async fn concurrent_request_of_a_different_type_does_not_disturb_a_pending_request() {
             let crypto_provider = NoEncryptionCryptoProvider;
@@ -1037,19 +1034,20 @@ mod tests {
             };
             tokio::time::sleep(Duration::from_millis(100)).await;
 
-            // Recover both request ids from the wire.
+            // Recover both requests from the wire, keyed by type, so we can answer each on the
+            // dedicated response topic it minted.
             let outgoing = communication_provider.outgoing().await;
-            let request_ids: HashMap<String, String> = outgoing
+            let requests: HashMap<String, RpcRequestMessage<serde_json::Value>> = outgoing
                 .iter()
                 .map(|message| {
                     let partial: RpcRequestMessage<serde_json::Value> =
                         serde_utils::from_slice(&message.payload)
                             .expect("Deserialization should not fail");
-                    (partial.request_type, partial.request_id)
+                    (partial.request_type.clone(), partial)
                 })
                 .collect();
 
-            let respond = |payload: Vec<u8>| {
+            let respond = |topic: String, payload: Vec<u8>| {
                 communication_provider.push_incoming(IncomingMessage {
                     payload,
                     source: Source::BrowserBackground { id: HostId::Own },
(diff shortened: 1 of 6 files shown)


crates/bitwarden-ipc/src/rpc/response_message.rs
```diff
@@ -19,9 +23,9 @@ pub struct OutgoingRpcResponseMessage<'a> {
 }
 
 impl<T> PayloadTypeName for IncomingRpcResponseMessage<T> {
-    const PAYLOAD_TYPE_NAME: &str = "RpcResponseMessage";
+    const PAYLOAD_TYPE_NAME: &str = RPC_RESPONSE_PAYLOAD_TYPE_NAME;
```
coroiu: Should we really implement `PayloadTypeName` for a generic type T? I guess we're kinda forced into it, right?

coroiu:
Refined (amended, force-pushed): dropped the vestigial `PayloadTypeName` impls on `IncomingRpcResponseMessage<T>` and `OutgoingRpcResponseMessage`. With per-request response topics, those consts were dead — the receive path never read them, and the handler set-then-overwrote the topic. Responses are now routed purely by the per-request topic: the requester subscribes untyped and deserializes the single response directly, and the handler builds the `OutgoingMessage` with the dedicated topic instead of going through `TypedOutgoingMessage`. The type system now reflects the truth — only requests carry a fixed payload-type routing topic. Side benefit: a malformed response now surfaces as `RpcError::ResponseDeserialization` rather than a vaguer `Receive(Typing)`.

crates/bitwarden-ipc/src/ipc_client.rs
```diff
@@ -308,14 +305,19 @@ fn handle_rpc_request<Crypto, Com, Ses>(
...
-                payload: response_message,
-                destination: incoming_message.source.into(),
-            }
-            .try_into()
-            .map_err(|e: serde_utils::SerializeError| HandleError::Serialize(e.to_string()))?;
+            // Publish the response directly on the dedicated topic the requester is subscribed to.
+            // Responses are routed by this per-request topic rather than a payload-type topic, so
+            // we build the `OutgoingMessage` here instead of going through
+            // `TypedOutgoingMessage` (which would derive the topic from
+            // `PayloadTypeName`).
```
coroiu: This comment only makes sense in the context of the PR that removed the usage of `TypedOutgoingMessage`. As soon as we merge this PR it won't make sense anymore. Please remove this and then modify CLAUDE.md so you remember not to make these kinds of comments in the future

crates/bitwarden-ipc/src/ipc_client.rs
```diff
@@ -985,9 +987,9 @@ mod tests {
                     tab_id: 9001,
                     document_id: "doc-1".to_string(),
                 },
-                topic: Some(
-                    IncomingRpcResponseMessage::<TestRequest>::PAYLOAD_TYPE_NAME.to_owned(),
-                ),
+                // The requester subscribes to the dedicated topic it minted, so the response must
+                // be published there rather than on the shared response payload topic.
```
coroiu: Again, weird comment, there is no shared topic because you removed it

crates/bitwarden-ipc/src/ipc_client.rs
```diff
@@ -1140,10 +1209,9 @@ mod tests {
                 serde_utils::from_slice(&outgoing_messages[0].payload)
                     .expect("Deserialization should not fail");
 
-            assert_eq!(
-                outgoing_messages[0].topic,
-                Some(IncomingRpcResponseMessage::<TestResponse>::PAYLOAD_TYPE_NAME.to_owned())
-            );
+            // The response must be published on the dedicated topic carried by the request, not on
+            // the shared response payload topic.
```
coroiu: Again, there is no shared topic

crates/bitwarden-ipc/src/ipc_client_ext.rs
```diff
@@ -90,14 +92,25 @@ pub trait IpcClientExt: IpcClient {
...
+            // Mint a dedicated response topic for this request and subscribe to it before sending.
+            // The handler echoes the response onto this topic, so this subscription only ever
+            // receives the single response addressed to this request. That removes the need to
+            // correlate on `request_id` and, crucially, means a deserialization error here is
+            // genuinely a malformed response to *this* request rather than an unrelated concurrent
+            // response.
+            //
+            // The subscription is untyped: responses are routed by this per-request topic, not by a
+            // payload-type topic, so there is no reason for the response types to implement
+            // `PayloadTypeName`. We deserialize the single response directly below.
```
coroiu: Typed messages aren't used so no need to mention them. I'm going to stop commenting on this and just let you do a sweep of the entire code and make sure we DON'T have any more of these comments

crates/bitwarden-ipc/src/rpc/request_message.rs
coroiu: question: Do we still need partial messages in this rewrite?

crates/bitwarden-ipc/src/ipc_client_ext.rs
```diff
@@ -90,49 +92,44 @@ pub trait IpcClientExt: IpcClient {
...
-                .await?;
+
+            // Each request gets its own response topic, subscribed to before sending. The handler
+            // publishes its reply there, so this subscription receives exactly one message: the
+            // response to this request. A deserialization failure is therefore unambiguously a
+            // malformed response to this request.
+            let response_topic = format!("{RPC_RESPONSE_PAYLOAD_TYPE_NAME}:{request_id}");
+            let mut response_subscription = self.subscribe(Some(response_topic.clone())).await?;
 
             let request_payload = RpcRequestMessage {
```
quexten: Suggestion: Why not move the generation of the response topic, and the request id generation into a `make` / `::new` impl of RpceRequestMessage?

Seems way simpler to have: `RpcRequestMessage::new(request)` or `RpcRequestMessage::from(request)`

quexten (dismissed):

coroiu:
@quexten sorry about the force push, Claude did that :( 

@claude can you re-review?

crates/bitwarden-ipc/src/rpc/request_message.rs
```diff
@@ -46,14 +53,14 @@ pub struct RpcRequestMessage<T> {
     pub request: T,
     pub request_id: String,
     pub request_type: String,
+    /// Dedicated topic the response should be published on. See
+    /// [`RpcRequestPayload::response_topic`].
+    pub response_topic: String,
```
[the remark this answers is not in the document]
coroiu: Confirmed and accepted as a hard break. This RPC request/response API has no released consumers — it is entirely behind feature flags and has never shipped — so there is no older peer to skew against and `discover` cannot hang against one in the field. Rationale recorded in the PR description's Breaking Changes section. No migration or timeout added, by design (no deployed consumers to protect).

crates/bitwarden-ipc/src/ipc_client_ext.rs
coroiu: Done — added `RpcRequestMessage::new(request)` that mints the request id and dedicated response topic and sets `request_type` from the request's `NAME`. `request()` now calls it and subscribes to `request_payload.response_topic`. Thanks for the nudge.

coroiu:
@claude please re-review this PR. Changes since your last pass:
- Dropped the `PayloadTypeName` impls on the RPC message types; both request and response sends now build `OutgoingMessage` directly with an explicit topic (`RPC_REQUEST_PAYLOAD_TYPE_NAME` for requests, the per-request topic for responses).
- Added `RpcRequestMessage::new(request)` to encapsulate request-id + response-topic generation.
- The wire-break / `discover` hang you flagged is accepted as a hard break: this RPC API has no released consumers (all behind feature flags, never shipped), so there is no peer to skew against. Rationale is in the Breaking Changes section.

quexten (approved):

dereknance (approved):

coroiu:
Confirmed discover RPC requests still work, merging

PR #1411, opened 2026-08-25
[PM-8313] Add WASM bindings for FIDO2 authenticator operations

blackwood:
## 🎟️ Tracking

[PM-8313/PM-42179](https://bitwarden.atlassian.net/browse/PM-42179)

## 📔 Objective

Adds WASM bindings for FIDO2 so the browser's authenticator can move into the
SDK (PM-8313). Three pieces:

### 1. New binding layer — `crates/bitwarden-fido/src/wasm/`

Bindings for `Fido2Authenticator`, `Fido2CredentialStore` and
`Fido2UserInterface`, exposed through `PlatformClient::fido2()`.

This is a wrapper over the existing crate, not a refactor of it — **no existing
Rust caller changes behavior.** The shape mirrors the UniFFI bridge
(`crates/bitwarden-uniffi/src/platform/fido2/mod.rs`) method for method.

The host traits require `Send + Sync` while `wasm_bindgen` only hands back
thread-bound `JsValue`s, so each callback is wrapped in a `ThreadBoundRunner`.
Every callback uses `#[wasm_bindgen(method, catch)]`, so a JavaScript throw
arrives as `Err` rather than trapping the WebAssembly instance.

`is_verification_enabled` reads as an untyped `JsValue` and is validated, not
typed as `bool` at the boundary: a host that implements it as a method (the
shape the Rust trait and UniFFI use) would otherwise hand back a function
object that the boolean ABI silently coerces to `false`, skipping user
verification with no error. Anything that isn't a boolean is logged and treated
as **enabled**, so a mistake costs a prompt rather than a weakened ceremony.

### 2. `MakeCredentialResult` gains `public_key` and `public_key_algorithm`

Filled by the new `public_key_der_and_algorithm` helper
(`crates/bitwarden-fido/src/types.rs:496`). The helper calls the same
`passkey::authenticator::public_key_der_from_cose_key` that
`passkey-client::register` uses, so the two levels can't drift apart.

The WebAuthn-level `getPublicKey()` form was the one field the CTAP-level
result didn't carry. The fields are appended **after** `extensions` because
adding a field changes the arity of the generated positional constructors.

### 3. Known downstream impact

`MakeCredentialResult` is a UniFFI `Record`, so the two new fields break three
Swift test-fixture construction sites in the iOS repo. No production Swift
constructs it, and Android doesn't use the type.

See the Breaking Changes section for the TypeScript side.

## 🚨 Breaking Changes

**1. What changed in the client interface**

`PlatformClient` gains a `fido2(): Fido2Client` method. Anything that satisfies
`PlatformClient` with an object literal now fails to compile until it supplies
the new method. Two specs in clients do exactly that:

- `libs/common/src/platform/services/sdk/default-sdk.service.spec.ts:312`
- `libs/common/src/platform/services/sdk/register-sdk.service.spec.ts:169`

Nothing existing changed shape — the break is purely additive.

**2. Why the change was necessary**

FIDO2 credential provider operations had no WASM surface at all; the CTAP layer
was only reachable through UniFFI. Moving the browser's authenticator into the
SDK (PM-8313) needs `make_credential` / `get_assertion` callable from
TypeScript, and the entry point has to hang off `PlatformClient` alongside
`state()` and `load_flags()`, matching how UniFFI already exposes it at
`crates/bitwarden-uniffi/src/platform/mod.rs:45`.

**3. Migration steps for client developers**

If you mock `PlatformClient` with an object literal, add the method:

```ts
client.platform.mockReturnValue({
  state: jest.fn().mockReturnValue(mock()),
  load_flags: jest.fn().mockReturnValue(mock()),
  fido2: jest.fn().mockReturnValue(mock()),   // add this
  free: mock(),
  [Symbol.dispose]: jest.fn(),
});
```

Bump @bitwarden/sdk-internal and @bitwarden/commercial-sdk-internal in the
same PR as the mock change. They can't be split: clients main pins
0.2.0-main.1034, whose typings have no fido2(), so adding the property on
its own trips TypeScript's excess-property check (TS2353) and turns main red. 

No runtime migration — existing calls are unaffected.

4. Paired client PR

[bitwarden/clients#23240](https://github.com/bitwarden/clients/pull/23240) (PM-8313/sdk-fido2-host-traits) — already carries
both mock fixes; needs the version bump once this merges.

### DIFF
--- a/crates/bitwarden-fido/src/authenticator.rs
+++ b/crates/bitwarden-fido/src/authenticator.rs
@@ -2,6 +2,7 @@ use std::sync::Mutex;
 
 use bitwarden_core::Client;
 use bitwarden_crypto::CryptoError;
+use bitwarden_error::bitwarden_error;
 use bitwarden_vault::{CipherError, CipherView, EncryptError, VaultClientExt};
 use itertools::Itertools;
 use passkey::{
@@ -36,7 +37,7 @@ pub enum GetSelectedCredentialError {
 
 #[allow(missing_docs)]
 #[derive(Debug, Error)]
-#[cfg_attr(feature = "uniffi", derive(uniffi::Error), uniffi(flat_error))]
+#[bitwarden_error(flat)]
 pub enum MakeCredentialError {
     #[error(transparent)]
     PublicKeyCredentialParameters(#[from] PublicKeyCredentialParametersError),
@@ -44,13 +45,15 @@ pub enum MakeCredentialError {
     UnknownEnum(#[from] UnknownEnumError),
     #[error("Missing attested_credential_data")]
     MissingAttestedCredentialData,
+    #[error("The created credential's public key algorithm is not supported")]
+    UnsupportedPublicKeyAlgorithm,
     #[error("make_credential error: {0}")]
     Other(String),
 }
 
 #[allow(missing_docs)]
 #[derive(Debug, Error)]
-#[cfg_attr(feature = "uniffi", derive(uniffi::Error), uniffi(flat_error))]
+#[bitwarden_error(flat)]
 pub enum GetAssertionError {
     #[error(transparent)]
     UnknownEnum(#[from] UnknownEnumError),
@@ -66,7 +69,7 @@ pub enum GetAssertionError {
 
 #[allow(missing_docs)]
 #[derive(Debug, Error)]
-#[cfg_attr(feature = "uniffi", derive(uniffi::Error), uniffi(flat_error))]
+#[bitwarden_error(flat)]
 pub enum SilentlyDiscoverCredentialsError {
     #[error(transparent)]
     Cipher(#[from] CipherError),
@@ -80,7 +83,7 @@ pub enum SilentlyDiscoverCredentialsError {
 
 #[allow(missing_docs)]
 #[derive(Debug, Error)]
-#[cfg_attr(feature = "uniffi", derive(uniffi::Error), uniffi(flat_error))]
+#[bitwarden_error(flat)]
 pub enum CredentialsForAutofillError {
     #[error(transparent)]
     Cipher(#[from] CipherError),
@@ -180,6 +183,9 @@ impl<'a> Fido2Authenticator<'a> {
             .attested_credential_data
             .ok_or(MakeCredentialError::MissingAttestedCredentialData)?;
         let credential_id = attested_credential_data.credential_id().to_vec();
+        let (public_key, public_key_algorithm) =
+            public_key_der_and_algorithm(&attested_credential_data.key)
+                .ok_or(MakeCredentialError::UnsupportedPublicKeyAlgorithm)?;
         let extensions = response.unsigned_extension_outputs.into();
 
         Ok(MakeCredentialResult {
@@ -187,6 +193,8 @@ impl<'a> Fido2Authenticator<'a> {
             attestation_object,
             credential_id,
             extensions,
+            public_key,
+            public_key_algorithm,
         })
     }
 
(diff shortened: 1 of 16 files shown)


blackwood:
See https://github.com/bitwarden/clients/pull/23240

crates/bitwarden-fido/src/wasm/credential_store.rs
```diff
@@ -0,0 +1,93 @@
...
+"#;
+
+#[wasm_bindgen]
+extern "C" {
+    /// JavaScript implementation of the credential store consulted during a FIDO2 ceremony.
+    #[wasm_bindgen(js_name = Fido2CredentialStore, typescript_type = "Fido2CredentialStore")]
+    pub type RawJsFido2CredentialStore;
+
+    /// `rip_id` is misspelled to match [Fido2CredentialStore::find_credentials]. Renaming it here
+    /// would silently stop matching the JavaScript object.
```
iinuwa, line 26: It might be good to go ahead and correct this now than perpetuate it.

crates/bitwarden-fido/src/wasm/user_interface.rs
```diff
@@ -0,0 +1,222 @@
...
+        }),
+        Err(error) => {
+            tracing::error!(
+                ?error,
+                "is_verification_enabled threw; treating verification as enabled"
+            );
+            true
+        }
+    }
+}
```
iinuwa, line 142: This feels fragile: This implementation, for example, would make it so that storing 2FA WebAuthn credentials in Bitwarden would always require user verification, which would make it 3FA.

Have we considered documenting this difference between WASM API and the UniFFI API and making a unit test to confirm that the API is called correctly from TypeScript?

Alternative, we set this value to true unconditionally in all our clients, and I don't see a reason to disable user verification in the future. This has tripped up multiple developers, and this method is only here to satisfy passkey-rs's API. I think we should just set this in the bitwarden-fido SDK internally, so there is no ambiguity for other clients. I think we can start with this WASM API and eventually push it further down into the SDK and update the Android and iOS implementations.)

crates/bitwarden-fido/src/authenticator.rs
```diff
@@ -850,9 +880,16 @@ mod tests {
     /// authenticator, we have disabled the configuration.
     /// When we implement PRF, this test should be updated to test that PRF _is_
     /// evaluated when PRF extension input is received.
```
iinuwa, line 882: I can't tell if this is a GitHub display issue, but it looks like this comment is not attached to the right method anymore

iinuwa (commented):
I'm not familiar with WASM bindings, but I left a few comments from the FIDO API perspective

crates/bitwarden-fido/src/wasm/user_interface.rs
blackwood: 47631e37

The WASM adapter now always reports true; host property removed

crates/bitwarden-fido/src/wasm/credential_store.rs
blackwood: renamed, except for UniFFi, because the name would have to change on the Swift client there, too. JS only sees the position, not the name.

c7c1e403

crates/bitwarden-fido/src/authenticator.rs
```diff
@@ -850,9 +880,16 @@ mod tests {
...
-    #[tokio::test]
-    async fn test_prf_is_not_evaluated() {
+    async fn create_test_client() -> Client {
```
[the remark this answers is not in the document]
blackwood: b341f278

PR #1413, opened 2026-08-26
[PM-41447] Add destination filtering to shared unlock

quexten:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-41447

## 📔 Objective

Let a shared unlock peer declare which client kinds (browser, desktop, web) it shares unlock state with. This is necessary to implement the setting for enabling / disabling shared unlock.

Filtering applies in both directions: a client not in the set is never synced to, and its incoming syncs are dropped on arrival. The set defaults to empty, so a peer neither sends nor accepts anything until the client calls `setDestinations`.

## 🚨 Breaking Changes

`SharedUnlockPeer` now syncs to nothing by default. Clients must call `setDestinations` after construction to restore syncing.

### DIFF
--- a/crates/bitwarden-shared-unlock/src/peer.rs
+++ b/crates/bitwarden-shared-unlock/src/peer.rs
@@ -344,10 +344,6 @@ impl<D: SharedUnlockDriver + Send + Sync + 'static> SharedUnlockPeer<D> {
     }
 
     async fn sync_user_to(&self, user_id: UserId, target: &SyncTarget) {
-        if !self.is_destination(user_id, &target.endpoint) {
-            return;
-        }
-
         if !self.may_sync_user_to(user_id, target).await {
             return;
         }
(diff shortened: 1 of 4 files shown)


crates/bitwarden-wasm-internal/integration-tests/tests/unlock/shared-unlock.test.ts
```diff
@@ -73,7 +73,10 @@ async function setupPair(options: {
 }) {
   init_sdk();
 
-  const [followerBackend, leaderBackend] = makeMockTransportPair("DesktopMain", "DesktopRenderer");
+  const [followerBackend, leaderBackend] = makeMockTransportPair(
+    { BrowserBackground: { id: "Own" } },
+    "DesktopRenderer",
```
quexten, line 78: There is a draft PR for a larger test suite which will include a test for the "no destinations set should not share" case (not this PR).

crates/bitwarden-shared-unlock/src/peer.rs
eligrubb, line 367: should `self.is_destination` be checked here too?

eligrubb, line 398: should `self.is_destination` be checked here as well?

quexten: Covered in https://github.com/bitwarden/sdk-internal/blob/fa6a5a39d4abb98ef36bdcf4a8e23670cdeb6f36/crates/bitwarden-shared-unlock/src/peer.rs#L368. Sync-targets just returns the list, used by `sync_user`, which then calls `sync_user_to`, which uses `may_sync_user_to`.

quexten: It was theoretically handled, since sync_user_to called it (next to may_sync_user_to), but i moved the is_destination check into may_sync_user_to, which is easier to think about, nice catch.

eligrubb (approved):

PR #1371, opened 2026-08-10
[PM-41798] 1Password access module

detunized:
## 🎟️ Tracking

[VULN-819](https://bitwarden.atlassian.net/browse/VULN-819) 

## 📔 Objective

Adds a direct 1Password client to `bitwarden-importers`. It logs in with the master password and
Secret Key and downloads every vault the account can open.

It is a Rust port of the 1Password module from the C# `password-manager-access` library.

What it does:

- SRP login with the Secret Key, TOTP as the second factor
- Registers the device when the server asks for it
- Downloads and decrypts all vaults and items
- Returns each item as its decrypted overview and details, the way 1Password sends them

Tests cover the crypto with known answer vectors from the C# tests, and the HTTP flows with
wiremock. I also ran it against a real account.

Not included:

- No SSO, no service accounts
- TOTP only, no Duo or WebAuthn
- Nothing in the SDK calls this yet, so the module has `#[allow(dead_code, unused_imports)]`
  for now

README.md has the remaining notes and open questions.

The next PR will introduce a 1P to BW conversion/importer layer.

Use https://github.com/detunized/onepassword-access-cli to drive this library to test against a live account.

## 🚨 Breaking Changes

None. Nothing consumes the new module yet.

[VULN-819]: https://bitwarden.atlassian.net/browse/VULN-819?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #1418, opened 2026-08-27
[PM-42599] [PM-41798] 1Password access: redact Debug output and fix crypto dep ownership

detunized:
## 🎟️ Tracking

Follow-up to #1371, addressing review feedback on that PR.

## 📔 Objective

Route `aes-gcm`, `hkdf` and `pbkdf2` renovate updates to Key Management instead of Tools; `pbkdf2`
is declared only in `bitwarden-crypto` and is not a dependency of this crate at all.

Drop `Debug` from the types holding decrypted vault data, so a stray `{:?}` or `dbg!` is a compile
error rather than a cleartext dump of passwords and SSH keys. Also drops the unused
`Display for ItemCategory`.

## 🚨 Breaking Changes

None. The access module is exported only under the `test-utils` feature and has no WASM or UniFFI
bindings.


harr1424 (dismissed):

itsadrago:
@quexten I added you for review because this is a dependency move from Tools to KM. Let us know if you have any concerns

itsadrago (approved):

harr1424 (approved):

quexten (approved):
Nice, thanks for the renovate fixes.

quexten:
@itsadrago No concerns, ty for the ping!

addisonbeck (approved):
approved the platform owned renovate file

PR #1419, opened 2026-08-27
[PM-39460] Add `UserCryptoManagementClient.within_v2_migration_grace_period` method

eligrubb:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-39460

## 📔 Objective

When v1 encryption users become eligible to upgrade to v2 encryption, we don't want to immediately prompt users to enter their master password until necessary. Users regularly enter their master password for other reasons (login, unlock) and the encrypted migration v1 -> v2 can take place then. In order to allow time for this to happen, this PR introduces a new client-callable SDK method that tracks if a user is within the grace period before we prompt them. This is done by storing the timestamp the first time the grace period is checked, and only return that a user is not within the grace period once the timestamp is 2 weeks or older. This is done via the new `v2_encrypted_migrations_grace_period_start` state in the KM state bridge. Once this PR is merged, a corresponding clients PR will introduce `v2_encrypted_migrations_grace_period_start` to clients and use the new `UserCryptoManagementClient.within_v2_migration_grace_period` method.

`V2EncryptedMigrationsGracePeriodStart` is a newly introduced wrapper type for `DateTime<Utc>`. This is necessary because `DateTime<Utc>` doesn't implement the `into_wasm_abi, from_wasm_abi` attributes, which are required for the state bridge macro.

`Cargo.toml` is updated with a new version minimum for the `chrono` dependency. This is because the `chrono::TimeDelta` type used in this PR was introduced in `0.4.32` and the const version we use introduced after.

## 🚨 Breaking Changes

Yes. This PR introduces new KM state bridge state and client's state bridge implementations will need to be updated to support this. For the `clients` repository, the PR to fix this is [clients#22653](https://github.com/bitwarden/clients/pull/22653). Similar PRs will be necessary for `android` and `ios`. The PRs should add support for new state bridge value `v2_encrypted_migrations_grace_period_start`. The new value should be persisted to disk and *not* cleared on logout or lock.

### DIFF
--- a/crates/bitwarden-user-crypto-management/src/v2_migration_grace_period.rs
+++ b/crates/bitwarden-user-crypto-management/src/v2_migration_grace_period.rs
@@ -1,0 +1,216 @@
+//! Grace period gating for the v2 encrypted migrations master-password prompt.
+//!
+//! A user gets two weeks before the client asks for the master password. The window is anchored by
+//! the state bridge value `v2_encrypted_migrations_grace_period_start`, which this module is the
+//! only consumer of. The first query starts the clock.
+
+use bitwarden_core::key_management::V2EncryptedMigrationsGracePeriodStart;
+use bitwarden_error::bitwarden_error;
+use chrono::{TimeDelta, Utc};
+use thiserror::Error;
+#[cfg(feature = "wasm")]
+use wasm_bindgen::prelude::*;
+
+use crate::UserCryptoManagementClient;
+
+/// How long the prompt stays suppressed after the window opens.
+const GRACE_PERIOD: TimeDelta = TimeDelta::weeks(2);
+
+/// Errors returned by the v2 migration grace period check.
+#[derive(Debug, Error)]
+#[bitwarden_error(flat)]
+pub enum V2MigrationGracePeriodError {
+    /// The grace period starting timestamp lives in client-managed state, which needs a bridge.
+    #[error("No state bridge registered, the v2 migration grace period is not supported")]
+    StateBridgeNotRegistered,
+}
+
+#[cfg_attr(feature = "wasm", wasm_bindgen)]
+#[cfg_attr(feature = "uniffi", uniffi::export(async_runtime = "tokio"))]
+impl UserCryptoManagementClient {
+    /// Returns whether the user is still inside the two-week v2 migration grace period.
+    ///
+    /// `true` means the master-password prompt must **not** be shown.
+    ///
+    /// The first call has a side effect: when no timestamp is stored, it writes
+    /// the current timestamp and returns `true`. The window is therefore two
+    /// weeks from the first call, not from login. A client that never calls
+    /// this method never opens the window.
+    ///
+    /// The SDK never clears the timestamp. Clearing is the client's decision
+    /// and resets the window. The timestamp is persisted so a user who logs out
+    /// often cannot avoid the prompt indefinitely.
+    ///
+    /// Returns an error when no state bridge is registered.
+    pub async fn within_v2_migration_grace_period(
+        &self,
+    ) -> Result<bool, V2MigrationGracePeriodError> {
+        let state_bridge = self.client.km_state_bridge();
+        if !state_bridge.is_bridge_registered() {
+            return Err(V2MigrationGracePeriodError::StateBridgeNotRegistered);
+        }
+
+        match state_bridge
+            .get_v2_encrypted_migrations_grace_period_start()
+            .await
+        {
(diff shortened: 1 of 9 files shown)


crates/bitwarden-user-crypto-management/src/v2_migration_permission.rs
```diff
@@ -0,0 +1,217 @@
...
+use chrono::{TimeDelta, Utc};
+use thiserror::Error;
+use tracing::info;
+#[cfg(feature = "wasm")]
+use wasm_bindgen::prelude::*;
+
+use crate::UserCryptoManagementClient;
+
+/// How long the prompt stays suppressed after the window opens.
+const GRACE_PERIOD: TimeDelta = TimeDelta::weeks(2);
```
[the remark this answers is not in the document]
eligrubb: Updated floor for `chrono` to `version = ">=0.4.34, <0.5"` - `chrono::TimeDelta` is available in all versions after this one.

quexten:
@eligrubb Could we add to the breaking changes section "v2_encrypted_migrations_grace_period_start should be persisted to disk". Mobile should auto-generate PR's via a claude automation if we add that.

crates/bitwarden-user-crypto-management/src/v2_migration_grace_period.rs
```diff
@@ -0,0 +1,216 @@
...
+    #[error("No state bridge registered, the v2 migration grace period is not supported")]
+    StateBridgeNotRegistered,
+}
+
+#[cfg_attr(feature = "wasm", wasm_bindgen)]
+#[cfg_attr(feature = "uniffi", uniffi::export(async_runtime = "tokio"))]
+impl UserCryptoManagementClient {
+    /// Returns whether the user is still inside the two-week v2 migration grace period.
+    ///
+    /// `true` means the master-password prompt must **not** be shown.
```
quexten: Thought: (not for this PR obviously): We should consider moving the migrations entirely to the SDK at some point. This function feels too low-level to expose.

crates/bitwarden-user-crypto-management/src/v2_migration_grace_period.rs
```diff
@@ -0,0 +1,216 @@
...
+    /// often cannot avoid the prompt indefinitely.
+    ///
+    /// Returns an error when no state bridge is registered.
+    pub async fn within_v2_migration_grace_period(
+        &self,
+    ) -> Result<bool, V2MigrationGracePeriodError> {
+        let state_bridge = self.client.km_state_bridge();
+        if !state_bridge.is_bridge_registered() {
+            return Err(V2MigrationGracePeriodError::StateBridgeNotRegistered);
+        }
```
quexten:
```suggestion
```

IMO we can skip this check. only IOS does not have the bridge registered, and the PR is open so this should be fine to drop.

crates/bitwarden-user-crypto-management/src/v2_migration_grace_period.rs
```diff
@@ -0,0 +1,216 @@
...
+    /// the current timestamp and returns `true`. The window is therefore two
+    /// weeks from the first call, not from login. A client that never calls
+    /// this method never opens the window.
+    ///
+    /// The SDK never clears the timestamp. Clearing is the client's decision
+    /// and resets the window. The timestamp is persisted so a user who logs out
+    /// often cannot avoid the prompt indefinitely.
+    ///
+    /// Returns an error when no state bridge is registered.
+    pub async fn within_v2_migration_grace_period(
```
quexten: Nit: It is a bit unexpected by this name that the side effect is present. We could do:

```
pub async fn request_permission_to_migrate_to_v2() -> Result<MigrationPermission, ..>

enum MigrationPermission {
   Wait,
   Migrate
}
```

which is easier to match on compared to a boolean.

However, please consider this optional feedback.

quexten (dismissed):

crates/bitwarden-user-crypto-management/src/v2_migration_grace_period.rs
eligrubb: I like this, renamed and changed the return type to match.

crates/bitwarden-user-crypto-management/src/v2_migration_grace_period.rs
eligrubb: Sounds good, check removed.

quexten (approved):

coroiu (approved):
Might wanna move core/../key_management under KM ownership, otherwise approved

PR #1433, opened 2026-08-31
PM-39040: add a rotation module to the bitwarden-pam crate

Hinton:
## 🎟️ Tracking

[PM-39040](https://bitwarden.atlassian.net/browse/PM-39040)

Stacked on #1384 (`pam/uat`). Paired `bitwarden/clients` PR to follow — see *Follow-up* below.

## 📔 Objective

Adds a `rotation` module to `bitwarden-pam`, porting the Admin Console's credential-rotation surface out of the web client and into the SDK: the domain enums, server DTO mapping, all 22 admin endpoints, the two pieces of shared business logic, request validation, and the access-connector registration crypto.

The web client's rotation layer was written against a *planned* server contract that shipped differently, so this is not a like-for-like move. The differences are the point of the PR as much as the relocation is:

| Web client today | What the server actually exposes |
|---|---|
| `/organizations/{orgId}/rotation/...` | `/organizations/{orgId}/access-connectors/...` |
| "rotation daemon" | "access connector" (kept server-side naming here; the standalone agent is still the *rotation daemon*) |
| `PUT .../target-systems/{id}/name` **+** `/policy` | one `PUT .../target-systems/{id}`, returning `204` |
| `DELETE .../target-systems/{id}` | **no delete route exists** — `disable` is how a target is retired |
| `PUT .../configs/{id}/settings` **+** `/account` | one `PUT .../configs/{id}` |

Every rotation call in the web client would 404 today, so this is what makes the feature work rather than a refactor.

### Registration crypto

This is the part worth reviewing closely. The web client derived its key with HKDF info `"pam-rotation-daemon"`, while the daemon re-derives from its token with `"sm-access-token"`. A connector registered from the web could therefore never decrypt the payload carrying the organization key — it would derive a different key and fail to authenticate.

This module uses the daemon's constants, which are also Secrets Manager's and match `bitwarden_core::auth::AccessToken`. `ConnectorToken` parses the format immediately beside where `RegistrationSecrets::into_token` mints it, so the two halves cannot drift apart again — splitting them across crates is what allowed the original divergence. The round-trip test walks a connector's entire path: mint → token string → parse → re-derive → decrypt → assert the organization key comes back verbatim.

On the `key` field, where the daemon's `examples/register.rs` ("CONTRACT C4") encrypts the raw seed's base64 and SM encrypts the derived key's base64, this follows **SM** — which is also what the web client does today. Nothing reads that field during registration; it matters only for organization-key rotation. Flagging it for whoever owns C4.

`bitwarden-rotation-daemon` carries a duplicate token parser it can now drop.

### Notes for review

- Every enum has an `Unknown` variant so a newer server never fails a whole list; writing one back is refused with `RotationError::UnrecognizedVariant`.
- `TargetSystemCreateRequest` is a `method`-tagged union rather than a struct of optionals, so the automatic/manual field combinations the server rejects are unrepresentable.
- `RotationConfigActions` derives all five action flags in one call because they are interdependent (pause/resume are complements; rotate-now and record-manual are decided by the same field). Unloaded target status fails closed.
- The server stays authoritative on the minimum rotation interval — deliberately not duplicated here.
- `PamClient`'s existing `key_store` gains a second consumer, so its doc comment now names both.

200 tests, clippy clean, builds for `wasm32-unknown-unknown` with the `wasm` feature.

## 🚨 Breaking Changes

None. Purely additive: a new `pam().rotation()` accessor and its types. No existing signature changes, so the TypeScript surface only gains members.

## Follow-up

Not in this PR — the clients-side migration, which deletes `rotation.ts`, `requests/`, `responses/`, both API services and `helpers/` (~1,500 lines) and rebinds `RotationApiService` to an SDK-backed service. Three UI changes are forced by the real contract and need product input: the target-system delete action has no endpoint to call, and both the target-system and config edit screens currently make two writes where the server takes one.

Registration crypto moving into the SDK is relevant to `@bitwarden/team-key-management-dev`.

[PM-39040]: https://bitwarden.atlassian.net/browse/PM-39040?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/bitwarden_license/bitwarden-pam/src/rotation/registration.rs
+++ b/bitwarden_license/bitwarden-pam/src/rotation/registration.rs
@@ -63,8 +63,8 @@ const DERIVE_INFO: &str = "sm-access-token";
 const TOKEN_VERSION: &str = "0";
 
 /// The token's client-kind segment. The connector builds its OAuth `client_id` as
-/// `daemon.<api_key_id>`, so this segment and that prefix have to agree.
-const TOKEN_CLIENT_KIND: &str = "daemon";
+/// `access-connector.<api_key_id>`, so this segment and that prefix have to agree.
+const TOKEN_CLIENT_KIND: &str = "access-connector";
 
 /// The locally-derived half of a registration, held between generating the key material and
 /// assembling the token around the server's response.
@@ -82,7 +82,7 @@ pub(super) struct RegistrationSecrets {
 impl RegistrationSecrets {
     /// Assembles the one-time connector token from the server's response and the local seed.
     ///
-    /// Format: `0.daemon.<api-key-id>.<client-secret>:<b64-seed>`.
+    /// Format: `0.access-connector.<api-key-id>.<client-secret>:<b64-seed>`.
     ///
     /// Consumes `self` so the seed cannot be used to mint a second token for the same connector.
     pub(super) fn into_token(self, api_key_id: Uuid, client_secret: &str) -> String {
@@ -106,7 +106,7 @@ pub enum ConnectorTokenInvalidError {
     /// The version segment was not `0`.
     #[error("Is the wrong version")]
     WrongVersion,
-    /// The client-kind segment was not `daemon`.
+    /// The client-kind segment was not `access-connector`.
     #[error("Has the wrong prefix")]
     WrongPrefix,
     /// The API key id segment was not a UUID.
@@ -133,7 +133,8 @@ pub enum ConnectorTokenInvalidError {
 /// first place. The rotation daemon carries its own copy of this parser and should converge on this
 /// one.
 pub struct ConnectorToken {
-    /// The API key identifier. The connector's OAuth `client_id` is `daemon.<api_key_id>`.
+    /// The API key identifier. The connector's OAuth `client_id` is
+    /// `access-connector.<api_key_id>`.
     pub api_key_id: Uuid,
     /// The OAuth client secret. Redacted from [`fmt::Debug`].
     pub client_secret: String,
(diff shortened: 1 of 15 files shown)


bitwarden_license/bitwarden-pam/src/rotation/registration.rs
```diff
@@ -0,0 +1,406 @@
...
+
+        let derived_key = ctx.derive_shareable_key(seed, DERIVE_NAME, Some(DERIVE_INFO))?;
+
+        // The payload hands the connector the organization key itself, so it is encrypted under the
+        // derived key - the one secret only the token holder can reproduce.
+        #[allow(deprecated)]
+        let organization_key_b64 = ctx
+            .dangerous_get_symmetric_key(organization_key)?
+            .to_base64();
+        let payload = serde_json::json!({ "encryptionKey": organization_key_b64.to_string() });
```
quexten, line 231: This is very not great. 

We 
1. couple permanent configurations to the organization key which we may need to rotate in the future, and this becomes a blocking stable api.
2. Introduce more logic based on old cryptographic apis

If the goal here is to:
"Give cryptographic access to the user key to the rotation daemon" then can I recommend just copying the design for invite cryptography? You get a secret you can give the rotation daemon, and it can access the organization key, but the organization key can be rotated, and it uses the modern "safe" cryptography apis: https://github.com/bitwarden/sdk-internal/blob/61a35d8ce7600d7f45e72ea4195deee85224768d/crates/bitwarden-organization-crypto/src/invite.rs#L9

bitwarden_license/bitwarden-pam/src/rotation/registration.rs
```diff
@@ -0,0 +1,406 @@
...
+//!
+//! A connector runs unattended, outside any user's session, so it cannot unlock a vault the way a
+//! client does. Registration therefore hands it the organization key up front, wrapped so that only
+//! the holder of one operator-provisioned token can unwrap it:
+//!
+//! 1. A fresh 16-byte seed is generated. Base64-encoded, it becomes the `:` suffix of the token.
+//! 2. A symmetric key is derived from that seed via
+//!    [`derive_shareable_key`](bitwarden_crypto::derive_shareable_key), using [`DERIVE_NAME`] and
+//!    [`DERIVE_INFO`].
+//! 3. `encryptedPayload` - `{"encryptionKey":"<org key b64>"}` - is encrypted under the derived
```
quexten, line 11: > Not in this PR — the clients-side migration, which deletes rotation.ts, requests/, responses/, both API services and helpers/ (~1,500 lines) and rebinds RotationApiService to an SDK-backed service. Three UI changes are forced by the real contract and need product input: the target-system delete action has no endpoint to call, and both the target-system and config edit screens currently make two writes where the server takes one.

> Registration crypto moving into the SDK is relevant to @bitwarden/team-key-management-dev.

Blocking: If we are asked to maintain it, then we should review it before merge. Please add us as reviewers whenever you open the PR.

bitwarden_license/bitwarden-pam/src/rotation/registration.rs
```diff
@@ -0,0 +1,406 @@
...
+        organization_id: OrganizationId,
+    ) -> Result<RegistrationSecrets, RotationError> {
+        let organization_key = SymmetricKeySlotId::Organization(organization_id);
+        let mut ctx = self.key_store.context_mut();
+
+        if !ctx.has_symmetric_key(organization_key) {
+            return Err(RotationError::MissingOrganizationKey);
+        }
+
+        let seed: Zeroizing<[u8; 16]> = generate_random_bytes();
```
quexten, line 220: Nit/thought: We should probably make get_random_bytes private. We have the bitwarden-random crate now which should be used.

quexten (dismissed):
Marking this as blocking for crypto; sent a dm to discuss. Let's get this unblocked.

bitwarden_license/bitwarden-pam/src/rotation/registration.rs
Hinton: Out of scope for the PA which this is aimed at. We can either replicate the SM approach (Which I believe is a variant of the invite flow) or design it using asymm.

bitwarden_license/bitwarden-pam/src/rotation/registration.rs
Hinton: No ask yet since this is not the final api.

PR #1434, opened 2026-09-01
Clarify usage permissions for cryptographic keys

quexten:
## 📔 Objective

Clarify ownership so that claude based review can appropriately point out the requirement for KM to review new usages of it's keys.

### DIFF
--- a/.claude/rules/crypto-usage.md
+++ b/.claude/rules/crypto-usage.md
@@ -7,24 +7,29 @@ paths:
 # Using key-management-owned keys
 
 Keys owned by `@bitwarden/team-key-management-dev`:
+
 - User Key
 - Master Key
 - User Private Key
 - User Signature Key
 - Organization Key
-- Organization Private Key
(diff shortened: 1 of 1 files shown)


.claude/rules/crypto-usage.md
```diff
@@ -0,0 +1,42 @@
+---
+paths:
+  - "crates/**"
+  - "bitwarden_license/**"
```
SaintPatrck, line 4: ❓ Does Claude really need this information _every time_ a file in these paths is touched? That includes writes **and** reads. Nearly everything lives under `crates/**`. That means this is likely injected during every session working in the repo, regardless of what it's doing.

Is there a specific action or point in time this can be more closely scoped to? It reads like a code review step.

quexten: @patrick-bitwarden It should just be code review, but yes it should be invoked on any change in any file under `crates/**` (only when reviewing). Do we have documentation on how to scope a rule for just code review?

PR #1436, opened 2026-09-01
[PM-32955] Upgrade jni to 0.22 and rustls-platform-verifier to 0.7

coroiu:
## 🎟️ Tracking

- Jira: [PM-32955](https://bitwarden.atlassian.net/browse/PM-32955)
- Supersedes the Renovate range-widening PR #803 (that PR only widened the version range and did not perform the upgrade; #803 can be closed once this lands).

## 📔 Objective

Perform the actual `jni` `0.21 → 0.22` upgrade for `bitwarden-uniffi`'s Android support, rather than just widening the range.

`jni`'s major version is coupled to `rustls-platform-verifier`: `init_verifier` passes a jni `Env` into `rustls_platform_verifier::android::init_with_env`, so both crates must resolve to the same jni major. This PR bumps `rustls-platform-verifier` to `0.7` (which depends on jni `0.22`), which unifies the dependency tree on a single `jni 0.22.4`.

Changes:
- **Workspace `Cargo.toml`:** `rustls-platform-verifier` `0.6.2 → 0.7.0`.
- **`crates/bitwarden-uniffi/Cargo.toml`:** `jni` `>=0.21, <0.22` → `>=0.22, <0.23`. The floor is `0.22` because the ported code no longer compiles against `0.21`.
- **`crates/bitwarden-uniffi/src/android_support.rs`** ported to the jni 0.22 API:
  - `JNI_OnLoad` now receives a raw `*mut jni::sys::JavaVM` and wraps it with `JavaVM::from_raw` (the `JavaVM` wrapper is no longer FFI-transparent).
  - `JavaVM::from_raw` is now infallible.
  - Thread attachment uses the closure-based `attach_current_thread`.
  - `init_verifier` borrows a `jni::Env` and builds method/type signatures with the `jni_str!` / `jni_sig!` macros instead of raw `&str` signatures.
- **`Cargo.lock`** regenerated — `jni 0.21.1` and `rustls-platform-verifier 0.6.2` removed; a single `jni 0.22.4` is now shared by `bitwarden-uniffi` and `rustls-platform-verifier 0.7.0`.
- **`.github/renovate.json5`:** added `jni` to the "reqwest + TLS crates" group (so it updates in lockstep with rustls-platform-verifier) and to the Platform team review list.

## 🦮 Reviewer guidelines

Verified locally with the same toolchain CI uses:

```
cross build -p bitwarden-uniffi --target aarch64-linux-android
```

This is required because `android_support.rs` is Android-target-only, so host `cargo check` cannot catch API breakage. The build succeeds and produces `libbitwarden_uniffi.so`.

[PM-32955]: https://bitwarden.atlassian.net/browse/PM-32955?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


coroiu:
@claude review please

coroiu:
Thanks for the thorough review. Answers to the two questions, then the smaller notes.

## 🟠 Android CI — now green

The `startup_failure` was on an earlier run. The latest run builds all four targets successfully, including the 32-bit `armv7-linux-androideabi` and `i686-linux-android` you flagged as most likely to surface pointer-width issues:

- ✅ Build (aarch64-linux-android)
- ✅ Build (armv7-linux-androideabi)
- ✅ Build (i686-linux-android)
- ✅ Build (x86_64-linux-android)
- ✅ Check for Android breaking changes

So the one blocking item is resolved by CI, not just my local aarch64 build.

## 🔵 `>=0.22` floor — verified correct, all APIs are in 0.22.0

I checked the jni-rs CHANGELOG. Every API this port uses landed in the initial **0.22.0** release, not a later patch:

- `jni_str!` / `jni_sig!` macros — 0.22.0
- closure-based `attach_current_thread(&mut Env)` — 0.22.0
- `Env` / `EnvUnowned` split — 0.22.0
- `MethodSignature` — 0.22.0
- `.l()` predates 0.22.4 (0.22.4 only *added* the `.into_object()` alias, which I don't use)

So a consumer resolving `jni 0.22.0` still has everything. On top of that, `rustls-platform-verifier 0.7.0` pins its own jni dep at `0.22.4`, so the effective tree floor is 0.22.4 regardless of what this crate declares — there's no reachable resolution where a downstream consumer gets `0.22.0`. I'll keep `>=0.22, <0.23` since it's accurate and matches how the other android-only deps here are declared.

## 🔵 Grouping only solves half the coupling — agreed

You're right: grouping guarantees same-PR, not a real port. With the `>=x, <y` range + `non-pinned` strategy, the next jni major still opens as a range-widening PR (another #803), just with rustls-platform-verifier alongside. As you note, the Android build fails loudly on the widened range so it won't merge silently. I'll add a note to PM-32955 about whether android-only deps should switch to caret versions so Renovate emits a real bump — that's a versioning-policy decision worth making deliberately rather than sneaking into this PR.

## 🔵 dep-ownership lint — acknowledged, leaving as-is

Correct that `scripts/dep-ownership.ts` only scans `[workspace.dependencies]`, so crate-level `jni` is invisible to it either way. Moving `jni` into workspace deps just to get lint coverage isn't worth it for a single android-only dep; the Platform reviewer entry is harmless and strictly better than nothing.

## 🔵 "permanent attachment" comment — fixed

Fair catch. Updated the comment to name the scoped counterpart explicitly so it's self-evidently describing 0.22 semantics rather than surviving stale from the port:

    // `attach_current_thread` requests a permanent attachment (unlike
    // `attach_current_thread_for_scope`), keeping the thread attached beyond the callback so
    // later JNI calls stay cheap. The borrowed `Env` is only valid inside the callback.

## 🔵 `JavaVM` spelling — intentional

Keeping the explicit `*mut jni::sys::JavaVM` on the `JNI_OnLoad` signature specifically to disambiguate from `jni::JavaVM` two lines down, which is exactly the reason you noted.

## Note on the remaining red check

`Check dependencies maximum versions` is failing, but it's a pre-existing failure on `main` (a `tsify` `into_wasm_abi`/`from_wasm_abi` deprecation-as-error in `bitwarden-logging`) — the last three `main` commits fail the same check. Not introduced by this PR.

dani-garcia (approved):

PR #1437, opened 2026-09-01
feat(crypto): expose get_key_id_for_symmetric_key for uniffi

quexten:
## 📔 Objective

Uniffi doesn't surface the existing function. Mobile needs the function for V2 biometrics / never lock migrations so this PR exposes it. We can refactor later.


crates/bitwarden-uniffi/src/crypto.rs
```diff
@@ -113,4 +113,15 @@ impl CryptoClient {
     pub fn get_upgraded_user_key(&self, upgrade_token: Option<V2UpgradeToken>) -> Result<B64> {
         Ok(self.0.get_upgraded_user_key(upgrade_token)?)
     }
+
```
quexten, line 116: @coroiu Should KM own this client? 🤔

dani-garcia (approved):

crates/bitwarden-uniffi/src/crypto.rs
mzieniukbw: For sure, yes, it's CryptoClient

mzieniukbw (approved):

crates/bitwarden-uniffi/src/crypto.rs
dani-garcia: we should probably just move it into a KM owned folder yeah

quexten: Will make a follow-up PR monday.

PR #1445, opened 2026-09-02
[PM-39982] Expose two Send decryption calls to support temporary item…

mcamirault:
… sharing

## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-39982

## 📔 Objective

We want to keep the decryption logic for the new Item-type Sends centralized in the SDK, and to that end I'm exposing two temporary functions that allow us to decrypt both Send and SendAccess objects (the latter being for the recipient flow). These functions should become unnecessary as we transition the clients to fully use the SDK and operate more on SendViews than Sends.

## 🚨 Breaking Changes

There should be none, this is exposing two new functions

### DIFF
--- a/crates/bitwarden-send/src/access.rs
+++ b/crates/bitwarden-send/src/access.rs
@@ -457,13 +457,17 @@ impl SendClient {
         get_file_download_data(&config.api_client, &file_id, &access_token).await
     }
 
-    /// Decrypt a [`SendAccessResponse`] into a [`SendAccessView`] using a [`SendAccessKey`].
+    /// Decrypt a [`SendAccessResponse`] into a [`SendAccessView`].
+    ///
+    /// `key_b64` is the URL-safe-base64 send key from the trailing segment of the send URL
+    /// fragment (16 bytes when decoded) — the same form [`SendAccessKey::from_url_b64`] accepts
+    ///
     /// This is a temporary function to support the transition to fully using the SDK for Send logic
     pub fn decrypt_send_access(
-        key: SymmetricCryptoKey,
+        key_b64: String,
         response: SendAccessResponse,
     ) -> Result<SendAccessView, SendAccessDecryptError> {
-        let access_key = SendAccessKey::from_url_b64(key.to_base64().to_string().as_str())?;
+        let access_key = SendAccessKey::from_url_b64(key_b64.as_str())?;
         access_key.decrypt_response(response)
     }
 }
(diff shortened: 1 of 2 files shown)


crates/bitwarden-send/src/access.rs
```diff
@@ -450,6 +450,18 @@ impl SendClient {
         let config = self.client.internal.get_api_configurations();
         get_file_download_data(&config.api_client, &file_id, &access_token).await
     }
+
+    /// Decrypt a [`SendAccessResponse`] into a [`SendAccessView`] using a [`SendAccessKey`].
+    /// This is a temporary function to support the transition to fully using the SDK for Send logic
+    #[allow(missing_docs)]
+    pub fn decrypt_send_access(
+        key_b64: &str,
```
quexten: Nit: You can use symmetric crypto key here rather than key.

crates/bitwarden-send/src/access.rs
```diff
@@ -450,6 +450,18 @@ impl SendClient {
         let config = self.client.internal.get_api_configurations();
         get_file_download_data(&config.api_client, &file_id, &access_token).await
     }
+
+    /// Decrypt a [`SendAccessResponse`] into a [`SendAccessView`] using a [`SendAccessKey`].
+    /// This is a temporary function to support the transition to fully using the SDK for Send logic
+    #[allow(missing_docs)]
```
quexten: Do you need the allow missing docs? You have docs above?

crates/bitwarden-send/src/send_client.rs
```diff
@@ -58,6 +58,17 @@ pub struct SendClient {
     pub(crate) client: Client,
 }
 
+#[allow(missing_docs)]
+#[cfg_attr(feature = "wasm", wasm_bindgen)]
+impl SendClient {
+    /// Decrypt a [`Send`] into a [`SendView`].
+    /// This is a temporary function to support the transition to fully using the SDK for Send logic
+    #[allow(missing_docs)]
```
quexten: Do you need allow missing docs? you have docs written above?

quexten (dismissed):
Looks ok, I'll note that you can make your life simpler by just using a symmetric crypto key across the FFI bountary, that auto-converts from b64.

crates/bitwarden-send/src/send_client.rs
mcamirault: Copy-paste error, fixed

crates/bitwarden-send/src/access.rs
mcamirault: Copy-paste error, fixed

crates/bitwarden-send/src/access.rs
mcamirault: ~~Done~~ AI brought up a concern about mismatched lengths with this, so we'll keep it as a String for now

crates/bitwarden-send/src/access.rs
```diff
@@ -450,6 +450,18 @@ impl SendClient {
...
+    pub fn decrypt_send_access(
+        key_b64: &str,
+        response: SendAccessResponse,
+    ) -> Result<SendAccessView, SendAccessDecryptError> {
+        let access_key = SendAccessKey::from_url_b64(&key_b64)
+            .map_err(|_| SendAccessDecryptError::Crypto(bitwarden_crypto::CryptoError::Decrypt))?;
```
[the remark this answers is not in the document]
mcamirault: Fixed

crates/bitwarden-send/src/access.rs
```diff
@@ -450,6 +450,18 @@ impl SendClient {
...
+        access_key.decrypt_response(response)
```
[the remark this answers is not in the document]
mcamirault: Fixed

itsadrago (approved):

PR #1448, opened 2026-09-03
Pin tsify

dani-garcia:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-42985

## 📔 Objective

The latest versions of tsify deprecate the use of from_wasm_abi and to_wasm_abi, as they introduce memory leaks on failures. The expected migration path is to use the new `Ts<T>` wrapper, but that requires changes to every WASM exposed function. 

Instead, we'll very likely migrate to a custom wasm_export macro that does the correct thing automatically, but that's a very large mechanical diff, so for now let's just reduce the tsify range to avoid CI failures.

https://github.com/madonoharu/tsify/blob/main/CHANGELOG.md#v057

### DIFF
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -143,6 +143,7 @@
         "tracing",
         "tracing-subscriber",
         "tsify",
+        "tsify-macros",
         "uniffi",
         "uniffi_core",
         "uniffi_macros",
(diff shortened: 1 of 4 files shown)


djsmith85 (approved):
Approving. Could you please create ticket to keep track and once picked up to decide if we migrate to `Ts<T>` or the custom wasm_export macro you mentioned. 

Please link to this PR within the ticket

dani-garcia:
I was finishing a POC of the migration yesterday before I created a ticket, but it's done now: https://bitwarden.atlassian.net/browse/PM-42985

I've linked this PR there too.

PR #1454, opened 2026-09-04
Extract AnyMap from the state registry

dani-garcia:
## 📔 Objective

Part of some BEEEP work I'm doing on improving the state handling. The registry's client-managed repositories are stored in a hand-rolled `RwLock<HashMap<TypeId, Box<dyn Any + Send + Sync>>>` with the locking and downcasting inlined at each use. This couples them together making it harder to reason and reuse. 

This PR splits the map logic into a separate type with its own tests.


djsmith85 (approved):
Really neat abstraction!

PR #1455, opened 2026-09-04
[PM-43254] Decouple settings from the repository API

dani-garcia:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-43254

## 📔 Objective

Update `Setting` to be a first class citizen rather than rely on `Repository`. Instead, the `Setting` type now builds upon a trait, which is currently implemented by directly wrapping the DB like `Repository` does. This will open up the possibility of having client-managed settings in the future, but it's not part of this PR.

Also added a blanket `Persist` trait to avoid duplicating the `Serialize+Deserialize+Send+Sync+'static` bounds everywhere.

### DIFF
--- a/crates/bitwarden-core/src/platform/state_client.rs
+++ b/crates/bitwarden-core/src/platform/state_client.rs
@@ -1,7 +1,7 @@
 use std::sync::Arc;
 
 use bitwarden_state::{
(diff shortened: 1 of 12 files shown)


harr1424 (approved):
Tools-owned change looks good

djsmith85 (approved):

PR #1456, opened 2026-09-05
[PM-35238] Add policy state and sync handlers

eliykat:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-35238

## 📔 Objective

Add policy state and sync-handler infrastructure to the SDK, so organization policies can be
persisted and are kept up to date on every sync.

Excludes updating any business logic to actually consume the state — that will be done in a
follow-up PR.

Changes:

- **Repository registration** — register `Policy` as a `bitwarden-state` repository item
  (`register_repository_item!`), add the `Add(Policy::data())` SDK-managed migration, and add the
  `PolicyRepository` entry to the `create_client_managed_repositories!` macro. Mirrors
  Cipher/Folder/Send.
- **Sync handler** — new `PolicySyncHandler` (mirroring `FolderSyncHandler`) plus the
  `PolicyResponseModel` → `Policy` and API `PolicyType` → domain `PolicyType` conversions. `on_sync`
  prefers `policies_new` and falls back to the legacy `policies` field, converts each entry, logs and
  skips any that fail to parse (including unknown/forward-compat policy types), and persists the rest
  via `replace_all`. Registered in `PasswordManagerClient::new_with_sync`.
- **Integration tests** — wire the policy repository into the WASM integration-test local state
  (repository, `seedVault` support, registration).

## 🚨 Breaking Changes

The `create_client_managed_repositories!` macro now emits a `policy` field on the generated
`Repositories` interface, so clients that register client-managed repositories must provide it (or
`null`). Client-side plumbing will be fixed up in each consuming repository using its own tooling.


dereknance (approved):
✅ Platform

PR #1462, opened 2026-09-09
fix(crypto): tolerate unparseable EncStrings across FFI boundary

quexten:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-42212

## 📔 Objective

A malformed `EncString` errored at parse time. In the WASM bindings that failure happened while
moving arguments across the FFI boundary, before the called method ran, so one bad item took down
an entire batch decryption instead of being reported as a per-item failure. This is a problem in vault decryption, because it expects to handle failures on a per item basis (see the test that's added).

`FromStr` now maps anything matching no known format to `EncString::Unparseable`, which keeps the
raw value so it round-trips unchanged. Decryption, unwrapping and `to_buffer` reject it with a
dedicated `CryptoError::UnparseableEncString`, so the failure surfaces where it can be attributed.
Call sites that must reject malformed input up front use the new `EncString::parse_strict`.

`from_buffer` (binary format) stays strict.

## 🚨 Breaking Changes

New `EncString` enum variant. Exhaustive matches on `EncString` in consumers need a new arm.

### DIFF
--- a/crates/bitwarden-crypto/src/enc_string/symmetric.rs
+++ b/crates/bitwarden-crypto/src/enc_string/symmetric.rs
@@ -24,6 +24,12 @@ const TS_CUSTOM_TYPES: &'static str = r#"
 export type EncString = Tagged<string, "EncString">;
 "#;
 
+/// Numerical representation of each [EncString] variant, as used in both the string and the binary
+/// format.
+const AES256_CBC_TYPE: u8 = 0;
+const AES256_CBC_HMAC_TYPE: u8 = 2;
+const COSE_ENCRYPT0_TYPE: u8 = 7;
+
 /// # Encrypted string primitive
 ///
 /// [EncString] is a Bitwarden specific primitive that represents a symmetrically encrypted piece of
@@ -80,6 +86,11 @@ pub enum EncString {
     Cose_Encrypt0_B64 {
         data: Vec<u8>,
     },
+    /// A string that matches no known [EncString] forma. This could be an unimplemented
+    /// variant, or a malformed variant such as an encstring with invalid length of iv/mac.
+    Unparseable {
+        raw: String,
+    },
 }
 
 #[cfg(feature = "wasm")]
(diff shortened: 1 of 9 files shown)


crates/bitwarden-wasm-internal/integration-tests/tests/crypto/unparseable-enc-string.test.ts
```diff
@@ -0,0 +1,86 @@
...
+
+    // The item key protects every field, so an unparseable one fails the whole cipher.
+    const malformed: Cipher = { ...valid, key: encstring(MALFORMED_IV) };
+    return { ciphers, valid, malformed };
+  };
+
+  it("reports the malformed cipher as a failure and still decrypts the rest", async () => {
+    const { ciphers, valid, malformed } = await setup();
+
+    const result = await ciphers.decrypt_list_full_with_failures([valid, malformed]);
```
quexten, line 70: Previously this would throw since the parse fails.

crates/bitwarden-crypto/src/enc_string/symmetric.rs
```diff
@@ -134,66 +145,70 @@ impl TryFrom<wasm_bindgen::JsValue> for EncString {
 impl FromStr for EncString {
     type Err = CryptoError;
 
+    /// Never returns an error: anything that does not match a known format becomes
+    /// [EncString::Unparseable]. Use [EncString::parse_strict] to reject those instead.
     fn from_str(s: &str) -> Result<Self, Self::Err> {
```
dani-garcia, line 150: Non-blocking improvement: I wonder if we should provide some `From<&str>` and `From<String>` implementations, to allow conversion without having the parent do error handling. In the future it might be nice if the API bindings returned the `EncString` type directly for example.

dani-garcia (dismissed):
This looks great, thanks!

crates/bitwarden-crypto/src/enc_string/symmetric.rs
quexten: Agree. Will leave it to follow-up, but:
`In the future it might be nice if the API bindings returned the EncString type directly for example.` is something I'd very much like to enable.

dani-garcia: Yeah, it's something I've looked into the past a few times. The server already has the `[EncryptedString]` annotation, and I've already created a filter to make sure the data makes it into the schema: https://github.com/bitwarden/server/blob/main/src/SharedWeb/Swagger/EncryptedStringSchemaFilter.cs#L14

With that done, it's a single config like to get openapi to emit a custom type in the rust code, unfortunately the fact that server has a single EncString annotation while the SDK splits it based on Symmetric/Asymmetric makes it a bit hard as it is.

quexten: NVM I see i missed a test and failed to run test before sending the PR :( 
I'll add your suggestion now since I'll have to get re-approval anyways.

dani-garcia (approved):

eligrubb (approved):
Looks good!

PR #1464, opened 2026-09-09
[PM-43258] Add the 1Password converter, logins only

detunized:
## 🎟️ Tracking

Follow-up to the 1Password access module (#1371). First of four PRs that turn the downloaded
vaults into an import, split to keep each one small.

## 📔 Objective

Adds `convert.rs`, mapping downloaded 1Password vaults onto the importer's `ParsedImport`:

- Vaults become folders, items keep their title and note.
- Logins get username, password and deduplicated website addresses.
- Every other category becomes a plain secure note for now. Typed mappings and custom fields
  follow in the next PRs.

Nothing is wired into `ImporterClient` yet and no bindings change. Tests replay a captured
throwaway account through the production download path: the fixtures are the server responses,
re-encrypted with a fixed session key and served from wiremock.

## 🚨 Breaking Changes

None.

### DIFF
--- a/.aikido
+++ b/.aikido
@@ -1,4 +1,7 @@
-# Test vectors are derived from a throwaway account, no real secrets there.
+# Test data and test-only code from a throwaway account, no real secrets there.
 exclude:
   paths:
     - crates/bitwarden-importers/src/importers/onepassword/access/fixtures
+    - crates/bitwarden-importers/src/importers/onepassword/fixtures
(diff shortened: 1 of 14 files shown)


itsadrago (dismissed):

coroiu (dismissed):
Approving the `.aikido` changes

itsadrago (approved):

coroiu (approved):
No new platform changes

PR #1482, opened 2026-09-15
[PM-35238] refactor(policies): reconcile policy domain naming

eliykat:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-35238

## 📔 Objective

Fix naming conventions in the policy crate. This is a blocker to introducing state, because that will spread the naming into client apps even more and make it harder to change later.

Changes:
- `PolicyView` struct -> `Policy`. This is the basic domain object that will be stored in the repository, matching `Cipher` and `Folder` conventions. It also adds a newtype for its Id.
- `Policy` trait -> `PolicyDefinition`. Avoids a conflict with the `Policy` struct, and reflects its purpose as defining how a policy type works.
- `ResolvedPolicyView<P>` -> `PolicyView<P>`. "Resolved" is unclear - this is simply a view object.
- `EnforcedPolicy<P>` -> `PolicyDecision<P>`. A policy may or may not be enforced, so `EnforcedPolicy.enforced == false` is confusing. `PolicyDecision` clearly communicates that it's the outcome of the decision engine.

After this is merged, we should be able to easily introduce state for the `Policy` struct.

## 🚨 Breaking Changes

Client apps currently use the `PolicyView` struct, which is now renamed to `Policy`. It also uses a newtype uuid for `Policy.Id`.


JaredScar (dismissed):
Yay for rename/restructuring

quexten:
> Fix naming conventions in the policy crate. This is a blocker to introducing state, because that will spread the naming into client apps even more and make it harder to change later.

Just want to note for cases like this you can work with wasm bindgen / tsify annotations to change the FFI exposed name, without having to refactor rust yet. We do this for SymmetricCryptoKey (symmetrickey) until we get around to renaming the hundreds of callers in rust.

quexten (dismissed):

djsmith85 (dismissed):

eliykat:
Holding until `rc` is cut just to be safe.

eliykat:
Resolved conflicts and cleaned up some old test naming. Relying on the Claude skill to handle the breaking changes as they are simple renames. Should be good to go.

djsmith85 (approved):

JaredScar (approved):

PR #1483, opened 2026-09-15
[PM-42581] feat: Add OrganizationUsersClient for staged-invite invite endpoints

sven-bitwarden:
## 🎟️ Tracking
[PM-42581](https://bitwarden.atlassian.net/browse/PM-42581)

## 📔 Objective
This SDK work is in support of a followup for Milestone 2 on project-invite-link. We need the ability to invite staged users, which requires backend, SDK, and frontend support.

Respecting the the bitwarden-organizations README, a new crate was introduced for organization users. This sub-organization resource will handle management for org users, and is now outfitted with an OrganizationUserclient, callable from the frontend.

> [!WARNING]
> I am not at all a Rustacean, relying on Claude and advice from fellow AC devs, following precedence set by the other invite link endpoints.

[PM-42581]: https://bitwarden.atlassian.net/browse/PM-42581?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


crates/bitwarden-organizations/src/lib.rs
```diff
@@ -14,6 +15,8 @@ use uuid::Uuid;
 #[cfg(feature = "wasm")]
 use wasm_bindgen::prelude::wasm_bindgen;
 
+uuid_newtype!(pub OrganizationUserId);
```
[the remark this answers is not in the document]
sven-bitwarden: Already verified locally

BTreston (dismissed):
as far as I can tell, LGTM

crates/bitwarden-organization-users/README.md
eliykat, line 1: I think you're right to make a new crate here, specifically for member management by admins (rather than PM user sync data or foundational type definitions). But we should have a more descriptive name. I suggest `bitwarden-member-administration`, but open to suggestions.

crates/bitwarden-pm/src/lib.rs
```diff
@@ -210,6 +212,11 @@ impl PasswordManagerClient {
         self.0.invite_link()
     }
 
+    /// Organization member operations
```
eliykat: Should note that this is specifically for administration/management.

crates/bitwarden-organization-users/src/organization_users_client.rs
```diff
@@ -0,0 +1,353 @@
...
+
+#[cfg_attr(feature = "wasm", wasm_bindgen)]
+impl OrganizationUsersClient {
+    /// Sends invites to the given staged members, promoting them to invited.
+    ///
+    /// The server handles each member separately and reports the outcome per member: a member
+    /// that is no longer staged is skipped with an error on its own row rather than failing the
+    /// batch, so a stale selection degrades to a partial send. The call fails as a whole when
+    /// the caller may not manage members or when the organization cannot add the seats the
+    /// invites need, because seats are reserved once for the entire set.
```
eliykat: I think that this rustdoc comment (and the other ones in this struct) are overly specific about the behavior of the server. I think something like this is sufficient:

> Sends invites to the given staged members, promoting them to invited and consuming a seat.
>
> Returns an Err if the entire request fails. Otherwise returns Ok containing success or failure information for each user.

crates/bitwarden-organization-users/src/organization_users_client.rs
```diff
@@ -0,0 +1,353 @@
...
+                Some(bulk_request(organization_user_ids)),
+            )
+            .await?;
+
+        OrganizationUserBulkResponse::from_list(response).map_err(Into::into)
+    }
+
+    /// Re-sends the invitation to a single invited member.
+    ///
+    /// Unlike [`bulk_reinvite`](Self::bulk_reinvite), the server answers with no body, so the
```
eliykat: `the server answers with no body` -> good indication that you're documenting the wrong thing.

crates/bitwarden-organization-users/src/organization_users_client.rs
```diff
@@ -0,0 +1,353 @@
...
+        let request = bulk_request(vec![member(MEMBER_A), member(MEMBER_B)]);
+
+        let ids: Vec<OrganizationUserId> = request
+            .ids
+            .into_iter()
+            .map(OrganizationUserId::new)
+            .collect();
+        assert_eq!(ids, vec![member(MEMBER_A), member(MEMBER_B)]);
+        assert!(request.default_user_collection_name.is_none());
+    }
```
eliykat: Not a requirement.

crates/bitwarden-organization-users/src/organization_users_client.rs
```diff
@@ -0,0 +1,353 @@
...
+
+impl OrganizationUsersClientExt for Client {
+    fn organization_users(&self) -> OrganizationUsersClient {
+        OrganizationUsersClient::from_client(self)
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use bitwarden_api_api::apis::ApiClient;
```
eliykat: Given this client is largely passthrough for the api and isn't doing any business logic of its own, I'm not sure how useful these tests are. I'd be tempted to just remove them. @quexten what do you think?

eliykat (changes requested):
Pretty straightforward, just some minor feedback.

I also asked about this general pattern in `#sdk` just to make sure we're on the right track.

crates/bitwarden-organization-users/README.md
sven-bitwarden: The README on the organizations folder suggested a new create per sub-organization-domain, like policies. I kinda of assumed we'd have a crate-per-entity, almost, though it leaves overlapping concerns subjective on where they go. Happy with the new name for sure though

crates/bitwarden-organization-users/src/organization_users_client.rs
sven-bitwarden: thing does like making up requirements... yeeted.

crates/bitwarden-organization-users/src/organization_users_client.rs
sven-bitwarden: 🪓 'd

crates/bitwarden-organization-users/src/organization_users_client.rs
sven-bitwarden: Agree, cross-repo behavior documentation probably not good.

crates/bitwarden-organization-users/README.md
sven-bitwarden: I suppose my only concern is discoverability/consistency. It's "easy" to have a pattern of `server: FooBarController`, `sdk: foo_bar_client`. If we start contextualizing and changing FooBar between the two, folks have to double-check they're looking at the right one.

crates/bitwarden-organization-users/src/organization_users_client.rs
quexten: I tend to agree. Do these use-cases require no state management by the way?

Cargo.toml
```diff
@@ -52,6 +52,7 @@ bitwarden-logging = { path = "crates/bitwarden-logging", version = "=3.0.0" }
 bitwarden-logging-macro = { path = "crates/bitwarden-logging-macro", version = "=3.0.0" }
 bitwarden-managed-settings = { path = "crates/bitwarden-managed-settings", version = "=3.0.0" }
 bitwarden-managed-settings-types = { path = "crates/bitwarden-managed-settings-types", version = "=3.0.0" }
+bitwarden-member-administration = { path = "crates/bitwarden-member-administration", version = "=3.0.0" }
```
quexten, line 55: Thought: I wonder how small we want these crates to be in the end. I don't think there is a guideline for this, but too small composition means you don't have good abstractions (i.e high coupling low cohesion).

No action needed

crates/bitwarden-member-administration/src/organization_users_client.rs
```diff
@@ -0,0 +1,116 @@
...
+    /// A required field was missing from the server response.
+    #[error(transparent)]
+    MissingField(#[from] MissingFieldError),
+}
+
+/// Client for administering the members of an organization. Every operation requires the caller
+/// to be permitted to manage the organization's members.
+#[cfg_attr(feature = "wasm", wasm_bindgen)]
+#[derive(FromClient)]
+pub struct OrganizationUsersClient {
```
quexten: I find the name a bit confusing. Maybe "OrganizationUsersManagementClient" is more appropriate? "OrganizationUsersAdministrationClient"?

If I have an API that doesn't manage but still does actions on organization users, such as getting a list of the users available when sharing collections, this I would expect under a (separate from this) client called organization users.

crates/bitwarden-member-administration/src/organization_users_management_client.rs
```diff
@@ -0,0 +1,116 @@
...
+    pub(crate) api_configurations: Arc<ApiConfigurations>,
+}
+
+#[cfg_attr(feature = "wasm", wasm_bindgen)]
+impl OrganizationUsersClient {
+    /// Sends invites to the given staged members, promoting them to invited and consuming a seat.
+    ///
+    /// Returns an `Err` if the entire request fails. Otherwise returns `Ok` containing success or
+    /// failure information for each member.
+    pub async fn send_staged_invites(
```
quexten, line 40: No action nedeed but I wanted to note that if you want camel-case in wasm, you can get it `#[wasm_bindgen(js_name = sendStagedinvites)]` which looks more natural from JS.

crates/bitwarden-member-administration/src/organization_user_bulk_response.rs
```diff
@@ -0,0 +1,113 @@
...
+    pub id: OrganizationUserId,
+    /// Why the operation failed for this member. Absent when it succeeded.
+    pub error: Option<String>,
+}
+
+impl OrganizationUserBulkResponse {
+    /// Maps a bulk list response into one outcome per member.
+    ///
+    /// A missing list is treated as empty, matching how the clients parse list responses.
+    pub(crate) fn from_list(
```
quexten: Noting, the more canonical way to write this is with an impl of the `From` trait, so `impl From<Vec<OrganizationUserBulkResponse>> for OrganizationUserBulkResponse { ...}`.

crates/bitwarden-organization-users/README.md
BTreston: The client in this regard uses `MemberActions` for naming. We could do:
 `MemberActionsService` => `member_actions_client`(instead of `OrganizationUserApiService`on the client) => `MembersController`. 
That seems to fit into existing context better than the current naming for the API service while also being intuitive. I think it also gives us a good pathway for some techdebt work to move a well scoped set of existing code into the SDK.

crates/bitwarden-member-administration/src/organization_users_management_client.rs
sven-bitwarden: @BTreston I didn't address this here - figured if we wanted to do it, we'd want it everywhere - thoughts?

crates/bitwarden-member-administration/src/organization_users_client.rs
sven-bitwarden: Done, and agreed.

Cargo.toml
sven-bitwarden: Perhaps a crate-per-MSA-module? Not sure how unwieldy that gets. I'm worried about name drift between the crate name and the actual API the crate consumes

crates/bitwarden-member-administration/src/organization_users_management_client.rs
BTreston: I didn't really have any preference. Honestly, I almost like the snake_case in the wasm bindings more because I can tell at a glance from the consumer in the client that we're now in SDK land (in addition to the sdk client boilerplate you need).

crates/bitwarden-organization-users/src/organization_users_client.rs
eliykat: No, Admin Console is mostly direct API calls, very little state management (unless we need keys :)

quexten (approved):

Cargo.toml
quexten: I think MSA should have an input here, but there is crates that are shared between different parts of the services. Let's say organization crypto core is imported by multiple msa services. Invite link may be split by "acceptor" (in password manager? maybe in core identity?) and creator in AC only.

eliykat: For what it's worth, I expect "member administration" to be a reasonably sized domain:
- fetching
- inviting, confirming
- reinviting
- removing, revoking
- editing

could all go in this crate.

crates/bitwarden-organization-users/README.md
eliykat: @sven-bitwarden My thinking was that "member administration" is self-explanatory if you're writing admin functionality for managing members. It focuses on the business domain being implemented rather than the server controller it's interacting with (which should be abstracted from the SDK consumer). 

On the other hand, I do see that `OrganizationUser` keeps one name for the same thing everywhere, and avoids an `OrganizationUser` -> `Member` rename. I can live with either.

More important I think is that we clearly disambiguate it from end-user sync data, which also represents membership/orgUsers (that's what the `organizations` crate currently is and that is not very clear). I think `actions` is too broad for this.

crates/bitwarden-member-administration/src/organization_users_management_client.rs
```diff
@@ -0,0 +1,127 @@
...
+
+/// Client for administering the members of an organization.
+#[cfg_attr(feature = "wasm", wasm_bindgen)]
+#[derive(FromClient)]
+pub struct OrganizationUsersManagementClient {
+    pub(crate) api_configurations: Arc<ApiConfigurations>,
+}
+
+#[cfg_attr(feature = "wasm", wasm_bindgen)]
+impl OrganizationUsersManagementClient {
```
eliykat, line 35: (non-blocking) still not sure about naming as I would expect this to match the crate. "member administration" and "organization user management" mean the same thing, we should pick one. (And I think I'd prefer consistency over either particular name.) But it's not worth holding up the PR for, let's just keep discussing it in the team.

eliykat (approved):

dani-garcia (approved):

PR #1487, opened 2026-09-16
[PM-43843] Add domain method to SDK

JaredScar:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-43843

## 📔 Objective

Implement a new endpoint in backend to handle prefilling the invite link of domains input


BTreston (dismissed):
LGTM

dereknance (dismissed):
LGTM!

In case it's helpful in the future, there's a GH workflow to [regenerate](https://github.com/bitwarden/sdk-internal/tree/main/crates/bitwarden-api-api#updating-bindings-after-a-server-api-change) the API bindings without needing to wait for the regular 2 week occurrence.

crates/bitwarden-api-api/src/apis/organization_domain_api.rs
```diff
@@ -122,6 +128,27 @@ impl OrganizationDomainApi for OrganizationDomainApiClient {
...
+    async fn get_all_mini<'a>(
+        &self,
+        org_id: uuid::Uuid,
+    ) -> Result<models::OrganizationDomainMiniResponseModelListResponseModel, Error> {
+        let local_var_configuration = &self.configuration;
+
+        let local_var_client = &local_var_configuration.client;
+
+        let local_var_uri_str = format!(
+            "{}/organizations/{orgId}/domain/mini",
```
eliykat, line 140: As discussed in our sync - this is not the right way to call the server. Merge your server PR first, then run https://github.com/bitwarden/sdk-internal/actions/workflows/update-api-bindings.yml and merge the PR it generates. Then update your branch from `main` and call it as a method.

eliykat (changes requested):

crates/bitwarden-organizations/src/organization_domains_client.rs
```diff
@@ -0,0 +1,172 @@
...
+        let response = self
+            .api_configurations
+            .api_client
+            .organization_domain_api()
+            .get_all_mini(organization_id.into())
+            .await?;
+
+        Ok(response
+            .data
+            .unwrap_or_default()
```
eliykat, line 75: We generally use the `require!` macro here, e.g. `require!(response.data)`. That throws if the value is `None`, which is more appropriate where you are definitely expecting a value from the server. (as opposed to the server sending a potential null value where you'd allow it)

crates/bitwarden-organizations/src/organization_domains_client.rs
```diff
@@ -0,0 +1,172 @@
...
+/// without granting them visibility into its SSO configuration.
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq)]
+#[cfg_attr(feature = "wasm", derive(Tsify), tsify(into_wasm_abi, from_wasm_abi))]
+#[serde(rename_all = "camelCase")]
+pub struct ClaimedDomain {
+    /// The claimed domain name, for example `example.com`.
+    pub domain_name: String,
+    /// Whether the organization has proven ownership of the domain via its DNS TXT record.
+    pub verified: bool,
+}
```
eliykat, line 35: I do not expect consumers to ask for "domains that are claimed but not verified" - that is a distinction made within the feature (e.g. the admin interface to manage claimed domains), but not something that outside consumers know about (e.g. pre-filling domains in the invite link dialog or for Send). Generally domains that have not been verified are not used or useful - it's a trap only.

Instead of returning the `ClaimedDomain` model out to consumers, let's filter it for them: return a `Vec<String>` of claimed **and verified** domains only. This makes the interface easier to use and it means that consumers don't have to know internals of the claimed domains feature.

crates/bitwarden-organizations/src/organization_domains_client.rs
eliykat, line 1: Should be in its own crate, per the README of this crate:

> Contains organization-related data types and structures used across the SDK.
> 
> More specific organization resources (such as policies or collections) should be located in their
own crates.

eliykat (changes requested):

crates/bitwarden-organizations/src/organization_domains_client.rs
eliykat: That said, @quexten you have expressed concern about the size of our crates and this would be creating another one - would you prefer that we start grouping organization admin functionality within a single crate (i.e. nested rather than flat structure)? Open to discussion.

eliykat (changes requested):
Please also add TS integration tests. (see `crates/bitwarden-wasm-internal/integration-tests`)

crates/bitwarden-organization-domains/src/organization_domains_client.rs
quexten: This is very general advice, but low-coupling high cohesion still applies. You generally want features within a module (crate in rust) to work together, and have a single API surface. In case the split forces you to expose internals which are not supposed to be public, it's a bad split. If you do really intend to have all of these types be public API, then that seems more reasonable. However, I would still recommend thinking about whether you can keep things in a single crate, unless there is a good reason to split it out. Having them in separate crates makes the code hard to follow.

This is up to you though.

@eliykat does admin functionality have to be a commercial only feature by the way?

PR #1496, opened 2026-09-17
[PM-42946] Autotype Encrypted IPC

coltonhurst:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-42946

## 📔 Objective

Adds encrypted IPC channels to `sdk-internal` for use with the Bitwarden Desktop Autotype GA implementation.


neuronull (dismissed):

crates/bitwarden-wasm-internal/integration-tests/tests/autotype/autotype-set-enabled-ipc.test.ts
```diff
@@ -0,0 +1,116 @@
...
+ * in-memory transports.
+ *
+ * Only `main` registers the handler, matching the one-directional channel — the
+ * renderer sends, main applies.
+ */
+async function setupClientPair(driver: AutotypeDriver = makeMockAutotypeDriver()) {
+  init_sdk();
+
+  const [mainBackend, rendererBackend] = makeMockTransportPair();
+  const main = IpcClient.newWithSdkInMemorySessions(mainBackend);
```
quexten, line 42: Sweet! I didn't realize this would come in handy for other teams too, but it seems to <3

crates/bitwarden-autotype/src/wasm.rs
```diff
@@ -0,0 +1,147 @@
...
+
+#[wasm_bindgen]
+extern "C" {
+    /// JavaScript implementation of the platform Autotype operations.
+    #[wasm_bindgen(js_name = AutotypeDriver, typescript_type = "AutotypeDriver")]
+    pub type RawJsAutotypeDriver;
+
+    /// Enable or disable autotype. Returns if setting the enabled state was successful or not.
+    #[wasm_bindgen(method, catch)]
+    async fn set_autotype_enabled(
```
quexten, line 34: Thought: I wonder whether we should add native desktop-native support for using the IPC via rust, rather than having to shuffle from DN to JS main to sdk to renderer, just have DN -> (sdk) renderer.

coltonhurst: This could potentially help future progress, so it's not a bad idea. Right now though, we still need it in the main process. One reason why is we need it for setting, enabling, and disabling the shortcut you use to activate autotype, which is set in the main process with Electron's keyboard shortcut API.

PR #1497, opened 2026-09-17
Raise the global lease ceiling to one year and enforce it where rules are written

abergs:
## 🎟️ Tracking

No ticket; found while reviewing the access-window picker work. Happy to file one.

## 📔 Objective

The admin rule editor offers lease caps up to 7 days plus a "No cap" option, but every one of them was narrowed to 24h. The ceiling turned out to have five clamp points across three repos, two of them here, so raising it on the server alone would have changed nothing.

`MAX_REQUEST_ACCESS_WINDOW_SECONDS` 24h → one year, which fixes both:

- `access_requests/models.rs`: `AccessPreCheckView::try_from` re-clamped the maximum the server had already computed, so a 7-day rule cap was narrowed to 24h before the client ever saw it (read path).
- `access_requests/validate.rs`: `validate_at` refused a longer duration or window before the wire, so a multi-day request never reached the server at all (write path).

One thing beyond the constant: **`access_rules/validate.rs` now refuses a lease duration past the ceiling at rule write.** Stored unchecked, such a value is echoed back to the admin console verbatim and then narrowed on every read, so the rule page promises access the server never grants. All three rule-configurable durations are checked, including the extension length, which the lease paths never measure against the ceiling. It bounds each configured value, not the cumulative length of a repeatedly extended lease.

## 📸 Screenshots

n/a

---

Ships with bitwarden/server#8378 and bitwarden/clients#23263. This PR and the server one **must land together**: either alone leaves the effective ceiling at 24h.


abergs:
Thanks — addressed the question and can resolve the `null` note.

**❓ should the rule-write check also bound `max_extension_duration_seconds`?** Yes, and it now does (pushed), matching the paired change in bitwarden/server#8378. It is the one rule-configurable duration the lease paths never measure against the ceiling — the server’s `RequestLeaseExtensionCommand` compares only against the rule’s own value. The three bounds are now checked together. Caveat written into the comment: this bounds each configured *increment*, not the cumulative length of a lease extended repeatedly, which has never been bounded here.

**Note on `None → LEGACY_MAX_REQUEST_ACCESS_WINDOW_SECONDS`.** You flagged it as resting on a modern server never sending `null` for `maxDurationSeconds`, and worth a second look if the paired server PR could emit `null`. It cannot: `AccessPreCheckResponseModel.MaxDurationSeconds` is a non-nullable `public int`, assigned unconditionally from `AccessPreCheckQuery`, so the field is always serialized. The `None` arm is reachable only from a server predating the bounds fields — which is exactly the case it is there for. Your reading of the contract is right and the assumption holds.

**On CI.** `Check Style`, `doc`, `fmt` and `windows-2022` fail here, and all four are pre-existing on `pam/uat` rather than from this PR:

- `fmt` flags five comment rewraps, none in code this PR touches — `entra.rs`, `cipher.rs`, and three `bitwarden-pam` lines (`access_requests/models.rs:322` is the pre-existing `slot_frees_at` doc, not the code changed here)
- `doc` fails on an unresolved `encrypt_cipher_password` link in `bitwarden-rotation-daemon`
- `Check Style` is the aggregate of those two

For a baseline: #1438 on the same branch fails those **plus** `clippy`, `dylint`, `Coverage`, `macOS-14` and `ubuntu-24.04`, and carries the identical typescript/android breaking-change result. This branch is green on clippy and dylint. I have left the pre-existing `fmt`/`doc` failures alone rather than reach into other teams’ files from a PAM ceiling change; happy to do a separate cleanup PR if that is wanted.

abergs:
Retracting my previous comment about `LEGACY_MAX_REQUEST_ACCESS_WINDOW_SECONDS` — the constant is gone.

I had added it to keep the absent-bounds fallback at 24h for a server predating those fields, and defended it above. That premise does not hold: PAM ships server, SDK and clients together behind a flag, so a server old enough to omit the bounds is not a deployment we support. Carrying a constant for it was speculative compatibility.

`AccessPreCheckView::try_from` is back to the simple form, with absent and non-positive both resolving to the ceiling, which is what the server’s own `EffectiveMax` does for a rule storing no cap:

```rust
let max_duration_seconds = positive_u32(response.max_duration_seconds)
    .map_or(MAX_REQUEST_ACCESS_WINDOW_SECONDS, |max| {
        max.min(MAX_REQUEST_ACCESS_WINDOW_SECONDS)
    });
```

That also drops the three-way `match` and its explanatory comment, and restores `pre_check_view_falls_back_when_bounds_are_absent` / `pre_check_view_treats_non_positive_bounds_as_absent` to their original expectations. Net: 18 fewer lines than the previous push, and one fewer constant to keep in step. 289 tests pass, clippy and fmt clean.

PR #1498, opened 2026-09-18
fix(ipc): add CLI endpoint addressed by host id

quexten:
## Tracking
https://bitwarden.atlassian.net/browse/PM-44029

## 📔 Objective

Adds `Endpoint::Cli` / `Source::Cli` (keyed by `HostId`, since each CLI invocation is a separate process) and `SharedUnlockClient::Cli`.

We want the CLI to be correctly reported, now that we have biometric unlock implemented, and shared unlock pending.

Note on breaking changes: We will attempt the new claude auto-fix flow on this one.

## 🚨 Breaking Changes

New `Cli` variants on the `Endpoint`, `Source`, and `SharedUnlockClient` TS unions.

This clients new error:
```
  [angular] libs/common/src/platform/ipc/ipc-session-repository.ts:75:7 - error TS2367: This comparison appears to be unintentional because the types '{ Cli: { id: HostId; }; }' and 'string' have no overlap.
  [angular]
  [angular] 75   if (endpoint === "DesktopRenderer") {
```

is fixed by removing: 
```
  if (endpoint === "DesktopRenderer") {
    return "DesktopRenderer";
  }
```

and adding a CLI case.


coroiu (approved):
Oh hey, CLI, cool!

PR #1504, opened 2026-09-18
Fix flaky passphrase test

dani-garcia:
## 📔 Objective

This test was failing as some of the dictionary words contain the substring `space` (`backspace`, `aerospace`, `myspace`). 

The test's goal was to ensure that the literal CLI separator argument `space` was converted into an actual ` `. I've updated the test to split on space and check that we have the expected number of words, which should ensure that the actual literal space character was used. I've also checked that all the words are non-empty and alphanumeric, in case a ` ` leaked into the word somehow, but that should never happen.

https://github.com/bitwarden/sdk-internal/actions/runs/35354132776/job/105629176982


djsmith85 (approved):
Changes look good.
❓ @dani-garcia: Should this test suite be owned by @bitwarden/team-tools-dev

dani-garcia:
Yeah it should, but currently all the tests are at the root of the folder, so it's cumbersome to set the ownership. I'll see if i can move them to subfolders and assign them teams there.

PR #1507, opened 2026-09-22
feat(tests): build out the identity server and cover password login

quexten:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-41875

## 📔 Objective

Implements login for integration tests.

## 🚨 Breaking Changes

None — test-only.


crates/bitwarden-wasm-internal/integration-tests/tests/auth/login.test.ts
```diff
@@ -0,0 +1,136 @@
...
+      const prelogin = await client.get_password_prelogin(vector.account.email);
+      const response = await login(client, vector.account.email, vector.account.password, prelogin);
+
+      // The account is now reachable with the issued token, and the unlock data it came back with
+      // is the account's own.
+      const authenticated = response.Authenticated;
+      expect(authenticated.accessToken).toEqual(vector.account.userId);
+
+      const unlock = authenticated.userDecryptionOptions.masterPasswordUnlock;
+      expect(unlock?.salt).toEqual(vector.account.email);
```
quexten: This will probably be not true in the future? Should we add separate salts to the test vectors?

PR #1512, opened 2026-09-23
feat(state): dev-only debug browse over the SDK state registry

coroiu:
## 🎟️ Tracking

Successor to the introspection POCs (#1446 / #1447). This scopes the work down to a single, platform-owned capability so it can land without cross-team review: the auth (login-method) and crypto (key-store) debug handles from #1447 are intentionally left out, and the framework from #1446 is gone.

## 📔 Objective

Add a dev-only, feature-gated debug-capability tree that browses the SDK's state registry generically. Every registered repository (client- and SDK-managed) is addressable by its string type name and key:

- `types()`: names of every registered repository
- `list(type)`: a repository's values as JSON (values only; the repository API lists values without keys)
- `get(type, key)` / `set(type, key, value)`: keyed access

The unique value here is reaching registered state that the public, type-safe API does not expose generically, for automated tooling. Read/structural access to the object graph is better served by logging, so this deliberately stays narrow.

**Shape**
- Each registered type captures monomorphized get/set/list shims at its registration point (`register_client_managed` and the migration path), dispatched generically by name. No `inventory`, no specialization.
- The shim machinery lives in its own `bitwarden-state::debug` module; `registry.rs` keeps only the `debug` field, its initialization, and the two registration calls, so the registry's core stays readable.
- The typed handle, `StateDebug`, is authored in `bitwarden-state` (where the registry lives) and carries its own `#[wasm_bindgen]` surface. It wraps an `Arc<StateRegistry>`; `InternalClient` exposes a gated `state_registry()` accessor to hand it out, mirroring `managed_profile_handle`.
- `PasswordManagerClient::debug()` (bitwarden-pm) is a thin root that routes to `StateDebug`. Each debug handle is `#[wasm_bindgen]` in its owning crate; `bitwarden-wasm-internal` only re-exports them, the way it already does for the other sub-clients (no wrapper module).
- Exposed to TS as a typed tree, discoverable through the generated `.d.ts` and console autocomplete: `client.debug().state().types()`, `.list(type)`, `.get(type, key)`, `.set(type, key, value)`.
- All behind a dev-only `debug-capabilities` feature per crate; build wasm with `build.sh -d`. Never compiled into production.

## 🚨 Breaking Changes

None. This adds a `FromStr` bound to `RepositoryItem::Key` (needed to parse the string key). Every current key type (`String` and the `uuid_newtype!` ids) already implements `FromStr`, so all existing registrations continue to compile, and the trait cannot be implemented outside the `register_repository_item!` macro. No public API changes when the feature is off.


crates/bitwarden-core/src/debug.rs
```diff
@@ -0,0 +1,67 @@
...
+        let Ok(value) = serde_json::from_str::<Value>(&value_json) else {
+            return;
+        };
+        self.client
+            .internal
+            .state_registry
+            .debug_set(&type_name, &key, value)
+            .await;
+    }
+}
```
coroiu: This should be moved to `bitwarden-state`

crates/bitwarden-wasm-internal/src/debug.rs
```diff
@@ -0,0 +1,56 @@
...
+//! autocomplete rather than a runtime crawl. Compiled only under the
+//! `debug-capabilities` feature.
+
+use wasm_bindgen::prelude::*;
+
+/// Root of the debug-capability tree. Reached via `PasswordManagerClient.debug()`.
+#[wasm_bindgen]
+pub struct DebugClient(bitwarden_pm::debug::DebugClient);
+
+impl DebugClient {
```
coroiu: same here, we don't create separate wasm clients and we don't put them in wasm-internal

crates/bitwarden-wasm-internal/integration-tests/tests/state-debug.test.ts
```diff
@@ -0,0 +1,30 @@
...
+  it("lists the SDK-managed repositories on the memory-backed client", () => {
+    // The wasm/mobile bindings build a memory-backed registry with no
+    // client-managed repositories wired up here, so before the fix `types()`
+    // was empty. The SDK-managed set can only appear once `debug()` registers
+    // the migration shims.
+    const client = makePasswordManagerClient(makeStateBridge()) as unknown as DebugCapable;
+
+    if (typeof client.debug !== "function") {
+      return; // Not a debug-capabilities build; nothing to exercise.
+    }
```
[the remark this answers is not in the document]
coroiu: adressed

PR #1513, opened 2026-09-23
test(ipc): add behavior test suite (native + wasm boundary)

coroiu:
## 🎟️ Tracking

No ticket. Exploratory review of the IPC framework fundamentals (lifetimes, concurrency across the Rust/WASM/JS divide), turned into a test suite. Draft for review.

## 📔 Objective

Encodes the properties the IPC framework is meant to guarantee as tests, and turns the currently-violated ones into a tracked, reproducible checklist.

**Scheme:** a passing test means a behavior holds. An `#[ignore]`d native test (or a skipped wasm test) documents a behavior that is currently violated, with the reason string explaining the gap and where it lives. Default `cargo test` / `jest` stay green; the gaps are opt-in.

### Native (`crates/bitwarden-ipc/src/behavior_tests.rs`)

Guaranteed (passing):
- Concurrent same-type requests each receive their own response
- Concurrent encrypted sends all decrypt (no nonce reuse under concurrency)
- Framework does not filter incoming by destination (pins the contract: the backend must)
- Subscriptions do not replay pre-subscription messages
- A request returns when its cancellation token fires

Documented gaps (ignored; run `cargo test -p bitwarden-ipc --lib -- --ignored` to see them fail):
1. A failing session store panics instead of returning a `SendError` (`crypto_provider.rs:212` and the other `.expect()`s). Latent today: production uses in-memory sessions, which never fail.
2. One slow handshake serializes all sends behind `crypto_state_guard` (two sends to different peers take ~4s instead of ~2s). Best fit for the "hangs under concurrency on the hubs" symptom.

### WASM boundary (`crates/bitwarden-wasm-internal/integration-tests/tests/ipc/`)

Guaranteed (passing), across the real Rust/WASM/JS divide:
- Many concurrent requests correlate correctly
- A later request completes after an earlier send throws (recoverable)
- A "Destination unreachable" send is non-fatal
- An `AbortSignal` cancels a pending request and leaves the client running

Documented gap (skipped, executable spec): a rejecting client-managed session store traps the wasm module (`RuntimeError: unreachable`) and wedges the request forever (the request's own `AbortSignal` does not cancel it). Boundary counterpart of native gap 1. Skipped rather than run because triggering the trap cannot be observed cleanly by jest; the runnable proof is the native ignored test.

### How to run
- Native: `cargo test -p bitwarden-ipc --lib` (36 passing, 2 ignored). Gaps: append `-- --ignored`.
- Boundary: from `crates/bitwarden-wasm-internal/integration-tests`, `npm test` (runs against the prebuilt `npm/` package).

### Not covered (known holes)
Inbound handshake save racing outside the lock, `CryptoInvalidated` reset, end-to-end rehandshake, concurrent `start()` TOCTOU, handle-drop stopping the loop, the `discover` RPC, and GC/finalizer lifetime of JS objects held by Rust. Left out as either moot in the production singleton or not deterministically testable here.

## 🚨 Breaking Changes

None. Test-only: two new test modules plus a two-line `#[cfg(test)]` module registration in `lib.rs`. No production code changed.


crates/bitwarden-ipc/src/invariant_tests.rs
coroiu: What's with the numbering? Is `B1` missing? Just remove it all together

crates/bitwarden-ipc/src/invariant_tests.rs
```diff
@@ -0,0 +1,438 @@
...
+
+/// G1 — INVARIANT (delivery under load): a burst to one subscriber is delivered without silent
+/// loss. Debatable as a hard guarantee (the transport is a deliberately bounded broadcast), but
+/// captured because it is the "messages disappear under concurrency" failure mode.
+#[tokio::test]
+#[ignore = "INVARIANT VIOLATED (by design choice): the internal transport is a bounded \
+            tokio::broadcast channel (CHANNEL_BUFFER_CAPACITY = 50 internal; 20 WASM backend; 10 \
+            test backend). A burst larger than the buffer is silently dropped for a lagging \
+            receiver (RecvError::Lagged). Not exercised by production today. Decide whether no-loss \
+            is a wanted invariant before un-ignoring."]
```
coroiu: Remove this test, we don't guarantee this.

crates/bitwarden-wasm-internal/integration-tests/tests/ipc/ipc-invariants.test.ts
coroiu: Naming should follow this convention: https://contributing.bitwarden.com/contributing/testing/unit/naming-conventions

PR #1514, opened 2026-09-23
[PM-43030] Implement error variants of InviteLinkError into SDK

JaredScar:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-43030

## 📔 Objective

This introduces new error variants of InviteLinkError into the SDK for use of Milestone 3 for Invite Link


crates/bitwarden-organization-invite-link/src/accept_error.rs
ike-kottlowski: Auth may need to branch off an error code from this:`"email_not_verified"` but it's not mapped here.

Is there plans to map it in the future? Or would the mapping be done in the `confirm_error` flow?

crates/bitwarden-organization-invite-link/src/lib.rs
```diff
@@ -1,15 +1,7 @@
 #![doc = include_str!("../README.md")]
 
-mod error;
-mod invite_link_admin_client;
+mod accept_error;
```
JimmyVo16: Since this is being used for all three endpoints and not just `accept`, maybe `server_error` or `invite_link_error` to reflect that. Doesn't have to be these, just some suggestions that came to mind.

crates/bitwarden-organization-invite-link/src/invite_link_client.rs
```diff
@@ -1,108 +1,246 @@
...
+    MissingField(#[from] MissingFieldError),
+    /// A value was present but malformed and could not be parsed.
+    #[error("Failed to parse `{0}`")]
+    ParseFailure(&'static str),
+    /// The account-recovery public key returned by the server does not match the organization
+    /// public key bound into the invite.
+    #[error("Account recovery public key does not match the invite's bound organization key")]
+    RecoveryKeyMismatch,
+
+    // Server-reported failures when accepting or confirming an invite link, mapped from the
```
eliykat: Let's split this into its own error enum, `AcceptInviteLinkError`, containing only the error variants returned by `accept_and_optionally_confirm`. This is the pattern elsewhere, e.g. `CreateFolderError`. It prevents a large enum with errors that most of its endpoints don't use.

crates/bitwarden-organization-invite-link/src/accept_error.rs
eliykat: We should try to avoid free functions (i.e. functions not in an `impl` block).

Once `AcceptInviteLinkError` is its own enum, these can be moved inside its `impl` block. e.g. `map_err(AcceptInviteLinkError::from_api_error)`.

Note that I did consider the `From` trait, but that would automatically convert it for all api calls in this method, which we don't want - we only want to opt into it expressly for the accept/confirm endpoints.

crates/bitwarden-organization-invite-link/src/accept_error.rs
```diff
@@ -0,0 +1,205 @@
...
+        return InviteLinkError::Api(error);
+    };
+
+    match content.status.as_u16() {
+        404 => InviteLinkError::LinkNotFound,
+        400 => match first_validation_code(&content.message) {
+            Some(code) => from_code(code),
+            None => InviteLinkError::Api(error),
+        },
+        _ => InviteLinkError::Api(error),
```
eliykat: You can do a typed comparison rather than `as_u16`. See `InviteLinkAdminClient.get` on `main`:

https://github.com/bitwarden/sdk-internal/blob/main/crates/bitwarden-organization-invite-link/src/invite_link_admin_client.rs#L49

